PR #10, opened 2016-10-01
Header styling

christophehenry:


client/.bootstraprc
```diff
@@ -70,7 +70,7 @@ styles:
   # Core CSS
   scaffolding: true
   type: true
-  code: false
+  code: true
```
Chocobozzz: Where is it used?

client/.bootstraprc
```diff
@@ -93,7 +93,7 @@ styles:
   alerts: true
   progress-bars: true
   media: true
-  list-group: false
+  list-group: true
```
Chocobozzz: Where is it used?

Chocobozzz:
Merged in the styling branch. https://github.com/Chocobozzz/PeerTube/commits/styling

Issue #19, opened 2016-11-27
Add ability for an administrator to remove any video


Chocobozzz:
For now, only the user that uploaded the video can remove it. The idea is the pod administrator should have the possibility to remove any video (remote or from another user).

- [ ] Add ability to remove a remote video (the video is remote, we do not notify friends)
- [ ] Add ability to remove a video of another user that is on the same pod (so the video is local and we need to notify the video removal to our friends)

Green-Star:
Hey @Chocobozzz , I would like to work on this feature but I have some problems understanding the expected behaviour.

> Add ability to remove a remote video (just on the pod, do not notify friends)

If I understand well, admin will be able to "hide" a remote video, i.e., remove it on the local pod but the video will still be reachable on the friend pods (and on its friends).

> Add ability to remove a video of another pod user (on the pod and on pod friends, so notify them)

Completely remove a video, i.e. remove it on the pod and on our friends too.

Is that correct ?

Chocobozzz:
Yes this is correct :)

Chocobozzz:
Implemented: https://github.com/Chocobozzz/PeerTube/commit/198b205c10dba362b9ae1ef6895b29d7e0dd685f

PR #61, opened 2017-04-20
Add ability for an administrator to remove any video

Green-Star:
Should close issue #19.

At first I simply thought of granting the possibility for an admin to delete (from the DB) any video on the pod.
I had some problems when updating a deleted video information (since it was deleted ...). So I thought of another implementation. An admin can delete any video on its pod and it can blacklist a remote video. Hence, the video won't show up in the video list but there is no problem when updating the video information because it is still in the DB. (plus we can add a feature in the admin part to manage blacklisted videos or even "unblacklist" a video to see it on the video list again).

I, thus, split this feature in two parts : 

### Give an adminstrator the abilty to delete a video on the pod 
I simply modified the `isRemovableBy` method on the client side and the `videoRemove` checks on the server side.
I've also moved this feature on the `video-watch` component rather than the previous `video-list`.

Users can only delete their videos on a local pod.
Admins can't delete a remote video on the pod, but they can "hide" it thanks to a new feature : blacklisting a video.

### Blacklist a video 
This feature is on the `video-watch` component (to match the remove video feature).
Only admins can blacklist a video, and admins can only blacklist a remote video ; which means the video won't be displayed in the videos list.

I add a new relation : BlacklistedVideos.
When an admin click on the blacklist button, a `post` request is made on `/video/:videoId/blacklist`.
This route creates a new row in the `BlacklistedVideos` table. This table stores the local and remote id of the video (as key information) and the pod id of the video (and maybe some other information) only on information purpose.

I've modified the `listForApi` method of the `video` model to exclude videos in the `BlacklistedVideos` from the videos being listed.
(I had to use a `sequelize.litteral` to execute the sub-query on `BlacklistedVideos` table but I might find a more sequelize solution to avoid hard-dependency if needed).

I modified the hook deletion of a video to delete the corresponding row in `BlacklistedVideos` (if it exists).

### DIFF
--- a/server/controllers/api/videos.js
+++ b/server/controllers/api/videos.js
@@ -93,11 +93,14 @@ router.get('/:id',
   validatorsVideos.videosGet,
   getVideo
 )
+
 router.delete('/:id',
   oAuth.authenticate,
   validatorsVideos.videosRemove,
+  addVideoToBlacklist,
   removeVideo
 )
+
 router.get('/search/:value',
   validatorsVideos.videosSearch,
   validatorsPagination.pagination,
@@ -621,4 +624,26 @@ function reportVideoAbuse (req, res, finalCallback) {
     logger.info('Abuse report for video %s created.', videoInstance.name)
     return finalCallback(null)
   })
+}
+
+
+function addVideoToBlacklist (req, res, next) {
+  const videoInstance = res.locals.video
+
+  if (videoInstance.isOwned() === false) {
+    db.BlacklistedVideo.create({
+      remoteId: videoInstance.remoteId,
+      localId: videoInstance.localId,
+      remotePodId: videoInstance.Author.Pod.id
+    })
+    .asCallback(function (err) {
+      if (err) {
+	logger.error('Errors when blacklisting video ', { error: err })
+
+	return next(err)
+      }
+    })
+  }
+
+  next()
 }
--- a/server/controllers/api/videos.js
+++ b/server/controllers/api/videos.js
@@ -639,7 +639,7 @@ function addVideoToBlacklist (req, res, next) {
   if (videoInstance.isOwned() === false) {
     db.BlacklistedVideo.create({
       remoteId: videoInstance.remoteId,
-      localId: videoInstance.localId,
+      localId: videoInstance.id,
       remotePodId: videoInstance.Author.Pod.id
     })
     .asCallback(function (err) {
@@ -647,7 +647,10 @@ function addVideoToBlacklist (req, res, next) {
 	logger.error('Errors when blacklisting video ', { error: err })
 	return next(err)
       }
+
+      return res.type('json').status(204).end()
     })
+    return
   }
 
   next()
--- a/server/controllers/api/videos.js
+++ b/server/controllers/api/videos.js
@@ -97,7 +97,6 @@ router.get('/:id',
 router.delete('/:id',
   oAuth.authenticate,
   validatorsVideos.videosRemove,
-  addVideoToBlacklist,
   removeVideo
 )
 
@@ -109,6 +108,13 @@ router.get('/search/:value',
   pagination.setPagination,
   search.setVideosSearch,
   searchVideos
+)
+
+router.post('/:id/blacklist',
+  oAuth.authenticate,
+  admin.ensureIsAdmin,
+  validatorsVideos.videosBlacklist,
+  addVideoToBlacklist
 )
 
 // ---------------------------------------------------------------------------
@@ -639,7 +645,6 @@ function addVideoToBlacklist (req, res, next) {
     .asCallback(function (err) {
       if (err) {
 	logger.error('Errors when blacklisting video ', { error: err })
-
 	return next(err)
       }
     })
--- a/server/controllers/api/videos.js
+++ b/server/controllers/api/videos.js
@@ -635,22 +635,13 @@ function reportVideoAbuse (req, res, finalCallback) {
 function addVideoToBlacklist (req, res, next) {
   const videoInstance = res.locals.video
 
-  if (videoInstance.isOwned() === false) {
-    db.BlacklistedVideo.create({
-      remoteId: videoInstance.remoteId,
-      localId: videoInstance.id,
-      remotePodId: videoInstance.Author.Pod.id
-    })
-    .asCallback(function (err) {
-      if (err) {
-        logger.error('Errors when blacklisting video ', { error: err })
-        return next(err)
-      }
-
-      return res.type('json').status(204).end()
-    })
-    return
-  }
-
-  next()
+  db.BlacklistedVideo.create({
+    videoId: videoInstance.id
+  })
+  .asCallback(function (err) {
+    if (err) {
+      logger.error('Errors when blacklisting video ', { error: err })
+      return next(err)
+    }
+  })
 }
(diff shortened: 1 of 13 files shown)


server/middlewares/validators/videos.js
```diff
@@ -159,3 +156,40 @@ function checkVideoExists (id, res, callback) {
...
+
+      if (res.locals.video.Author.userId !== res.locals.oauth.token.User.id) {
+        return res.status(403).send('Cannot remove video of another user')
+      }
+    }
+
+    // If we reach this comment, we can delete the video
+  })
+
+  callback()
```
Chocobozzz: The callback should be below the *If we reach this comment, we can delete the video* comment.

server/middlewares/validators/videos.js
```diff
@@ -159,3 +156,40 @@ function checkVideoExists (id, res, callback) {
...
+}
+
+function videosBlacklist (req, res, next) {
+  req.checkParams('id', 'Should have a valid id').notEmpty().isUUID(4)
+
+  logger.debug('Checking videosBlacklist parameters', { parameters: req.params })
+
```
Chocobozzz: You should check the video is not owned too.

server/controllers/api/videos.js
```diff
@@ -622,3 +631,26 @@ function reportVideoAbuse (req, res, finalCallback) {
     return finalCallback(null)
   })
 }
+
+function addVideoToBlacklist (req, res, next) {
+  const videoInstance = res.locals.video
+
+  if (videoInstance.isOwned() === false) {
```
Chocobozzz: This verification should be placed in the videoToBlacklist validator.

server/controllers/api/videos.js
```diff
@@ -622,3 +631,26 @@ function reportVideoAbuse (req, res, finalCallback) {
...
+    })
+    .asCallback(function (err) {
+      if (err) {
+        logger.error('Errors when blacklisting video ', { error: err })
+        return next(err)
+      }
+
+      return res.type('json').status(204).end()
+    })
+    return
```
Chocobozzz: Remove it

server/controllers/api/videos.js
```diff
@@ -622,3 +631,26 @@ function reportVideoAbuse (req, res, finalCallback) {
...
+  }
+
+  next()
```
Chocobozzz: You don't need this.

server/models/video.js
```diff
@@ -540,7 +547,12 @@ function listForApi (start, count, sort, callback) {
       },
 
       this.sequelize.models.Tag
-    ]
+    ],
+    where: {
```
Chocobozzz: You need to put these where clause in the `searchAndPopulateAuthorAndPodAndTags` method too

server/models/video.js
```diff
@@ -755,3 +767,31 @@ function generateImage (video, videoPath, folder, imageName, size, callback) {
...
+    // If an error occured, stop here
+    if (err) {
+      logger.error('Error when fetching video from blacklist.', { error: err })
+
+      return callback(err)
+    }
+
+    // If we found the video, remove it from the blacklist
+    if (video) {
+      video.destroy().asCallback(function (err) {
```
Chocobozzz: `video.destroy().asCallback(callback)`

Just let caller function handle error logging.

server/models/video-blacklist.js
```diff
@@ -0,0 +1,110 @@
+'use strict'
+
+const modelUtils = require('./utils')
+
+// ---------------------------------------------------------------------------
+
+module.exports = function (sequelize, DataTypes) {
+  const BlacklistedVideo = sequelize.define('BlacklistedVideo',
```
Chocobozzz: You don't need `remoteId` and `localId` (so we don't define columns explicitly in this table). This table will just be composed by two columns: `id` and `videoId`. The first is implicit and the second will be declared in the `associate` method.

server/models/video-blacklist.js
```diff
@@ -0,0 +1,110 @@
...
+        }
+      }
+    },
+    {
+      indexes: [
+        {
+          fields: [ 'remoteId' ]
+        },
+        {
+          fields: [ 'localId' ],
```
Chocobozzz: `localId` -> `videoId`

server/models/video-blacklist.js
```diff
@@ -0,0 +1,110 @@
...
+        allowNull: false,
+        validate: {
+          isUUID: 4
+        }
+      }
+    },
+    {
+      indexes: [
+        {
+          fields: [ 'remoteId' ]
```
Chocobozzz: Remove it

server/models/video-blacklist.js
```diff
@@ -0,0 +1,110 @@
...
+    localId: this.localId,
+    remotePodId: this.remotePodId,
+    createdAt: this.createdAt
+  }
+}
+
+// ------------------------------ STATICS ------------------------------
+
+function associate (models) {
+  this.belongsTo(models.Pod, {
```
Chocobozzz: The `BlacklistedVideos` table just belongs to the `Videos` table (to create the `videoId` foreign key). So:

    this.belongsTo(models.Video, {
      foreignKey: 'videoId',
      onDelete: 'cascade'
    })

server/models/video-blacklist.js
```diff
@@ -0,0 +1,110 @@
...
+  })
+}
+
+function loadById (id, callback) {
+  return this.findById(id).asCallback(callback)
+}
+
+function loadByPod (remotePodId, callback) {
+  const query = {
+    where: {
```
Chocobozzz: With ES6 you can do:

     where: {
       remotePodId
     }

server/models/video.js
```diff
@@ -303,6 +303,12 @@ function afterDestroy (video, options, next) {
         return callback()
       }
     )
+  } else {
```
Chocobozzz: You don't need this because we put a `onDelete: 'cascade'` on the `Videos`/`BlacklistedVideos` relation.

server/models/video.js
```diff
@@ -755,3 +767,31 @@ function generateImage (video, videoPath, folder, imageName, size, callback) {
     })
     .thumbnail(options)
 }
+
+function removeFromBlacklist (video, callback) {
+  // Find the blacklisted video
+  video.sequelize.models.BlacklistedVideo.loadById(video.remoteId, function (err, video) {
```
Chocobozzz: `db.BlacklistedVideo`

`loadByVideoId(video.id, ...`

server/controllers/api/videos.js
Green-Star: I have left `return res.type('json').status(204).end()` because I get an error on the client when blacklisting a video otherwise. To be more precise, everything works fine, the video _is_ blacklisted but the notification shows an error status.

server/controllers/api/videos.js
```diff
@@ -643,5 +643,7 @@ function addVideoToBlacklist (req, res, next) {
       logger.error('Errors when blacklisting video ', { error: err })
       return next(err)
     }
+
+    return res.type('json').status(204).end()
```
Green-Star: I have left `return res.type('json').status(204).end()` because I get an error on the client when blacklisting a video otherwise. To be more precise, everything works fine, the video _is_ blacklisted but the notification shows an error status.

Chocobozzz: 👍

Issue #20, opened 2016-11-27
Add ability to remove a particular friend


Chocobozzz:
If an administrator notices that a particular friend send bad videos or spams... it would be interesting to remove it throught the interface.

PR #66, opened 2017-05-20
Remove one pod

Green-Star:
Should close issue #20.

I've just used the `quitFriends` method and adapt it to remove one friend.

I can see one problem with this method, though : It's not really DRY. Maybe we should add one method in the `server/lib/friends.js` to remove just one friend and call it 
- in the removeFriend method
- while iterating over the friend list in quitFriends 
?

### DIFF
--- a/client/src/app/+admin/friends/friend-list/friend-list.component.html
+++ b/client/src/app/+admin/friends/friend-list/friend-list.component.html
@@ -2,7 +2,7 @@
   <div class="content-padding">
     <h3>Friends list</h3>
 
-    <ng2-smart-table [settings]="tableSettings" [source]="friendsSource"></ng2-smart-table>
+    <ng2-smart-table [settings]="tableSettings" [source]="friendsSource" (delete)="removeFriend($event)"></ng2-smart-table>
 
     <a *ngIf="hasFriends()" class="btn btn-danger pull-left" (click)="quitFriends()">
       Quit friends
--- a/client/src/app/+admin/friends/friend-list/friend-list.component.ts
+++ b/client/src/app/+admin/friends/friend-list/friend-list.component.ts
@@ -71,8 +71,7 @@ export class FriendListComponent {
 
         this.friendService.quitFriends().subscribe(
           status => {
-            this.notificationsService.success('Sucess', 'Friends left!')
-
+            this.notificationsService.success('Success', 'Friends left!');
             this.friendsSource.refresh()
           },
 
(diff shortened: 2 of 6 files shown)


server/lib/friends.js
```diff
@@ -222,6 +223,55 @@ function quitFriends (callback) {
   })
 }
 
+function removeFriend (podId, callback) {
+  // Stop pool requests
+  requestScheduler.deactivate()
+
+  waterfall([
+    function flushRequests (callbackAsync) {
+      requestScheduler.flush(err => callbackAsync(err))
```
Chocobozzz: You don't need this. If the pod doesn't exist in our database anymore we won't do requests on it.

server/lib/friends.js
```diff
@@ -222,6 +223,55 @@ function quitFriends (callback) {
...
+    },
+
+    function flushVideoQaduRequests (callbackAsync) {
+      requestVideoQaduScheduler.flush(err => callbackAsync(err))
```
Chocobozzz: Same here: https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/request-video-qadu.js#L82

server/lib/friends.js
```diff
@@ -222,6 +223,55 @@ function quitFriends (callback) {
...
+    },
+
+    function getPod (callbackAsync) {
+      return db.Pod.load(podId, callbackAsync)
```
Chocobozzz: Please create a middleware validator which checks the pod exists and put it in the `req` object (you could take example of videos https://github.com/Chocobozzz/PeerTube/blob/develop/server/middlewares/validators/videos.js#L91).

server/lib/friends.js
```diff
@@ -222,6 +223,55 @@ function quitFriends (callback) {
...
+    },
+
+    function announceIQuitThisFriend (pod, callbackAsync) {
+      const requestParams = {
+        method: 'POST',
+        path: '/api/' + constants.API_VERSION + '/remote/pods/remove',
+        sign: true
+      }
+
+      requestParams.toPod = pod
```
Chocobozzz: You could put this directly in the object initialization.

server/lib/friends.js
```diff
@@ -222,6 +223,55 @@ function quitFriends (callback) {
...
+      requests.makeSecureRequest(requestParams, function (err) {
+        if (err) {
+          logger.error('Some errors while quitting friend.', { err: err })
```
Chocobozzz: `logger.error('Some errors while quitting friend %d.', pod.host, { err: err })` ?

PR #71, opened 2017-06-21
Fix test

Green-Star:
Some tests timeouted when I tried to run them on my computer.
I've just increase/add some timeouts to pass them.

### DIFF
--- a/server/tests/api/check-params/videos.js
+++ b/server/tests/api/check-params/videos.js
@@ -392,6 +392,8 @@ describe('Test videos API validator', function () {
       const attach = {
         'videofile': pathUtils.join(__dirname, '..', 'fixtures', 'video_short.webm')
       }
+      this.timeout(10000)
+
       requestsUtils.makePostUploadRequest(server.url, path, server.accessToken, data, attach, function () {
         attach.videofile = pathUtils.join(__dirname, '..', 'fixtures', 'video_short.mp4')
         requestsUtils.makePostUploadRequest(server.url, path, server.accessToken, data, attach, function () {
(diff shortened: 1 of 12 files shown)


server/tests/api/single-pod.js
```diff
@@ -187,14 +187,19 @@ describe('Test a single pod', function () {
   })
 
   it('Should have the views updated', function (done) {
-    videosUtils.getVideo(server.url, videoId, function (err, res) {
-      if (err) throw err
+    this.timeout(5000)
 
-      const video = res.body
-      expect(video.views).to.equal(1)
+    setTimeout(videosUtils.getVideo,
```
Chocobozzz: Are you sure this is necessary?

Green-Star: The failed on my laptop (video.views was equal to 0) so I add this timeout. 1000ms may be enough tough.

Chocobozzz: Please move the setTimeout in the previous tests (and try with 500ms):

```
 videosUtils.testVideoImage(server.url, 'video_short.webm', video.thumbnailPath, function (err, test) {
        if (err) throw err
        expect(test).to.equal(true)

        // Wait the async views increment
        setTimeout(done, 500)
})
```

So you don't need to modify the `Should have the views updated` test.

server/tests/api/check-params/videos.js
```diff
@@ -392,6 +392,8 @@ describe('Test videos API validator', function () {
       const attach = {
         'videofile': pathUtils.join(__dirname, '..', 'fixtures', 'video_short.webm')
       }
+      this.timeout(10000)
```
Chocobozzz: Move this to the top

server/tests/api/friends-basic.js
```diff
@@ -143,10 +143,13 @@ describe('Test basic friends', function () {
 
   it('Should not be allowed to make friend again', function (done) {
     const server = servers[1]
+    this.timeout(10000)
```
Chocobozzz: Move this to the top + \n

PR #76, opened 2017-07-06
Remove one pod

Green-Star:
Should close issue #20.

I've just used the `quitFriends` method and adapt it to remove one friend.

I can see one problem with this method, though : It's not really DRY. Maybe we should add one method in the `server/lib/friends.js` to remove just one friend and call it 
- in the removeFriend method
- while iterating over the friend list in quitFriends 
?

### DIFF
--- a/server/lib/friends.ts
+++ b/server/lib/friends.ts
@@ -242,7 +242,7 @@ function fetchRemotePreview (pod: PodInstance, video: VideoInstance) {
   return request.get(REMOTE_SCHEME.HTTP + '://' + host + path)
 }
 
-function removeFriend (podId, callback) {
+function removeFriend (podId: number, callback: (err: Error) => void) {
   // Stop pool requests
   requestScheduler.deactivate()
 
@@ -261,13 +261,13 @@ function removeFriend (podId, callback) {
 
     function announceIQuitThisFriend (pod, callbackAsync) {
       const requestParams = {
-        method: 'POST',
-        path: '/api/' + constants.API_VERSION + '/remote/pods/remove',
-        sign: true
+        method: 'POST' as 'POST',
+        path: '/api/' + API_VERSION + '/remote/pods/remove',
+        sign: true,
+        toPod: pod
       }
 
-      requestParams.toPod = pod
-      requests.makeSecureRequest(requestParams, function (err) {
+      makeSecureRequest(requestParams, function (err) {
         if (err) {
           logger.error('Some errors while quitting friend.', { err: err })
           // Continue anyway
@@ -280,7 +280,7 @@ function removeFriend (podId, callback) {
     function removePodFromDB (pod, callbackAsync) {
       pod.destroy().asCallback(callbackAsync)
     }
-  ], function (err) {
+  ], function (err: Error) {
     // Don't forget to re activate the scheduler, even if there was an error
     requestScheduler.activate()
 
@@ -317,6 +317,7 @@ export {
   hasFriends,
   makeFriends,
   quitFriends,
+  removeFriend,
   removeVideoToFriends,
   sendOwnedVideosToPod,
   getRequestScheduler,
--- a/server/lib/friends.ts
+++ b/server/lib/friends.ts
@@ -243,9 +243,6 @@ function fetchRemotePreview (pod: PodInstance, video: VideoInstance) {
 }
 
 function removeFriend (pod: PodInstance) {
-  // Stop pool requests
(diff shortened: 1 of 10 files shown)


server/lib/friends.ts
```diff
@@ -229,6 +229,28 @@ function sendOwnedVideosToPod (podId: number) {
     })
 }
 
+function removeFriend (pod: PodInstance) {
+  // Stop pool requests
+  requestScheduler.deactivate()
```
Chocobozzz: I'm not sure we need to stop the scheduler

server/tests/api/friends-advanced.js
```diff
@@ -25,6 +25,27 @@ describe('Test advanced friends', function () {
...
 
+  function removeFriend (podNumber, podNumberToRemove, callback) {
+    const server = servers[podNumber - 1]
+    const serverToRemove = servers[podNumberToRemove - 1]
+
+    getFriendsList(podNumber, function (err, res) {
+      if (err) throw err
+
+      let friendsList = res.body.data
+      let podIdToRemove = -1
```
Chocobozzz: `podIdToRemove = friendsList.find(friend => friend.host === serverToRemove.host)`

server/tests/api/friends-basic.js
```diff
@@ -198,6 +198,78 @@ describe('Test basic friends', function () {
...
+      // Pod 1 quits pod 2
+      function (next) {
+        const server = servers[0]
+
+        // Get pod 2 id so we can query it
+        podsUtils.getFriendsList(server.url, function (err, res) {
+          if (err) throw err
+
+          const result = res.body.data
+          let podId
```
Chocobozzz: Same as above, use `.find`

server/tests/utils/pods.js
```diff
@@ -90,6 +91,27 @@ function quitFriends (url, accessToken, expectedStatus, end) {
...
+  request(url)
+    .delete(path)
+    .set('Accept', 'application/json')
+    .set('Authorization', 'Bearer ' + accessToken)
+    .expect(expectedStatus)
+    .end(function (err, res) {
+      if (err) throw err
+
+      // Wait for the request between pods
+      setTimeout(end, 1000)
```
Chocobozzz: `setTimeout` is not needed

server/lib/friends.ts
```diff
@@ -242,6 +242,24 @@ function fetchRemotePreview (pod: PodInstance, video: VideoInstance) {
   return request.get(REMOTE_SCHEME.HTTP + '://' + host + path)
 }
 
+function removeFriend (pod: PodInstance) {
+  const requestParams = {
+    method: 'POST' as 'POST',
+    path: '/api/' + API_VERSION + '/remote/pods/remove',
+    sign: true,
```
Chocobozzz: You don't need this anymore

server/lib/friends.ts
```diff
@@ -242,6 +242,24 @@ function fetchRemotePreview (pod: PodInstance, video: VideoInstance) {
...
+    toPod: pod
+  }
+
+  return makeSecureRequest(requestParams)
+    .then(() => pod.destroy())
+    .then(() => {
+      logger.info('Removed friend.')
+    })
+    .catch(err => {
+      logger.error('Some errors while quitting friend %s (id: %d).', pod.host, pod.id, { err: err })
```
Chocobozzz: `logger.error('Some errors while quitting friend %s (id: %d).', pod.host, pod.id, err)`

server/middlewares/validators/pods.ts
```diff
@@ -58,9 +58,33 @@ function podsAddValidator (req: express.Request, res: express.Response, next: ex
...
+        if (!pod) {
+          logger.error('Cannot find pod %d.', req.params.id)
+          return res.sendStatus(404)
+        }
+
+        res.locals.pod = pod
+        return next()
+      })
+      .catch(err => {
+        logger.error('Cannot load pod %d.', req.params.id, { error: err })
```
Chocobozzz: `logger.error('Cannot load pod %d.', req.params.id, err)`

server/lib/friends.ts
Green-Star: We don't need to post a remove request to the pod we're quitting ?

Chocobozzz: You don't need the `sign` parameter (it was removed recently, now all requests are automatically signed).

Green-Star: Okay :+1:

Issue #63, opened 2017-04-28
Add blacklisted videos management


Green-Star:
In the admin part, we could add a new component to manage blacklisted videos.

In this component, admins can (at least) :
- [ ] list the blacklisted videos on the pod.
- [ ] remove a video from the blacklist (hence makes it available again in the videos list).

For the blacklist list, it might be useful to modify the blacklist model to add informations about the blacklisted videos (I think it would be nice to store at least the admin name who did the blacklist, and maybe add a reason for the blacklist too ?).
In the list, we should display informations about videos.

Chocobozzz:
Implemented: https://github.com/Chocobozzz/PeerTube/commit/792dbaf07f83fbe3f1d209cd9edf190442c7d2f3

Thanks @Green-Star!

PR #84, opened 2017-09-01
Handle blacklist

Green-Star:
Should close issue #63.

With this feature, admins will be able to list the blacklisted videos on the pod (and get some information about the video). They will also be able to remove a video from the blacklist and make it available on the pod again.

I've add a new component on the admin part to manage the blacklist.


client/src/app/shared/blacklist/blacklist.model.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.name = hash.name
+    this.uuid = hash.uuid
+    this.description = hash.description
+    this.duration = hash.duration
+    this.views = hash.views
+    this.likes = hash.likes
+    this.dislikes = hash.dislikes
+    this.nsfw = hash.nsfw
+
+    if (hash.createdAt) {
```
Chocobozzz: Why is is optional?

client/src/app/shared/blacklist/blacklist.model.ts
```diff
@@ -0,0 +1,49 @@
...
+      this.createdAt = hash.createdAt
+    }
+    if (hash.updatedAt) {
```
Chocobozzz: Why is is optional?

server/controllers/api/index.ts
```diff
@@ -19,6 +20,7 @@ apiRouter.use('/remote', remoteRouter)
 apiRouter.use('/request-schedulers', requestSchedulerRouter)
 apiRouter.use('/users', usersRouter)
 apiRouter.use('/videos', videosRouter)
+apiRouter.use('/blacklists', blacklistsRouter)
```
Chocobozzz: I think `/blacklist` is better because we always have only one blacklist.

server/tests/api/video-blacklist-management.js
```diff
@@ -0,0 +1,235 @@
...
+      videoBlacklistsUtils.getSortedBlacklistedVideosList(servers[0].url, servers[0].accessToken, '-id', function (err, res) {
+        if (err) throw err
+
+        expect(res.body.total).to.equal(2)
+
+        const videos = res.body.data
+        expect(videos).to.be.an('array')
+        expect(videos.length).to.equal(2)
+
+        const result = res.body.data.slice(0).sort((a, b) => {
```
Chocobozzz: Try to use https://lodash.com/docs/4.17.4#sortBy

And why do you use slice?

server/tests/api/video-blacklist-management.js
```diff
@@ -0,0 +1,235 @@
...
+      videoBlacklistsUtils.getSortedBlacklistedVideosList(servers[0].url, servers[0].accessToken, '-name', function (err, res) {
+        if (err) throw err
+
+        expect(res.body.total).to.equal(2)
+
+        const videos = res.body.data
+        expect(videos).to.be.an('array')
+        expect(videos.length).to.equal(2)
+
+        const result = res.body.data.slice(0).sort((a, b) => {
```
Chocobozzz: Same here

server/tests/api/video-blacklist-management.js
```diff
@@ -0,0 +1,235 @@
...
+      videoBlacklistsUtils.getSortedBlacklistedVideosList(servers[0].url, servers[0].accessToken, 'createdAt', function (err, res) {
+        if (err) throw err
+
+        expect(res.body.total).to.equal(2)
+
+        const videos = res.body.data
+        expect(videos).to.be.an('array')
+        expect(videos.length).to.equal(2)
+
+        const result = res.body.data.slice(0).sort((a, b) => {
```
Chocobozzz: Same here

server/tests/api/video-blacklist-management.js
```diff
@@ -0,0 +1,235 @@
...
+
+        expect(videos).to.deep.equal(result)
+
+        done()
+      })
+    })
+  })
+
+  describe('When removing a blacklisted video', function () {
+    it('Should not have any video in videos search on pod 1', function (done) {
```
Chocobozzz: "in videos list"

server/tests/api/video-blacklist-management.js
```diff
@@ -0,0 +1,235 @@
...
+
+    it('Should remove a video from the blacklist on pod 1', function (done) {
+      series([
+        // Get one video in the blacklist
+	function (next) {
+          videoBlacklistsUtils.getSortedBlacklistedVideosList(servers[0].url, servers[0].accessToken, '-name', function (err, res) {
+	    if (err) throw err
+
+	    servers[0].videoToRemove = res.body.data[0]
+	    servers[0].blacklist = res.body.data.splice(1)
```
Chocobozzz: Why splice?

shared/models/videos/video-blacklist.model.ts
```diff
@@ -3,3 +3,18 @@ export interface BlacklistedVideo {
   videoId: number
   createdAt: Date
 }
+
+export interface RestBlacklistedVideo {
```
Chocobozzz: We put in `shared` objects that are shared with the client so the formated videos. Please only export one interface that include all fields sent by the server. 

If you need more fields or methods in the client you can create another model file like we do for the Video model for example: https://github.com/Green-Star/PeerTube/blob/a9c5666773a2ac0e3b06ed6b3e717519b9f561e3/client/src/app/videos/shared/video.model.ts

server/controllers/api/blacklists.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+  })
+
+  return {
+    total: formatedList.length,
+    data: formatedList
+  }
+}
+
+function formatBlacklistObject (blacklist: BlacklistedVideoInstance, video: VideoInstance): RestBlacklistedVideoInstance {
```
Chocobozzz: Should be in the BlacklistVideo model like in Video for example https://github.com/Green-Star/PeerTube/blob/a9c5666773a2ac0e3b06ed6b3e717519b9f561e3/server/models/video/video.ts#L417

server/controllers/api/blacklists.ts
```diff
@@ -0,0 +1,92 @@
...
+
+function removeVideoFromBlacklistController (req: express.Request, res: express.Response, next: express.NextFunction) {
+  const entry = res.locals.blacklistEntryToRemove as BlacklistedVideoInstance
+
+  removeVideoFromBlacklist(entry)
+    .then(() => res.sendStatus(204))
+    .catch(err => next(err))
+}
+
+function formatBlacklistForRest (resultList): ResultList<RestBlacklistedVideoInstance> {
```
Chocobozzz: Use `getFormattedObjects` -> https://github.com/Green-Star/PeerTube/blob/a9c5666773a2ac0e3b06ed6b3e717519b9f561e3/server/helpers/utils.ts

server/tests/api/video-blacklist-management.js
Green-Star: Because in the test I want to check if the two arrays have the same the content. I don't want to check if they have the same reference so I'm not using `expect(videos).equals(result)`.

To be sure the `videos` array have the same content as `result`, while being a different object, I make a sorted copy of the `res.body.data` object (which is the result I expect in the test). And `slice(0)` is the fastest way to make a copy of an array in Javascript according to many benchmarks.

server/tests/api/video-blacklist-management.js
Green-Star: Same here

server/tests/api/video-blacklist-management.js
Green-Star: Same here

server/tests/api/video-blacklist-management.js
Green-Star: I think it's a mistake. It should be `slice(1)`. I will correct it ;)

server/tests/api/video-blacklist-management.js
Chocobozzz: 👍

Green-Star: I will use the [orderBy](https://lodash.com/docs#orderBy) method because it allows to specify sort orders. Also, this method creates a new array so there's no need to use `slice(0)` anymore

Chocobozzz:
Beautiful 👍

PR #87, opened 2017-09-07
Replace Diaspora reference with Patchwork

WildYorkies:
Diaspora is not decentralized. It is federated (AKA distributed). Patchwork (on the Secure Scuttlebutt protocol) is a lot closer to the social media version of PeerTube.

Was just bugging me a bit :)

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,8 @@ Want to see in action?
 
 We can't build a FOSS video streaming alternatives to YouTube, Dailymotion, Vimeo... with a centralized software. One organization alone cannot have enought money to pay bandwith and video storage of its server.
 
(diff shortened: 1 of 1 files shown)


Chocobozzz:
In fact for now PeerTube is distributed (all pods are communicating with all other pods of the network), but in the future administrators will be able to blacklist some pods (just hide their videos in fact) with a blacklist/whitelist system. There may be many different videos between pods that are in the same network.

So it would seem decentralized but under the hood it will stay a distributed protocol. I think I'll need to s/decentralized/distributed/ docs.

What is your opinion?

WildYorkies:
Oh, I didn't realize that! Makes sense to me.

Issue #88, opened 2017-09-08
PeerTube runs out of memory during client compilation


Green-Star:
I've recently updated my local develop from commit `e7dbeae8d915cdf4470ceb51c2724b04148b30b5` to 
`0cb11df7f5b6ae78e6694f160bdce0da4f0376db`
Since this update, Peertube runs OOM when I try to build the client part. Everything was running fine before the update.
The server part compiles without any troubles.

I build the client using `npm run build:client`
The stack trace I get at the end of the compilation : 
```
<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x346d1fa3fa99 <JS Object>
    1: DoJoin(aka DoJoin) [native array.js:~129] [pc=0x1ef5910abff9] (this=0x346d1fa04241 <undefined>,w=0x37371dcf3169 <JS Array[798]>,x=798,N=0x346d1fa04281 <true>,J=0x200dc5c68431 <String[6]\: ,\n    >,I=0x3e31b55bb729 <JS Function ConvertToString (SharedFunctionInfo 0x346d1fa5dbc9)>)
    2: Join(aka Join) [native array.js:180] [pc=0x1ef5741841b2] (this=0x346d1fa04241 <undefined>,w=0x37371dcf...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [node]
 2: 0x10ca9dc [node]
 3: v8::Utils::ReportApiFailure(char const*, char const*) [node]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [node]
 5: v8::internal::Factory::NewRawTwoByteString(int, v8::internal::PretenureFlag) [node]
 6: v8::internal::Runtime_StringBuilderJoin(int, v8::internal::Object**, v8::internal::Isolate*) [node]
 7: 0x1ef5741060c7
Aborted
```

PR #89, opened 2017-09-08
Client: Fix Javascript OOM error during webpack compilation

Green-Star:
Linked to issue #88.

Javascript running OOM during webpack compilation seem to be an known issue :
https://github.com/webpack/webpack/issues/1914

The last comment of this issue (https://github.com/webpack/webpack/issues/1914#issuecomment-326937497) fixed it by disabling the ModuleConcatenationPlugin. I did the same on PeerTube and the client part compiles on my PC 🎉


Chocobozzz:
It's a pitty :/ 

I hope we will be able to reactivate it because it optimizes the build a lot.

PR #97, opened 2017-10-12
Fix some inconsistency in link

### LINKED ISSUE #96: More meaningful error when uploading a video in a not supported format
```
There have been validation errors: [ { param: 'videofile', msg: 'Invalid value', value: undefined } ]
```

I was having this error and couldn't find why. But reading FAQ.md I found that Peertube doesn't support .mkv format. We should perhaps tell the user what format we currently support and handle the error to display a meaningful message.

(I am willing to do a PR to fix this issue)

flyingrub:
Some link could appear with double // in url.
e.g : "POST /api/v1/videos//upload HTTP/1.1"

A simple PR for now, meanwhile trying to fix #96

### DIFF
--- a/client/src/app/videos/shared/video.service.ts
+++ b/client/src/app/videos/shared/video.service.ts
@@ -52,13 +52,13 @@ export class VideoService {
       nsfw: video.nsfw
     }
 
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks!

PR #98, opened 2017-10-13
Handle express-validator error on the client side and fix #96

### LINKED ISSUE #96: More meaningful error when uploading a video in a not supported format

flyingrub:


server/middlewares/validators/videos.ts
```diff
@@ -23,7 +23,8 @@ import {
 } from '../../helpers'
 
 const videosAddValidator = [
-  body('videofile').custom((value, { req }) => isVideoFile(req.files)).withMessage('Should have a valid file'),
+  body('videofile').custom((value, { req }) => isVideoFile(req.files))
+                   .withMessage('This file is is not supported. Are you sure it is of the following type : ' + CONSTRAINTS_FIELDS.VIDEOS.EXTNAME),
```
Chocobozzz: Typo (is is)

"Check it is of the following type..."

client/src/app/shared/rest/rest-extractor.service.ts
```diff
@@ -45,8 +45,18 @@ export class RestExtractor {
       errorMessage = err.error.message
       console.error('An error occurred:', errorMessage)
     } else if (err.status !== undefined) {
-      const body = err.error
-      errorMessage = body ? body.error : 'Unknown error.'
+      // A server-side error occurred.
+      let body = JSON.parse(err.error)
```
Chocobozzz: Add a "TODO: remove when https://github.com/angular/angular/pull/19477#issuecomment-335795151 is fixed"

Chocobozzz:
Thanks!

PR #108, opened 2017-10-25
hide error message in https too

qsypoq:
it's the shortest method to do it (while creating another `if `or `else if` statement add more code.)


Chocobozzz:
Thanks!

PR #129, opened 2017-11-28
Fix webpack link in client documentation

sticmac:
Webpack link used to lead to webpack 1 documentation.
Fixed it to reach lastest documentation instead.


Chocobozzz:
👍

PR #144, opened 2017-12-04
Add a PeerTube development environment Dockerfile compatible with Janitor.

jankeromnes:
This should allow us to set up a Docker Hub automated build containing an up-to-date PeerTube development environment.

This image is based on [ubuntu-dev](https://github.com/JanitorTechnology/dockerfiles#ubuntu-dev), so it should be really convenient and efficient to use for developers, and it's also compatible with the https://janitor.technology service.

### DIFF
--- a/support/docker/dev/Dockerfile
+++ b/support/docker/dev/Dockerfile
@@ -1,0 +1,20 @@
+FROM janx/ubuntu-dev
+MAINTAINER Jan Keromnes <janx@linux.com>
+
+# Install PeerTube's dependencies.
+# Packages are from https://github.com/Chocobozzz/PeerTube#dependencies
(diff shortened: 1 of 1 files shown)


support/docker/dev/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Install PeerTube's dependencies.
+# Packages are from https://github.com/Chocobozzz/PeerTube#dependencies
+RUN sudo apt-get update -q && sudo apt-get install -qy \
+ ffmpeg \
+ postgresql \
+ openssl
+
+# Download PeerTube's source code.
+RUN git clone -b master https://github.com/Chocobozzz/PeerTube /home/user/PeerTube
```
Chocobozzz: I think it should be better to clone the `develop` branch :)

Issue #106, opened 2017-10-23
Add user setting to not play automatically videos


imtube:
Some customers don't want the video to autoplay, is there any way to make it not play automatically ?

zeealiawan:
can you please guide to install it on centos, i contact the author by email but still no response. if you can help to do so?

Chocobozzz:
@zeealiawan Please don't spam issues.

@imtube Maybe a setting would be interesting.

Letiteuf55:
Maybe a user setting will be better than a server setting.

Chocobozzz:
Agreed. I updated the title.

dedesite:
Hi, I start working on this issue and add a migration file for adding an "autoPlayVideo" column in User table but notice that there was no devDependencie for `sequelize-cli`, so how do you apply those migration ?

Another problem : don't know why but `npm run dev:client` doesn't work anymore. I clean up my `node_modules` dir and rebuild everything but when I go to `localhost:3000/` I got nothing.
No errors in the console, just the regular debug print of the SQL requests. 

For now, I can work with `npm run dev:server` but it's not optimal.

dedesite:
I found that we should manually update `LAST_MIGRATION_VERSION` in `server/initializers/constants.ts.ts`

dedesite:
My first _real_ PR, tell me if it is what you wanted :)

Chocobozzz:
Implemented: https://github.com/Chocobozzz/PeerTube/pull/159

Thanks @dedesite

dedesite:
Glad I could help ! Wish I had more time to work on PeerTube :).

PR #159, opened 2017-12-18
Enh #106 : Add an autoPlayVideo user attribute

dedesite:
!!! WARNING : I was not able to run the tests on my machine. It uses a different approach to handle databse connexion and didn't find where to configure it... please run the tests before and don't kill me if they don't pass... and also help me running my tests :)

- create a migration file to add a boolean column in user table
- add autoPlayVideo attribute everywhere it is needed (both on client and server side)
- add tests
- add a way to configure this attribute in account-settings
- use the attribute in video-watch component to actually autoplay or not the video

Note: it was quite a pain to include a simple attribute to the user. User's data structure is declared multiple times and checks are also done at different places. It feels like things could be made implicit and not always explicit (ie copy the entire objet instead of each attributes) but I'm new to the project, I may miss some important constraints.

### DIFF
--- a/client/src/app/account/account-settings/account-details/account-details.component.html
+++ b/client/src/app/account/account-settings/account-details/account-details.component.html
@@ -9,6 +9,12 @@
   <div *ngIf="formErrors['displayNSFW']" class="alert alert-danger">
     {{ formErrors['displayNSFW'] }}
   </div>
+  <br/>
+  <input
+    type="checkbox" id="autoPlayVideo"
+    formControlName="autoPlayVideo"
+  >
+  <label for="autoPlayVideo">Automatically plays video</label>
 
   <input type="submit" value="Save" [disabled]="!form.valid">
 </form>
--- a/client/src/app/account/account-settings/account-details/account-details.component.ts
+++ b/client/src/app/account/account-settings/account-details/account-details.component.ts
@@ -31,7 +31,8 @@ export class AccountDetailsComponent extends FormReactive implements OnInit {
 
   buildForm () {
     this.form = this.formBuilder.group({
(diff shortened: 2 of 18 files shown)


dedesite:
I realize that this PR is quite "big", tell me if you want it to be cut into several commits.

Chocobozzz:
Wonderful!

Tests configurations are defined in `config/test-1.yaml`, `config/test-2.yaml`... and they inherit `config/test.yaml` and `config/default.yaml`. So just `createdb peertube_test{1,2,3,4,5,6}` and this should work.

Yes, the user attributes are replicated in different places:
 * In the database model
 * In the shared interface
 * In the client that implements the interface to represent a classic user
 * In the client that implements the interface to represent a authenticated user

I think I could try to refractor this, but there will still be at least 3 files (database model, interface and client model that implement some methods only used by the client).

Chocobozzz:
There is just a small issue with the lint (https://travis-ci.org/Chocobozzz/PeerTube/jobs/318334735). Please fix it and we will able to merge :)

dedesite:
Hello, I fixed the tslint issues but still struggle to launch the tests. I created the databases but the error is about user used to connect to the db. It takes my unix name and I use `postgres` user.
I'm not very familiar with PostgreSql though...

Thank you for your reply about the user attribute, I have some ideas about that but that'll come in another time, for now let's fix some issues :).

Chocobozzz:
Maybe the travis configuration could help you: https://github.com/Chocobozzz/PeerTube/blob/develop/.travis.yml#L31

Don't hesitate to come on IRC (#peertube on freenode)

PR #175, opened 2018-01-06
change nginx config to fix deprecation of a old module

ReK2Fernandez:
just a small change, so it does not byte people when they actually stop showing a warning and instead just fails


Chocobozzz:
Thanks!

Issue #194, opened 2018-01-12
Show a message on client when JavaScript is disabled


Chocobozzz:
We use Angular, so the front end is in full JavaScript. We could show a message to users that disabled JavaScript explaining what are the alternatives (other front-end, desktop applications...).

Chocobozzz:
Fixed: https://github.com/Chocobozzz/PeerTube/commit/f41e76cd4a627fc3f25a995a9732c359639a73b2

Thanks @rigelk

PR #200, opened 2018-01-15
Show a message on client when JavaScript is disabled

rigelk:
Following with #194, that's a (for now text-only) message to explain:
- the need for JS
- the alternatives existing (none for now)
- the extent to which our code can be trusted

Maybe we should stick with something simpler?

### DIFF
--- a/client/src/index.html
+++ b/client/src/index.html
@@ -20,6 +20,15 @@
   <!-- 3. Display the application -->
   <body>
 
+    <noscript>
+      <p>You are blocking Javascript, and we totally get that. However this endpoint uses Angular, so the front end is in full JavaScript and won't work without it.
+      </br></br>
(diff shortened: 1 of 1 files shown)


Chocobozzz:
I think this is fine!

Thanks @rigelk!

Zig-03:
Maybe a noobish remark, but FOSS videos are possible to watch even with all the JS disabled ([example from LibrePlanet videos](https://media.libreplanet.org/)), so I believe this would be the end-goal for this issue? If yes, that'd be amazing. No idea though if it'd work for PeerTube, due to it's complex architecture.

If that'd be possible it would probably take some time until it could be done. So in the meantime, maybe [this](https://www.gnu.org/software/librejs/) could be the solution? If it's easy to implement ([the instructions are here](https://www.gnu.org/software/librejs/free-your-javascript.html)) then it would sure be welcoming more FOSS enthusiasts, if it'd take too much time to implement - then perhaps it might not be such a great idea.

PR #204, opened 2018-01-16
Add Janitor configuration file for PeerTube.

jankeromnes:
This `janitor.json` file should allow us to add PeerTube to https://janitor.technology.

### DIFF
--- a/support/docker/dev/janitor.json
+++ b/support/docker/dev/janitor.json
@@ -1,0 +1,34 @@
+{
+  "name": "PeerTube",
+  "description": "Federated (ActivityPub) video streaming platform using P2P (BitTorrent) directly in the web browser with WebTorrent and Angular.",
+  "icon": "https://janitor.technology/img/peertube.svg",
+  "docker": {
+    "image": "chocobozzz/peertube-dev"
+  },
+  "ports": {
+    "22": {
+      "label": "SSH",
+      "proxy": "none"
+    },
+    "8088": {
+      "label": "VNC",
+      "proxy": "https"
+    },
+    "8089": {
+      "label": "Cloud9",
+      "proxy": "https"
+    },
+    "9000": {
+      "label": "PeerTube",
+      "proxy": "https",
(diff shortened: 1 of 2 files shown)


support/docker/dev/janitor.json
```diff
@@ -0,0 +1,34 @@
...
+      "label": "VNC",
+      "proxy": "https"
+    },
+    "8089": {
+      "label": "Cloud9",
+      "proxy": "https"
+    },
+    "9000": {
+      "label": "PeerTube",
+      "proxy": "https",
```
Chocobozzz: PeerTube will be behind https? Because we would need to update the configuration: https://github.com/Chocobozzz/PeerTube/blob/develop/config/default.yaml#L5

support/docker/dev/janitor.json
```diff
@@ -0,0 +1,34 @@
...
+      "preview": true
+    }
+  },
+  "scripts": {
+    "Start PeerTube": "npm start",
```
Chocobozzz: If https -> `NODE_CONFIG='{ "webserver": { "https": true } }' npm start` should work

jankeromnes:
@Chocobozzz please have a look, especially at the exposed `ports` and the `scripts` section (which will populate Janitor's `Run` button with useful one-click workflows for PeerTube containers).

support/docker/dev/janitor.json
jankeromnes: `"proxy": "https"` actually means that our HTTPS web proxy will first authenticate users (and verify that they are allowed to access the requested container) and then connect to PeerTube's HTTP port in the container to serve content.

support/docker/dev/janitor.json
jankeromnes: Thanks, but it's better if PeerTube starts in HTTP. The port won't be accessible publicly, to protect developer containers from unwanted visitors, and requests will only be routed there by our authenticated proxy.

PR #207, opened 2018-01-16
Fix Janitor configuration;

bnjbvr:
Hey @jankeromnes, can you help me please? It's almost working, but I couldn't find out how to start the postgresql daemon from supervisord. Any idea here?

edit: starting from supervisord as I did doesn't work, for some reason (?).

### DIFF
--- a/scripts/watch/client.sh
+++ b/scripts/watch/client.sh
@@ -2,4 +2,4 @@
 
 cd client || exit -1
 
-npm run ng -- server --hmr --host localhost --port 3000
+npm run ng -- server --hmr --host 0.0.0.0 --port 3000
--- a/support/docker/dev/Dockerfile
+++ b/support/docker/dev/Dockerfile
@@ -15,10 +15,19 @@ WORKDIR /home/user/PeerTube
 # Configure Cloud9 IDE to use PeerTube's source directory as workspace (-w).
 RUN sudo sed -i "s/-w \/home\/user/-w \/home\/user\/PeerTube/" /etc/supervisord.conf
 
+# Install dependencies.
+RUN yarn install --pure-lockfile
+
 # Configure Janitor for PeerTube.
 ADD janitor.json /home/user/
 RUN sudo chown user:user /home/user/janitor.json
 
 # Configure and build PeerTube.
-RUN yarn install \
- && npm run build
+ADD create_user.sql /tmp/
+RUN sudo service postgresql start && \
+    sudo -u postgres psql --file=/tmp/create_user.sql
+
+ADD supervisord.conf /tmp/supervisord-extra.conf
+RUN cat /tmp/supervisord-extra.conf | sudo tee -a /etc/supervisord.conf
+
+EXPOSE 3000 9000
--- a/support/docker/dev/janitor.json
+++ b/support/docker/dev/janitor.json
@@ -10,6 +10,11 @@
       "label": "SSH",
       "proxy": "none"
     },
+    "3000": {
+      "label": "PeerTube web app",
+      "proxy": "https",
+      "preview": true
+    },
     "8088": {
       "label": "VNC",
       "proxy": "https"
@@ -19,13 +24,12 @@
       "proxy": "https"
     },
     "9000": {
-      "label": "PeerTube",
-      "proxy": "https",
-      "preview": true
+      "label": "PeerTube API",
+      "proxy": "https"
     }
   },
   "scripts": {
-    "Start PeerTube": "npm start",
+    "Start PeerTube": "npm run dev",
     "Build PeerTube": "npm run build",
     "Run tests": "npm test",
     "Update source code": "git pull --rebase origin",
--- a/support/docker/dev/supervisord.conf
+++ b/support/docker/dev/supervisord.conf
@@ -1,0 +1,3 @@
+[program:postgresql]
+user = user
+command = sudo -u postgres /usr/lib/postgresql/9.5/bin/postgres -D /var/lib/postgresql/9.5/main -c config_file=/etc/postgresql/9.5/main/postgresql.conf
--- a/support/docker/dev/create_user.sql
+++ b/support/docker/dev/create_user.sql
@@ -1,0 +1,3 @@
+create database peertube_dev;
+create user peertube password 'peertube';
+grant all privileges on database peertube_dev to peertube;


jankeromnes:
Wow, super cool! Thanks for working on this fix. 😄 

I also don't see anything wrong with your approach. Have you looked at `sudo docker logs <your-test-container>`? This will display the supervisor logs with any errors postresql might throw.

bnjbvr:
Actually, scratch my issue: by cargo culting https://github.com/allisson/docker-postgresql/blob/master/supervisord.conf I've found a way that works.

bnjbvr:
This is ready for review:

- @jankeromnes can you have a look at the Janitor changes, please?
- @Chocobozzz too if you want; there's also a change to the watch:client command so that the front web server is exposed to all interfaces (viz. to the docker sub network). I think it's pretty safe.

Thanks!

support/docker/dev/Dockerfile
```diff
@@ -15,10 +15,22 @@ WORKDIR /home/user/PeerTube
...
+# Install dependencies.
+RUN yarn install
+
 # Configure Janitor for PeerTube.
 ADD janitor.json /home/user/
 RUN sudo chown user:user /home/user/janitor.json
 
 # Configure and build PeerTube.
-RUN yarn install \
- && npm run build
```
jankeromnes: Nit: Why do you no longer pre-build PeerTube?

support/docker/dev/Dockerfile
```diff
@@ -15,10 +15,22 @@ WORKDIR /home/user/PeerTube
...
+ADD create_user.sql /tmp/
+RUN sudo service postgresql start && \
+    sudo -u postgres psql --file=/tmp/create_user.sql && \
+    sudo rm /tmp/create_user.sql
```
jankeromnes: Nit: Even if you delete files here, they'll remain part of the image history, since they were `ADD`ed above. I suggest leaving this file somewhere useful, or running these steps with inline psql commands as opposed to a file.

Also, if this is really necessary to get PeerTube to work, why isn't it in a `Makefile` or in the `package.json` NPM scripts?

support/docker/dev/janitor.json
```diff
@@ -19,13 +19,17 @@
       "proxy": "https"
     },
     "9000": {
-      "label": "PeerTube",
+      "label": "PeerTube API",
+      "proxy": "https"
+    },
+    "3000": {
```
jankeromnes: Super-nit: Please sort ports from smallest to largest (i.e. `"3000"` before `"9000"`).

support/docker/dev/janitor.json
```diff
@@ -19,13 +19,17 @@
...
+      "label": "PeerTube front",
```
jankeromnes: Nit: What about "web app" instead of "front"? I'd find that more explicit/transparent.

support/docker/dev/janitor.json
```diff
@@ -19,13 +19,17 @@
...
       "proxy": "https",
       "preview": true
     }
   },
   "scripts": {
-    "Start PeerTube": "npm start",
+    "Start PeerTube": "npm run dev",
```
jankeromnes: Interesting, why did you need to switch this? What's the difference between `npm start` and `npm run dev`?

support/docker/dev/supervisord.conf
```diff
@@ -0,0 +1,3 @@
+[program:postgresql]
+user = user
+command = sudo -u postgres /usr/lib/postgresql/9.5/bin/postgres -D /var/lib/postgresql/9.5/main -c config_file=/etc/postgresql/9.5/main/postgresql.conf
```
jankeromnes: Nit: This command is very long. Why not make it shorter by using `user = postgres` like in the supervisor config example you found?

support/docker/dev/Dockerfile
```diff
@@ -15,10 +15,22 @@ WORKDIR /home/user/PeerTube
...
+
+ADD supervisord.conf /tmp/supervisord-extra.conf
+RUN cat /tmp/supervisord-extra.conf | sudo tee -a /etc/supervisord.conf && \
+    sudo rm /tmp/supervisord-extra.conf
+
+EXPOSE 3000
+EXPOSE 9000
```
jankeromnes: Nit: You can expose multiple ports in one instruction [like so](https://github.com/JanitorTechnology/dockerfiles/blob/6d3e7835c90fafe76a2e8463b4d5104ac0a39a9d/ubuntu-dev/ubuntu-dev.dockerfile#L224).

scripts/watch/client.sh
```diff
@@ -2,4 +2,4 @@
 
 cd client || exit -1
 
-npm run ng -- server --hmr --host localhost --port 3000
+npm run ng -- server --hmr --host 0.0.0.0 --port 3000
```
jankeromnes: This looks a bit scary, because you're exposing something to the world that was limited to `localhost` before. Could you please explain why this is needed? Is this a development server?

support/docker/dev/Dockerfile
Chocobozzz: `npm run dev` will build the client on the fly

support/docker/dev/janitor.json
Chocobozzz: `npm start` just serves start the built server and serves compiled client files.

`npm run dev` builds the server, run it on port 9000, builds the client on the fly and serves it on port 3000

support/docker/dev/Dockerfile
```diff
@@ -15,10 +15,22 @@ WORKDIR /home/user/PeerTube
 # Configure Cloud9 IDE to use PeerTube's source directory as workspace (-w).
 RUN sudo sed -i "s/-w \/home\/user/-w \/home\/user\/PeerTube/" /etc/supervisord.conf
 
+# Install dependencies.
+RUN yarn install
```
Chocobozzz: `RUN yarn install --pure-lockfile`

scripts/watch/client.sh
bnjbvr: It's not scary, it's only in development mode: note that this just means exposing the dev web server on the Docker network interface, which is needed if we want to expose the service outside of the container. Keeping it this way (the risk seems nonexistent to me).

support/docker/dev/Dockerfile
bnjbvr: What @Chocobozzz said. Closing.

support/docker/dev/Dockerfile
bnjbvr: Re: files: well, these are tiny, so I don't seem to care too much ;) I'll just let them in /tmp in this case.

Re: Makefile/package.json docs: it depends a lot on your database install / version; it's more of an external dependency that something mandated by PeerTube, so IMO it makes sense to keep it separate.

support/docker/dev/janitor.json
bnjbvr: If I wanted to be picky, I'd say the API also uses web technologies and is an app, so it's also a web app ;) But yeah, it probably makes more sense overall :)

support/docker/dev/supervisord.conf
bnjbvr: I tried it the way you suggest, it didn't work, I don't have much time digging up why it didn't work, so meh 🤷‍. Happy to review a follow-up change to this if you're in the mood of making your PR thereafter.

bnjbvr:
Thanks for the reviews, @Chocobozzz @jankeromnes ! Mind to have another look, please? (this works on my machine; since the basic configuration works, and we still have no guarantees that absolute links will work in Janitor anyways, I'd like to **not** invest too much time on this)

Chocobozzz:
Seems good for me. Waiting @jankeromnes validation.

Even if it does not work yet, we can merge and fix it in another PR :)

PR #213, opened 2018-01-17
Add config for running with Docker Compose and on Docker Swarm

djmaze:
This allows running a local or production server, either with Docker Compose or on a Docker Swarm cluster. I think this makes deployment (and also development) much easier, so I hope this is deemed valuable.

I have the swarm stack successfully working on a cluster (currently without federation though). 

I made some of the main settings configurable via environment variables on the way. This can of course been expanded upon.

nginx is notably missing from this setup and I am not sure if the default settings of the traefik loadbalancer are enough for this to work reliably as a webseed / webtorrent tracker. Any advice welcome. (Also, practice will show.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,5 +19,5 @@ ENV NODE_ENV production
 ENV NODE_CONFIG_DIR ./docker
 
 CMD ["npm", "start"]
-VOLUME ["/data"]
+VOLUME ["/usr/src/app/data"]
 EXPOSE 9000
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM node:8
+
+# install yarn
+RUN apt-get update && apt-get install -y apt-transport-https \
+ && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+ && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
+ && apt-get update && apt-get install -y yarn \
+ && rm /var/lib/apt/lists/* -fR
+
+# install ffmpeg
+RUN echo "deb http://ftp.uk.debian.org/debian jessie-backports main" >/etc/apt/sources.list.d/jessie-backports.list \
+ && apt-get update \
+ && apt-get -y install ffmpeg \
+ && rm /var/lib/apt/lists/* -fR
+
+WORKDIR /usr/src/app
+
+COPY package.json yarn.lock ./
+COPY client/package.json client/yarn.lock client/
+RUN yarn install --pure-lockfile
+
+COPY . ./
+
+RUN npm run build
+
+ENV NODE_ENV production
+ENV NODE_CONFIG_DIR ./docker
+
+CMD ["npm", "start"]
+VOLUME ["/data"]
+EXPOSE 9000
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,7 @@
-FROM node:8
-
-# install yarn
-RUN apt-get update && apt-get install -y apt-transport-https \
- && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
- && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
- && apt-get update && apt-get install -y yarn \
- && rm /var/lib/apt/lists/* -fR
+FROM node:8-stretch
 
 # install ffmpeg
-RUN echo "deb http://ftp.uk.debian.org/debian jessie-backports main" >/etc/apt/sources.list.d/jessie-backports.list \
- && apt-get update \
+RUN apt-get update \
  && apt-get -y install ffmpeg \
  && rm /var/lib/apt/lists/* -fR
 
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ BitTorrent) inside the web browser, as of today.
 
 You can quickly get a server running using Docker. You need to have [docker](https://www.docker.com/community-edition) and [docker-compose](https://docs.docker.com/compose/install/) installed.
 
-Also, you need to run a Traefik loadbalancer using Docker Compose. You can start one locally with this [example compose config](https://gist.github.com/djmaze/72f0565715c59712ce191b41d3c377da).
+For this example configuration, you also need to run a Traefik loadbalancer using Docker Compose. You can start one locally with this [example compose config](https://gist.github.com/djmaze/72f0565715c59712ce191b41d3c377da).
 
 Example for running a peertube server locally:
 
--- a/README.md
+++ b/README.md
@@ -136,6 +136,22 @@ BitTorrent) inside the web browser, as of today.
   * OpenSSL (cli)
   * FFmpeg
 
+## Run using Docker
+
+You can quickly get a server running using Docker. You need to have [docker](https://www.docker.com/community-edition) and [docker-compose](https://docs.docker.com/compose/install/) installed.
+
+Example for running a server locally:
+
+```bash
+sudo \
+  PEERTUBE_HOSTNAME=peertube.lvh.me \
+  PEERTUBE_ADMIN_EMAIL=test@example.com \
+  PEERTUBE_TRANSCODING_ENABLED=true \
+  docker-compose up app
+```
+
+(Get the initial root user password from the program output.)
+
 ## Production
 
 See the [production guide](support/doc/production.md).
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -136,6 +136,24 @@ BitTorrent) inside the web browser, as of today.
   * OpenSSL (cli)
   * FFmpeg
 
+## Run using Docker
+
+You can quickly get a server running using Docker. You need to have [docker](https://www.docker.com/community-edition) and [docker-compose](https://docs.docker.com/compose/install/) installed.
+
+Also, you need to run a Traefik loadbalancer using Docker Compose. You can start one locally with this [example compose config](https://gist.github.com/djmaze/72f0565715c59712ce191b41d3c377da).
```
rigelk: I would rephrase so as not let the reader misunderstand Træfik as a hard requirement of PeerTube, but just as a requirement in this example configuration.

djmaze: Right, I rephrased it slightly.

For a local dev environment, we could even go completely without a reverse proxy in front. But I like this approach since it is usable almost out-of-the box for production as well.

Dryusdan:
Hey
I noticed two concerns about your image : You run PeerTube as root. This is not very safe because you can exit the container thank to root and a PeerTube break (not exist actually but possible later)
You have got lot of layer, it's can possible to reduce them and reduce weight :)

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+    networks:
+      - backend
+
+volumes:
+  app_data:
+  db_data:
+
+networks:
+  backend:
+  traefik:
```
bnjbvr: Can you remove the networks section, please? It seems this is useful only for the docker swarm compose file.

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      NODE_CONFIG: '{"webserver": {"https": false, "port": 80}}'
+      PEERTUBE_HOSTNAME:
+      PEERTUBE_ADMIN_EMAIL:
+      PEERTUBE_SIGNUP_ENABLED:
+      PEERTUBE_TRANSCODING_ENABLED:
+    labels:
+      traefik.enable: "true"
+      traefik.frontend.rule: "Host:${PEERTUBE_HOSTNAME}"
+      traefik.docker.network: traefik_traefik
+      traefik.port: "9000"
```
bnjbvr: (see comment below) I think these can be removed too, in the non-swarm compose file.

docker/production.yaml
```diff
@@ -0,0 +1,55 @@
+listen:
+  port: 9000
+
+# Correspond to your reverse proxy "listen" configuration
+webserver:
+  https: true
+  hostname: undefined
```
bnjbvr: Will the yaml parser consider this to be a string with the value "undefined" or the real undefined value? 🤔

README.md
```diff
@@ -136,6 +136,24 @@ BitTorrent) inside the web browser, as of today.
   * OpenSSL (cli)
   * FFmpeg
 
+## Run using Docker
+
+You can quickly get a server running using Docker. You need to have [docker](https://www.docker.com/community-edition) and [docker-compose](https://docs.docker.com/compose/install/) installed.
+
+For this example configuration, you also need to run a Traefik loadbalancer using Docker Compose. You can start one locally with this [example compose config](https://gist.github.com/djmaze/72f0565715c59712ce191b41d3c377da).
```
bnjbvr: Maybe we could remove the need for the load balancer here? Does it make sense that this is something end users would decide to add by themselves?

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# install ffmpeg
+RUN echo "deb http://ftp.uk.debian.org/debian jessie-backports main" >/etc/apt/sources.list.d/jessie-backports.list \
+ && apt-get update \
+ && apt-get -y install ffmpeg \
+ && rm /var/lib/apt/lists/* -fR
+
+WORKDIR /usr/src/app
+
+COPY package.json yarn.lock ./
```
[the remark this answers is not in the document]
Dragnucs: It is preferable to run Yarn after successful Docker deployment. This would keep containers light.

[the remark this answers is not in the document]
kaiyou: Disagreeing with both comments. Downloading a binary release is very diffrent from building the whole thing and has very different properties. For instance, one might wish to build from source and have a repeatable result.

Also, running yarn after deployment keeps the image light but definitely not the container as files are still stored whenever the thing is running. Also, you then lose the interesting stateless property of the image.

I would maybe go for multiple Dockerfile, at least one building from source and one embedding the latest release.

Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM node:8
```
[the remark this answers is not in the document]
kaiyou: Also, configuration files could be centralized in a ``support/docker/config`` folder, in order to be used by multiple Dockerfiles for various flavors

kaiyou:
Currently, the image does not set a specific uid. I know that user namespaces are intended to mitigate these issues in most cases, but currently this could be considered a vulnerability: any capabilities assigned to the container can be used by the node application, which is most certainly unnecessary.

docker/production.yaml
```diff
@@ -0,0 +1,55 @@
...
+# Your database name will be "peertube"+database.suffix
+database:
+  hostname: 'db'
+  port: 5432
+  suffix: '_prod'
+  username: 'postgres'
+  password: 'postgres'
+
+# From the project root directory
+storage:
```
kaiyou: If these are relative path, then the volume at ``/data`` will not contain them. Thus, no data will be persistent.

Dockerfile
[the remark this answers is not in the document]
djmaze: Agreeing with @kaiyou. I was hoping the repository owner might add an [automated build](https://docs.docker.com/docker-hub/builds/) on the Docker hub, thus automatically building a new release for every commit. We could also use Git tags in order to let the Docker hub build and tag images for every release. So it makes sense to build from source.

Concerning the yarn comment, it is desirable to have an image which can be deployed seamlessly on a Docker host / a Docker swarm cluster. In order for that to work, we need a ready-made, complete image. As mentioned by @kaiyou, in practice this does not save space at all.

That said, we could have a different Dockerfile for development and production. In my opinion, this doesn't help much though. In either case, you only need to rebuild the image when dependencies change. (The current code from your working directory can be bind-mounted in development.)

kaiyou:
I am preparing an Alpine-based image, based both on https://github.com/TeDomum/docker-peertube and this PR. I was thinking about the directory structure and ways to simplify things.

We could maybe have ``support/docker`` contain a ``config`` directory with the ``custom-environment-variables.yaml`` and other per-environment configuration files like ``production.yaml``, then have a Dockerfile per flavor: ``Dockerfile.stetch``, ``Dockerfile.alpine``, ``Dockerfile.dev``, etc.

Dockerfile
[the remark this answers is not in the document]
djmaze: Good catch! Adjusted.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# install yarn
```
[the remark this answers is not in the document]
djmaze: Well spotted.. Removed it.

docker/production.yaml
djmaze: The latter I think. Got that from the [node-config documentation](https://github.com/lorenwest/node-config/wiki/Common-Usage#using-config-values), although I may have misunderstood it.

kaiyou: It will set it as the undefined JavaScript object and fail configuration validation if the value is not overloaded by an environment variable.

djmaze: :+1: That's what I wanted.

Dockerfile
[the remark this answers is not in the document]
kaiyou: Afaik, Docker automated builds do not support setting the Dockerfile path, but the build root path instead, which is not compatible with the current structure. Still, an automated build could be delegated to Travis or any CI platform :)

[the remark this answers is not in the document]
bnjbvr: Yes it does now, actually (if we're talking about setting the Dockerfile path in hub.docker.com). I know since I've set it up, for one project of mine ;)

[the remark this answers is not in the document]
kaiyou: I think that setting the "Dockerfile location" in a Docker hub build will set the Docker build context, not the Dockerfile path, which prevents from adding files in parent directories and will make ``COPY . ./`` invalid.

Chocobozzz:
Merging this PR in https://github.com/Chocobozzz/PeerTube/pull/225

@djmaze could you review the other PR? :)

Chocobozzz:
@djmaze I [removed your documentation](https://github.com/Chocobozzz/PeerTube/commit/4df6a1b8e1ec98355667cb872511d27035bfaf5a#diff-c2b523528190f2f5331286bb55e7ed8bL319) about docker swarm because the `docker-stack.example.yml` was missing.

Feel free to open another PR :)

PR #220, opened 2018-01-19
Basic api documentation #7

DimitriGilbert:
basic api documentation with info I could gather from the code, It can go way further, but i'd have to improve my knowledge of the codebase.

Also, even if convenient, ApiDoc seems to be a bit limited compared to OpenApi for example.
I noticed there where existing tool to transform ApiDoc to OpenApi, it might be a good way to have both and see which one is most useful.


Chocobozzz:
@DimitriGilbert I agree, OpenAPI seems to be better. Could you update your PR? If not, I can do it.

Anyway thanks a lot for your work! :)

DimitriGilbert:
here you go, api documentation in openApi 2.0 . I didn't use the 3.0 version as most tool are not ready for it and there is nice tools to convert from 2.0 to 3.0 anyway :)

As a side note, you should review the generated doc as I'm not entirely sure of the responses ( especially concerning comments). I use swagger editor for that : http://editor.swagger.io
Also I didn't put most of the bodies in post request as I was not sure what was expected (first time working on a express project so it'll get better with time ^^ )

rigelk:
LGTM − but maybe you could squash your commits prior to the final merge?

DimitriGilbert:
a hard refresh of the editor did spew that to (guess it's a bit buggy if you leave it open for too long...), it's corrected now.
I did give a look at squashing things, but i'm not that comfortable with it as i never really used it, my git usage was always to get the full history unmodified as much as possible but I guess it makes more sense on big team in an open source software. If you happen to have a good link on how to use that i'd be interested, otherwise I'll add it to my "must research" list :)

Chocobozzz:
@DimitriGilbert Don't worry I'll do it myself.

FYI https://git-scm.com/book/en/v2 is always a good resource when you're looking for git support :)

Chocobozzz:
This is an amazing work, thanks @DimitriGilbert!

Issue #196, opened 2018-01-12
Allow to cancel an upload


tcitworld:
Aside the progress bar should be a cancel button

Chocobozzz:
Implemented by @DimitriGilbert https://github.com/Chocobozzz/PeerTube/commit/8c4890cbfe3b5c742ed8756d7a0fd2f533534c6a

PR #221, opened 2018-01-19
Issue #196 : Allow to cancel an upload

DimitriGilbert:
The button isn't fancy but i did not find directive concerning visual elements. Tell me what to change it for, i'll be happy to do so.


client/src/app/videos/+video-edit/video-add.component.ts
```diff
@@ -27,6 +27,7 @@ export class VideoAddComponent extends FormReactive implements OnInit {
 
   isUploadingVideo = false
   videoUploaded = false
+  videoUpload = null
```
bnjbvr: From a quick skim, isn't `videoUpload` redundant with `isUploadingVideo`, and thus we could remove the latter by testing `!!videoUpload`

DimitriGilbert: As I do not know the code base very well yet, I didn't want to introduce unforeseen side-effect. Also, isVideoUploading is a boolean flag while videoUpload (very bad name but I was out of idea) is the observable returned by the subscription to the HttpClient request and I was not feeling comfortable testing on it (the upload is finished but for some reason, something did not go entirely as planned for example).
I'm not against it though. 
What do maintainers thinks about that ?

Chocobozzz: Just rename it to `videoUploadObservable` and it should be okay :)

client/src/app/videos/+video-edit/video-add.component.ts
```diff
@@ -93,6 +94,16 @@ export class VideoAddComponent extends FormReactive implements OnInit {
     return this.form.valid
   }
 
+  uploadCancel () {
```
Chocobozzz: `cancelUpload`?

DimitriGilbert:
I did do the requested change, I also added a few paths to the gitignore, they seem to be created on the fly and contain user related content. Would you like me to open another pull request for that  and delete the changes on this one ?

Chocobozzz:
@DimitriGilbert No it's fine. The directories are automatically created according to your configuration: https://github.com/Chocobozzz/PeerTube/blob/develop/config/default.yaml#L19 

Thanks for your work! :)

DimitriGilbert:
You're welcome :)

PR #235, opened 2018-01-23
cleanup travis by using nix

stephengroat:
just an idea, no worries if it's not what you're looking for!


Chocobozzz:
I fact, we need a specific ffmpeg version for the tests (because we compare byte per byte thumbnails generated by the server). But thanks anyway 👍

PR #267, opened 2018-01-31
update yarn installation command

valvin1:
on ubuntu `sudo -u` doesn't set homedirectory of the targeted user althought debian does. it requires the option `-H`. I think with this option it works in both case.


Chocobozzz:
Thanks!

PR #273, opened 2018-02-05
Include Web Manifest

lucaspontoexe:
Now we need to fix the icons path, which shouldn't be hard.


rigelk:
why did you add a "GooglePlayStore" icon?

lucaspontoexe:
It is not necessary. I just forgot to delete.

Chocobozzz:
I think `client/src/assets/images/icons/Thumbs.db` should be deleted too?

I'm sorry but I'm not familiar with manifests, what do you mean with "Now we need to fix the icons path, which shouldn't be hard."?

lucaspontoexe:
> "Now we need to fix the icons path, which shouldn't be hard."

At the time I was writing that, I was kind of guessing where the images were stored.

PR #277, opened 2018-02-06
Expose SMTP and instance settings in the Docker environment variables

kaiyou:
This exposes missing settings inside the Docker image and through environment variables.


Chocobozzz:
Thanks! Forgot to update these files sorry.

PR #281, opened 2018-02-07
Upgrade.sh

svnet-libre:
Bash script for upgrading Instance


rigelk:
Hi @svnet-libre - LGTM, although you might want to add at least [bash's strict mode](https://dev.to/thiht/shell-scripts-matter): `set -e` at the beginning.

svnet-libre:
Thanks @rigelk 20233874dffaa3c904991936fd68727126d99964

Chocobozzz:
Please put this script in `scripts/` folder and update the `production.md` documentation so we don't have to maintain two different shell commands.

PR #287, opened 2018-02-11
Precisions and security enhancements to the production guide

rigelk:
- added precisions and suggestions about how to generate Let's Encrypt certificates. Users have reported their installations didn't work when the problem came from missing certificates (false positives).
- security defaults of Nginx follow the basic robustness principle "be conservative in what you send, be liberal in what you accept", which isn't enough with modern security standards, so we should be picky with the cipher suites we use, among other things. Extra comments (especially for the TLS1.3 protocol support parameter) make the requirement of a recent Nginx installation obvious, and the downgrade alternative remains clear to the system administrator.

All in all, we should aknowledge users will most often copy and paste the configuration files. Making them secure by default may force a few users to read their configuration, but on the long run we are making the fediverse more secure.

Since I've come to modify a bit the Nginx config in `support/doc/production.md`, I've merged it with the template so that they stay consistent.

### DIFF
--- a/support/doc/production.md
+++ b/support/doc/production.md
@@ -70,6 +70,8 @@ configuration.
 
 ### Webserver
 
+We only provide official configuration files for Nginx.
+
 Copy the nginx configuration template:
 
 ```
@@ -83,9 +85,7 @@ It should correspond to the paths of your storage directories (set in the config
 $ sudo vim /etc/nginx/sites-available/peertube
 ```
 
-If you want to set https with Let's Encrypt please follow the steps of [this guide](https://www.digitalocean.com/community/tutorials/how-to-secure-nginx-with-let-s-encrypt-on-ubuntu-16-04).
-
-An example of the nginx configuration could be:
+Your Mileage May Vary, but what follows is an example of configuration for nginx with a certificate made via `certbot` ([other utilities exist](https://letsencrypt.org/docs/client-options/)):
 
 ```
 server {
@@ -104,11 +104,28 @@ server {
   listen [::]:443 ssl http2;
   server_name peertube.example.com;
 
-  # For example with Let's Encrypt
+  # For example with Let's Encrypt (you need a certificate to run https)
   ssl_certificate      /etc/letsencrypt/live/peertube.example.com/fullchain.pem;
   ssl_certificate_key  /etc/letsencrypt/live/peertube.example.com/privkey.pem;
-  ssl_trusted_certificate /etc/letsencrypt/live/peertube.example.com/chain.pem;
-
+  
+  # Security hardening (as of 11/02/2018)
+  ssl_protocols TLSv1.3, TLSv1.2;# TLSv1.3 requires nginx >= 1.13.0 else use only TLSv1.2
+  ssl_ciphers 'ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256';
+  ssl_prefer_server_ciphers on;
+  ssl_ecdh_curve secp384r1; # Requires nginx >= 1.1.0
+  ssl_session_timeout  10m;
+  ssl_session_cache shared:SSL:10m;
+  ssl_session_tickets off; # Requires nginx >= 1.5.9
+  ssl_stapling on; # Requires nginx >= 1.3.7
+  ssl_stapling_verify on; # Requires nginx => 1.3.7
+  resolver $DNS-IP-1 $DNS-IP-2 valid=300s;
+  resolver_timeout 5s;
+  add_header Strict-Transport-Security "max-age=63072000; includeSubDomains; preload";
+  add_header X-Frame-Options DENY;
+  add_header X-Content-Type-Options nosniff;
+  add_header X-XSS-Protection "1; mode=block";
+  add_header X-Robots-Tag none;
+  
   access_log /var/log/nginx/peertube.example.com.access.log;
   error_log /var/log/nginx/peertube.example.com.error.log;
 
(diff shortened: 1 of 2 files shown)


support/doc/production.md
```diff
@@ -104,11 +102,29 @@ server {
...
-  # For example with Let's Encrypt
+  # For example with Let's Encrypt (you need a certificate to run https)
   ssl_certificate      /etc/letsencrypt/live/peertube.example.com/fullchain.pem;
   ssl_certificate_key  /etc/letsencrypt/live/peertube.example.com/privkey.pem;
-  ssl_trusted_certificate /etc/letsencrypt/live/peertube.example.com/chain.pem;
-
+  ssl_dhparam /etc/nginx/dhparam.pem; # openssl dhparam -out /etc/nginx/dhparam.pem 4096
+  
+  # Security hardening (as of 11/02/2018)
+  ssl_protocols TLSv1.3;# Requires nginx >= 1.13.0 else use TLSv1.2
```
Chocobozzz: I don't think it's a good idea: https://caniuse.com/#feat=tls1-3

support/doc/production.md
```diff
@@ -182,8 +198,19 @@ server {
...
+```
+
+```
+$ sudo certbot --authenticator standalone --installer nginx --pre-hook "systemctl stop nginx" --post-hook "systemctl start nginx"
+```
+
+Remember your certificate will expire in 90 days, and thus needs renewal.
 
-Activate the configuration file:
+Activate the configuration file :
```
Chocobozzz: Please remove the space

support/doc/production.md
```diff
@@ -182,8 +198,19 @@ server {
...
 ```
 
+To generate the certificate for your domain as required to make https work, you have two alternatives:
+
+```
+$ sudo certbot --authenticator standalone certonly -d peertube.example.com && openssl dhparam -out /etc/nginx/dhparam.pem 4096
+```
+
+```
+$ sudo certbot --authenticator standalone --installer nginx --pre-hook "systemctl stop nginx" --post-hook "systemctl start nginx"
```
Chocobozzz: `post-hook "systemctl reload nginx"` should be enough

tcitworld: do `nginx -t && systemctl reload nginx` so that it won't reload the server config if it failed to apply it.

support/doc/production.md
tcitworld: `TLSv1.3, TLSv1.2` would be fine.

support/doc/production.md
rigelk: A mishap :)

support/doc/production.md
```diff
@@ -70,6 +70,8 @@ configuration.
 
 ### Webserver
 
+We maintain a few example configurations in `support/doc/`, for different web servers. Here we depict the installaton procedure with the one of Nginx.
```
Chocobozzz: We only support nginx (and we don't want to support other webservers) so I don't think this is useful :)

support/nginx/peertube
```diff
@@ -1,27 +1,49 @@
...
-  ssl_certificate      /etc/letsencrypt/live/domain.tld/fullchain.pem;
-  ssl_certificate_key  /etc/letsencrypt/live/domain.tld/privkey.pem;
-  ssl_trusted_certificate /etc/letsencrypt/live/domain.tld/chain.pem;
+  listen [::]:443 ssl http2;
+  server_name peertube.example.com;
+
+  # For example with Let's Encrypt (you need a certificate to run https)
+  ssl_certificate      /etc/letsencrypt/live/peertube.example.com/fullchain.pem;
+  ssl_certificate_key  /etc/letsencrypt/live/peertube.example.com/privkey.pem;
+  ssl_dhparam /etc/nginx/dhparam.pem; # openssl dhparam -out /etc/nginx/dhparam.pem 4096
```
Chocobozzz: I think we'll need to put `openssl dhparam -out /etc/nginx/dhparam.pem 4096` in the installation guide

support/nginx/peertube
```diff
@@ -1,27 +1,49 @@
...
-  server_name domain.tld;
+  listen [::]:80;
+  server_name peertube.example.com;
 
-  access_log /var/log/nginx/peertube_access.log;
-  error_log /var/log/nginx/peertube_error.log;
+  access_log /var/log/nginx/peertube.example.com.access.log;
+  error_log /var/log/nginx/peertube.example.com.error.log;
 
-  location /.well-known/acme-challenge/ { allow all; }
```
Chocobozzz: Don't remove it because you'll not be able to create a certificate with certbot

support/nginx/peertube
tcitworld: I think DH for key exchange is already depreciated (only ECDH is the correct way nowdays), so no need for dhparams anyway.

support/doc/production.md
```diff
@@ -104,11 +104,29 @@ server {
...
   ssl_certificate      /etc/letsencrypt/live/peertube.example.com/fullchain.pem;
   ssl_certificate_key  /etc/letsencrypt/live/peertube.example.com/privkey.pem;
-  ssl_trusted_certificate /etc/letsencrypt/live/peertube.example.com/chain.pem;
-
+  ssl_dhparam /etc/nginx/dhparam.pem; # openssl dhparam -out /etc/nginx/dhparam.pem 4096
+  
+  # Security hardening (as of 11/02/2018)
+  ssl_protocols TLSv1.3, TLSv1.2;# TLSv1.3 requires nginx >= 1.13.0 else use only TLSv1.2
+  ssl_prefer_server_ciphers on;
+  ssl_ciphers ECDHE-RSA-AES256-GCM-SHA512:DHE-RSA-AES256-GCM-SHA512:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384:!3DES;
```
tcitworld: Remove DHE since it's much slower than ECDHE and all modern clients use it.
Checkout https://mozilla.github.io/server-side-tls/ssl-config-generator/

support/doc/production.md
```diff
@@ -136,7 +154,7 @@ server {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 
     # For the video upload
-    client_max_body_size 8G;
+    client_max_body_size 2G;
```
tcitworld: Why ?

support/doc/production.md
rigelk: Didn't know that :)

support/doc/production.md
rigelk: Because it was in `support/nginx/peertube`, which I consider canonical.

support/doc/production.md
rigelk: Indeed! I didn't think about removing it, which makes sense for elliptic curve computations.

Chocobozzz:
Thanks @rigelk!

Issue #168, opened 2018-01-05
Link to a specific comment


tcitworld:
The « x min ago » text should link to the comment.

Aluriak:
I've always found that behavior weird and unexpected.

Why not a 🔗 icon appearing when hovering on the comment ?
Would be much more explicit and easier to understand.

I would expect clicking/hovering the «x min ago» giving me the exact date instead. Because the «x min ago» soon became «x days ago», then «x years ago». Not really helpful when trying to get a precise date of a conversation.
(unless the time is implemented using something like [hyperloglog](https://en.wikipedia.org/wiki/HyperLogLog) ?)

Chocobozzz:
Implemented by @jonathanraes: https://github.com/Chocobozzz/PeerTube/commit/d5b53822ae7e1660cebe3a35be5ce76ea73dc1b9

PR #297, opened 2018-02-16
 Issue #168: youtube-like marking of comments

jonathanraes:
Uses GET parameters to mark comments similar to youtube, allowing to send a comment link to someone else.
I agree with Aluriak that the « x min ago » text is not a logical place to provide the link, even though this is the case in youtube. (#168)
Right now I included a link next to the `Delete` and `Reply` links, but i'm open to changing that.

### DIFF
--- a/client/src/app/videos/+video-watch/comment/video-comment.component.html
+++ b/client/src/app/videos/+video-watch/comment/video-comment.component.html
@@ -2,6 +2,7 @@
   <img [src]="getAvatarUrl(comment.account)" alt="Avatar" />
 
   <div class="comment">
+    <span class="marked-comment" *ngIf="comment.marked">Marked comment</span>
     <div class="comment-account-date">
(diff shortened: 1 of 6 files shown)


Chocobozzz:
Awaiting a better solution I prefer to put the link in the " x min ago " text :)

tcitworld:
Using a `GET` parameter isn't really sexy. Couln't we have a dedicated route ?

Chocobozzz:
An anchor tag would be useful, but we can't for now: https://github.com/angular/angular/issues/6595

Maybe we could just use [params](https://angular.io/api/router/ActivatedRouteSnapshot#params) for now (like in the videos list).

jonathanraes:
I'll look into using a route and params

jonathanraes:
How about this?

Chocobozzz:
Nice! Thanks @jonathanraes!

PR #304, opened 2018-02-20
Create CODE_OF_CONDUCT.md

rigelk:
We want to be inclusive to the largest number of participants, with the most varied and diverse backgrounds possible. As such, we should take position as to how we will defend our contributors. Better settle on how we handle dire situations before they happen.

I saw some projects of Framasoft already used the Contributor Covenant CoC, and it's apparently used by a large number of projects, so I figured out we could use something of the like. Again, it's not just a text. We should be ready to enforce it too, so better agree on it :-)

Nota: the code of conduct isn't clear on the disclosure of an incident per se. That should be investigated further.
Nota bis: maybe we should leave the PR open some time and call for discussion over it?

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,85 @@
+# Contributor Covenant Code of Conduct
+
+## Our pledge
+
+In order to create an open and welcoming environment, we as contributors and maintainers pledge to make participation in our project and our community a harassment-free experience for everyone, regardless of:
+
+* age
+* body size
+* disability
+* ethnicity
+* gender identity and expression
+* level of experience
+* education
+* socio-economic status
+* nationality
+* personal appearance
+* ethnicity
+* religion
+* sexual identity and orientation
+
+## Our standards
+
+A supplemental goal of this Code of Conduct is to increase open source and culture citizenship by encouraging participants to recognize and strengthen the relationships between our actions and their effects on our community.
+
+Communities mirror the societies in which they exist and positive action is essential to counteract the many forms of inequality and abuses of power that exist in society.
+
+If you see someone who is making an extra effort to ensure our community is welcoming, friendly, and encourages all participants to contribute to the fullest extent, please recognize their efforts.
+
+You can contribute to creating a positive environment in many ways. For example you can:
+
+* use welcoming and inclusive language
+* be respectful of differing viewpoints and experiences
+* accept constructive criticism gracefully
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,85 @@
...
+* gender identity and expression
+* level of experience
+* education
+* socio-economic status
+* nationality
+* personal appearance
+* ethnicity
+* religion
+* sexual identity and orientation
+
```
framabot: You should add `etc`, because right now, you can harasse someone on, for ex., its editor (Vi vs Emacs flamewar).

ldidry: You should add `etc`, because right now, you can harasse someone on, for ex., its editor (Vi vs Emacs flamewar).

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,85 @@
...
+You should not:
+
+* use sexualised language or imagery
+* make unwelcome sexual advances
+* troll, and make insulting or derogatory comments
+* make personal or political attacks
+* harass others, in public or private
+* publish others' private information, such as a physical or electronic address, without explicit permission
+* engage in any other conduct which could reasonably be considered inappropriate in a professional setting
+
```
ldidry: Same thing: don't create a finite list of bad behaviors, you can always find a bad behavior that is not in the list and thus say "You can't ban me, I respect your CoC".

So, it's better to add a `etc`.

CODE_OF_CONDUCT.md
rigelk: One could argue this is covered by the `religion` clause ;-)

More seriously, I do think flamewars over tools should be covered. They happen often enough on some forums.

CODE_OF_CONDUCT.md
rigelk: I will definitely add an `etc.`, and I see your point. However, listing a few bad behaviours contextualises concepts, and is useful to explicit things. I don't say we should create a long list, but at least the most significant/recurring cases should be covered.

ldidry: Sure, having a list of examples is good, and covering the most obvious cases is good. And `etc` avoids the long list.

CODE_OF_CONDUCT.md
ldidry: Well, it's covered by the `etc` :wink: .

Issue #299, opened 2018-02-17
Renew token with import-youtube to avoid 401 error


LecygneNoir:
Hello !

* **PeerTube version or commit:** : 0.24
* **Browser name/version:** N/A
* **NodeJS version:** Node js v8.9.4 on Debian 9
* **PostgreSQL version:** N/A

I am using the import-youtube.js script to import the entire content of my Youtube channel

The script works smoothly, but after uploading some videos (depending the time ellapsed), the script gets a 401 unauthorized from Peertube, and crash:

```
(node:18380) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: expected 200 "OK", got 401 "Unauthorized"
```

@Chocobozzz investigation shows that this is a token issue: it expires during the long process of import, so Peertube server reject connection.

I think at least two ways to handle this, do not know which could be best:
 - authenticate for each upload
 - catch the 401 and re-authenticate to retry upload

What do you think is best? (or other ideas)

Thanks!

Chocobozzz:
The second option is the best ;)

LecygneNoir:
Do you know if there is a way to invalidate an existing token? To be able to test the fix easily :-)
Thanks!

Chocobozzz:
You can change the [token lifetime](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/constants.ts#L39) for you tests

LecygneNoir:
Thank you, I will see if I can do something for that, to increase my dev comp on js before trying to contribute to core (I prefer to avoid breaking something important, mmmh :joy: )

PR #306, opened 2018-02-20
 add try/catch to manage token expiration when importing lot of youtube videos

LecygneNoir:
Hello,

Here is a contribution that should fix https://github.com/Chocobozzz/PeerTube/issues/299
It retries login to the peertube server when getting one 401 error at first upload.
I have tested with a token lifetime of 100 secondes, ti works smoothly so it should works with extrem configurations :-)
Renew works for each videos

It could probably be increased if we need to manage more than one retry or add better error handling (disconnection, password changed, quota exceeded?)

I am open to discuss these points with pleasure ;-)

Regards,


Chocobozzz:
Thanks!

PR #308, opened 2018-02-21
Upgrade.sh : Add diff command for config update

svnet-libre:


Chocobozzz:
Please add some context (eg `echo "Differences in configuration files..."`)

scripts/upgrade.sh
```diff
@@ -22,6 +22,8 @@ ln -s /var/www/peertube/versions/peertube-${VERSION} /var/www/peertube/peertube-
 cd  /var/www/peertube/peertube-latest
 yarn install --production --pure-lockfile 
 cp /var/www/peertube/peertube-latest/config/default.yaml /var/www/peertube/config/default.yaml
+
+echo "Updating Configuration File..."
```
Chocobozzz: Well, it's not really updating the configuration file

svnet-libre: it was to make elegant ;)

PR #318, opened 2018-02-28
Support different video playback rates

jonathanraes:
Allow video playback on 0,5, 1, 1.25, 1.5 and 2x speed.

I didnt create an issue for this as it was a pretty trivial addition, feel free to close if it is undesired.


Chocobozzz:
Nice! I'll put it in a setting button I think because the control bar begins to be full :)

PR #322, opened 2018-03-03
Remove X-Frame options in nginx config

valvin1:
`X-Frame-Options DENY;` doesn't permit sharing using iframe

if present we get the following message in browser : 
`
Load denied by X-Frame-Options: https://peertube.valvin.fr/videos/embed/b2cc97ef-dc05-400b-94f8-43ea61237718 does not permit framing.`

### DIFF
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -38,7 +38,6 @@ server {
   # resolver_timeout 5s;
 
   add_header Strict-Transport-Security "max-age=63072000; includeSubDomains; preload";
-  add_header X-Frame-Options DENY;
   add_header X-Content-Type-Options nosniff;
   add_header X-XSS-Protection "1; mode=block";
   add_header X-Robots-Tag none;


rigelk:
I introduced it as part of the modern security recommendations in web servers, but arguably this particular protection is not needed on embedded videos.

I'd rather have the protection disabled for the `/videos/embed` route only.

valvin1:
@rigelk oh right, it might be better. i'm looking how to do it.

Yesterday when apply this nginx configuration I had to disable this part in order to make things working : 

```
ssl_session_timeout  10m;
ssl_session_cache shared:SSL:10m;
ssl_session_tickets off; # Requires nginx >= 1.5.9
```

but i don't know if I miss something on my side or if there is incompatibility somewhere.

valvin1:
it looks like we have to apply X-Frame-Options DENY in `/` directly and create a `/videos/embed` section without it.

I've [read](https://developer.mozilla.org/fr/docs/HTTP/Headers/X-Frame-Options) that the option ALLOW-FROM is not supported in Chrome. Which was the option I was looking for.

I'm not sure it is required to setup this options in static section.

Chocobozzz:
Merging this, do not hesitate to open another PR to re add this header on other routes :)

anmol26s:
@Chocobozzz  @rigelk 
Why was **selective route permission** commit reverted?

rigelk:
Because it simply didn't work as expected :/

anmol26s:
In YunoHost its disable by default.
I have added x-frame header by:
	
`add_header X-Frame-Options https://__DOMAIN__/videos/embed;

where __DOMAIN__ is the domain name for the Peertube instance.

https://github.com/YunoHost-Apps/peertube_ynh/blob/iframe_fix/conf/nginx.conf#L1
`

rigelk:
I see. We should add something along the lines of

```
add_header X-Frame-Options DENY;
add_header X-Frame-Options https://DOMAIN/videos/embed;
```

then.

anmol26s:
Most probably this should work.

rigelk:
That's what I thought too in 7f9a388acedc4ec1307ef20b5347a16feb98c2fe :sweat_smile: 

So I'm a bit wary of the changes, but I'll get to try that at least.

anmol26s:
> That's what I thought too in 7f9a388 sweat_smile

But this time its working for me in YunoHost. :joy:

anmol26s:
@rigelk 
This will not work. I don't know why it was working on my browsers. I tested it on my firerfox mobile browser and its not working there. :(
We need to remove X-Frame-Options rather then adding some url to it for the /video/embed.

For yunohost this its bigger issue because the **add_header X-Frame-Options "SAMEORIGIN";**  is present by default on all the new domains created by yunohost. And the header needs to be hidden by apps nginx conf (some way to overwrite this header or change its value to allow for /videos/embed).

I have tried to hide it by proxy_hide_header,proxy_set_header, add_header and few other ways but its not working. So the user will have to edit the nginx domain core file to remove this header after installing PeerTube through YunoHost. And this will get override every time YunoHost gets updated.

rigelk:
@anmol26s please check the PR, as I have essentially moved the implementation from nginx to express (internal). Would be awesome if you could test that!

PR #332, opened 2018-03-11
Update the production Docker image and compose file, initial travis support

### LINKED ISSUE #276: Synchronize configuration before the example and the Docker image
Currently the Docker image uses a separate ``production.yaml`` with sane default production values.
Unfortunately, that means that any new configuration setting needs to be mirrored in that file; this causes updates on ``develop`` to break the image because configuration cannot be found when loading the file.

I suggest we use the default configuration file in the image while customizing it a bit for production during build.

kaiyou:
This pull requests updates the Docker production files, including:
 - the build files for alpine and Debian, using the default configuration file as a base which fixes #276 
 - the compose file, fixing some typos
 - the travis build file for building and pushing Docker images.

Obviously, the Travis part should be extensively tested before anything is merged. If required, I can PR these commit separately.


Chocobozzz:
We'll use docker hub to build automatically our images like we do with the dev image: https://hub.docker.com/r/chocobozzz/peertube-dev/

kaiyou:
I do not believe Docker hub offers the possibility to specify a dockerfile path without changing the build context.

I discovered however that you can do so on Docker Cloud. If you wish to go that way I can of course revert changes to the Travis config.

Chocobozzz:
Yes please remove it

.travis.yml
```diff
@@ -55,3 +58,18 @@ after_failure:
...
+  - TAG=`echo $TRAVIS_BRANCH | sed 's,^master$,latest,'`
+  - docker build -t "${IMAGE_NAME}:alpine-$TAG" -f support/docker/production/Dockerfile.alpine
+  - docker build -t "${IMAGE_NAME}:stretch-$TAG" -f support/docker/production/Dockerfile.stretch
+  - docker tag "${IMAGE_NAME}:stretch-$TAG" "${IMAGE_NAME}:$TAG"
+
+deploy:
+  provider: script
+  script: docker push "${IMAGE_NAME}:$TAG" && docker push "${IMAGE_NAME}:alpine-$TAG" && docker push "${IMAGE_NAME}:stretch-$TAG"
+  on:
+    all_branches: true
```
rigelk: Shouldn't it be only on master in the end? (or is it only there while you are testing?)

Chocobozzz:
Closing due to inactivity. Don't hesitate to reopen :)

PR #340, opened 2018-03-14
ppa for ffmpeg in dependencies documentation

GammaNu:


rigelk:
Hi @GammaNu and thanks for pointing out the need for a more updated version of `ffmpeg` on Ubuntu platforms.

As much as possible, I would refrain from adding new third-party repositories. It isn't just a matter of malware, but also of how much I trust that repository maintainer and how regularly updated it is. With that being said, I checked that particular maintainer/repository and it seems fine.

support/doc/dependencies.md
```diff
@@ -20,10 +20,17 @@
 ```
 $ sudo apt update
 $ sudo apt install nginx ffmpeg postgresql openssl g++ make redis-server git
-$ ffmpeg -version # Should be >= 3.x 
+$ ffmpeg -version # Should be >= 3.x
 $ g++ -v # Should be >= 5.x
 ```
 
+if you still have a 2.x ffmpeg version on ubuntu-server :
```
rigelk: "If you still have a 2.x version of `ffmpeg` on Ubuntu:"

Chocobozzz: I fixed it myself: https://github.com/Chocobozzz/PeerTube/commit/30274d67ebd47a5cbbd57ea1cb14fb904e5e4445

Thanks for the review @rigelk

Issue #356, opened 2018-03-16
Display 122 minutes-long videos as 2:02:00


conferences-gesticulees:
* **PeerTube version or commit: 1.0.0-alpha.8**

When browsing videos longer than 60 minutes, their length is displayed only as `MMM:SS` when it should be `H:MM:SS`.

PR #360, opened 2018-03-17
Update video duration string to show hours when duration greater than or equal 60min

lucas-dclrcq:
Simple change to displayed video duration : show it as ```HH:MM:SS``` instead of  ```MMM:SS``` when duration is greater than or equal 60min.

Fixes: #356


tcitworld:
Not sure we need hours padding here, it takes place for nothing.

lucas-dclrcq:
Question : for now I'm displaying ```00:``` for hours when the duration is less than 60 min. Is it correct or do we want to only show the hours part when duration is greater than 60 min ?

It's would be less consistent across thumbnails, but the user would only see relevant numbers (Youtube shows video duration that way).

lucas-dclrcq:
@tcitworld ok i'll remove hours padding

tcitworld:
Yeah, hours should also be hidden when not useful.

PR #372, opened 2018-03-21
Use alpine images with docker-compose

Angristan:
Works the same, but lighter. I think it's a good idea to add them in the `docker-compose.yml` example.

### DIFF
--- a/support/docker/production/docker-compose.yml
+++ b/support/docker/production/docker-compose.yml
@@ -36,7 +36,7 @@ services:
     restart: "always"
 
   postgres:
-    image: postgres:10
+    image: postgres:10-alpine
     environment:
       POSTGRES_USER: postgres_user
       POSTGRES_PASSWORD: postgres_password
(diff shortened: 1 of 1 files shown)


Angristan:
Maybe we should also put the peertube:alpine image in the docker-compose.yml if you don't see any drawback.

rigelk:
@Angristan that would be fine by me. @bnjbvr ?

bnjbvr:
@Angristan Are you using the Alpine docker image? I've seen you mentioned a few segfaults on Mastodon when using PeerTube, so if you're using Alpine (which uses musl, for which native programs might need recompilation), I'd rather avoid it in the meanwhile and keep on using the Debian image instead.

Angristan:
Yep, let's keep the Debian image for peertube for now 😄

bnjbvr:
@Angristan Just to confirm: was it actually the Peertube alpine image that got the segfaults, or the postgres or redis alpine ones?

Angristan:
Yes: 
d80ea2168cbf4e73

No issue with peertube:stretch with alpine postgres/redis

rigelk:
Then I guess we can merge this PR as it stands.

PR #383, opened 2018-03-23
Fix #382 generic error when password is too long

### LINKED ISSUE #382: Generic error when password is too long
https://peertube.moe/signup

`Password cannot be more than 255 characters long.`

`Should have a valid password`

jomo:
I believe this fixes #382.

To be fair, I didn't set up a development environment to actually test this as it would have taken much longer than writing the actual fix. Please review carefully.

### DIFF
--- a/client/src/app/shared/forms/form-validators/user.ts
+++ b/client/src/app/shared/forms/form-validators/user.ts
@@ -22,10 +22,15 @@ export const USER_EMAIL = {
   }
 }
 export const USER_PASSWORD = {
-  VALIDATORS: [ Validators.required, Validators.minLength(6) ],
+  VALIDATORS: [
+    Validators.required,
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks @jomo!

client/src/app/shared/forms/form-validators/user.ts
```diff
@@ -22,10 +22,15 @@ export const USER_EMAIL = {
   }
 }
 export const USER_PASSWORD = {
-  VALIDATORS: [ Validators.required, Validators.minLength(6) ],
+  VALIDATORS: [
+    Validators.required,
+    Validators.minLength(6),
+    Validators.maxLength(255)
```
Dragnucs: Why can't a password be more than 255 chars length?

PR #398, opened 2018-03-24
Clean shell scripts

zapashcanon:
Hi,

I switched all the compatible scripts to `#!/bin/sh`, I added `set -eu` on all of them, which allowed me to remove a lot of `|| exit -1`.

Then I fixed some stuff spotted by ShellCheck. A few are remaining:

```sh
In ./scripts/release.sh line 7:
  PGID=$(ps -o pgid= $$ | grep -o "[0-9]*")
         ^-- SC2009: Consider using pgrep instead of grepping ps output.
```

I think this one is a false positive in this case. Can't find a way to do the same using `pgrep` (but I didn't searched a lot). We can leave it like this, and maybe add a shellcheck directive to disable this warning on this line if you want.

```sh
In ./scripts/release.sh line 44:
  [[ "$0" = "$BASH_SOURCE" ]] && exit 0
             ^-- SC2128: Expanding an array without an index only gives the first element.
```

I don't see where does `$BASH_SOURCE` come from, and I don't understand what's the goal of this line: why changing `$0` just before doing `exit 0` ? Depending on this, there's various way to fix the warning...

Last:

```sh
In ./scripts/danger/clean/dev.sh line 9:
  NODE_ENV=$(test npm run ts-node -- --type-check "scripts/danger/clean/cleaner")
  ^-- SC2034: NODE_ENV appears unused. Verify it or export it.


In ./scripts/watch/server.sh line 5:
NODE_ENV=$(test concurrently -k \
^-- SC2034: NODE_ENV appears unused. Verify it or export it.


In ./scripts/dev/client.sh line 5:
NODE_ENV=$(test concurrently -k \
^-- SC2034: NODE_ENV appears unused. Verify it or export it.


In ./scripts/dev/index.sh line 5:
NODE_ENV=$(test concurrently -k \
^-- SC2034: NODE_ENV appears unused. Verify it or export it.
```

I'm also not sure about what I did when changing all `NODE_ENV=test [...]` to `NODE_ENV=$(test [...])` as I don't see where's `NODE_ENV` used. ShellCheck can't find it either...

### DIFF
--- a/scripts/danger/clean/dev.sh
+++ b/scripts/danger/clean/dev.sh
@@ -1,8 +1,10 @@
 #!/bin/bash
 
+set -eu
+
 read -p "This will remove all directories and SQL tables. Are you sure? (y/*) " -n 1 -r
 echo
 
 if [[ "$REPLY" =~ ^[Yy]$ ]]; then
-  NODE_ENV=test npm run ts-node -- --type-check "./scripts/danger/clean/cleaner"
+  NODE_ENV=test npm run ts-node -- --type-check "scripts/danger/clean/cleaner"
 fi
--- a/scripts/dev/client.sh
+++ b/scripts/dev/client.sh
@@ -1,4 +1,6 @@
-#!/usr/bin/env sh
+#!/bin/sh
+
+set -eu
 
 NODE_ENV=test concurrently -k \
   "npm run watch:client" \
--- a/scripts/dev/index.sh
+++ b/scripts/dev/index.sh
@@ -1,4 +1,6 @@
-#!/usr/bin/env sh
+#!/bin/sh
+
+set -eu
 
 NODE_ENV=test concurrently -k \
   "npm run watch:client" \
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,11 +1,13 @@
 #!/bin/bash
 
+set -eu
+
 shutdown() {
   # Get our process group id
-  PGID=$(ps -o pgid= $$ | grep -o [0-9]*)
+  PGID=$(ps -o pgid= $$ | grep -o "[0-9]*")
 
   # Kill it in a new new process group
-  setsid kill -- -$PGID
+  setsid kill -- -"$PGID"
   exit 0
 }
 
@@ -16,7 +18,7 @@ if [ -z "$1" ]; then
   exit -1
 fi
 
-if [ -z $GITHUB_TOKEN ]; then
+if [ -z "$GITHUB_TOKEN" ]; then
   echo "Need GITHUB_TOKEN env set."
   exit -1
 fi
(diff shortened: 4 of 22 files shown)


scripts/danger/clean/dev.sh
```diff
@@ -1,8 +1,10 @@
 #!/bin/bash
 
+set -eu
+
 read -p "This will remove all directories and SQL tables. Are you sure? (y/*) " -n 1 -r
 echo
 
 if [[ "$REPLY" =~ ^[Yy]$ ]]; then
-  NODE_ENV=test npm run ts-node -- --type-check "./scripts/danger/clean/cleaner"
+  NODE_ENV=$(test npm run ts-node -- --type-check "scripts/danger/clean/cleaner")
```
Chocobozzz: We set the `test` value to env variable `NODE_ENV`, that will be used by our NodeJS script so I think you should remove this :)

scripts/dev/client.sh
```diff
@@ -1,5 +1,7 @@
-#!/usr/bin/env sh
+#!/bin/sh
 
-NODE_ENV=test concurrently -k \
+set -eu
+
+NODE_ENV=$(test concurrently -k \
```
Chocobozzz: Here too

scripts/dev/index.sh
```diff
@@ -1,5 +1,7 @@
-#!/usr/bin/env sh
+#!/bin/sh
 
-NODE_ENV=test concurrently -k \
+set -eu
+
+NODE_ENV=$(test concurrently -k \
```
Chocobozzz: Here too

scripts/release.sh
```diff
@@ -44,32 +44,32 @@ then
   [[ "$0" = "$BASH_SOURCE" ]] && exit 0
 fi
 
-cd ./client || exit -1
-npm version --no-git-tag-version --no-commit-hooks $1 || exit -1
+( cd client
```
Chocobozzz: What's the point of using `(`? :)

rigelk: It helps with operations in different directories, by changing it only in the scope of a subshell. http://www.tldp.org/LDP/abs/html/subshells.html

Chocobozzz: :+1:

rigelk:
To add to the correction of shell scripts, it might be interesting to add ShellCheck as a test step in our continuous integration suite.

Chocobozzz:
@zapashcanon :+1:

scripts/danger/clean/dev.sh
zapashcanon: OK, so actually, if I understand correctly what you're doing, I think my changes are correct.

zapashcanon:
@Chocobozzz, can you also answer my question about `BASH_SOURCE` in my first post ?

Chocobozzz:
@zapashcanon Sorry, I don't remember why I did this. I think you can remove it.

scripts/danger/clean/dev.sh
Chocobozzz: No, because with your version NODE_ENV will contain the result of the `test npm run ts-node...` script. We just want NODE_ENV to contain the string 'test'.

zapashcanon:
@Chocobozzz, you mean removing the whole line ? Or just the `exit 0` thing ? In the second case, I still need to know where is `BASH_SOURCE` used to fix the warning.

Chocobozzz:
@zapashcanon I mean you can replace `[[ "$0" = "$BASH_SOURCE" ]] && exit 0` by `exit 0`

I don't remember why I put this `[[ "$0" = "$BASH_SOURCE" ]] &&``

scripts/danger/clean/dev.sh
zapashcanon: Ah, OK, get it. :)

zapashcanon:
Fixed everything, it should be ready to merge now. Will do another PR to add Shellcheck as a part of the CI.

zapashcanon:
The CI failed, looking quickly at the logs, I don't see where it comes from. Does someone have any idea ?

rigelk:
Looking at the CI logs, nothing specific to your code has failed, so it's fine :) (not sure this test in particular is super stable)

PR #415, opened 2018-03-26
truncating long video names in import-videos.js (fixes #392)

### LINKED ISSUE #392: Importing video with a long name does not seems to work
latest
https://tbs.okhin.fr
* **Browser name/version:**
CLI
* **NodeJS version:**
v9.9.0

Using the import-videos.js script with an arabic name, the video should be created
There's a 500 Error, saying that the name of the video is invalid

* **Browser console log if useful (Gist/Pastebin...):**
* **Server log if useful (Gist/Pastebin...):**
https://gist.github.com/okhin/fd2de289beb61b00cd79ec7d29a92d43

rigelk:


server/tools/import-videos.ts
```diff
@@ -133,7 +133,7 @@ async function uploadVideoOnPeerTube (videoInfo: any, videoPath: string, languag
   }
 
   const videoAttributes = {
-    name: videoInfo.title,
+    name: videoInfo.title.substring(0,CONSTRAINTS_FIELDS.VIDEOS.NAME.max-2)+(videoInfo.title.length>CONSTRAINTS_FIELDS.VIDEOS.NAME.max-2)?'..':'',
```
Chocobozzz: Use `truncate` from lodash:

Chocobozzz:
Thanks @rigelk :+1:

PR #416, opened 2018-03-26
per missing field error in upload.js (fixes #414)

### LINKED ISSUE #414: [upload.js] Url, username, password, name and input file are required.
PeerTube v1.0.0-beta.1 

* **Browser name/version:**
n/a

* **NodeJS version:**
nodejs : 8.10.0-1nodesource1

```
$ node dist/server/tools/upload.js --url "https://peertube.example.org" --user "user" --password "password" --category 1  --licence 2 --file "/mnt/folder/file with whitespace.mp4"
Url, username, password, name and input file are required.
```

Some indication of which argument is missing... In my case, I'd forgotten `--video-name` but though I was a hitting a but because of something else.

rigelk:


Chocobozzz:
Thanks!

PR #421, opened 2018-03-26
show quota in stats, display quota on the about page, fixes #405

### LINKED ISSUE #405: Display User quota on about page
* **PeerTube version or URL:** Beta1

* **What is the expected behaviour?** : Users should be able to check if an instance applies a quota before registering.
* **What do you see instead?** : The only way to discover if you will be able to upload some videos on a peertube instance is by registering on it and then go to your user settings.

rigelk:
As outlined by @Aldarone, the quota is one of the deciding factors when choosing an instance, and we should make it easy to spot.

For now I placed it in the about page as suggested (admittedly the wording could be improved…):

showing quota on the about page

It conveniently pulls data from the `API/server/stats` route, which now exposes the video quota defined in the config. This way, servers publicly disclose their baseline quota for potential newcomers to know.

### DIFF
--- a/client/src/app/about/about.component.html
+++ b/client/src/app/about/about.component.html
@@ -13,6 +13,14 @@
     <div class="section-title">Terms</div>
 
     <div [innerHTML]="termsHTML"></div>
+
+    <div *ngIf="userVideoQuota !== -1;else noQuota">
+      This instance provides a baseline quota of {{ userVideoQuota | bytes: 0 }} space for the videos of its users.
+    </div>
+    
+    <ng-template #noQuota>
+      This instance provides unlimited space for the videos of its users.
+    </ng-template>
   </div>
 
   <div id="p2p-privacy">
--- a/client/src/app/about/about.component.ts
+++ b/client/src/app/about/about.component.ts
@@ -23,6 +23,10 @@ export class AboutComponent implements OnInit {
     return this.serverService.getConfig().instance.name
   }
 
+  get userVideoQuota () {
+    return this.serverService.getConfig().user.videoQuota
+  }
+
   ngOnInit () {
     this.serverService.getAbout()
       .subscribe(
@@ -31,7 +35,7 @@ export class AboutComponent implements OnInit {
           this.termsHTML = this.markdownService.textMarkdownToHTML(res.instance.terms)
         },
 
-        err => this.notificationsService.error('Error', err)
+        err => this.notificationsService.error('Error getting about from server', err)
       )
   }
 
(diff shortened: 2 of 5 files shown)


Chocobozzz:
It would be better to put the quota in the `ServerConfig` model.

client/src/app/shared/misc/number-formatter.pipe.ts
```diff
@@ -5,12 +5,15 @@ import { Pipe, PipeTransform } from '@angular/core'
 @Pipe({ name: 'myNumberFormatter' })
 export class NumberFormatterPipe implements PipeTransform {
   private dictionary: Array<{max: number, type: string}> = [
-    { max: 1000, type: '' },
+    { max: 1000, type: 'B' },
```
Chocobozzz: Please remove it because we use this number formatter to count views (and we don't want to have 102B views).

You could add an optional `suffix: string` parameter in the transform method.

rigelk:
ah indeed that sounds more logical to put in `ServerConfig`.

client/src/app/shared/misc/number-formatter.pipe.ts
```diff
@@ -4,16 +4,19 @@ import { Pipe, PipeTransform } from '@angular/core'
 
 @Pipe({ name: 'myNumberFormatter' })
 export class NumberFormatterPipe implements PipeTransform {
-  private dictionary: Array<{max: number, type: string}> = [
-    { max: 1000, type: '' },
-    { max: 1000000, type: 'K' },
-    { max: 1000000000, type: 'M' }
+  private dictionary: Array<{max: number, magnitude: string, unit_suffix: string}> = [
+    { max: 1000, magnitude: '', unit_suffix: ' bytes' },
```
Chocobozzz: Please abort your modifications here

rhaamo:
I've updated to master but it looks like I have an issue with this quota display.

The web page display ```This instance provides a baseline quota of NaN space for the videos of its users. ``` (NaN space) and the API call ```/api/v1/config``` returns ```"user":{"videoQuota":"1073741824"}``` the quota as string instead of a number.

Is that a bug or I have something wrong elsewhere ?

Chocobozzz:
@rhaamo Sorry my bad. https://github.com/Chocobozzz/PeerTube/commit/be1fc4bc50811df522cbfef299da6a7540589f65

rhaamo:
Your fix works as expected thanks!
Except the cosmetic of ```quota of 1.0000000252621255 GB space``` but the bug itself is fixed ;)

Issue #418, opened 2018-03-26
Background highlight on the hover in left menu


PhieF:
* **PeerTube version or URL:** beta 1
* **Browser name/version:**
* **NodeJS version:**

Left menu should have a hover background, as well as top-left icon
for the left menu, I've successfully "fixed" it by adding 

```css

menu[_ngcontent-c2] .panel-block[_ngcontent-c2] a[_ngcontent-c2]{
height:100%;
line-height:40px;
padding-left: 26px;
display:block;
}
menu[_ngcontent-c2] .block-title[_ngcontent-c2] {
padding-left: 26px
}
menu[_ngcontent-c2] .panel-block[_ngcontent-c2] a[_ngcontent-c2]:hover {
background:var(  --less-dark-blue);
}
```

in css
here is what I get 

https://framapic.org/j0PYJBfACWEw/KumBKRbzF2LP.png

it's better on mobile phone, even if not perfect (a slide animation for the menu would make things smoother, but can be laggy)

Thanks ! :) :)

rigelk:
LGTM. Would you mind filling a PR? :)

Chocobozzz:
Implemented by @Ealhad :)

PR #425, opened 2018-03-27
Add a hover effect on left menu links (#418)

Ealhad:
- removed the `margin-left` on the `.panel-block` to put it on the children
instead
- squashed the `margin-bottom` of the links into their `height` (and rounded it
to 40px, feel free to correct me)
- centered vertically the content of the links
- added the `background-color` change on hover for the links

I put the same color as the `.logged-in-block`. Maybe it should be externalized
as a SCSS variable?

I also added a CSS transition. Is it alright?


Chocobozzz:
Thanks @Ealhad

PR #434, opened 2018-03-29
Modified scripts/release.sh to add tar.xz and provide archive signatures

rigelk:


Chocobozzz:
Thanks @rigelk! But do you really think adding a `tar.xz` archive is useful?

rigelk:
@Chocobozzz barebones *nix installs don't contain unzip (i.e. server installs), but they always contain tar. Since our audience is servers, I'd go for that.

Then xz provides a [good deal of compression](http://stephane.lesimple.fr/blog/2010-07-20/lzop-vs-compress-vs-gzip-vs-bzip2-vs-lzma-vs-lzma2xz-benchmark-reloaded.html) (a good example for that is [the Linux kernel which uses it](https://www.kernel.org/xz-by-default-and-json.html) and has reduced archive size by 30% compared to gzip).

scripts/release.sh
```diff
@@ -56,31 +59,47 @@ git tag -s -a "$version" -m "$version"
...
-
-rm "$directory_name"
-
-cd "PeerTube"
+# Creating the archives
+(
+  cd ..
+  ln -s "PeerTube" "$directory_name"
+  zip -r "PeerTube/$zip_name" "$directory_name/CREDITS.md" "$directory_name/FAQ.md" \
+                              "$directory_name/LICENSE" "$directory_name/README.md" \
```
Chocobozzz: Maybe we could find a way to refractor these lines

PR #439, opened 2018-03-30
Add publishedAt field for video model.

jlebras:
* New field added in the `video` table + migration script

* `publishedAt` updated to NOW when privacy changes from private to
  public/unlisted (default = NOW)

* Models updated to handle the new attribute

* Client interface updated to use `publishedAt` instead of `createdAt`
  except in My Account > My Videos view

When testing I had to create the peertube_testX databases because the clean:server:test task was trying to drop them and failed, did I miss a step ?.


Chocobozzz:
Yeah \o/ Thanks @jlebras 

Yes, to run unit tests we need peertube_test{1,2,3} databases. Maybe we should add it to the contributing guide.

PR #447, opened 2018-03-30
Misc. typos

luzpaz:
Found via `codespell -q 3`


rigelk:
Thanks a lot @luzpaz! I didn't know about `codespell` but that would be an insteresting addition to our continuous integration testing suite too.

PR #452, opened 2018-03-31
Remove HTML markup

Dragnucs:
Make README.md fully markdown.

Some services do not render HTML. A full markdown file is more suitable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,70 +1,32 @@
-<h1 align="center">
-  PeerTube
-</h1>
+# PeerTube
 
-<h4 align="center">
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Elements are not centered anymore, so I prefer to keep the html

Dragnucs:
They don't really need to be centred. A better looking page has its place in joinpeertube.org in my opinion.

Chocobozzz:
No sorry I prefer things the way they are now. If there is a way to center elements in markdown I would be happy to change though

PR #511, opened 2018-04-15
Use gosu in Dockerfile to fix /data permissions errors

floSoX:


support/docker/production/docker-entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+set -e
+
+# first arg is `-f` or `--some-option`
+# or first arg is `something.conf`
+if [ "${1#-}" != "$1" ] || [ "${1%.conf}" != "$1" ]; then
+        set -- npm "$@"
+fi
+
+# allow the container to be started with `--user`
+if [ "$1" = 'npm' -a "$(id -u)" = '0' ]; then
```
Chocobozzz: Why `-a "$(id -u)" = '0'`?

PR #514, opened 2018-04-17
#512

### LINKED ISSUE #512: Add a way to bind the server
Hello,

as PeerTube is setup behind a reverse proxy following the production documentation, it'd seem useful to be able to make it bind to 127.0.0.1:port instead of 0.0.0.0:port.
After searching I didn't find any way of changing that.

Regards,

daftaupe:
Hello,

this is an attempt to fix #512.

Regards,

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -1,5 +1,5 @@
 listen:
-  hostname: ''
+  hostname: '127.0.0.1'
   port: 9000
 
 webserver:
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -1,4 +1,5 @@
 listen:
+  hostname: ''
   port: 9000
 
 webserver:
--- a/config/test-1.yaml
+++ b/config/test-1.yaml
@@ -1,5 +1,4 @@
 listen:
-  hostname: ''
   port: 9001
 
 webserver:
--- a/config/test-1.yaml
+++ b/config/test-1.yaml
@@ -1,4 +1,5 @@
 listen:
+  hostname: ''
   port: 9001
 
 webserver:
--- a/server.ts
+++ b/server.ts
@@ -216,6 +216,6 @@ async function startApplication () {
 
   // Make server listening
   server.listen(port, hostname)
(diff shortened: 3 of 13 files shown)


config/default.yaml
```diff
@@ -1,4 +1,5 @@
 listen:
+  hostname: ''
```
Chocobozzz: By default bind on  `127.0.0.1`

config/test-1.yaml
```diff
@@ -1,4 +1,5 @@
 listen:
+  hostname: ''
```
Chocobozzz: We don't need this parameter in test configurations

server.ts
```diff
@@ -214,7 +215,7 @@ async function startApplication () {
   Redis.Instance.init()
 
   // Make server listening
-  server.listen(port)
-  logger.info('Server listening on port %d', port)
+  server.listen(port, hostname)
+  logger.info('Server listening on port %s:%d', hostname, port)
```
Chocobozzz: on ~port~ %s:%d

Chocobozzz:
Please update docker config too https://github.com/Chocobozzz/PeerTube/blob/develop/support/docker/production/config/production.yaml#L2

config/default.yaml
daftaupe: Alright, I did it for the production example, but I didn't want to break the default config.

server.ts
daftaupe: woops, typo :)

daftaupe:
I will do the updates later, thanks for the review.

config/test-1.yaml
daftaupe: Are you sure @Chocobozzz  ? I think that would fail to start in that case.

Chocobozzz: Test configurations inherit the `default` configuration, so if you do https://github.com/Chocobozzz/PeerTube/pull/514#discussion_r181984904 it should work :)

Chocobozzz:
Thanks @daftaupe

PR #526, opened 2018-04-26
feature: initial syndication feeds tests for instance-wide feeds

rigelk:
It's the very minimal set of tests. It's meant to be extended to other feeds (per account, channel, and with transcoding/multiple torrents) needs as they will evolve, and is in no way exhaustive. 

Since we depend on `pfeed` (which I maintain and gets tested more formally), I found it pointless in the end to test against an XSD model/JSON schema.


package.json
```diff
@@ -59,6 +59,7 @@
     "bittorrent-tracker": "^9.0.0",
     "bluebird": "^3.5.0",
     "body-parser": "^1.12.4",
+    "chai-xml": "^0.3.2",
```
Chocobozzz: It's a dev dependency :)

PR #532, opened 2018-05-02
By default, use the official Docker image with Docker Compose

floSoX:


Chocobozzz:
I think compose tries to fetch the hub image first, and if it does not find it, it builds it manually no?

Chocobozzz:
Okay, it seems you are right. Since there were conflicts on this PR I commented out the lines in https://github.com/Chocobozzz/PeerTube/commit/13a6b53655cdd7baba494abc3c9ff52788d4651e. Thanks!

PR #543, opened 2018-05-07
Add PEERTUBE_INSTANCE_CUSTOM_CSS and PEERTUBE_INSTANCE_CUSTOM_JS

fflorent:
This allow a user with custom CSS and JS to persist it accross container removal.

What do you think?
Florent

### DIFF
--- a/support/docker/production/config/custom-environment-variables.yaml
+++ b/support/docker/production/config/custom-environment-variables.yaml
@@ -55,3 +55,6 @@ instance:
   name: "PEERTUBE_INSTANCE_NAME"
   description: "PEERTUBE_INSTANCE_DESCRIPTION"
   terms: "PEERTUBE_INSTANCE_TERMS"
+  customizations:
+    css: "PEERTUBE_INSTANCE_CUSTOM_CSS"
+    javascript: "PEERTUBE_INSTANCE_CUSTOM_JS"


Chocobozzz:
I think we should put the configuration files in `/data/config` directory (`data` is a volume).

Ping @Angristan @bnjbvr

Angristan:
I have two volumes now:

```
    volumes:
      - ./data:/data
      - ./config/local-production.json:/app/support/docker/production/config/local-production.json
```

But yes that would be a good idea

fflorent:
Right, so `support/docker/production/config/*` should be moved where ? Or could we just bind `support/docker/production/config` to `./config` maybe (with an overlayed volume, not sure whether that's feasible or relevant)?

Chocobozzz:
I implemented it :) https://github.com/Chocobozzz/PeerTube/commit/fd5e57bbe2accbdb16b6aa65337c5ef44b5bd8fb

Angristan:
If I'm not mistaken, the CMD will run as the last USER, so now it's root 🤔

Chocobozzz:
@Angristan I'm sorry I don't understand. The `USER root` was just moved, so it should not change the previous behaviour.

Angristan:
Nevermind, I didn't see all the `gosu` stuff

bnjbvr:
(Just wanted to say good job!)

Angristan:
So what should I with my volumes now? I'm lost :(

Chocobozzz:
@Angristan Save your local configuration `local-production.json`, then declare `./config:/config` in volumes, then `docker-compose up` and finally copy your `local-production.json` inside the config folder.

Angristan:
Thanks!

I have this issue:

```
[undefined:443] 2018-05-09 22:17:44.063 info: Getting client tokens for host peertube.angristan.xyz is forbidden (expected undefined).
```

screenshot_10-05-2018_00-18-55

Could it be related?

fflorent:
@Angristan Have you renamed the `PEERTUBE_HOSTNAME` variable into `PEERTUBE_WEBSERVER_HOSTNAME` in the docker-compose.yml and/or in the `docker-compose up` command you invoke?

Angristan:
Indeed, thanks a lot @fflorent!

PR #545, opened 2018-05-07
Switch Arch Linux dependency to NodeJS LTS

daftaupe:
In order to avoid potential issue with new non-tested nodejs version that Arch Linux is shipping.


Chocobozzz:
Thanks @daftaupe

Issue #552, opened 2018-05-08
Feature request for import-videos.js : don't pass password as a command line argument


JohnXLivingston:
Hello everyone.
First of all, congrats for this amazing job !

I have a feature request for import-videos.js : is it possible to ask for password once the command line runs rather than passing it by argument ? So it will not be in bash-history or displayed on the screen

J.

JohnXLivingston:
See https://github.com/Chocobozzz/PeerTube/pull/557 for an implementation

PR #557, opened 2018-05-10
import-videos.js : prompt for password if not provided in CLI

JohnXLivingston:
As I suggested in issue #552 , here is an implementation for prompting for password when using the import-videos.js without providing one is the args. 
So you will neither display the password on screen, neither store it in the .bash_history


server/tools/import-videos.ts
```diff
@@ -23,29 +24,54 @@ program
...
+     properties: {
+       password: {
+         hidden:true,
+         required:true
+       }
+     }
+   }
+   prompt.get(schema, function(err, result) {
+     if (err) {
+       console.log(err.message)
```
Chocobozzz: You need to reject with err

server/tools/import-videos.ts
```diff
@@ -23,29 +24,54 @@ program
...
+     }
+     resolve(result.password)
+   })
+  })
+}
+
 async function run () {
+  if (
+    !user.password
```
Chocobozzz: All the if (including `user.password`) on the same line please

server/tools/import-videos.ts
```diff
@@ -23,29 +24,54 @@ program
...
+  ) {
+    user.password = await promptPassword();
```
Chocobozzz: Remove ;

server/tools/import-videos.ts
```diff
@@ -23,29 +24,54 @@ program
...
   maxBuffer: Infinity
 }
 
+async function promptPassword () {
+  return new Promise ( resolve => {
+   prompt.start()
+   const schema = {
+     properties: {
+       password: {
+         hidden:true,
```
Chocobozzz: space after `:`

server/tools/import-videos.ts
```diff
@@ -23,29 +24,54 @@ program
...
+         required:true
```
Chocobozzz: space after `:`

JohnXLivingston:
Ok, sorry. 
I'm not familiar with the standard code style. And my vim editor was not configured for typescript. I did it too quickly.
I'm going to fix this.


And what do you thing about an import button in the client ? Near to the upload button. Opening a form with one 'url' field ? On submit, it will create a planed job doing the import (to avoid a request timeout).
I never worked with angularJS, but I can try to implement this feature.

Chocobozzz:
It's planned (should be done in a few months).

server/tools/import-videos.ts
```diff
@@ -46,30 +46,29 @@ const processOptions = {
...
+         hidden: true,
+         required: true
        }
      }
    }
-   prompt.get(schema, function(err, result) {
+   prompt.get(schema, function (err, result) {
      if (err) {
        console.log(err.message)
+       return res();
```
Chocobozzz: You need to reject, not resolve

Chocobozzz:
There are some linting issues :) https://travis-ci.org/Chocobozzz/PeerTube/jobs/377640914

JohnXLivingston:
Should be good this time.
I'm an old fashioned dev, I'm discovering these tools (lint, nodejs, typescript, etc...)

PR #566, opened 2018-05-14
redis unix socket connection variant and patching kue

rigelk:
+ misc: fixed a small inconsistency in `.prettierrc.yml` that clashed with tslint's config of removing unecessary commas.

basically kue was not supporting anymore unix sockets. A small patch fixed it (https://github.com/Automattic/kue/issues/1017) but that meant patching kue, for which I used a small patch script. that places patches in a new `patches` directory.


Chocobozzz:
I'm sorry but I don't want to patch external packages. Just maintaining node dependencies is already a hard work (packages that break things on minor version change etc), if we begin to patch them manually it will be too much work.

This fix should be applied upstream

rigelk:
Then I guess all I can do is try to make push the existing PR upstream to kue.

PR #567, opened 2018-05-14
robots.txt config and route

### LINKED ISSUE #553: Adding a robots.txt file
It would be nice to include a robots.txt file in peertube. And offer to configure if the instance is allowed or not (to differenciate tests and productions servers

Best regards,
J.

rigelk:
resolves #553

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -90,6 +90,10 @@ instance:
   customizations:
     javascript: '' # Directly your JavaScript code (without <script> tags). Will be eval at runtime
     css: '' # Directly your CSS code (without <style> tags). Will be injected at runtime
+  # Robot.txt rules. To allow robots to crawl your instance and allow indexation of your site, remove the '/'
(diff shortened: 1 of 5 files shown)


config/default.yaml
```diff
@@ -90,6 +90,9 @@ instance:
   customizations:
     javascript: '' # Directly your JavaScript code (without <script> tags). Will be eval at runtime
     css: '' # Directly your CSS code (without <style> tags). Will be injected at runtime
+  robots: |
+    User-agent: *
+    Disallow: /
```
Chocobozzz: Do we really want disallow crawling by default?

rigelk: I figured that since we discover videos and instances without crawlers anyway, that's fine.

Chocobozzz: Maybe add a comment "to allow robots to crawl your instance..."?

PR #573, opened 2018-05-17
IPvX registration filtering via CIDR (supports whitelists and blacklists)

### LINKED ISSUE #558: Restricted access to the signup page
Hello,
It would be nice to be able to configure a few ip range allowed to access the signup page.
Signup page only, not login page (full access).

Use case, who can create an account:
- Family at home
- Collegues in office
- Students at school
- Customers in an ISP
...

Best regards,
Grays

rigelk:
resolves #558

### DIFF
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -75,6 +75,10 @@ admin:
 signup:
   enabled: false
   limit: 10 # When the limit is reached, registrations are disabled. -1 == unlimited
+  filters: 
+    cidr: # You can specify CIDR ranges to whitelist (empty = no filtering) or blacklist
+      whitelist: []
+      blacklist: []
 
 user:
   # Default value of maximum video BYTES the user can upload (does not take into account transcoded files).
--- a/server/helpers/utils.ts
+++ b/server/helpers/utils.ts
@@ -1,4 +1,6 @@
 import { Model } from 'sequelize-typescript'
+import * as ipaddr from 'ipaddr.js'
+const isCidr = require('is-cidr')
 import { ResultList } from '../../shared'
 import { VideoResolution } from '../../shared/models/videos'
 import { CONFIG } from '../initializers'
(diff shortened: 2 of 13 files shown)


server/helpers/utils.ts
```diff
@@ -48,6 +50,34 @@ async function isSignupAllowed () {
...
+  let matched: string
+
+  if (CONFIG.SIGNUP.FILTERS.WHITELIST.length > 0) {
+    excludeList.push('unknown')
+  }
+
+  if (addr.kind() === 'ipv4') {
+    const addrV4: ipaddr.IPv4 = ipaddr.IPv4.parse(ip)
+    let rangeList = {
+      whitelist: CONFIG.SIGNUP.FILTERS.WHITELIST.filter(cidr => isCidr.v4(cidr)).map(cidr => ipaddr.IPv4.parseCIDR(cidr)),
```
Chocobozzz: maybe add \n before .map

server/helpers/utils.ts
```diff
@@ -48,6 +50,34 @@ async function isSignupAllowed () {
...
+  const addr = ipaddr.parse(ip)
+  let excludeList: string[] = ['blacklist']
+  let matched: string
+
+  if (CONFIG.SIGNUP.FILTERS.WHITELIST.length > 0) {
+    excludeList.push('unknown')
+  }
+
+  if (addr.kind() === 'ipv4') {
+    const addrV4: ipaddr.IPv4 = ipaddr.IPv4.parse(ip)
```
Chocobozzz: I think addrV4 is inferred

server/helpers/utils.ts
```diff
@@ -48,6 +50,34 @@ async function isSignupAllowed () {
...
+    let rangeList = {
```
Chocobozzz: const

server/helpers/utils.ts
```diff
@@ -48,6 +50,34 @@ async function isSignupAllowed () {
...
+      whitelist: CONFIG.SIGNUP.FILTERS.WHITELIST.filter(cidr => isCidr.v4(cidr)).map(cidr => ipaddr.IPv4.parseCIDR(cidr)),
+      blacklist: CONFIG.SIGNUP.FILTERS.BLACKLIST.filter(cidr => isCidr.v4(cidr)).map(cidr => ipaddr.IPv4.parseCIDR(cidr))
+    }
+    matched = ipaddr.subnetMatch(addrV4, rangeList, 'unknown')
+  } else if (addr.kind() === 'ipv6') {
+    const addrV6: ipaddr.IPv6 = ipaddr.IPv6.parse(ip)
+    let rangeList = {
```
Chocobozzz: const

server/helpers/utils.ts
```diff
@@ -48,6 +50,34 @@ async function isSignupAllowed () {
   return totalUsers < CONFIG.SIGNUP.LIMIT
 }
 
+async function isSignupAllowedForCurrentIP (ip: string) {
+  const addr = ipaddr.parse(ip)
+  let excludeList: string[] = ['blacklist']
+  let matched: string
+
+  if (CONFIG.SIGNUP.FILTERS.WHITELIST.length > 0) {
+    excludeList.push('unknown')
```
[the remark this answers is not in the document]
rigelk: The `subnetMatch()` function matches IPs against ranges. If there is no matched range, it returns 'unknown'. I it matches against a range, it returns the range name.

See https://github.com/whitequark/ipaddr.js/blob/8e69afeb4053ee32447a101845f860848280eca5/lib/ipaddr.js#L34-L54

```js
  ipaddr.subnetMatch = function(address, rangeList, defaultName) {
    var k, len, rangeName, rangeSubnets, subnet;
    if (defaultName == null) {
      defaultName = 'unicast';
    }
    for (rangeName in rangeList) {
      rangeSubnets = rangeList[rangeName];
      if (rangeSubnets[0] && !(rangeSubnets[0] instanceof Array)) {
        rangeSubnets = [rangeSubnets];
      }
      for (k = 0, len = rangeSubnets.length; k < len; k++) {
        subnet = rangeSubnets[k];
        if (address.kind() === subnet[0].kind()) {
          if (address.match.apply(address, subnet)) {
            return rangeName;
          }
        }
      }
    }
    return defaultName;
  };
```

The thing is, if our whitelist is empty, we don't want to block unknown adresses (only blacklisted adresses). So we only add 'unknown' to the list of excluded ranges if there is a non-empty whitelist.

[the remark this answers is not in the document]
Chocobozzz: okidoki

config/production.yaml.example
```diff
@@ -75,6 +75,9 @@ admin:
 signup:
   enabled: false
   limit: 10 # When the limit is reached, registrations are disabled. -1 == unlimited
+  filters: # You can specify CIDR ranges to whitelist (empty = no filtering) or blacklist
```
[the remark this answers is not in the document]
rigelk: maybe `filters_cidr`
(could be more consistent if we ever implement another kind of filter, like `filters_name`)

Chocobozzz:
Nice work! Just lint issues and you can merge it :)

If you have time please add unit tests in the future.

Grayswandir:
Perfect, I just finished few tests, if you use a 'space', everybody can create an account (failed, but it's an error of config).
List is fine, comma works too.
Thank you,

rigelk:
@Grayswandir you mean that a ' ' character is not allowed by the config but could be interpreted as a CIDR?

Grayswandir:
@rigelk no, it was just a test to find the CIDR separator. I try a ' ', like you can find in /etc/hosts.allow file. But as it's not the way to write a list in Yaml language, it didn't work fine. Normal :)

rigelk:
Oh okay :) Thanks for testing the functionality by the way! Admittedly I only did basic testing.

PR #581, opened 2018-05-21
DB selection in redis

### LINKED ISSUE #579: How to define which redis database to use?
In the context of YunoHost hosting, I would like to tell PeerTube which Redis Database to use, so it doesn't create side-effects when another app is removed.

Is it possible?

rigelk:
resolves #579


Chocobozzz:
Nice work @rigelk!

PR #583, opened 2018-05-21
Configure path used by script upgrade.sh

jocelynj:
It is now possible to pass the path to upgrade on script, as Peertube could be installed in another path than /var/www/peertube.

I couldn't find a clean way to guess the path to use, so I've chosen instead to let administrator specify the path by hand.

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -2,26 +2,28 @@
 
 set -eu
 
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks @jocelynj for the PR and @rigelk for the review!

PR #592, opened 2018-05-27
Allow to setup transcoding in the docker image

kaiyou:
This is a very basic pull request to improve support for transcoding options in the Docker image.

### DIFF
--- a/support/docker/production/config/production.yaml
+++ b/support/docker/production/config/production.yaml
@@ -45,3 +45,9 @@ log:
 
 admin:
   email: null
+
+transcoding:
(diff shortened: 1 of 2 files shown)


support/docker/production/config/production.yaml
```diff
@@ -45,3 +45,9 @@ log:
 
 admin:
   email: null
+
+transcoding:
+  resolutions:
```
Chocobozzz: Could you remove it? I'm not sure we want to transcode to multiple resolutions by default (or at least it should be consistent with the main production.yaml file).

PR #600, opened 2018-05-28
update script peertube upgrade

OpenAlgeria:
correction bug upgrade


Chocobozzz:
`yarn install --production --pure-lockfile` already does this 

what was your issue?

PR #603, opened 2018-05-29
Use sysrc command on FreeBSD

norrist:
Run sysrc commands instead of directly editing /etc/rc.conf.
syntax errors in /etc/rc.conf can leave the system unbootable.


rigelk:
Thanks @norrist !

Issue #450, opened 2018-03-31
Improve discoverability of followed instances


pmiossec:
Feature request / idea of improvement:

* Being able to see somewhere the list of the instances followed by the current instance I'm on and propose links to navigate toward this instance
Example: `framatube.org` follow `peertube.datagueule.tv` but there is nowhere to see it and go to it easily

* From a video, display a link to be able to navigate toward the instance where the video has been published.

Example: if I play https://framatube.org/videos/watch/0d4a1ec1-9aa5-47e5-9e0d-3704b0ea7384 , I want to find a link to the `peertube.datagueule.tv` instance to easily be able to find all the *local* videos from this instance

Thanks a lot for what you are doing! :+1:

strypey:
For the same reason, it would be good for users to have access to the list of instances blocked by any given instance. Mastodon provides some useful precedents for the community politics of instances. When people are choosing their instance, they're essentially choosing their "local pub", the community of "regulars" they will see and interact with a lot, and the "publican" who is in charge of moderating behaviour (and pouring the drinks). Seeing who is banned from the pub and why, is as important in making that decision as seeing who the regulars are.

Jorropo:
Hi, as temporary solution you can use this script :)
It does this automaticaly.
https://github.com/Jorropo/PeerTube-autoDiscover

Chocobozzz:
Should be fixed by https://github.com/Chocobozzz/PeerTube/commit/a6dbbf03865a955caaedc3b12f3de3e386fe850f that adds a new section in the about page to discover who the instance follows and who are its followers.

PR #607, opened 2018-05-30
Improve discoverability by fixing helper and showcasing origin instance

### LINKED ISSUE #595: Display name being shown instead of actual username@address
beta.6
The actual username@address should display in the applicable "Get Help" menus instead of using the display name.
image

rigelk:
fixes #595
partial #450 implementation

### DIFF
--- a/client/src/app/videos/+video-watch/video-watch.component.html
+++ b/client/src/app/videos/+video-watch/video-watch.component.html
@@ -21,11 +21,11 @@
               {{ video.channel.displayName }}
             </a>
             <!-- Here will be the subscribe button -->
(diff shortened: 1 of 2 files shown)


client/src/app/videos/+video-watch/video-watch.component.html
```diff
@@ -21,14 +21,17 @@
...
           </div>
 
           <div class="video-info-by">
-            <a [routerLink]="[ '/accounts', video.by ]" title="Go the account page">
+            <a [routerLink]="[ '/accounts', video.by ]" title="Go to the account page">
               <span>By {{ video.by }}</span>
               <img [src]="video.accountAvatarUrl" alt="Account avatar" />
             </a>
+            <a *ngIf="!video.isLocal" href="https://{{video.account.host}}" title="Go to the instance of origin of this account">
+              <span>on {{ video.account.host }}</span>
```
Chocobozzz: The info is already in the `video.by`. Please remove this block.

If you want to have a link directly in the origin instance, put it in https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/%2Baccounts/accounts.component.html#L10

Issue #477, opened 2018-04-06
Environment variables for Docker Compose


floSoX:
Hello!

Thank you for your great job!
I started with PeerTube and its `docker-compose.yml`.
I wonder, to simplify the customisation, whether it's a good idea to generalise the environment variables usage in the `docker-compose.yml` file and add a file `.env` in this directory.

For example:
```yaml
version: "3.3"

services:

  peertube:
[...]
    environment:
      PEERTUBE_HOSTNAME: my.domain.tld
[...]
      PEERTUBE_DB_PASSWORD: postgres_password
[...]
  postgres:
[...]
      POSTGRES_PASSWORD: postgres_password
```
Became:
```yaml
version: "3.3"

services:

  peertube:
[...]
    environment:
      PEERTUBE_HOSTNAME: ${PEERTUBE_HOSTNAME}
[...]
      PEERTUBE_DB_PASSWORD: ${DB_PASSWORD}
[...]
  postgres:
[...]
      POSTGRES_PASSWORD: ${DB_PASSWORD}
```

And we find, in `.env` file:
```ini
PEERTUBE_HOSTNAME="my.domain.tld"
DB_PASSWORD="postgres_password"
```

If you think it's a good idea, I can submit a pull request.

Regards,

PR #617, opened 2018-06-02
Environment variables for Docker Compose

floSoX:
Re-Implementation of this discussion: https://github.com/Chocobozzz/PeerTube/issues/477


support/docker/production/.env
```diff
@@ -0,0 +1,15 @@
...
+PEERTUBE_WEBSERVER_PORT=443
+PEERTUBE_WEBSERVER_HTTPS="true"
+PEERTUBE_REDIS_HOSTNAME=redis
+PEERTUBE_SMTP_USERNAME=null
+PEERTUBE_SMTP_PASSWORD=null
+PEERTUBE_SMTP_HOSTNAME=null
+PEERTUBE_SMTP_PORT=25
+PEERTUBE_SMTP_FROM=noreply@peertube.domain.tld
+PEERTUBE_SMTP_TLS="true"
+PEERTUBE_ADMIN_EMAIL=admin@domain.tld
```
Chocobozzz: I think it would be better to add the comment from https://github.com/Chocobozzz/PeerTube/pull/617/files#diff-5003d902a2333dfdbcd79d688f2806daL38 directly in this file

Issue #560, opened 2018-05-12
Transcoding elsewhere and submit to Peertube


fflorent:
I would like to know if there are some ways to:
 - transcode videos on another computer (using ffmpeg);
 - upload the result of the transcoding to the Peertube server?

This way, as I have a much more efficient CPU, I could submit the different resolutions of a video more rapidly.

Does it sound to you like a reasonnable use case? Do you know if that's feasible with the current tools? Or should this lead to some dirty tricks in the database (I saw there is the "videoFile" table for that).

Cheers!
Florent

Chocobozzz:
We use the `ffmpeg` command, so you can create a script that accepts the same options than `ffmpeg` but sends the file on another server, transcode it, and fetch the result on the PeerTube server (just an example). There are some tools on the web that try to distribute `ffmpeg` transcoding

You should be able to change `ffmpeg` path with `FFMPEG_PATH` env.

fflorent:
Well, I meant on my local personal computer (which is time to time off).

PR #619, opened 2018-06-02
Add create-import-video-file-job command

fflorent:
This is a WIP but I request some feedback ;).

This PR is meant to provide a command to import a video file for a given video UUID. Here are some use case for such a command:
- You want to restore a corrupted video file ;
- You want to support another resolution for all your videos and the server don't have a powerful CPU → You realize that you'd better transcode (with the right options) on another machine, upload it to the server and import the video file (#560)  ;
- You uploaded a video with a low resolution and want to propose a higher one ;
- You'd like to update something on your video (fixing something wrong, give more details, etc) ;

On all these use case, you'd like to keep the meta-data on it (descriptions, number of views, comments, etc).

So I propose a new command:
```bash
$ node ./dist/scripts/create-import-video-file-job.js -v [videoUUID] -i [videoFile]
```

Where:
 - `videoUUID` is the UUID of an existing video
 - `videoFile` is the video which replaces an existing video file or add a new resolution for it 

If you think these use case are justified and correctly addressed, here are some TODO:

- [ ] write tests (yeah, I shamefully submitted without any of them whereas I usually do TDD);

And some questions:

- [x] Is `import-video-file` meaningful? Probably far from being perfect, suggestions welcome!
- [x] Is `video-file.ts` the right place to add the function `processVideoImport`? Not really sure;
- [x] Not sure of what should be done when the import is finished. I call the same callback for the transcoding job, but not really sure;

Nitpicking and bikeshedding are welcome! I am conscious lots of things should be improved and I'll take comments into account humbly! ;)

Cheers!
Florent


server/models/video/video.ts
```diff
@@ -1315,6 +1316,29 @@ export class VideoModel extends Model<VideoModel> {
...
+      resolution: (await getVideoFileResolution(inputFilePath)).videoFileResolution,
+      extname: extname(inputFilePath),
+      size: (await statPromise(inputFilePath)).size,
+      videoId: this.id
+    })
+
+    const outputPath = this.getVideoFilePath(newVideoFile)
+    const oldVideoFile = this.VideoFiles.find(videoFile => videoFile.resolution === newVideoFile.resolution)
+    if (oldVideoFile) {
+      oldVideoFile.destroy()
```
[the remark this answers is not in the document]
fflorent: OK, I think removing the oldVideoFile from the array should help. Also there is a momentary in which the video is not available (during the video copy).

server/initializers/constants.ts
```diff
@@ -74,6 +74,7 @@ const JOB_ATTEMPTS: { [ id in JobType ]: number } = {
   'activitypub-http-unicast': 5,
   'activitypub-http-fetcher': 5,
   'activitypub-follow': 5,
+  'video-file-import': 5,
```
Chocobozzz: 1 like video-file

server/models/video/video.ts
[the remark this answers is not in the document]
Chocobozzz: I think that just updating the size, infohash and extname should be better

fflorent:
@Chocobozzz Thanks for your feedback. I'll write the tests this afternoon!

fflorent:
@Chocobozzz Not sure what is the intent behind these steps:
> - Upload a video on server 2 (transcoded is enabled)
> - Import a vile
> - Check server 1 and 2 has this video with your imported file

Chocobozzz:
> Not sure what is the intent behind these steps:

To test if **replacing** an existing video file works (ie the infohash and size are updated)

fflorent:
Here is the work in progress on the tests. The second test doesn't pass for some reasons… https://github.com/Chocobozzz/PeerTube/pull/619/commits/4c998fef8736ec2b2887c85b554b28a5395846b7

I have two questions for now:
 - I am not sure what to pass to `overrideConfig`;
 - It seems like the extension is not updated in the magnet;

If anyone can give me some clue, that would be very appreciated :).
Thanks in advance,
Florent

server/tests/cli/create-import-video-file-job.ts
```diff
@@ -0,0 +1,110 @@
...
+  ServerInfo,
+  setAccessTokensToServers,
+  uploadVideo,
+  wait,
+  getVideo, flushAndRunMultipleServers, doubleFollow
+} from '../utils'
+
+function assertVideoProperties (video: VideoFile, resolution: number, extname: string) {
+  expect(video).to.have.nested.property('resolution.id', resolution)
+  expect(video).to.have.property('magnetUri').that.matches(new RegExp(`\\.${extname}$`))
```
Chocobozzz: Just `indexOf(str) !== -1` should be enought. Please test the `torrentUrl` and `fileUrl` too

Chocobozzz: Please check the `size` too

server/tests/cli/create-import-video-file-job.ts
```diff
@@ -0,0 +1,110 @@
...
+      const { data: videos } = (await getVideosList(server.url)).body
+      expect(videos).to.have.lengthOf(2)
+
+      let infoHashes: { [ id: number ]: string }
+
+      const video = videos.find(({ uuid }) => uuid = video2UUID)
+      const videoDetail: VideoDetails = (await getVideo(server.url, video.uuid)).body
+
+      console.log(videoDetail.files)
+      expect(videoDetail.files).to.have.lengthOf(2)
```
Chocobozzz: server 2 transcodes so files should have length of 4

server/tests/cli/create-import-video-file-job.ts
```diff
@@ -0,0 +1,110 @@
...
+      const [originalVideo, transcodedVideo] = videoDetail.files
+      assertVideoProperties(transcodedVideo, 320, 'webm')
```
Chocobozzz: transcoded videos are always mp4

server/tests/utils/server/servers.ts
```diff
@@ -35,7 +35,7 @@ interface ServerInfo {
   }
 }
 
-function flushAndRunMultipleServers (totalServers) {
+function flushAndRunMultipleServers (totalServers, configOverrideFn = (numServer: number) => undefined) {
```
Chocobozzz: please remove configOverrideFn

server/tests/utils/server/servers.ts
```diff
@@ -53,7 +53,7 @@ function flushAndRunMultipleServers (totalServers) {
         for (let j = 1; j <= totalServers; j++) {
           // For the virtual buffer
           setTimeout(() => {
-            runServer(j).then(app => anotherServerDone(j, app))
+            runServer(j, configOverrideFn(j)).then(app => anotherServerDone(j, app))
```
Chocobozzz: Please remove configOverrideFn

server/models/video/video.ts
```diff
@@ -1315,6 +1316,38 @@ export class VideoModel extends Model<VideoModel> {
...
+
+    const outputPath = this.getVideoFilePath(updatedVideoFile)
+    await copyFilePromise(inputFilePath, outputPath)
+
+    const currentVideoFile = this.VideoFiles.find(videoFile => videoFile.resolution === updatedVideoFile.resolution)
+    const isNewVideoFile = !currentVideoFile
+
+    if (!isNewVideoFile) {
+      if (currentVideoFile.extname !== updatedVideoFile.extname) {
+        await unlinkPromise(this.getVideoFilePath(currentVideoFile))
```
Chocobozzz: `this.removeFile(currentVideoFile)`

server/models/video/video.ts
```diff
@@ -1315,6 +1316,38 @@ export class VideoModel extends Model<VideoModel> {
...
+      }
+      currentVideoFile.set('size', updatedVideoFile.size)
+      updatedVideoFile = currentVideoFile
+    }
+
+    await this.createTorrentAndSetInfoHash(updatedVideoFile)
+
+    await updatedVideoFile.save()
+
+    if (isNewVideoFile) {
```
Chocobozzz: You need to delete the old one if it's not a new file

fflorent: Well, in this case, either:
 - the extension name has changed, so the file is removed (see above)
 - the extension name is the same, so is the whole file name. The old file is overridden by the `copyFilePromise` call

I think this cover all the cases :)

server/tests/utils/server/servers.ts
fflorent: How do I disable transcoding then for the first server? (if I should)

Chocobozzz:
You don't need to disable transcoding manually

See https://github.com/Chocobozzz/PeerTube/blob/develop/config/test-1.yaml#L31 and https://github.com/Chocobozzz/PeerTube/blob/develop/config/test-2.yaml#L28

fflorent:
So server[0] is automatically mapped to the test-1.yaml config, and server[1] to the test-2.yaml?

Florent

Chocobozzz:
Yes

PR #621, opened 2018-06-03
FAQ: Fix out of date summary with broken link + rephrase

tuxayo:
"Which" seems more appropriate than "what". (disclaimer: I'm not a native speaker though (french))

You can check here that the link works: https://github.com/Chocobozzz/PeerTube/blob/ff085ff62e09e7efc45f08b0c60d4012917da026/FAQ.md


rigelk:
Sounds indeed better. Thanks @tuxayo ! :)

PR #635, opened 2018-06-07
Update sudo command for yarn during upgrade

mkody:
Like #267.
Without this, Yarn will try to use `/root/.config/yarn` with the peertube user, which obviously doesn't have access.


Chocobozzz:
Nice catch! Thanks @mkody

PR #636, opened 2018-06-08
Enable default link behavior and accessibility for login page

SansPseudoFix:
Links wasn't really obvious and needed some accessibility.

### DIFF
--- a/client/src/app/login/login.component.html
+++ b/client/src/app/login/login.component.html
@@ -17,7 +17,7 @@
           or create an account
         </a>
 
-        <a i18n *ngIf="signupAllowed === false" href="https://joinpeertube.org/en/#getting-started" target="_blank" alt="Link to joinpeertube.org Getting started" title="Click here to see how to get started!" class="create-an-account">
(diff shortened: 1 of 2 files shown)


client/src/app/login/login.component.html
```diff
@@ -17,7 +17,7 @@
           or create an account
         </a>
 
-        <a i18n *ngIf="signupAllowed === false" href="https://joinpeertube.org/en/#getting-started" target="_blank" class="create-an-account">
+        <a i18n *ngIf="signupAllowed === false" href="https://joinpeertube.org/en/#getting-started" target="_blank" alt="Link to joinpeertube.org Getting started" title="Click here to see how to get started!" class="create-an-account">
```
Chocobozzz: `<a>` tags does not have an `alt` attribute in HTML specification

SansPseudoFix: Yep, it's exactly what I was looking for… I don't why I was sure they had! :/ Will fix it.

Chocobozzz:
Thanks!

Issue #574, opened 2018-05-17
Account link width on video page is too much


tcitworld:
* **PeerTube version or URL:** Peertube Beta 4
* **Browser name/version:** Firefox 61

The account link on the video page takes the full width of the container, which can cause misclicks:

![](https://framapic.org/WSbfD7Ntrwa7/UsaZtLaVsI99)

A quick and dirty fix I've found is to change `display: flex` on the link to `display: inline`.

rigelk:
Thanks @tcitworld for spotting the bug!

PR #665, opened 2018-06-15
Fixes account link width on video page is too much #574

sesn:
changed a tag from flex to inline to avoid misclicks in firefox


rigelk:
Thanks @sesn!

Issue #481, opened 2018-04-07
Use FLOSS formats for video transcoding


Sp3r4z:
Hello, I've seen Peertube transcode video after upload (really good idea, for security, standardization…) but, here I am: Peertube use MP4 & AAC, the first is under `patent encumbed` the second is a proprietary format… I think it's not really interesting and satisfying for a FLOSS project promoting proprietary formats.

I think something like: VP9+Opus in a WebM container, is really more appropriate for this project.

Let me know.

PS: I know that x264 and MP4 are more CPU-friendly since low-level CPU instruction sets exist for them ; but for me a FLOSS project should promote free (as speech) formats regardless.

Serkan-devel:
How can mkv-files be played on the web? It seems to be similar to webm

Sp3r4z:
@Serkan-devel what do you mean? Is your comment linked to the purpose of the issue?

Serkan-devel:
mkv is a FLOSS format, that I can only run on VLC but not on the web

Sp3r4z:
Right, I get that. But why talking about .mkv? I propose WebM container with VP9+Opus inside, because it's readable on Internet browsers (like Firefox), which is free (as speech and as beer).

rigelk:
@Serkan-devel Matroska is just a container standard, equivalent to WebM but unfit for web diffusion. 

The encoding time for VP9 is still 5-fold of H.264, and the upcoming AV1 is gonna be 20-fold (I know it's yet to be optimised but it probably won't be any lower than VP9). So there's that.

We could let administrators choose the encoding they want and leave the actual behaviour as default.

Sp3r4z:
@rigelk You're right, for the transcoding task but, here is my point: it requires less storage resources and bandwidth. I know, it's a balance between CPU and storage/bandwith, but for videos and streaming, I think it's more important than CPU usage.

As I said on this #479 in `mediainfo.txt` transcoding seems to pose some problems (30fps but the original was 1fps, worse sound quality… ).
Maybe just improve transcoding parameters, can be something really "in-between".

Chocobozzz:
We use H264/AAC because we use the [videostream](https://github.com/jhiesey/videostream/) package to stream data inside a `video` element. These codecs are the only ones compatible with all web browsers. It's currently not the case for [VP9](https://caniuse.com/#search=vp9).

rigelk:
[...]

PR #679, opened 2018-06-20
FFmpeg performance refactoring

rigelk:
*This PR has grown beyond its initial scope of improving the audio quality, but it keeps focused on this*

It brings:

- a toolbox namespace to analyse the audio stream of file via `ffprobe`.
- a bunch of presets chained together to provide sane defaults and future specifics for audio-focused instances.
- a much better bitrate by default, with support for good VBR if `libfdk_aac` is installed on the system (autodectecting it at runtime), and else tweaking the bitrate based on format/bitrate to save on disk space if possible.
- removing metadata (to the extent of what can be removed by ffmpeg) to prevent video metadata leakage. [beware this is just a convenience, not a guarantee]
- niceness (CPU cycles priority over other applications) tuning to avoid competing with the core PeerTube runtime.


server/helpers/ffmpeg-utils.ts
```diff
@@ -136,3 +134,18 @@ function getVideoFileStream (path: string) {
...
+/**
+ * Standard profile, with variable bitrate audio and faststart.
+ *
+ * As for the audio, quality '5' is the highest and ensures 96-112kbps/channel
+ * See https://trac.ffmpeg.org/wiki/Encode/AAC#fdk_vbr
+ */
+function standard (ffmpeg) {
+  ffmpeg
+    .videoCodec('libx264')
+    .audioCodec('libfdk_aac')
```
Chocobozzz: Maybe we could detect in the checker function that ffmpeg supports libfkd, and if not fallback to default aac encoder?

rigelk: If that's possible, that would be the best option. Any idea on how to do that?

*side note*: when we default back to the default aac encoder, we should also come back to CBR, as its support for VBR is infamous.

Chocobozzz: I think you can use this function https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/checker.ts#L50

See https://github.com/fluent-ffmpeg/node-fluent-ffmpeg#querying-ffmpeg-capabilities

Then store this information somewhere :)

rigelk: Any preference as to where to store the information? Because I think I shouldn't modify the CONFIG object just for that.

Chocobozzz: You can create a function in ffmpeg-utils that create an object with information we want, and cache it for future calls?

yarn.lock
```diff
@@ -1237,7 +1237,7 @@ chalk@^1.0.0, chalk@^1.1.1, chalk@^1.1.3:
     strip-ansi "^3.0.0"
     supports-color "^2.0.0"
 
-chalk@^2.0.0, chalk@^2.0.1, chalk@^2.1.0, chalk@^2.3.0, chalk@^2.3.1:
+chalk@^2.0.0, chalk@^2.0.1, chalk@^2.3.0, chalk@^2.3.1:
```
Chocobozzz: Please checkout this file

server/initializers/checker.ts
```diff
@@ -64,6 +67,16 @@ async function checkFFmpeg (CONFIG: { TRANSCODING: { ENABLED: boolean } }) {
       throw new Error('Unavailable encode codec ' + codec + ' in FFmpeg')
     }
   }
+
+  const optionalEncoders = [ 'libfdk_aac' ]
```
Chocobozzz: I would prefer a dedicated function that fetch available encoders, create an hashmap and cache it. This function would called by checkffmpeg. The main goal is to avoid the for block each time we want to check an encoder is supported + improve efficiency with a hashmap (~ O(1))

server/helpers/ffmpeg-utils.ts
```diff
@@ -136,3 +135,63 @@ function getVideoFileStream (path: string) {
...
+ */
+async function standard (ffmpeg) {
+  let _ffmpeg = ffmpeg
+    .foramt('mp4')
+    .videoCodec('libx264')
+    .outputOption('-level 3.1') // 3.1 is the minimal ressource allocation for our highest supported resolution
+    .outputOption('-movflags faststart')
+
+  if ((await checkFFmpeg(CONFIG)).includes('libfdk_aac')) {
+    _ffmpeg = _ffmpeg
```
Chocobozzz:
```
return _ffmpeg
      .audioCodec('libfdk_aac')
      .audioQuality(5)
```

So you don't need the else anymore

server/helpers/ffmpeg-utils.ts
```diff
@@ -136,3 +135,104 @@ function getVideoFileStream (path: string) {
...
+}
+
+/**
+ * Standard profile, with variable bitrate audio and faststart.
+ *
+ * As for the audio, quality '5' is the highest and ensures 96-112kbps/channel
+ * See https://trac.ffmpeg.org/wiki/Encode/AAC#fdk_vbr
+ */
+async function standard (ffmpeg) {
+  let _bitrate: Number = 384
```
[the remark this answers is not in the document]
rigelk: Ah right :x

server/initializers/checker.ts
```diff
@@ -64,6 +64,28 @@ async function checkFFmpeg (CONFIG: { TRANSCODING: { ENABLED: boolean } }) {
...
+  }
+
+  const Ffmpeg = require('fluent-ffmpeg')
+  const getAvailableEncodersPromise = promisify0(Ffmpeg.getAvailableEncoders)
+  const encoders = await getAvailableEncodersPromise()
+  const optionalEncoders = [ 'libfdk_aac' ]
+
+  for (const encoder of optionalEncoders) {
+    supportedOptionalEncoders.set(encoder,
+      Object.prototype.hasOwnProperty.call(encoders, encoder) ? true : false
```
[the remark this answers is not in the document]
rigelk: not sure how you see it…

server/helpers/ffmpeg-utils.ts
```diff
@@ -136,3 +135,104 @@ function getVideoFileStream (path: string) {
...
+      if (!audio) return _ffmpeg.noAudio()
+      if (selectAudioBitrate[audio['codec_name']]) {
+        _bitrate = selectAudioBitrate[audio['codec_name']](audio['bit_rate'])
+        if (_bitrate === -1) {
+          return _ffmpeg.audioCodec('copy')
+        }
+      }
+    })
+
+  return (await checkFFmpegEncoders()).get('libfdk_aac') ?
```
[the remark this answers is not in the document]
rigelk: `has` doesn't return the value (`true` if the encoder is supported), which is what we are looking for.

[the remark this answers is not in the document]
Chocobozzz: :+1:

rigelk:
@Chocobozzz do you have suggestions of types of unit tests to write in particular? I fail to see where to begin with…

Chocobozzz:
Upload some particular videos (with audio, no audio, high audio bitrate), wait transcoding and check with ffprobe for example if they were encoded as expected

Jorropo:
@rigelk why do you use mp4 ?
Webm make little more smaller file, webm is open, webm use less ram, webm decode/encode better on netbook, light client, ...
The unique advantage of mp4 is, mp4 is older and hardware decoding/encoding is more supported (but on some platform you need to buy a 10€ license and configure the driver for use this license (like with rpi 2, 3)) (you have only 1 platform where hardware decoding is allways disponible (if present), mobile, tablet, ... cause license is included with the product).

rigelk:
@Jorropo we use AVC/AAC for the lack of a better alternative (for now) ; please see #481 for ongoing discussion about that.

Jorropo:
@rigelk I think you should pr name :
- ffmpeg reborn ?
- ffmpeg lot of improvment ?
- ...
Cause there's more than just audio fixing :
- audio fixing
- transcoding rewrite
- niceness
- auto thread

rigelk:
Indeed :grin:

server/helpers/ffmpeg-utils.ts
```diff
@@ -149,3 +188,127 @@ function getVideoFileStream (path: string) {
...
+  _ffmpeg
+    .preset(veryfast)
+    .outputOption('-tune stillimage')
+}
+
+/**
+ * A toolbox to play with audio
+ */
+namespace audio {
+  export const get = (_ffmpeg, pos: number | string = 0) => {
```
rigelk: I'm having trouble dealing with the Promise and get stuck with strange timeouts, even if I catch them in my tests:

`Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.`

I might need help on that one @Chocobozzz @Jorropo

server/helpers/ffmpeg-utils.ts
```diff
@@ -1,10 +1,47 @@
...
+  const isAVC = await getVideoFileStream(path).then(s => {
+    return s['codec_name'] === 'h264' &&
+           s['bit_rate'] < 3000000
+  })
+  const isAAC = await audio.get(ffmpeg, path).then(s => {
+    return s['codec_name'] === 'aac'
+  })
+  console.log('got it until aac')
+  const hasFastStart = await new Promise<boolean>((res, rej) => {
+    ffmpeg(path, { stdoutLines: 0 })
```
rigelk: While this function is quite optional in the PR scope, it is interesting for the future and essentially bears the same problems as the other Promise-based functions (albeit relying on a call to `ffmpeg` instead of `ffprobe`).

It essentiall tries to mimick the command advised [here in the FFmpeg mailing list](http://ffmpeg.org/pipermail/ffmpeg-devel/2018-July/232693.html): `ffmpeg -i subfile,,start,0,end,10000,,:in.mp4 -v 56 2>&1 | grep -e 
"type:'moov'" -e "type:'mdat'" | head -1 | grep moov`.

To debug after the other functions.

server/helpers/ffmpeg-utils.ts
Jorropo: Sorry I'm not very good at typescript/angular (I only got 2 weeks of studing in my school) it's hard for me to understand but I have understand what is namespace but I don't understand how the new Promise at line 234 get res and rej ?
And how the promise is transmitted out side of this function ?

server/helpers/ffmpeg-utils.ts
Jorropo: You should change option to :
You forgot ' for stdoutLines (I'm not totally sure that is a must but it should be here) and niceness is better except if that only for check ffmpeg instalation
```
{ 'niceness': FFMPEG_NICE.THUMBNAIL, 'stdoutLines': 0 }
```

rigelk: @Jorropo unless an attribute has spaces or special characters, there shouldn't be a need to quote it.

As for the niceness, it's not helping the tests, nor is actually interesting in production since this is a short-lived command.

Chocobozzz: I don't see the point of having a function we are not using right now

server/helpers/ffmpeg-utils.ts
```diff
@@ -55,7 +92,7 @@ async function generateImageFromVideoFile (fromPath: string, folder: string, ima
 
   try {
     await new Promise<string>((res, rej) => {
-      ffmpeg(fromPath)
+      ffmpeg(fromPath, { 'niceness': FFMPEG_NICE.THUMBNAIL })
```
Chocobozzz: Why quotes?

client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
```diff
@@ -29,6 +29,7 @@ export class EditCustomConfigComponent extends FormReactive implements OnInit {
     { value: 50 * 1024 * 1024 * 1024, label: '50GB' }
   ]
   transcodingThreadOptions = [
+    { value: 0, label: 'auto (not optimized)' },
```
Chocobozzz: I don't really understand why not optimized? What value will ffmpeg choose?

server/tests/api/videos/video-transcoder.ts
```diff
@@ -91,6 +93,93 @@ describe('Test video transcoding', function () {
     expect(torrent.files[0].path).match(/\.mp4$/)
   })
 
+  it('Should transcode high bit rate mp3 to proper bit rate', async function () {
+    this.timeout(60000)
+
+    console.log('is streamable?')
```
Chocobozzz: Please remove

server/tests/api/videos/video-transcoder.ts
```diff
@@ -91,6 +93,93 @@ describe('Test video transcoding', function () {
...
+    const videoAttributes = {
+      name: 'my super name for server 2',
+      description: 'my super description for server 2',
+      fixture: 'video_short_mp3_256k.mp4'
+    }
+    const fixturePath = buildAbsoluteFixturePath(videoAttributes.fixture)
+    const isFixureStreamable = await isVideoFileStreamable(fixturePath)
+    expect(isFixureStreamable.isAAC).to.be.false
+
+    console.log('is streamable.')
```
Chocobozzz: Please remove

server/tests/api/videos/video-transcoder.ts
```diff
@@ -91,6 +93,93 @@ describe('Test video transcoding', function () {
...
+
+    const video = res.body.data[0]
+    const res2 = await getVideo(servers[1].url, video.id)
+    const videoDetails: VideoDetails = res2.body
+
+    expect(videoDetails.files).to.have.lengthOf(4)
+
+    const path = join(root(), 'test2', 'videos', video.uuid + '-240.mp4')
+    const probe = await audio.get(ffmpeg, path)
+      .catch(err => expect(err, 'Could not read the audio stream of the video file.').to.be.null)
```
Chocobozzz: What's the point of this test? To be sure `audio.get` does not throw an exception?

server/tests/api/videos/video-transcoder.ts
```diff
@@ -91,6 +93,93 @@ describe('Test video transcoding', function () {
     expect(torrent.files[0].path).match(/\.mp4$/)
   })
 
+  it('Should transcode high bit rate mp3 to proper bit rate', async function () {
+    this.timeout(60000)
+
+    console.log('is streamable?')
+
+    const videoAttributes = {
+      name: 'my super name for server 2',
```
Chocobozzz: You don't need a particular name for this test, you can remove it (same for the description and other tests)

server/tests/api/videos/video-transcoder.ts
```diff
@@ -91,6 +93,93 @@ describe('Test video transcoding', function () {
...
+
+    const res = await getVideosList(servers[1].url)
+
+    const video = res.body.data[0]
+    const res2 = await getVideo(servers[1].url, video.id)
+    const videoDetails: VideoDetails = res2.body
+
+    expect(videoDetails.files).to.have.lengthOf(4)
+    const path = join(root(), 'test2', 'videos', video.uuid + '-240.mp4')
+    expect(await audio.get(ffmpeg, path).catch(_ => done())).to.be.undefined
```
Chocobozzz: Same here, why do you catch?

server/tests/api/videos/video-transcoder.ts
rigelk: It was just to help debug.

client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
Jorropo: ffmpeg choose cpu thread count, and this is better to make cpu thread count, but this is better to do cpu thread count -1
Cause on all cpu that isn't high range x86, mid/high range x86_64 or arm v8 if you use all core at 100% your server can become unresponsive durring transcoding.
So all low x86_64, old x86 cpu, arm v6, risc need to keep a core at less than 95% or sometimes all will froze (except transcoding).
(I don't have number but I have a low x86_64 cpu on my laptop so I know, that happend a lot of times, I have to be aware of cpu usage, I have 2c/4t and if all the 4t become fully used my laptop freeze, and it become faster than you think)

But program is niced so this shouldn't appear ?
Yes, but nice require a little bit of cpu time to work but if a process take 100% how it can execute (there is some security for that not happend on last x86, and mid/high x86_64, and arm v8 but not for other (maybe but they didn't work very well (cause this things require a complex cpu))).

server/tests/api/videos/video-transcoder.ts
```diff
@@ -91,6 +93,76 @@ describe('Test video transcoding', function () {
...
+
+    const res = await getVideosList(servers[1].url)
+
+    const video = res.body.data[0]
+    const res2 = await getVideo(servers[1].url, video.id)
+    const videoDetails: VideoDetails = res2.body
+
+    expect(videoDetails.files).to.have.lengthOf(4)
+    const path = join(root(), 'test2', 'videos', video.uuid + '-240.mp4')
+    expect(await audio.get(ffmpeg, path)).to.be.undefined
```
rigelk: It is still giving me off a 

```
(node:20610) UnhandledPromiseRejectionWarning: AssertionError: expected { Object (index, codec_name, ...) } to be undefined
    at Context.<anonymous> (/home/rigelk/github/peertube/server/tests/api/videos/video-transcoder.ts:139:48)
    at fulfilled (/home/rigelk/github/peertube/server/tests/api/videos/video-transcoder.ts:5:58)
    at <anonymous>
(node:20610) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:20610) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

server/helpers/ffmpeg-utils.ts
```diff
@@ -149,3 +151,127 @@ function getVideoFileStream (path: string) {
...
+    // we make it consider the first input only
+    // if you pass a file path to pos, then ffprobe acts on that file directly
+    return new Promise<any>((res, rej) => {
+      _ffmpeg.ffprobe(pos, (err,data) => {
+        if (err) return rej(err)
+
+        if ('streams' in data) {
+          return res(data['streams'].find(stream => stream['codec_type'] === 'audio'))
+        } else {
+          rej(undefined)
```
Chocobozzz: `return rej(new Error('No audio stream found'))` or `return res(undefined)` depending on how you want to handle this case

server/helpers/ffmpeg-utils.ts
```diff
@@ -149,3 +151,127 @@ function getVideoFileStream (path: string) {
...
+    .videoCodec('libx264')
+    .outputOption('-level 3.1') // 3.1 is the minimal ressource allocation for our highest supported resolution
+    .outputOption('-b_strategy 1') // NOTE: b-strategy 1 - heuristic algorythm, 16 is optimal B-frames for it
+    .outputOption('-bf 16') // NOTE: Why 16: https://github.com/Chocobozzz/PeerTube/pull/774. b-strategy 2 -> B-frames<16
+    .outputOption('-map_metadata -1') // strip all metadata
+    .outputOption('-movflags faststart')
+  const _audio = await audio.get(localFfmpeg)
+
+  if (!_audio) {
+    console.log('no audio.')
```
Chocobozzz: `logger.debug` would be interesting (only if you are able to retrieve the file uuid or path)

server/tests/api/videos/video-transcoder.ts
Chocobozzz:
```
const streams = await audio.get(ffmpeg, path)
console.log(streams)
expect(stream).to.be.undefined
```

client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
rigelk: as for the label, wouldn't a "auto via ffmpeg" or "let ffmpeg choose" be better? This way, users would refer to the ffmpeg documentation.

server/helpers/ffmpeg-utils.ts
rigelk: I can get the absolute path via ffprobe. Anything else requires more than an independent preset.

Chocobozzz: Absolute path should be fine

client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
Chocobozzz: `Auto (via ffmpeg)` :+1:

client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
```diff
@@ -29,6 +29,7 @@ export class EditCustomConfigComponent extends FormReactive implements OnInit {
     { value: 50 * 1024 * 1024 * 1024, label: '50GB' }
   ]
   transcodingThreadOptions = [
+    { value: 0, label: 'auto (via ffmpeg)' },
```
Chocobozzz: Auto (via ffmpeg)

server/helpers/ffmpeg-utils.ts
```diff
@@ -149,3 +151,127 @@ function getVideoFileStream (path: string) {
...
+    // without position, ffprobe considers the last input only
+    // we make it consider the first input only
+    // if you pass a file path to pos, then ffprobe acts on that file directly
+    return new Promise<any>((res, rej) => {
+      _ffmpeg.ffprobe(pos, (err,data) => {
+        if (err) return rej(err)
+
+        if ('streams' in data) {
+          const audioStream = data['streams'].find(stream => stream['codec_type'] === 'audio')
+          if (audioStream) return res(audioStream)
```
Chocobozzz: Maybe returning `{ absolutePath: string, audioStreams?: any }` would be more consistent

server/helpers/ffmpeg-utils.ts
```diff
@@ -149,3 +151,127 @@ function getVideoFileStream (path: string) {
...
+  }
+
+  export namespace bitrate {
+    export const baseKbitrate = 384
+
+    const toBits = (kbits: number): number => { return kbits * 8000 }
+
+    export const aac = (bitrate: number): number => {
+      switch (true) {
+      case bitrate > toBits(384):
```
Chocobozzz: `case bitrate > toBits(baseKbitrate)`?

server/helpers/ffmpeg-utils.ts
```diff
@@ -149,3 +151,127 @@ function getVideoFileStream (path: string) {
...
+        return baseKbitrate
+      default:
+        return -1 // we interpret it as a signal to copy the audio stream as is
```
Chocobozzz: Like this kind of comment :+1:

server/helpers/ffmpeg-utils.ts
```diff
@@ -149,3 +151,127 @@ function getVideoFileStream (path: string) {
...
+      }
+    }
+
+    export const mp3 = (bitrate: number): number => {
+      switch (true) {
+      case bitrate <= toBits(192):
```
Chocobozzz: Is there particular reasons why we don't keep 192?

server/helpers/ffmpeg-utils.ts
rigelk: :+1:

server/helpers/ffmpeg-utils.ts
rigelk: a 192kbit/sec mp3 doesn't hold as much information as a 192kbit/sec aac. That's why, when using aac, we can go to lower kbit/sec

Chocobozzz: :+1: Could you add this as comment?

rigelk:
Thanks for your time and code @Chocobozzz @Jorropo

PR #681, opened 2018-06-20
[WIP] i18n: Add Polish translation

m4sk1n:


Chocobozzz:
Hello @m4sk1n, thanks but i18n is not ready in peertube. But if you want to begin to translates strings (that might change in the future), you can use our [translation platform](https://trad.framasoft.org/zanata/iteration/view/peertube/develop/languages/pl-PL?dswid=8115) (you need to create an account and join the polish language team first)

m4sk1n:
There’s no Polish language team, at least there wasn’t one an hour ago…

Chocobozzz:
We just created it :) https://trad.framasoft.org/zanata/language/view/pl-PL?dswid=6854

PR #705, opened 2018-06-22
Fix #639 providing magnet URI in player and download modal

### LINKED ISSUE #639: Magnet links for direct download
MediaGoblin and Vimeo (sometimes) have a button for downloading the video instead of streaming it, which is helpful for people with older computers, or patchy net connections, or who want to download it while they have free wifi, for watching at a time when they don't. It seems me that in keeping with the torrent-based streaming method PeerTube, the way to do direct downloads would be to provide a magnet link and a web seed (or 3), allowing users to download a video with their BitTorrent client. This ways, as with the WebTorrent streaming, users aren't just putting a strain on the server when they download, they can also contribute back to helping other users download popular videos.

If and when desktop BitTorrent clients start to become WebTorrent compatible, this would be a simple way for a user to help stream a video they really like to the PeerTube network too. They would download the video with the magnet link, and then seed it as normal, helping others user watch it on PeerTube *or* download it via the magnet link.

rigelk:


Chocobozzz:
Nice work thanks @rigelk!

PR #712, opened 2018-06-24
Specify Tor Browser instead of just Tor in the recommendation

### LINKED ISSUE #676: Concern about "About" page telling people to use Tor.
The "About" page contains this blurb:

> If you want to keep your public IP address private, please use a VPN or Tor.

However, Tor project [explicitly tells people to not use BitTorrent over it](https://www.torproject.org/docs/faq.html.en#FileSharing):

> File sharing (peer-to-peer/P2P) is widely unwanted in the Tor network, and exit nodes are configured to block file sharing traffic by default. Tor is not really designed for it, and file sharing through Tor slows down everyone's browsing. Also, Bittorrent over Tor is not anonymous! 

Should the "or Tor" part be removed?

rigelk:
Since Tor Browser disables WebRTC, the fallback to HTTP streaming will respect
Tor's terms of service. Anyway any recommendation to use Tor should point to
the Tor Browser, as it provides sane and secure defaults to use Tor.

Fixes #676

### DIFF
--- a/client/src/app/about/about.component.html
+++ b/client/src/app/about/about.component.html
@@ -44,7 +44,7 @@
 
     <p i18n>
       PeerTube uses the BitTorrent protocol to share bandwidth between users. It implies that your public IP address is stored in the public BitTorrent tracker of the video PeerTube instance as long as you're watching the video.
-      If you want to keep your public IP address private, please use a VPN or Tor.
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Please do not edit i18n files (source or target) because they will be overwritten

PR #720, opened 2018-06-25
Filter by category

PhieF:
Just added another filter : category

I needed it to display only musics on exode.me (not yet deployed) and I was wondering if it could be pushed to main peertube repo

Just a question by the way
do you know a way to force reload a component when parameters change : for example 
from
videos/trending

to 
videos/trending?category=1

I will have to refresh the page or go to another component in between


Chocobozzz:
Nice work! Would you want to add tests too? https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/videos/single-server.ts#L380

For the component reload, you need to subscribe to parameters changes in abstract video list but i can do it it you want

PhieF:
OK I will try to do it by myself, thanks !

PhieF:
ok ! thanks !

PR #735, opened 2018-06-27
 Added some translation to fr_FR 

Jorropo:
I know I should use zanata but this sentence isn't on zanata.


Chocobozzz:
Hi, thanks for the PR but the files you modified are automatically generated. I updated strings on zanata so you should be able to report your translation over there! :)

I there are missing translation strings, do not hesitate to create an issue so I fix that.

PR #741, opened 2018-06-28
Open CORS to allow in-browser apps to communicate w/ PeerTube instances

rezonant:
A small change to open up CORS access to the PeerTube instance REST API for all origins.
The net result is that PeerTube would begin including the following headers on all API calls except for the torrents and webseeds endpoints (which have their own CORS behavior):

```
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: Retry-After
```

PeerTube was already enabling CORS for test/development instances, turning it on for real instances will allow in-browser Javascript apps to access PeerTube instances without an intermediary server. This may also open up opportunities for optimization/resiliency within the PeerTube client SPA, as there is the potential for the client of one instance to be able to communicate with the REST API of another, should that be useful.

Security Impact: Minimal. The PeerTube REST API uses bearer tokens via the `Authorization` header to pass the auth token for the currently logged in user, as opposed to an auth cookie. This is the standard these days, and since a malicious actor must run JS on the instance's origin in order to gain access to the domain's `localStorage` data in order to correctly populate the `Authorization` header in the first place, basic XSS/XHR attacks are still prevented even with this turned on.

If PeerTube decides to enable SSR, `localStorage` no longer works as an authentication storage medium (let me know if it's unclear why not and I can explain further), and in my experience it is best to switch back to cookie storage. If this were to happen, it's important that the API not be changed to accept an authentication cookie. Continue using an explicitly-set `Authorization` header regardless of whether the app is running on client or server and do not accept cookies to authenticate users with the API. Sticking to `Authorization` will help to avoid exposing the potential for XSS in the future.


rezonant:
Removed semicolon to satisfy linter rule.
Let it be known that I believe a linter rule that enforces that you do not use semicolons when not necessary is a very bad linter rule.

Chocobozzz:
In this PR you enabled CORS on all server routes, including feed, tracker or static routes. If you just want to enable CORS on API routes, please add the middleware on `app.use(apiRoute, apiRouter)`

> Let it be known that I believe a linter rule that enforces that you do not use semicolons when not necessary is a very bad linter rule.

Devs have always a strong opinion on linters :) FYI we use typescript adapted [standardjs rules](https://standardjs.com/).

rezonant:
Ahh, I see. OK I'll have to dig into these routes to see if there's any use to CORSing them. If not, I'll push a revision to enable only for API routes.

DavidLibeau:
Is `static` routes are also included (ex: `https://videonaute.fr/static/thumbnails/87b8af73-ec0a-4b3b-8fb5-eff0f1b82061.jpg`) ?
Thanks

rezonant:
It does not. You can still use those resources with HTML though (ie img or background-image) or any other place where an opaque response is suitable.

DavidLibeau:
I cannot use it with my JS app (using Aframe.io). I'll open an issue then.

rezonant:
Yes applications which need access to the pixels, including putting them in canvas, webgl, or webvr would require non-opaque responses.

PR #742, opened 2018-06-28
Replace angular-cli patch with something less risky

rezonant:
As it sounds, this replaces the super-hairy angular patch that Peertube has now with something that doesn't need to modify any external modules.

It uses Typescript's path resolution settings available in `tsconfig.json`'s `paths` section. Using `paths` I remap all of the modules that are not available on the web. Changeset includes a trivial `global` polyfill and the polyfills PeerTube depends on.

All polyfills brought in with this PR are the ones Webpack uses:
https://github.com/webpack/node-libs-browser#readme

Status: In testing. A video upload and playback smoke test looks good.


rezonant:
I was on the wrong branch :-S -- The production build succeeds with this solution.

rezonant:
did a video upload/playback test and things look to be working.

all polyfills in this PR are the ones used by webpack:
https://github.com/webpack/node-libs-browser#readme

Chocobozzz:
Thanks @rezonant

PR #748, opened 2018-06-28
Improve left menu

ebrehault:
- Add a transition when opening/closing the left menu
- Suppress the scroll to top when opening/closing the left menu.

Note: regarding the scroll to top, I did see the commit about it https://github.com/Chocobozzz/PeerTube/commit/a01f107bc436250706d4bc765f45335ee15b8e80#diff-30532d215cbc6539c241db5bdfc9b60eR60 
It is not clear to me how it improved the mobile version, but it is clearly annoying for the desktop one.
Still, I assume it has been done for a good reason, so I would be happy to talk about it in order to find the best solution on this matter.

### DIFF
--- a/client/src/app/app.component.html
+++ b/client/src/app/app.component.html
@@ -18,7 +18,7 @@
   </div>
 
   <div class="sub-header-container">
-    <div *ngIf="isMenuDisplayed" class="title-menu-left">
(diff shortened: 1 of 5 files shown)


ebrehault:
rebase needed

PR #749, opened 2018-06-28
Improve left menu

ebrehault:
- Add a transition when opening/closing the left menu
- Suppress the scroll to top when opening/closing the left menu.

Note: regarding the scroll to top, I did see the commit about it https://github.com/Chocobozzz/PeerTube/commit/a01f107bc436250706d4bc765f45335ee15b8e80#diff-30532d215cbc6539c241db5bdfc9b60eR60 
It is not clear to me how it improved the mobile version, but it is clearly annoying for the desktop one.
Still, I assume it has been done for a good reason, so I would be happy to talk about it in order to find the best solution on this matter.

### DIFF
--- a/client/src/app/app.component.ts
+++ b/client/src/app/app.component.ts
@@ -115,7 +115,6 @@ export class AppComponent implements OnInit {
   }
 
   toggleMenu () {
-    window.scrollTo(0, 0)
     this.isMenuDisplayed = !this.isMenuDisplayed
   }
 }
--- a/client/src/sass/application.scss
+++ b/client/src/sass/application.scss
@@ -286,11 +286,9 @@ table {
   }
 }
 
-// On small screen, menu is absolute
 @media screen and (max-width: 600px) {
   .menu-wrapper {
     width: 100% !important;
-    position: absolute !important;
     z-index: 10000;
   }
 


rezonant:
Hrm, normally that sort of thing is done to handle devices that don't support fixed positioning, but it seems like the only way to activate the menu is at the top of the view, right?

ebrehault:
Oh right, the menu is positioned in absolute for small devices! https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/sass/application.scss#L293

Do we really need that? `position: fixed;` seems to be widely supported https://caniuse.com/#feat=css-fixed

rezonant:
Well I mean, if you can only open the menu from the top of the scroll area anyway, is it really helping?

ebrehault:
oh the menu button is in a fixed top position, so the user sees it permanently even if if he scrolled down, so yes, it does help (I think)

ebrehault:
I have amended my commit to remove the `position: absolute` on small devices

Chocobozzz:
Thanks for this PR @ebrehault!

The way you handle the menu without the scroll is much better!

But I don't really like the transition. We see some icons disappear, texts move... It's not pleasant, and could add unnecessary overhead on mobile devices.

ebrehault:
ok, I'll remove it then, no problem

ebrehault:
@Chocobozzz done

rezonant:
As to the menu on the top being fixed: If mo I'll doesn't support fixed positioning, it won't be fixed...

Chocobozzz:
Thanks @ebrehault 

@rezonant I never seen a mobile device that does not support fixed positioning? Could you elaborate?

PR #758, opened 2018-06-30
filter to Cap the maximum resolution to the screen resolution

### LINKED ISSUE #746: Cap the maximum resolution to the screen resolution
*note*: adapted from https://framacolibri.org/t/choix-automatique-de-la-resolution-video/3204

Right now the resolution is selected automatically in order to fit the maximum bandwidth the connection allows. However it doesn't make sense to watch a 1080p video on a 720p screen, for instance.

Adding a check detecting the current window size and adapting the resolution if necessary would fit the use and allow to avoid consuming more bandwidth than necessary.

ns23:
fixes #746


rezonant:
and this check runs every 3s, a dom width check every 3s is nominal. looks good from here

Chocobozzz:
Thanks @ns23 for the PR, and @rezonant for the review!

PR #761, opened 2018-07-01
Update production.md

baileyrd:
When I followed the production guide, one issue that I came across dealt with the fact that the default configuration file was still linked under sites-enabled.  Updated the production document to add this.


rigelk:
@baileyrd wouldn't it be beyond the scope of just *adding a service to Nginx*? We kind of assume Nginx is more or less setup already.

Chocobozzz:
Thanks for the PR @baileyrd but as @rigelk said, we expect nginx is setup already.

PR #764, opened 2018-07-02
make edits in documents

melongbob:
FAQ.md
- "the name" -> "a name"
- "re install" -> "reinstall"
- Changed the sentence structure to accomodate the flow of "objective -> method to achieve the objective -> reason why the objective is desired", rather than "how we are going to do it -> what we are trying to do -> why it is important"

README.md
- "It's the reason" -> "That is why" 
- "in it" -> "on it"
- "This program is free software" -> "This is a free software": reduce redundancy (program and software), and introduce a proper article
- "redistribute it and/or modify it" -> "redistribute and/or modify it"
- "version 3 of the License, or (at your option) any later version" -> "version 3 of the License or (at your option) later": reduce use of two "version" in one sentence (version 3 or later implies any version higher than 3 is acceptable).

--------------------------------------
Last three changes in README.md was reverted

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ Here are some simple schemes:
 
 Copyright (C) 2018 PeerTube Contributors
 
-This program is free software: you can redistribute and/or modify
+This program is free software: you can redistribute it and/or modify
 it under the terms of the GNU Affero General Public License as published
 by the Free Software Foundation, either version 3 of the License or
 (at your option) later.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -177,10 +177,10 @@ Here are some simple schemes:
 
 Copyright (C) 2018 PeerTube Contributors
 
-This program is free software: you can redistribute it and/or modify
+This is a free software: you can redistribute and/or modify
```
rigelk: This is actually from the official GNU text: https://www.gnu.org/licenses/gpl-3.0.en.html

melongbob: Yes, of course. I should have realized it was from the official GNU website.
Thank you :)

PR #770, opened 2018-07-05
Add a precision in the FAQ regarding why we are mainly using GitHub for now

rigelk:
The question was legitimatly raised in the latest AMA on r/freesoftware and we
should explain why most of the project's activity is still happening on GitHub.


tcitworld:
@rigelk I would have added a paragraph on the fact than most developers already have a Github account, that it's an issue with Gitlab until it's federated. :)

Chocobozzz:
Federation for the win \o/

rezonant:
https://gitlab.com/gitlab-org/gitlab-ee/issues/4517

https://github.com/git-federation/gitpub

Issue #468, opened 2018-04-03
Controllable embedded player


henke37:
The ability to control the embedded player from the hosting page.

The [youtube player api](https://developers.google.com/youtube/iframe_api_reference) is a good example to follow, keeping the same function names allows for easy migration of legacy code.

**EDIT:** Additional features https://github.com/Chocobozzz/PeerTube/issues/468#issuecomment-466179176

nvivant:
Yes. For example, we would love to have a feature to start the embedded video at a given time (https://peertube.xyz/videos/watch/c20a3f35-d1b3-4865-8dff-25593151ffe2?start=120 to start playing it at 120 seconds).

henke37:
While it would be nice to be able to specify the start time like that, this is about more dynamic control using javascript.

Chocobozzz:
@nvivant See https://github.com/Chocobozzz/PeerTube/commit/f37bad639b36d35c29a464dc52123a1e7c9cd28a for your start parameter

fredsif:
Nice!
I agree with henke37; dynamic control of the player akin to the Youtube API for iframe would be a cool addition to the platform.

nahuely:
I agree with the playback feature, mostly for API hooks of seeking and pause for the embedded player

nahuely:
I meant dynamic control, sorry coz the mistake, cheers

rezonant:
Ah ha, yep this covers what I was thinking. Is anyone already working on this?

rigelk:
@rezonant not that I know of.

raphaelbastide:
What kind of options do users already have for iframe parameters? Is it possible to hide the title or at least the subtitle? If it is not the case should I open a new issue?
Youtube et al have commercial reasons to limit the customization of video players. I think Peertube can really interest people wanting advanced customization options, e.g a minimal player interface. What do you think?

rezonant:
@raphaelbastide 
The current embed options are `autoplay` which defaults to false, and `start` which can be set to a specific time to start playback at some point within the movie.

My PR adds `muted`, `loop`, `controls`, and `api` which enables the API over `postMessage`

Chocobozzz:
@raphaelbastide It's a very interesting idea. Would you want to create a separate issue and list the things you would like to customize?

raphaelbastide:
@Chocobozzz Here it is https://github.com/Chocobozzz/PeerTube/issues/790

mrx1983:
[...]

PR #776, opened 2018-07-06
[WIP] Ability to programmatically control embeds

rezonant:
## Synopsis
Uses [JSChannel](https://github.com/mozilla/jschannel) to provide RPC semantics on top of `window.postMessage` enabling a website embedding a PeerTube video to programmatically control it. Currently covers commands including `play`, `pause`, `seek`, `setVolume`, `setResolution` and `setPlaybackRate`. Also lets you query these things with `getVolume`, `getResolution`, `getResolutions`, `getPlaybackRate` and `getPlaybackRates`. 

Also emits a number of events including `pause`, `play`, `playbackStatusUpdate` (which continuously sends back playback position, whether paused/playing etc) and `playbackStatusChange` which fires only when the pause/play state changes. Also provides `resolutionUpdate` which returns a set of resolution objects representing the resolutions PeerTube offers. 

## Build changes

This adds a new entrypoint to the Webpack config (which currently just builds the embed package). The entrypoint is located at `client/src/standalone/player`. This endpoint exports a `PeerTubePlayer` class which a convenient frontend to the JSChannel that is established by the embed when the API is enabled.

## Usage
To enable the API, include `api=1` in the query parameters of the embed object, import the PeerTube player API JS module and:

```typescript
let player = new PeerTubePlayer(playbackElement, optionalScopeHere)
await player.ready
// ready to use!
```

## Integrated Testing

You can also change any PeerTube embed URL to replace `/embed/` with `/test-embed/` and visit that URL in the browser. You get a simple test page that lets you demo the various ways you can control a PeerTube embed without the builtin controls.

## Misc changes

To accomplish it's goal, this PR also refactors the embed to be a Typescript/ES6 class. It also extends the available embed attributes to include `loop`, `controls`, and `muted` which can also be used without the API. 

## Status
* Mostly complete, ready for feedback. 
* Needs documentation
* Examine player packaging (CommonJS, UMD, ES6, web, etc) and distribution (NPM, etc)


rezonant:
Addresses #468

client/src/standalone/player/definitions.ts
```diff
@@ -0,0 +1,9 @@
+
+export interface EventHandler<T> {
+    (ev : T) : void
+}
+
+export type PlayerEventType = 
+    'playbackStateChanged'
+    | 'ready'
```
rezonant: update to match the current set of events

client/src/standalone/player/events.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public addListener<T>(name : string, handler : EventHandler<T>) {
+        if (!this.eventRegistrations[name]) {
+            console.warn(`PeerTube: addEventListener(): The event '${name}' is not supported`)
+            return false;
+        }
+
+        this.eventRegistrations[name].registrations.push(handler);
+        return true;
```
rezonant: semicolons everywhere

client/src/standalone/player/player.ts
```diff
@@ -0,0 +1,133 @@
...
+    'playbackStatusChange',
+    'resolutionUpdate'
+]
+
+export interface PeerTubeResolution {
+    id : any
+    label : string
+    src : string
+    active : boolean
+}
```
rezonant: should go in `definitions.ts`

client/src/standalone/player/player.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        this.constructChannel()
+        this.prepareToBeReady()
+    }
+
+    private eventRegistrar : EventRegistrar = new EventRegistrar()
+    private channel : Channel.MessagingChannel
+    private currentPosition = 0
+
+    public destroy() {
```
rezonant: more method-level docs

client/src/standalone/player/player.ts
```diff
@@ -0,0 +1,133 @@
...
+        return await this.sendMessage<void, number>('getPlaybackRate')
+    }
+
+    async setPlaybackRate(rate : number) {
+        await this.sendMessage('setPlaybackRate', rate)
+    }
+}
+
+// put it on the window as well as the export
+window['PeerTubePlayer'] = PeerTubePlayer;
```
rezonant: a rather naive implementation, what about UMD, etc

client/src/assets/player/peertube-player.ts
```diff
@@ -29,10 +29,15 @@ function getVideojsOptions (options: {
...
   startTime: number
-  theaterMode: boolean
+  theaterMode: boolean,
+  controls?: boolean,
+  muted?: boolean,
+  loop?: boolean
 }) {
   const videojsOptions = {
-    controls: true,
+    controls: options.controls !== undefined ? options.controls : true,
```
dedesite: This can be manage via ES6 parameters default value.

ie : `controls: boolean = true,`

client/src/standalone/player/player.ts
```diff
@@ -0,0 +1,182 @@
...
+                error: error => reject(error)
+            })
+        })
+    }
+
+    /**
+     * Tell the embed to start/resume playback
+     */
+    async play() {
+        await this.sendMessage('play')
```
Chocobozzz:
```
play () {
  return this.sendMessage('play')
}
```

client/src/standalone/videos/embed.ts
```diff
@@ -17,100 +17,301 @@ import 'core-js/es6/set'
...
+      this.channel.notify({ method: 'playbackStatusChange', params: 'paused' })
+    })
+
+    // PeerTube specific capabilities
+
+    this.player.peertube().on('autoResolutionUpdate', () => this.loadResolutions())
+    this.player.peertube().on('videoFileUpdate', () => this.loadResolutions())
+  }
+
+  private resolutions = null
```
Chocobozzz: miss typing

client/src/standalone/videos/embed.ts
```diff
@@ -17,100 +17,301 @@ import 'core-js/es6/set'
...
+
+  private loadResolutions() {
+    let resolutions = []
```
Chocobozzz: miss typing

client/src/standalone/player/player.ts
rezonant: I think it is better to always await instead of silently passing the promise back. This prevents a later modification like:

```
play() {
   try {
      return this.sendMessage('play');
   } catch (e) {
      // handle exception, but this will never ever fire
   }
}

File: client/src/assets/player/peertube-player.ts (hunk 6 above)
Comment (rezonant) [reply client/src/assets/player/peertube-player.ts]: I'm not sure what you mean. This is explicitly defaulting `controls` to be `true` only when `options.controls` is exactly undefined. It is not the same to just set it to true here. Also the syntax you provide is for defining classes but this is a raw object.

File: client/src/standalone/videos/embed.ts (hunk 8 above)
Comment (rezonant) [reply client/src/standalone/videos/embed.ts]: good catch :-D

File: client/src/standalone/player/player.ts (hunk 7 above)
Comment (Chocobozzz) [reply client/src/standalone/player/player.ts]: don't know why you would do such thing, we only `try` when we know the method implementation or when we do an `await` inside the block, but whatever it's not very important

Comment (rezonant) [reply client/src/standalone/player/player.ts]: Say for instance we decide we need to throw an exception out of the `play()` that is inside the iframe. It triggers the `error` handler on JS channel and we throw it--- but say we want to have a proper error class to represent said type. But the object coming out of the channel will always be a plain object. So perhaps we do something like:

```
play() {
   try {
      return this.sendMessage('play');
   } catch (e) {
      if (e.type == 'unplayable') {
           throw new UnplayableError(e);
      }

      throw e;
   }
}
```

Now, the dev has made the mistake I am talking about with a concrete use case. Does this make more sense?

Comment (Chocobozzz) [reply client/src/standalone/player/player.ts]: :+1:

File: client/src/assets/player/peertube-player.ts (hunk 6 above)
Comment (dedesite) [reply client/src/assets/player/peertube-player.ts]: Sorry, it could be written like that : `controls = true,`.

For me, that's why default parameters values is for, if controls is undefined or not specified (so undefined), it will be set to true. Otherwise, it'll take whatever value you pass. It's just syntactic sugar to do the same thing.

**Comment by rezonant** [comment]
Screenshot of `test-embed` tool for convenience:
embed-playground-2

**Comment by rezonant** [comment]
I think this is pretty much ready, except for some packaging work. I'm thinking that distributing the `player.js` that goes with the Embed API via NPM makes sense for ease of use in projects consuming the API, does this sound good?

File: client/src/assets/player/peertube-player.ts (hunk 6 above)
Comment (rezonant) [reply client/src/assets/player/peertube-player.ts]: No, if I wrote `controls = true` it would be permanently set to `true` regardless of the intention of the caller to this function, (as well as give me a syntax error since this is an object literal, not a class definition). Are you reading the red line as the added line? The line as proposed in this PR is: 

```typescript
    controls: options.controls !== undefined ? options.controls : true,
```

When the incoming `options` object has no `controls` property defined, the value of the `controls` key in the resulting `videoJsOptions` becomes the default value that I specify, `true`. But when it _is_ defined in the incoming `options` object, _even if it is false_, it will not be overridden to be `true`, because that would sort of defeat the purpose of an option.

rigelk:
@rezonant sounds good to me, at least.

client/src/assets/player/peertube-player.ts
dedesite: Sorry, I misread the code and write to quickly (and my suggestions were totally wrong). I didn't saw that it was not parameters destructuring but type annotation (though it was obvious because you use `options.contol`). I agree with you it is outside the scope of this PR.

,

Chocobozzz:
It's hard to review , but I saw some linting issues (that travis does not catch). Let's merge this PR and I'll fix them directly in develop

Thanks again @rezonant

PR #777, opened 2018-07-06
[Issue 641] Add 'My Videos' shortcut to profile dropdown 

LeoMouyna:
Currently it's in the profile dropdown as suggested by @Chocobozzz but if it doesn't fit with the need we can add some links on left navbar bellow the profile name.

#[WIP] Show video options from list views for owner
Can quickly access to delete and update options for owned video via video list view, but options appear at bottom left of the page (you have to scroll down to see its). I might need some help to understand where dropdownMenu position is set in scss files.

FAQ and README are older than the ones in current dev branch.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -14,7 +14,6 @@
 - [I want to change my domain name, how can I do that?](#i-want-to-change-my-domain-name-how-can-i-do-that)
 - [Should I have a big server to run PeerTube?](#should-i-have-a-big-server-to-run-peertube)
 - [Can I seed videos with my classic BitTorrent client (Transmission, rTorrent...)?](#can-i-seed-videos-with-my-classic-bittorrent-client-transmission-rtorrent)
-- [Why host on GitHub and Framagit?](#why-host-on-github-and-framagit)
 - [Are you going to use the Steem blockchain?](#are-you-going-to-use-the-steem-blockchain)
 - [Are you going to support advertisements?](#are-you-going-to-support-advertisements)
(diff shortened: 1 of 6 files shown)


Chocobozzz:
Hi @LeoMouyna! Thanks for the PR!

But I'm not sure that adding a dropdown to every video miniature owned by the user is a good idea:
 * It adds code and links in another place, while we already have these "quick links" in the watch page
 * It adds a dropdown element/ng if check to every video miniature, and it could be heavy since can have many video miniatures in the page

For the README and the FAQ, you'll have to checkout them because your commit is currently removing @rigelk's modifications.

LeoMouyna:
Hi @Chocobozzz !

Thanks for your feedback !
"Browsing your private videos the same way you browse public videos will be a nice enhancement !" Attracted my attention and adding the dropdown was the first thing I imagine to keep all my videos page features. If a checkbox is added to the left top video miniature corner besides this dropdown menu, it may match with the need.
I didn't want to create a new specific video miniature for my videos component so I worked on the component used in public presentation.
But as you said, ng if check could be heavy if many video miniatures are displayed.

So I should remove the dropdown to the PR to match with the first need: Direct access to your private video from the main left pane. I will also update readme and faq files.
But what can I do to browse private videos the same way public videos are browsed ? Create a specific private video miniature with dropdown and checkbox ? If you have other suggestions let me know !

Chocobozzz:
Maybe this feature could be covered by the advanced search on which we would specify the author?

Chocobozzz:
Closing due to inactivity, feel free to ask for reopening when you'll have time!

PR #795, opened 2018-07-11
rewording download option and reorder options

rigelk:
".torrent" is a common name for users familiar with BitTorrent, and less
impressive than "metainfo" or no information as opposed to the magnet option,
less commonly known but rather just clicked on most sites.


Chocobozzz:
:+1:

PR #812, opened 2018-07-15
Selective route permission to use embeds, fixes #322 in a better way

rigelk:
note: this time it should work :sweat_smile:

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -52,7 +52,25 @@ app.set('trust proxy', CONFIG.TRUST_PROXY)
 // Security middlewares
 app.use(helmet({
   frameguard: {
-    action: 'deny'
+    action: 'deny' // we only allow it for /videos/embed, see server/controllers/client.ts
+  },
+  dnsPrefetchControl: {
+    allow: true
+  },
+  contentSecurityPolicy: {
+    directives: {
+      fontSrc: ["'self'"],
+      frameSrc: ["'none'"],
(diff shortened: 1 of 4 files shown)


support/nginx/peertube
```diff
@@ -38,6 +38,8 @@ server {
   # resolver_timeout 5s;
 
   add_header Strict-Transport-Security "max-age=63072000; includeSubDomains; preload";
+  add_header X-Frame-Options DENY;
+  add_header X-Frame-Options https://peertube.example.com/videos/embed;
```
rezonant: If people forget to update this, the default position will be to deny embeds. I think the default should be to allow embeds. This header should be applied dynamically in the codebase not in the server configuration.

rigelk: Should be corrected.

server.ts
```diff
@@ -124,7 +132,10 @@ const apiRoute = '/api/' + API_VERSION
 app.use(apiRoute, apiRouter)
 
 // Services (oembed...)
-app.use('/services', servicesRouter)
+app.use('/services', servicesRouter, helmet.frameguard({
+  action: 'allow-from',
+  domain: CONFIG.WEBSERVER.URL
+}))
```
rezonant: The regular HTML embed is served from /embed, this is the oembed part

rigelk:
@rezonant I've added a CSP since helmet makes adding one quite easy. Could you have a look at it?

PR #820, opened 2018-07-18
Adding initial support for nodeinfo

### LINKED ISSUE #528: Expose node information following the nodeinfo convention
Following some discussion in Pleroma about reporting basic info about a server, there has been some interest for the [nodeinfo](https://github.com/jhass/nodeinfo/blob/master/schemas/2.0/example.json) standard. They seem to be already used by diaspora*, Friendica and Hubzilla, to name a few. 

While I know we already have reporting of such info, maybe it could help play nice with other server implementations at some point?

rigelk:
closes #528

### DIFF
--- a/server/controllers/static.ts
+++ b/server/controllers/static.ts
@@ -1,11 +1,16 @@
 import * as cors from 'cors'
 import * as express from 'express'
-import { CONFIG, STATIC_DOWNLOAD_PATHS, STATIC_MAX_AGE, STATIC_PATHS } from '../initializers'
+import { CONFIG, STATIC_DOWNLOAD_PATHS, STATIC_MAX_AGE, STATIC_PATHS, ROUTE_CACHE_LIFETIME } from '../initializers'
 import { VideosPreviewCache } from '../lib/cache'
+import { cache } from '../middlewares/cache'
 import { asyncMiddleware, videosGetValidator } from '../middlewares'
 import { VideoModel } from '../models/video/video'
 import { VideosCaptionCache } from '../lib/cache/videos-caption-cache'
+import { UserModel } from '../models/account/user'
+import { VideoCommentModel } from '../models/video/video-comment'
+import { HttpNodeinfoDiasporaSoftwareNsSchema20 } from '../models/nodeinfo'
 
+const packageJSON = require('../../../package.json')
 const staticRouter = express.Router()
 
 staticRouter.use(cors())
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -13,10 +13,15 @@
       "es2016",
       "es2017"
     ],
+    "typeRoots": [
+      "node_modules/@types",
```
Chocobozzz: Why did you need this?

tsconfig.json
```diff
@@ -13,10 +13,15 @@
...
+      "server/models"
+    ],
     "types": [
       "node",
       "chai-xml",
-      "chai-json-schema"
+      "chai-json-schema",
+      "nodeinfo"
```
Chocobozzz: And this?

server/controllers/static.ts
```diff
@@ -95,6 +103,45 @@ async function getVideoCaption (req: express.Request, res: express.Response) {
...
+async function generateNodeinfo (req: express.Request, res: express.Response, next: express.NextFunction) {
+  const { totalVideos: totalVideos } = await VideoModel.getStats()
+  const { totalLocalVideoComments: totalLocalVideoComments } = await VideoCommentModel.getStats()
+  const { totalUsers } = await UserModel.getStats()
+
+  const json: HttpNodeinfoDiasporaSoftwareNsSchema20 = {
+    version: '2.0',
+    software: {
+      name: 'PeerTube',
+      version: '1.0.0-beta.9'
```
Chocobozzz: Fetch the version from package.json (see config.ts controller)

server/controllers/static.ts
```diff
@@ -70,6 +73,11 @@ staticRouter.get('/robots.txt', (req: express.Request, res: express.Response) =>
   return res.send(CONFIG.INSTANCE.ROBOTS)
 })
 
+// nodeinfo service
+staticRouter.use('/.well-known/nodeinfo',
+  asyncMiddleware(generateNodeinfo)
```
Chocobozzz: Maybe we could add cache here

server/controllers/static.ts
rigelk: Oh. It's so simple it hurts :')

tsconfig.json
rigelk: I thought defining explicitly the typeRoots would require to redefine the existing ones.

server/initializers/constants.ts
```diff
@@ -574,6 +576,36 @@ export {
...
+  ms:           1,
+  second:       1000,
+  minute:       60000,
+  hour:         3600000,
+  day:          3600000 * 24,
+  week:         3600000 * 24 * 7,
+  month:        3600000 * 24 * 30
+}
+
+export function parseDuration (duration, defaultDuration) {
```
Chocobozzz: Miss type

server/initializers/constants.ts
```diff
@@ -574,6 +576,36 @@ export {
...
+      let unit = split[2].replace(/s$/i,'').toLowerCase()
+      if (unit === 'm') {
+        unit = 'ms'
+      }
+
+      return (len || 1) * (timeTable[unit] || 0)
+    }
+  }
+
+  return defaultDuration
```
Chocobozzz: Maybe add an error log, this should not happen

server/controllers/static.ts
```diff
@@ -95,6 +109,48 @@ async function getVideoCaption (req: express.Request, res: express.Response) {
...
 
+async function generateNodeinfo (req: express.Request, res: express.Response, next: express.NextFunction) {
+  const { totalVideos } = await VideoModel.getStats()
+  const { totalLocalVideoComments } = await VideoCommentModel.getStats()
+  const { totalUsers } = await UserModel.getStats()
+
+  const json: HttpNodeinfoDiasporaSoftwareNsSchema20 = {
+    version: '2.0',
+    software: {
+      name: 'PeerTube',
```
rhaamo: https://github.com/jhass/nodeinfo/blob/master/schemas/2.0/schema.json#L35
The software->name element shouldn't have uppercase chars

rhaamo:
I'm not sure to understand the routing, ```/.well-known/nodeinfo``` seems to returns the whole nodeinfo json ?

From https://github.com/jhass/nodeinfo/blob/master/PROTOCOL.md it should return a list of links to the actual nodeinfo and associated version

rigelk:
@rhaamo I just realised that 30min ago… Should be fixed now.

server/controllers/static.ts
rigelk: Thanks! I missed that one.

rigelk:
Update: the caching function has trouble caching the response due to the `content-type` of the response, which Express definitely doesn't like (it parses the profile URI as a charset…).

Right now, the cache makes further requests timeout as it cannot return a value that could not be saved, but is marked as cached.

I'll try fixing the cache function.

rigelk:
Update: the caching function fails at caching the response due to its `content-type`, which Express parses wrong (it doesn't support `profile=` URIs apparently).

Issue #826, opened 2018-07-19
Log panel for admins / modo


FirePowi:
* **PeerTube version or URL:** https://queertube.org 1.0.0bêta.9
* **Browser name/version:** Firefox 52.9

* **What is the wanted behaviour?**
When I go in « Administration » I want to have a "log" tab (With "Users","Manage Follows","Video abuses",… tabs).
What I want in it : 

    - New comments on videos : `[Comment] $author said "$commentFirst100digits…" in comment to video [$nameOfTheVideo]($linkToTheVideo) [*link*]($linkToTheComment) [administrationButtons][extendButton]`. When you click on "Extend" button you have : 
`[Comment]
From : $author
On : [$nameOfTheVideo]($linkToTheVideo)
Comment : $fullComment
[administrationButtons]
[shortenButton]`

    - New posted video : `[Video] $author posted [$nameOfTheVideo]($linkToTheVideo) [administrationButtons] [extendButton]`
Extend : 
`[Video]
From : $author
[$embeddedVideo]
[administrationButtons]
[shortenButton]`

This is open to discussion.

Chocobozzz:
FYI RSS feeds can be used for now to get these information

FirePowi:
Ok thanks.
I'll search for it :-)

Nautigsam:
I think [audit logs](https://www.martinfowler.com/eaaDev/AuditLog.html) are what we want here. These are just another level of logs but they are emitted whenever a user do some writing action. They may be stored in a file different from other logs.

I checked the code of existing logger, and it seems that we could create a custom level with `winston`.

After this step, the admin would already have a clearer look on what is going on on his/her instance. Step 2 could be to send these logs to the front, but I think there is a lot to discuss before doing it.

realityfabric:
There should definitely be audit logs for mod/admin actions at the very least. It's hard to hold people accountable if you have multiple moderators and no way to know who did what and when it happened.

Comprehensive logs would be good, too. But logs specifically for actions taken by mods and admins would be especially useful for instances with more than 2 moderators.

Chocobozzz:
Done in https://github.com/Chocobozzz/PeerTube/commit/566c125d6eee3bd907404523d94e1e0b5e403a46

Issue #825, opened 2018-07-19
Blacklist local videos


FirePowi:
* **PeerTube version or URL:** 1.0.0-beta.9

* **What is the wanted behaviour?**
The possibility for admin and modo to blacklist local videos so that they no more can be reached for a moment.
The goal is to give the possibility to moderate content, ask some things to add to description,… Before removing the video from the blacklist.

Chocobozzz:
Implemented: https://github.com/Chocobozzz/PeerTube/commit/191764f30b0a812bf3a9dbdc7daf1d5afe25e12a

PR #837, opened 2018-07-23
WIP: add audit logs

Nautigsam:
This is a WIP to implement audit logs. This is an attempt to make a first step towards a log panel inside admin tools (see #826 for original request).
I already added audit logs for videos and comments but I wanted to get some review of the implementation before generalizing it to other entities.

As an example, here is the log I get when I add a video:
```json
{"level":"audit","message":"{\"user\":1,\"domain\":\"videos\",\"action\":\"create\",\"video-tags\":[],\"video-uuid\":\"6edd2620-3c68-4e43-9cd5-f7b3a1797d03\",\"video-id\":3,\"video-createdAt\":\"2018-07-23T11:15:49.244Z\",\"video-updated
At\":\"2018-07-23T11:15:49.244Z\",\"video-publishedAt\":\"2018-07-23T11:15:48.203Z\",\"video-description\":null,\"video-duration\":287,\"video-isLocal\":true,\"video-name\":\"SketchFinal\",\"video-thumbnailPath\":\"/static/thumbnails/6ed
d2620-3c68-4e43-9cd5-f7b3a1797d03.jpg\",\"video-previewPath\":\"/static/previews/6edd2620-3c68-4e43-9cd5-f7b3a1797d03.jpg\",\"video-nsfw\":false,\"video-waitTranscoding\":true,\"video-account-id\":2,\"video-account-uuid\":\"f357b1f7-d34a
-4940-9966-d79d1ed9ad50\",\"video-account-name\":\"root\",\"video-channel-id\":1,\"video-channel-uuid\":\"e711fb0d-7dfa-49c2-8fb1-5599927fd672\",\"video-channel-name\":\"e711fb0d-7dfa-49c2-8fb1-5599927fd672\",\"video-support\":null,\"vid
eo-commentsEnabled\":true}","timestamp":"2018-07-23T11:15:49.286Z","label":"localhost:9000"}
```

Yes, that's pretty big. I can actually make an even more strict filter if you think some fields are useless. I also had a hard time deciding how I could identify the user so I user the userId from the OAuth token but maybe there is a better way?


server/controllers/api/videos/index.ts
```diff
@@ -298,6 +302,7 @@ async function updateVideo (req: express.Request, res: express.Response) {
   }
 
   try {
+    let videoInstanceUpdated
```
Chocobozzz: Why?

Chocobozzz:
I think we would want to audit remote objects too (coming from the federation) (see `/server/lib/activitypub/process` directory).

>  I can actually make an even more strict filter if you think some fields are useless

``` 
const videoKeysToKeep = [
  'tags',
  'category',
  'licence',
  'language',
  'privacy',
  'description',
  'name',
  'nsfw',
  'account-uuid',
  'account-name',
  'channel-uuid',
  'channel-name',
  'support'
]

const commentKeysToKeep = [
  'text',
  'video-name',
  'account-uuid',
  'account-name'
]
```

Should be enough for now.

> I also had a hard time deciding how I could identify the user so I user the userId from the OAuth token but maybe there is a better way?

Prefer using the account username + hostname (if remote): `res.locals.oauth.token.User.Account.Actor.preferredUsername + '@' + res.locals.oauth.token.User.Account.Actor.getHost()` (please create a method for this below `getWebfingerUrl` in actor model)

server/controllers/api/videos/index.ts
Nautigsam: I don't have strong opinion on this change. I'm sure there is a better way. I could simply make the log inside the transaction.

Chocobozzz: Oh, did not see the log was outside! But yes put it just below `federateVideoIfNeeded` (+ logger.info too please)

Nautigsam:
I pushed some corrections according to your remarks. I will now use it in as many entities as I can. 

> I think we would want to audit remote objects too

I tend to disagree with you here. I think the use case is more like "As an admin, I use audit logs to monitor writing actions of users connected on my instance". I'm afraid that logging every actions from remote instances will add a lot of noise for actions the admin can not do anything about.

Chocobozzz:
> I tend to disagree with you here. I think the use case is more like "As an admin, I use audit logs to monitor writing actions of users connected on my instance". I'm afraid that logging every actions from remote instances will add a lot of noise for actions the admin can not do anything about.

I understand but remote objects are as important as you local objects since they are displayed on your instance, with almost no differences with your local objects. But let's ping @firePowi

Chocobozzz:
Very nice work @Nautigsam! Is it still a WIP?

Nautigsam:
Some logs are missing because of a bug I had. I needed some refactoring in another place. If the core module and the current uses seem ok, then you can merge it I guess.

Chocobozzz:
:+1: Thanks @Nautigsam. Let's merge it now to avoid conflicts.

> Some logs are missing because of a bug I had.

What is the bug and what logs are missing?

Nautigsam:
> What is the bug and what logs are missing?
I got a bug when I added logs for videos blacklists. I will try again but it's not trivial to test as it needs a second instance running (#825 could be handy in that case).

Captions logs are missing. I also wanted to log when an admin add instances to follow, but I don't yet fully understand that part and maybe it would require some refactoring.

FirePowi:
Sorry for late respond (wasn't around)

So I might more agree with @Nautigsam → Local log.

If someone from another instance leave a comment on the local instance → log it
If someone from another instance publish a video on that another instance → do not log it / leave the choice to admin to log it or not [options takes time so… Maybe first one]
If someone from my instance leave a comment on another instance → log it

That's what I wanted but don't know if it's what everyone want. So I just give my point of view, give yours, let's discuss ?

Chocobozzz:
We could add an enum field inside the object (`local`, 'to_remote`, `from_remote`); it would cover all the needs if we give the ability on the client side to filter the entries.

PR #850, opened 2018-07-25
Updated travis.yml git depth 1

Jorropo:
We don't do any git operation except include last commit name, so we only need last commit, that will make travis-ci test faster.


rigelk:
Thanks @Jorropo !

PR #852, opened 2018-07-26
[DO NOT MERGE][Server>Middlewares>Follows] Changing error code 400 to 500

Jorropo:
Follows on http server is a server error in the configuration not an error of the client.

I'm fixing follows api missing doc but I would know how I can easly edit the html doc file ?


Chocobozzz:
> Follows on http server is a server error in the configuration not an error of the client.

No, it's a client error that wants to follow a server he can not follow -> bad parameter.

>I'm fixing follows api missing doc but I would know how I can easly edit the html doc file ?

https://github.com/Chocobozzz/PeerTube/blob/develop/.github/CONTRIBUTING.md#write-documentation

Chocobozzz:
@Jorropo After thinking, you are right. Please could you do another merge request?

PR #856, opened 2018-07-26
[FIX] doc: correct URL to PeerTube project

mart-e:
Instead of the French translation page


Chocobozzz:
Maybe https://trad.framasoft.org/iteration/view/peertube/develop would be better

Chocobozzz:
Thanks @mart-e!

PR #862, opened 2018-07-26
Added redis-server.service to After

Jorropo:
On very very slow server, peertube can generate error if redis isn't started.


Chocobozzz:
Thanks @Jorropo

PR #866, opened 2018-07-27
[Server>Middlewares>Follows] Changing error code 400 to 500

Jorropo:
Follows on http server is a server error in the configuration not an error of the client.

@Chocobozzz as you ask in #852 this is reopen.


Chocobozzz:
Thanks @Jorropo

PR #878, opened 2018-07-28
Fix routing for search in API docs

### LINKED ISSUE #871: Routing for search are incorrect in API docs
The [API docs indicate](https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/api/openapi.yaml#L515) `/videos/search` but [it is coded](https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/utils/search/videos.ts#L19) as `/search/videos`. Which one is good and which needs to be fixed?

DavidLibeau:
Resolve #871


rigelk:
Thanks @DavidLibeau !

PR #880, opened 2018-07-28
adding ORM pool configuration

### LINKED ISSUE #518: Expose database connection pool configuration
When I start following peertube account from Mastodon instance my peertube instance throws this error 

error: Cannot create remote actor and check signature. {
  "err": "TimeoutError: ResourceRequest timed out\n    at ResourceRequest._fireTimeout (/var/www/peertube/versions/peertube-v1.0.0-beta.3/node_modules/generic-pool/lib/ResourceRequest.js:62:17)\n    at Timeout.bound (/var/www/peertube/versions/peertube-v1.0.0-beta.3/node_modules/generic-pool/lib/ResourceRequest.js:8:15)\n    at ontimeout (timers.js:482:11)\n    at tryOnTimeout (timers.js:317:5)\n    at Timer.listOnTimeout (timers.js:277:5)"

PeerTube v1.0.0-beta.3
* **Browser name/version:**
Safari 11.1
* **NodeJS version:**
v18.11.1

Expected behaviour is be able to follow peertube instance's account from Mastodon

Mentioned above error.

Thanks to point me to the right direction to solve this.

rigelk:
Resolves #518

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -22,6 +22,8 @@ database:
   suffix: '_dev'
   username: 'peertube'
   password: 'peertube'
+  pool:
+    max: 5
 
 # You can also specify a 'socket' path to a unix socket but first need to
 # comment out hostname and port
--- a/server/initializers/database.ts
+++ b/server/initializers/database.ts
@@ -32,6 +32,7 @@ const username = CONFIG.DATABASE.USERNAME
 const password = CONFIG.DATABASE.PASSWORD
 const host = CONFIG.DATABASE.HOSTNAME
 const port = CONFIG.DATABASE.PORT
+const poolMax = CONFIG.DATABASE.POOL.MAX
 
 const sequelizeTypescript = new SequelizeTypescript({
   database: dbname,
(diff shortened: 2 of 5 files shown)


config/default.yaml
```diff
@@ -22,6 +22,7 @@ database:
   suffix: '_dev'
   username: 'peertube'
   password: 'peertube'
+  pool: 5
```
Chocobozzz: `pool.max` or `poolMax` would be a better name

server/initializers/database.ts
```diff
@@ -40,6 +41,11 @@ const sequelizeTypescript = new SequelizeTypescript({
   port,
   username,
   password,
+  pool: {
+    max: poolMax,
+    idle: 30000,
+    acquire: 60000
```
Chocobozzz: Where those values come from?

rigelk: You're right, I should use the default pooling config: https://github.com/sequelize/sequelize/blob/63d4aaa3298e7a8a9b12d96deef1b8d015af712a/lib/dialects/abstract/connection-manager.js#L10-L17

Chocobozzz:
Thanks!

PR #887, opened 2018-07-29
fix pg_restore command for post-migrations

### LINKED ISSUE #886: Suggested pg_restore command does not work after migrations
The suggested command to [restore the database](https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/production.md#things-went-wrong) doesn't work if there are migrations.

The reasons is that the restore script only knows about tables/constraints prior to migration. So many of the commands within the `-c` (clean) part of the script will fail due to breaking of newly introduced constraints.

The option that will always work is to drop and recreate the database. This must be done by the superuser (postgres) which is the sensible option anyways since we are executing the commands locally as a user in sudo group. 

This is what should be used:

```bash
sudo -u postgres pg_restore -c -C -d postgres $SQL_BACKUP_PATH
```

Key parts to explain:

* `-c` + `-C`: When clean and create are combined the database will be dropped and recreated followed by everything else.
* `-d postgres`: This is just the database to connect to (something is required) while the actual database being restored is defined in the `.bak` file.

I will put in a pull request with the changes. Also for the actual back up part, the current script is fine but we can also use `-u postgres` so we're not entering/passing another password.

Here is the error after restoring to `v1.0.0-beta.9` from `v1.0.0-beta.10.pre.2`:

```
pg_restore: [archiver (db)] Error while PROCESSING TOC:
pg_restore: [archiver (db)] Error from TOC entry 2202; 1259 29750 INDEX video_name peertube
pg_restore: [archiver (db)] could not execute query: ERROR:  index "video_name" does not exist
    Command was: DROP INDEX public.video_name;

pg_restore: [archiver (db)] Error from TOC entry 2200; 1259 29748 INDEX video_id_privacy_state_wait_transcoding peertube
pg_restore: [archiver (db)] could not execute query: ERROR:  index "video_id_privacy_state_wait_transcoding" does not exist
    Command was: DROP INDEX public.video_id_privacy_state_wait_transcoding;

pg_restore: [archiver (db)] Error from TOC entry 2204; 2606 29714 CONSTRAINT video_pkey peertube
pg_restore: [archiver (db)] could not execute query: ERROR:  cannot drop constraint video_pkey on table public.video because other objects depend on it
DETAIL:  constraint videoCaption_videoId_fkey on table public."videoCaption" depends on index public.video_pkey
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
    Command was: ALTER TABLE ONLY public.video DROP CONSTRAINT video_pkey;

pg_restore: [archiver (db)] Error from TOC entry 205; 1259 29592 TABLE video peertube
pg_restore: [archiver (db)] could not execute query: ERROR:  cannot drop table public.video because other objects depend on it
DETAIL:  constraint videoCaption_videoId_fkey on table public."videoCaption" depends on table public.video
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
    Command was: DROP TABLE public.video;

```

joshmorel:
Resolves #886 

The `pg_restore` command fix is required to properly restore after any migrations have been performed on upgrade.

The tweak of the `pg_dump` command is optional as the result is equivalent to the original command. The benefit is it avoids the need to enter both the sudo password and peertube database user password. Because the `upgrade.sh` script is meant to be run entirely as the `peertube` user it makes sense as-is.

### DIFF
--- a/support/doc/production.md
+++ b/support/doc/production.md
@@ -221,7 +221,7 @@ Make a SQL backup
 ```
 $ SQL_BACKUP_PATH="backup/sql-peertube_prod-$(date -Im).bak" && \
     cd /var/www/peertube && sudo -u peertube mkdir -p backup && \
-    sudo pg_dump -U peertube -W -h localhost -F c peertube_prod -f "$SQL_BACKUP_PATH"
+    sudo -u postgres pg_dump -F c peertube_prod | sudo -u peertube tee "$SQL_BACKUP_PATH" >/dev/null
 ```
 
 Fetch the latest tagged version of Peertube:
@@ -305,8 +305,8 @@ Change `peertube-latest` destination to the previous version and restore your SQ
 
 ```
 $ OLD_VERSION="v0.42.42" && SQL_BACKUP_PATH="backup/sql-peertube_prod-2018-01-19T10:18+01:00.bak" && \
-    cd /var/www/peertube && unlink ./peertube-latest && \
+    cd /var/www/peertube && sudo -u peertube unlink ./peertube-latest && \
     sudo -u peertube ln -s "versions/peertube-$OLD_VERSION" peertube-latest && \
-    pg_restore -U peertube -W -h localhost -c -d peertube_prod "$SQL_BACKUP_PATH"
+    sudo -u postgres pg_restore -c -C -d postgres "$SQL_BACKUP_PATH" && \
     sudo systemctl restart peertube
 ```


support/doc/production.md
```diff
@@ -221,7 +221,7 @@ Make a SQL backup
 ```
 $ SQL_BACKUP_PATH="backup/sql-peertube_prod-$(date -Im).bak" && \
     cd /var/www/peertube && sudo -u peertube mkdir -p backup && \
-    sudo pg_dump -U peertube -W -h localhost -F c peertube_prod -f "$SQL_BACKUP_PATH"
+    sudo -u postgres pg_dump -F c peertube_prod | sudo -u peertube tee "$SQL_BACKUP_PATH" >/dev/null
```
Chocobozzz: Why do you prefer to use `tee` command instead of `-f`?

Chocobozzz:
Nice catch @joshmorel, I thought the `-c` command removed the constraints too :( Thanks!

support/doc/production.md
joshmorel: Since the `pg_dump` command is run as the `postgres` the file can't be written where `postgres` doesn't have permissions. The only option is to pipe the dump stdout to `tee` (either just with `sudo` - `root` will be owner, or as `peertube`)

joshmorel:
Your welcome. You are right that `-c` removes constraints - but only for those existing when the dump file was created. So if a table is added since, say `videoCaption`, the `pg_restore` won't drop it or any of the its constraints referencing other tables.

I must admit the whole command is not super intuitive. I had to experiment a bit to figure it out - the docs, while comprehensive, were not quite enough. `pg_restore` will actually show the series of SQL commands that will be run by omitting the `-d dbname` option.

support/doc/production.md
Chocobozzz: :+1:

Chocobozzz:
Thanks!

PR #904, opened 2018-08-03
Use the config file values in upgrade.sh script

jfoucry:
If the postgresql database is not hosted on the same host as Peertube, the upgrade.sh script failed.
I added a small parse_yaml,sh function file (found  [here}(https://gist.github.com/pkuczynski/8665367) and use it in the upgrade.sh script. I print the database password to avoid looking for it during ages and resign user from upgrade.

### DIFF
--- a/scripts/parse_yaml.sh
+++ b/scripts/parse_yaml.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+parse_yaml() {
+   local prefix=$2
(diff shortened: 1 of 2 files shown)


Chocobozzz:
i'm sorry but this is unmaintainable... if your database is on another server, just use your own upgrade script, or do a PR that adds some parameters to the script

PR #932, opened 2018-08-19
correct password reset disabled binding

joshmorel:
Fixed by changing the logic - if the form is not valid or the confirmed password does not match the button will be disabled. Considered a refactor with shared module between change and reset passwords but determined the effort did not match the value.

All testing done manually.


Chocobozzz:
Thanks @joshmorel!

Issue #930, opened 2018-08-19
Login not working on development Instance


Nutomic:
* **PeerTube version or URL:** v1.0.0-beta.10
* **Browser name/version:** Firefox 61, Chromium 68 on latest Manjaro
* **NodeJS version:** v10.9.0

I set up a local development instance according to the [contribututing instructions](https://github.com/Chocobozzz/PeerTube/blob/develop/.github/CONTRIBUTING.md#develop). Btw it seems they are missing instructions to enable the required SQL extensions (I did that too). Compilation seems fine, the only errors I can see in the terminal are these, but they don't seem to break anything:
```
[localhost:9000] 2018-08-18 18:57:29.958 error: Error in Redis client.
[1] [0] [nodemon] app crashed - waiting for file changes before starting...
```

* **What is the expected behaviour?**
I can load Peertube normally in my browser and login as admin.

* **What do you see instead?**
When I open the page, it shows a red box with "Error: Unknown error" in the bottom right. Additionally, trying to login with root/test results in an error "Unknown error".

* **Browser console log if useful (Gist/Pastebin...):**
After loading the page, I get multiple of these warnings and errors in the browser console:
https://gist.github.com/Nutomic/bb6993bb7a0b8ed3911ababeebbefced

Submitting any username/password results in the following messages:
https://gist.github.com/Nutomic/2660b1fec86c66f29e230236c153f709

* **Server log if useful (Gist/Pastebin...):**
nothing I could find here

Jorropo:
Have you installed an redis server and setup login and password for redis ?
If yes : is your dB id unique ?

Chocobozzz:
> Btw it seems they are missing instructions to enable the required SQL extensions (I did that too)

Could you make a PR? :)

There is an error in you redis configuration I think. Are you up to date with the `develop` branch? We fixed an issue with the logging system, but it seems there is still the bug on your version because we don't see what is the "Redis client error".

Nutomic:
I installed redis, but I hadn't created a [Peertube system user](https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/production.md#peertube-user), which I assume you are referring to. However, that didn't fix the problem for me. Which database ID do you mean, and how can I check it?

With the develop branch (or latest master), the build fails completely with a lot of errors:

https://gist.github.com/Nutomic/9625e5c5396213a7c7730f0f49b4bcde

[...]

PR #940, opened 2018-08-21
Adding missing instructions to Contributing.md

Nutomic:
See #930.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -56,6 +56,10 @@ $ cd PeerTube
 $ yarn install --pure-lockfile
 ```
 
+Note that development is done on the `develop` branch. If you want to hack on
+Peertube, you should switch to that branch. Also note that you have to repeat
+the `yarn install --pure-lockfile` command.
+
 Then, create a postgres database and user with the values set in the
 `config/default.yaml` file. For instance, if you do not change the values
 there, the following commands would create a new database called `peertube_dev`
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -67,6 +68,21 @@ Enter password for new role: peertube
...
+```
+
+Then enable extensions PeerTube needs:
+
+```
+$ sudo -u postgres psql -c "CREATE EXTENSION pg_trgm;" peertube_dev
+$ sudo -u postgres psql -c "CREATE EXTENSION unaccent;" peertube_dev
+```
+
 In dev mode, administrator username is **root** and password is **test**.
```
Jorropo: You should remove stars cause if like me you read this doc in a basic text editor you will be confused about the password.

.github/CONTRIBUTING.md
```diff
@@ -52,6 +52,7 @@ Then clone the sources and install node modules:
 
 ```
 $ git clone https://github.com/Chocobozzz/PeerTube
+$ git checkout develop
```
Jorropo: You don't need to checkout develop, if what you are testing is in an other branch you checkout the other branch.

.github/CONTRIBUTING.md
```diff
@@ -67,6 +68,21 @@ Enter password for new role: peertube
 # sudo -u postgres createdb -O peertube peertube_dev
 ```
 
+You also need to create a system user for Peertube:
```
Chocobozzz: Why did you need this peertube user? You should be able to dev on peertube with your regular user

Nutomic: Ah I misunderstood something, will remove this.

.github/CONTRIBUTING.md
Nutomic: Right, I'm adding a comment regarding this.

Nutomic:
Updated PR.

.github/CONTRIBUTING.md
```diff
@@ -56,6 +56,10 @@ $ cd PeerTube
 $ yarn install --pure-lockfile
 ```
 
+Note that development is done on the `develop` branch. If you want to hack on
```
Jorropo: Develop isn't really made on develop branch, in fact you should durring work on an other branch and merge into develop when finish.

Chocobozzz: I agree but to keep it as simple as possible, we don't really want to specify how the developer should work. If they want to use the git flow, they can, if they want to work directly on develop on their fork, they can too.

Chocobozzz:
Thanks @Nutomic

Issue #640, opened 2018-06-09
Allow setting a profile ID for channels


espectalll:
Just like user profiles are asked to have an username, video channels could be configured with a specific ID. This could be done when creating the channel and, if feasible and properly convenient, after doing so. The main purpose of this is to get easily human-readable channel URLs in order to make sharing and following a specific channel easier.

Chocobozzz:
Yes, but how to differentiate an account than a video channel?

If my account is `chocobozzz@peertube.cpy.re`, and I create a channel `squall@peertube.cpy.re` it's confusing. Moreover nobody will be able to register on `peertube.cpy.re` with the username `squall`.

espectalll:
is easy, you just don't need to. Block usernames which are already registered for both. Allow all of them. Or rather, if there is any need to differentiate, append a prefix/suffix for channels and/or accounts.

Chocobozzz:
Implemented: https://github.com/Chocobozzz/PeerTube/commit/8a19bee1a1ee39f973bb37429e4f73c3f2873cdb

PR #945, opened 2018-08-22
User subscriptions

Chocobozzz:
* Add ability to set the name of the channel
 * We will try to put more emphasis on channels in the future (instead of accounts)
 * Add subscriptions tab in the menu, to see videos from subscriptions
 * Add ability to subscribe to a video channel
 * Add subscriptions tab in user menu, to see subscribed video channels
 * Add some channels in search results
 * Add ability to search remote video or video channels (from a URL or handle) if allowed by the administrator (by default, only logged users can do that)
 * When we search a remote object, we fetch it and insert it in the database. If it is an actor, we process its outbox in a job (to create videos etc). Since we might not follow these actors, we need to refresh them sometimes


Chocobozzz:
@Jorropo yes for the `@app` in fact I use the auto import so sometimes it decides to use it, sometimes it doesn't ^^

> I have just a question (this isn't a problem of using name instead of uuid for channel (why can't we get conflict ?))

I'm sorry I don't understand the question. You fear conflicts on actor names?

Jorropo:
@Chocobozzz yeah, I fear that.

Chocobozzz:
@Jorropo I don't think so, we just detect conflicts when someone tries to create a user or a channel with an existing name. See https://github.com/Chocobozzz/PeerTube/issues/640

Chocobozzz:
Done, I updated the PR description.

PR #952, opened 2018-08-22
Add gzip support to the sample nginx configuration

### LINKED ISSUE #948: Sample nginx config doesn't include gzip
I'm not an expert in any of this, please let me know if I've done something wrong... but if this is just an omission, I can write a pull request.

The configuration from the readme's production guide does not include gzip transfer encoding, which afaict needs to be enabled in the server section of the nginx config.

In my instance, I've added this configuration to reduce the bandwidth and time to load the initial javascript bundle, from over a megabyte down to about 300K.

  gzip on;
  gzip_types text/plain text/css text/html application/javascript application/json;
  gzip_vary on;

edit: added gzip_vary

scanlime:
Without gzip explicitly enabled, load times suffer from transferring
over a megabyte of plaintext javascript. With gzip enabled, the bundle
is down to about 300K, and loads much faster.

This change does not enable gzip on files that are already compressed,
so images, fonts, and videos will be sent without the CPU overhead.

closes #948

### DIFF
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -37,6 +37,11 @@ server {
   # resolver $DNS-IP-1 $DNS-IP-2 valid=300s;
   # resolver_timeout 5s;
 
+  # Enable compression for JS/CSS/HTML and JSON, for improved client load times
+  gzip on;
+  gzip_types text/plain text/css text/html application/javascript application/json;
+  gzip_vary on;
+
   add_header Strict-Transport-Security "max-age=63072000; includeSubDomains; preload";
 
   access_log /var/log/nginx/peertube.example.com.access.log;
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -37,9 +37,11 @@ server {
   # resolver $DNS-IP-1 $DNS-IP-2 valid=300s;
   # resolver_timeout 5s;
 
-  # Enable compression for JS/CSS/HTML and JSON, for improved client load times
+  # Enable compression for JS/CSS/HTML bundle, for improved client load times.
+  # It might be nice to compress JSON, but leaving that out to protect against potential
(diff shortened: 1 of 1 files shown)


support/nginx/peertube
```diff
@@ -37,6 +37,11 @@ server {
   # resolver $DNS-IP-1 $DNS-IP-2 valid=300s;
   # resolver_timeout 5s;
 
+  # Enable compression for JS/CSS/HTML and JSON, for improved client load times
+  gzip on;
+  gzip_types text/plain text/css text/html application/javascript application/json;
```
Chocobozzz: Maybe we should disable gzip on `application/json`, to protect us from the [BREACH](http://breachattack.com/) attack.

Jorropo: @Chocobozzz I think you are right, and that will make app faster, but one other solution could be good is to sign with hmac, that will totaly protect us from replay attack and modification.

scanlime:
Not sure HMAC would protect against BREACH, that's about data leaking via
attacker-controlled plaintext incorporated into the reply. Sure, probably a
good idea to avoid any chance of that by omitting API from the compression.

And I agree SVG is a good idea too. I'll update the patch.

scanlime:
Looks like SVG is all being inlined in the CSS file, so that explains why I'm not seeing any separate SVG requests. 👌

support/nginx/peertube
scanlime: as I understand it, the issue here isn't necessarily about replay attacks (SSL itself protects us from that) but that when you compress then encrypt data, the size of the resulting packet can leak information about some of the encrypted plaintext if another part of the compressed and encrypted input was attacker-controlled.

Chocobozzz:
Thanks @scanlime!

PR #957, opened 2018-08-24
Display report reason in multiple lines

Nutomic:
Right now, it is impossible to read the entire report message if it has more than ~30 characters. The only way to read it is by inspecting the HTML. So I've adjusted the CSS to display the entire message in multiple lines in this case.


Chocobozzz:
Oup's, forgot to implement the abuse reason like the blacklist reason. Could you do that instead? (put the reason in the `rowexpansion` template)

Nutomic:
Okay I did that, plus some other adjustments (see the commit for details). Here's how it looks now:

peertube-moderation

I tried to put the "Reason" and "Moderation comment" labels into seperate table cells to make it align nicely, but I didn't get that working properly (colspan=2 is too little and colspan=3 is too much).

Chocobozzz:
@Nutomic Did you try to put them in different rows?

Nutomic:
I tried to put them on seperate rows, put then I couldn't get the columns to align properly (min-width doesn't work on `<td>` apparently. FYI I haven't done HTML/CSS in years, so it's all pretty rusty. 

I did set the min-width on the labels now. I'm only worried that it might break with translations.

moderation

Chocobozzz:
>  I'm only worried that it might break with translations.

Translations should be fine. Thanks!

PR #959, opened 2018-08-25
Update translation.md

Strubbl:
add hint, that the string yy-mm-dd shall not be translated, because otherwise you get translated strings in the date chooser like "JJ-August-TT 14:00" (German)

### DIFF
--- a/support/doc/translation.md
+++ b/support/doc/translation.md
@@ -33,3 +33,5 @@ For example:
 
 should be in french 
 ```<x id="INTERPOLATION" equiv-text="{{ video.publishedAt | myFromNow }}"/> - <x id="INTERPOLATION_1" equiv-text="{{ video.views | myNumberFormatter }}"/> vues```
+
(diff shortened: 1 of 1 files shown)


Nutomic:
I think you should adjust the date strings according to the language. Right now I'm seeing dates like `May 30, 2018, 9:30:56 AM` with language set to German. Or maybe there is a way to do this automatically in the code.

Strubbl:
Oh, yes you are right. But then only the letters are changed in the order, but shall not be translated

Nutomic:
It's the user list on the admin interface (in the latest release version).

peertube-user-list

Chocobozzz:
This `yy-mm-dd` string corresponds to the date format of the scheduled video input

screen_2018-08-26-19 04 17

So it should be translated too (see DateFormat section in https://www.primefaces.org/primeng/#/calendar)

Strubbl:
Shall i add a hint about this to the support/doc/translation.md ?

Chocobozzz:
:+1:

PR #964, opened 2018-08-25
Allow Docker admins to setup the default quota

kaiyou:
Simply add the default quota to the custom configuration variables.


rigelk:
Thanks @kaiyou !

PR #965, opened 2018-08-25
Keep audio lined up with the beginning of the video in ffmpeg

rigelk:


Jorropo:
Is there some e2e test related to uploading to change ?

Chocobozzz:
Travis fails

rigelk:
Just when I tried to correct it my setup crashed - let me try to fix that.

PR #977, opened 2018-08-28
add user account email verificiation

joshmorel:
Two remaining items related to this to (potentially) add in future PR:

1. Ability for admin to verify users manually via admin console
2. If SMTP is not configured properly the user will get a message indicating the email successfully sent. This is the same with reset-password. Only the server admin will see via logs. Not sure if this intended, but I could add some middleware to report on SMTP-related errors in the front-end. Let me know

Includes server and client code to:

* enable verificationRequired via custom config
* send verification email with registration
* ask for verification email
* verify via email
* prevent login if not verified and required
* conditional client links to ask for new verification email


Nutomic:
How does this handle existing, unverified users? It might be useful to send verification emails to them as well.

Error messages on the frontend about SMTP failures definitely sound useful. Maybe open an issue about it first?

rigelk:
@joshmorel 1 is good enough IMHO

joshmorel:
Hi @Chocobozzz - I'll address the changes tonight. There's one thing I'd like to clarify:

> Keep things simple, and don't try to change the past. The configuration only applies on present/future objects.

Before seeing this message, this morning I added functionality where existing non-verified users are verified if the admin enables verification via the admin config tab. Is this what you intend to occur with this statement? Or something else?

Chocobozzz:
> Before seeing this message, this morning I added functionality where existing non-verified users are verified if the admin enables verification via the admin config tab. Is this what you intend to occur with this statement? Or something else?

No, I would rather not check existing users when changing the configuration.

joshmorel:
So then would existing users would need to verify to login. They will get a option to get a verification email on next login with the existing code. Alternatively, I could also send verification emails when the setting is enabled.

joshmorel:
Okay - with this feedback I'll propose a very simple approach. 

When `requiresVerification` is enabled new users will be created with `verified=false`. When it is *not* enabled they will be created with `verified=null`. 

Only users with `verified=false` will be prevented from logging in. This way when the setting is changed no action will need to be triggered for existing users.

joshmorel:
I resolved all suggestions while implementing a process which should make switching between email verification required enabled & disabled simple.

Chocobozzz:
It's just perfect! Thanks @joshmorel!

PR #978, opened 2018-08-28
Replacing placeholder and feature table display improvement

rigelk:
There was a typo to A_r_mstrong so I took the opportunity to put a nameless
alternative.

I took the liberty to modify the table display so that it uses a bit better the
width of the screen.

### DIFF
--- a/client/src/app/login/login.component.scss
+++ b/client/src/app/login/login.component.scss
@@ -20,4 +20,8 @@ input[type=submit] {
 .create-an-account, .forgot-password-button {
   color: #000;
   cursor: pointer;
+  
+  &:hover {
+    text-decoration: underline !important;
+  }
 }
--- a/client/src/app/signup/signup.component.html
+++ b/client/src/app/signup/signup.component.html
@@ -4,63 +4,68 @@
     Create an account
   </div>
 
-  <my-instance-features-table></my-instance-features-table>
-
   <div *ngIf="error" class="alert alert-danger">{{ error }}</div>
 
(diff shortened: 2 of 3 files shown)


client/src/app/login/login.component.scss
```diff
@@ -20,4 +20,8 @@ input[type=submit] {
 .create-an-account, .forgot-password-button {
   color: #000;
   cursor: pointer;
+  
+  &:hover {
+    text-decoration: none;
```
Chocobozzz: I think it's important to keep it (and add it to the forgot password button) for accessibility

client/src/app/signup/signup.component.html
```diff
@@ -4,63 +4,68 @@
...
 
-  <my-instance-features-table></my-instance-features-table>
-
   <div *ngIf="error" class="alert alert-danger">{{ error }}</div>
 
-  <form role="form" (ngSubmit)="signup()" [formGroup]="form">
-    <div class="form-group">
-      <label for="username" i18n>Username</label>
+  <div class="clearfix">
+    <form class="float-left" role="form" (ngSubmit)="signup()" [formGroup]="form">
```
Chocobozzz: You may want to use flexboxs instead of float? Or bootstrap col-*.

client/src/app/login/login.component.scss
rigelk: Okay, either way it was just for consistency.

rigelk:
feature-demo

PR #994, opened 2018-08-31
Add more playback speed (0.75 and 1.25)

Glandos:
Those speeds are much more closer to the normal speed. They are useful for listening to human speeches.


Chocobozzz:
Thanks @Glandos

PR #1000, opened 2018-09-02
WIP: Add video recomandation by tags.

Jorropo:
Video recommendation in video watch is set by tags (it made a search with tagsOneOf).
# To do :
 - [X] Set by tags
 - [ ] Porting to new component
 - [ ] Don't search by tags if there is no tag
 - [ ] Fall back to last video uploaded if there isn't any video with corresponding tag


Jorropo:
Moved to #1001

Issue #352, opened 2018-03-16
Allow moderators or administrators to curate tags


conferences-gesticulees:
* **PeerTube version or commit: 1.0.0-alpha.8**

* **What is the expected behaviour?**

Tags will only be useful if videos about the same topic all use the same tag. It would be good to be able to list all tags and to be able to delete or “rename” a tag.

* **What do you see instead?**

There is currently no way to browse tags. Changing a tag in multiple videos requires to visit the metadata page for each video, which can be pretty tedious.

maxlath:
Following up on the conversation with @Chocobozzz at the JDLL and [on mastodon](https://framapiaf.org/@Chocobozzz/99761894332116480), here is how I think the task of deduplicating tags could be eased by the use of [Wikidata](https://wikidata.org) items:

Each [Wikidata item](https://www.wikidata.org/wiki/Wikidata:Glossary#Item) represents a concept: as a first approximation, 1 Wikipedia article => 1 Wikidata item.
Example:
* the city of London, a.k.a. Londres in French or Londra in Italian, is uniquely identified in Wikidata as [Q84](https://www.wikidata.org/entity/Q84)
* exoplanets, a.k.a. extrasolar planet, Экзапланета in Belarusian, or كوكب خارج المجموعة الشمسية in Arab have the unique id [Q44559](https://www.wikidata.org/entity/Q44559)

This means that you can search for any of those terms in any of those languages and be gently redirected to the uniquely identified concept:
* [search *London* in English](https://www.wikidata.org/w/api.php?action=wbsearchentities&language=en&format=json&search=London)
* [search *Londres* in French](https://www.wikidata.org/w/api.php?action=wbsearchentities&language=fr&format=json&search=Londres)
* [search *Londra* in Italian](https://www.wikidata.org/w/api.php?action=wbsearchentities&language=it&format=json&search=Londra)
* [search *exoplanet* in English](https://www.wikidata.org/w/api.php?action=wbsearchentities&language=en&format=json&search=exoplanet)
* [search *extrasolar planet* in English](https://www.wikidata.org/w/api.php?action=wbsearchentities&language=en&format=json&search=extrasolar%20planet)
* [search *Экзапланета* in Belarusian](https://www.wikidata.org/w/api.php?action=wbsearchentities&language=be&format=json&search=Экзапланета)
[...]

PR #1001, opened 2018-09-02
Add video recomandation by tags

Jorropo:
I all ready do that but before moving recommendation to an other service (#1000).
# To do :
## Client :
 - [X] Implement
 - [x] Do not load by tags if the is no tag set
### Then :
 - [X] Fix client tests

I didn't added new test cause all the logic about sorting similar videos is made by the server during the search.

### DIFF
--- a/client/src/app/videos/+video-edit/shared/video-edit.component.html
+++ b/client/src/app/videos/+video-edit/shared/video-edit.component.html
@@ -15,7 +15,7 @@
 
             <div class="form-group">
               <label i18n class="label-tags">Tags</label>
-              <my-help i18n-preHtml preHtml="Tags are used to suggest relevant recommendations.</br>Press Enter to add a new tag."></my-help>
+              <my-help i18n-preHtml preHtml="Tags could be used to suggest relevant recommendations.</br>Press Enter to add a new tag."></my-help>
               <tag-input
                 [validators]="tagValidators" [errorMessages]="tagValidatorsMessages"
                 formControlName="tags" maxItems="5" modelAsStrings="true"
--- a/client/src/app/videos/+video-edit/shared/video-edit.component.html
+++ b/client/src/app/videos/+video-edit/shared/video-edit.component.html
@@ -14,7 +14,8 @@
             </div>
 
             <div class="form-group">
-              <label i18n class="label-tags">Tags</label> <span i18n>(press Enter to add)</span>
+              <label i18n class="label-tags">Tags</label>
+              <my-help i18n-preHtml preHtml="Tags are used to suggest relevant recommendations.</br>Press Enter to add a new tag."></my-help>
               <tag-input
                 [validators]="tagValidators" [errorMessages]="tagValidatorsMessages"
                 formControlName="tags" maxItems="5" modelAsStrings="true"
--- a/client/src/app/videos/recommendations/recent-videos-recommendation.service.spec.ts
+++ b/client/src/app/videos/recommendations/recent-videos-recommendation.service.spec.ts
@@ -21,7 +21,7 @@ describe('"Recent Videos" Recommender', () => {
         { uuid: 'uuid2' }
       ]
       getVideosMock.mockReturnValueOnce(of({ videos: vids }))
-      const result = await service.getRecommendations('uuid1').toPromise()
+      const result = await service.getRecommendations({ uuid: 'uuid1' }).toPromise()
       const uuids = result.map(v => v.uuid)
       expect(uuids).toEqual(['uuid2'])
       done()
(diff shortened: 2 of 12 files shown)


Chocobozzz:
> ALWAYS lower tag (for counting "overwatch" and "Overwatch" as the same tag, and do faster search request than than do a case insensitive search)

I don't think it is useful, https://github.com/Chocobozzz/PeerTube/issues/352#issuecomment-377204232 would be better

ping @bradsk88

client/src/app/videos/recommendations/recent-videos-recommendation.service.ts
```diff
@@ -14,26 +17,36 @@ export class RecentVideosRecommendationService implements RecommendationService
...
+  private fetchPage (page: number, video: VideoDetails): Observable<Video[]> {
     let pagination = { currentPage: page, itemsPerPage: this.pageSize + 1 }
-    return this.videos.getVideos(pagination, '-createdAt')
-      .pipe(
+    if (video.tags.length === 0) {
+      return this.videos.getVideos(pagination, '-createdAt')
+        .pipe(
+          map(v => v.videos)
+        )
+    } else {
```
Chocobozzz: you don't need this `else` because you `return` in the previous if

Jorropo:
@Chocobozz :+1: wiki item is an other and surely better solution.

Jorropo:
@Chocobozzz the jest test fail but I'm not sure that my fault cause this fail in :
```/home/travis/build/Jorropo/PeerTube/client/node_modules/lodash-es/truncate.js:1```
See https://travis-ci.org/Jorropo/PeerTube/jobs/423816317.
Is it my fault ?

**Comment by Jorropo** [comment]
I have a big problem with jest (see )
I have found this issue https://github.com/kulshekhar/ts-jest/issues/494.
But after multiple try I can't solve it, (the problem is all ready know, this with lodash, the librairy isn't compatible by default and require a special configuration).

File: client/src/app/videos/recommendations/recent-videos-recommendation.service.spec.ts [hunk 2]
```diff
@@ -16,47 +17,61 @@ describe('"Recent Videos" Recommender', () => {
       service = new RecentVideosRecommendationService(videosService)
```
bradsk88: May want to give this a different name now that the behaviour is changing.  Maybe "RecentTaggedVideosService".  (Can probably drop the "Recommendation" for the sake of brevity, and because it's implied by the interface)

client/src/app/videos/recommendations/recent-videos-recommendation.service.ts
```diff
@@ -14,26 +17,35 @@ export class RecentVideosRecommendationService implements RecommendationService
...
   constructor (
-    @Inject(VideoService) private videos: VideosProvider
+    private videos: VideoService,
+    private searchService: SearchService
   ) {
   }
 
-  getRecommendations (uuid: string): Observable<Video[]> {
-    return this.fetchPage(1)
+  getRecommendations (video: VideoDetails): Observable<Video[]> {
```
bradsk88: As mentioned in [my PR](https://github.com/Jorropo/PeerTube/pull/6), I don't think using `VideoDetails` here is the right choice at this time because it's a very large object and we only use 2 of its properties.

I would recommend changing the signature to `(uuid: UUID, tags: string[])`.

Could also create a new class `RecommendationInfo` and pass that in, instead, if you think this function signature will swell greatly over time.

client/.babelrc
```diff
@@ -0,0 +1,5 @@
+{
+  "plugins": [
+    "transform-es2015-modules-commonjs"
```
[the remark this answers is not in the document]
bradsk88: AFAIK, we're actually not explicitly using babel.  Jest seems to use it under the hood, so its configuration is handled via the `.babelrc` file.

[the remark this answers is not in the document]
bradsk88: The alternate option would be to use `lodash`, instead of `lodash-es`.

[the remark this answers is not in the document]
bradsk88: Looks like we could also maybe use [jest-preset-angular](https://github.com/thymikee/jest-preset-angular) instead of leaning on the babel config.

client/src/app/videos/+video-edit/shared/video-edit.component.html
```diff
@@ -14,7 +14,8 @@
             </div>
 
             <div class="form-group">
-              <label i18n class="label-tags">Tags</label> <span i18n>(press Enter to add)</span>
+              <label i18n class="label-tags">Tags</label>
+              <my-help i18n-preHtml preHtml="Tags are used to suggest relevant recommendations.</br>Press Enter to add a new tag."></my-help>
```
Chocobozzz: *could be used to suggest...*

Jorropo: @Chocobozzz made !

PR #1011, opened 2018-09-04
Making password change erroring more friendly

Jorropo:
If you leave the form but the 2 password is different a big red box appears to warn you (no need to click on the button).

The submit button is desactivated if the 2 password isn't the same.

### DIFF
--- a/client/src/app/+my-account/my-account-settings/my-account-change-password/my-account-change-password.component.html
+++ b/client/src/app/+my-account/my-account-settings/my-account-change-password/my-account-change-password.component.html
@@ -6,6 +6,7 @@
   <input
     type="password" id="new-password" i18n-placeholder placeholder="New password"
     formControlName="new-password" [ngClass]="{ 'input-error': formErrors['new-password'] }"
+    (change)="validateNewPassword()"  (blur)="printAnError()"
   >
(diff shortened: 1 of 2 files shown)


Chocobozzz:
I don't really like that an error message is displayed while I'm currently filling the form

Jorropo:
@Chocobozzz fixed, now the error message is printed only when you leave the form.
One bug is staying but this is angular fault not mine (cause this is working for the button).
If you taped a wrong password and leave, the message appears but after you correct it, normally you don't have to leave the forms for make the message deasapear but you have to leave it.

Chocobozzz:
What's the difference with the previous behaviour? Before this PR, we display the error when the user clicks on the button, and now it's when he leaves the confirmation input field (in theory to click on the button). So I don't think this is really useful

Jorropo:
Beffore it were needed to click on the send button, now you can't click button,
I know that not the PR of the year, it just saves one click. But I do it cause I try to change my password, I type to fast on my keyboard but don't see it and push the button, then when I see the message I was thinking : "why it let me click on the button if it knows that will fail ?" so I change that.

Chocobozzz:
Ok i understand! :+1:

PR #1014, opened 2018-09-04
add theming via css custom properties

rigelk:
and a bonus dark color theme toggle


Chocobozzz:
I don't really we want a hacky dark theme for now: a proper theme system is better

rigelk:
@Chocobozzz What if I remove the dark theme part? We can still keep the CSS variables, don't you think?

Jorropo:
I think a dark theme is a very good idea, this is so simple that it could be simply erased when creating a real theme system.

PS: For the icon a think a sun can be good for light theme and a moon for dark theme, (I also think that icon could be in print current (so moon while dark and so sun while light))

Chocobozzz:
@rigelk sure

rigelk:
@Jorropo as you may have noticed I merged - after a discussion we settled on your point :)

As for the icon I found a middle ground - with no animation yet (because simplicity first)

PR #1018, opened 2018-09-05
Fix #1017

### LINKED ISSUE #1017: Missing instances on accounts if comming from video channel
Beta.11 [peertube.jorropo.ovh](https://peertube.jorropo.ovh/)
* **Browser name/version:**
Firefox 61.0.1 (ubuntu 18)

When you click on the account in video channel tab, you should see the account of the owner.
If the owner isn't on the same instances that you are viewing, you will fall in 404 cause the instances will not be here in the url (you will go on account/user/videos instead of account/user@instances/videos)
https://peertube.jorropo.ovh/videos/watch/be1ccbdb-1d69-40b4-af90-a57e65f9540c

@Chocobozzz can you assign it to me pls :) ?

Jorropo:
This fix #1017

### DIFF
--- a/client/src/app/+video-channels/video-channels.component.html
+++ b/client/src/app/+video-channels/video-channels.component.html
@@ -13,7 +13,7 @@
         </div>
         <div i18n class="actor-followers">{{ videoChannel.followersCount }} subscribers</div>
 
-        <a [routerLink]="[ '/accounts', videoChannel.ownerBy ]" i18n-title title="Go the owner account page" class="actor-owner">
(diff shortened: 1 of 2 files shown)


Chocobozzz:
This is the same result than `ownerBy`, and it seems to work on develop: https://peertube2.cpy.re/video-channels/182f2543-46a2-484c-98b4-a3d674096a22@peertube.mastodon.host/videos

Jorropo:
@Chocobozzz ok, now I understand, while my test I havn't the needed ram to run a server + client so I use my beta.11 server.
The bug came from here

PR #1026, opened 2018-09-07
add stopwords exclusion to the tag validation

### LINKED ISSUE #1025: Ignore stop words in tags (and search?)
A stop word is a commonly used word (such as "the") that a search/tag engine should been programmed to ignore. Right now PeerTube doesn't make the difference and I've already seen a "the" tag feature on top of the overview…

Very elaborate lists exist, but that's only for maximal accuracy. A simple list should be enough so long as it is multi-language. ie:

- https://github.com/fergiemcdowall/stopword (I compared the options on the link below and found this one to be the most updated/fitting)
- https://www.npmjs.com/search?q=stopwords

rigelk:
closes #1025

### DIFF
--- a/server/helpers/custom-validators/videos.ts
+++ b/server/helpers/custom-validators/videos.ts
@@ -1,6 +1,7 @@
 import { Response } from 'express'
 import 'express-validator'
 import { values } from 'lodash'
+import * as sw from 'stopword'
 import 'multer'
 import * as validator from 'validator'
 import { UserRight, VideoPrivacy, VideoRateType } from '../../../shared'
@@ -14,7 +15,7 @@ import {
   VIDEO_STATES
 } from '../../initializers'
 import { VideoModel } from '../../models/video/video'
-import { exists, isArray, isFileValid } from './misc'
+import { exists, isArray, isFileValid, isStopWord } from './misc'
 import { VideoChannelModel } from '../../models/video/video-channel'
 import { UserModel } from '../../models/account/user'
 import * as magnetUtil from 'magnet-uri'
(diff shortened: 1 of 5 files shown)


Chocobozzz:
Do we really want remove these tags from the search or the videos? If users want to search *the*, or add this kind of tag on their video, this is their choice.

Jorropo:
I think that can be good but only for search, (and not for search tag, only search query).

rigelk:
That being said, this PR only blocks tags from being added, not from being searched. As @Jorropo said, we should also filter out stopwords from the query.

server/helpers/custom-validators/videos.ts
```diff
@@ -60,6 +61,7 @@ function isVideoNameValid (value: string) {
 
 function isVideoTagValid (tag: string) {
   return exists(tag) && validator.isLength(tag, VIDEOS_CONSTRAINTS_FIELDS.TAG)
+                     && !isStopWord(tag)
```
Jorropo: I don't think for tag is good,
If user would create a tag for `the facebook` why we must impose `facebook` ?
And that maybe useless cause I think people are intelligent and will not put a useless word in tags.

rigelk: :laughing: https://peertube2.cpy.re/search?tagsOneOf=the

To answer your question, a `the facebook` tag would be deemed valid by the function. Having two seperate tags `the` and `facebook` would reject `the`.

rigelk:
@Jorropo no need to select it depending on the language ; the lists have been crafted to have no collision and are too small to impact the performance.

Chocobozzz:
> I don't agree. Tags have to be meaningful, and it's not for no reason that stopwords exist: they are they few words that add no meaning.

It's not our job to say if a tag is meaningful or not. If someone uses `THE`, maybe it represents something we don't really understand (a band or anything else). Keeping the limit to 5 tags will help to have meaningful tag. Moreover you will have to choose the right stopwords list depending on the tag language... it's a lot of work.

Issue #921, opened 2018-08-13
Keyboard-only Navigation Issues


Fidgetcetera:
* **PeerTube version or URL:** v1.0.0-beta.10.pre.3 / https://dialup.express 
* **Browser name/version:** Mozilla Firefox for Ubuntu, 61.0.1 64 bit
* **NodeJS version:** Unclear if that's relevant at all.

* **What is the expected behaviour?** Navigating the site with a keyboard should clearly highlight what is currently selected, and be usable for all features of the site.
* **What do you see instead?** 

The sidebar button is seemingly not interactable or highlightable by using tab/shift+tab to navigate the page. Most of the sidebar links have no indication of being highlighted when tabbed to instead of moused over. 

In addition, the player controls have highlighting upon being tabbed to, but it is difficult to see for some buttons. When the settings menu is opened, it is not immediately navigable to by tabbing, and if it is possible to reach, we were not able to before the settings menu automatically closed over time.

rigelk:
I'm planning on using [:focus-visible](https://drafts.csswg.org/selectors-4/#the-focus-visible-pseudo) (with a [polyfill](https://wicg.github.io/focus-visible/demo/) for non-Firefox browsers), like describe [here](https://nelo.is/writing/styling-better-focus-states/).

rigelk:
This has mostly been fixed on the main video listing views. Please request a re-opening (or open a new issue) with more specific pain points in the interface that need fixing :slightly_smiling_face:

PR #1031, opened 2018-09-08
Use focus-visible polyfill to improve keyboard navigation

rigelk:
Only the homepage is concerned, but it should have decent keyboard
navigation support now.

related to #921


Chocobozzz:
Thanks @rigelk

PR #1033, opened 2018-09-08
Japanese Translated File

h3zjp:
It is a file translated into Japanese.


Chocobozzz:
Hi @h3zjp, thank you very much! But as it is [explained in the documentation](https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/translation.md#translation), could you report your modifications on our translation platform? (https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/translation.md#how-to)

h3zjp:
Is that so!
I did not read the Translation page.
I will update Zanata.

PR #1038, opened 2018-09-10
further enhance tools install explanation

### LINKED ISSUE #998: Provide clarity on use of tools.md and `--production` flag
* **Browser name/version:** n/a
* **NodeJS version:** n/a

* **What is the expected behaviour?** Can use tools `tools/import-videos` or other scripts regardless of which doc I follow.

On the instance I'm helping with we've been using the the `import-videos` tool. We upgraded to 1.0.0-beta.11 using `yarn install --production --pure-lockfile` and it stopped working. My friend who administrating the instance doesn't have any node background so it was not obvious at all why it wasn't working.

After I took a look it became clear that the issue was `supertest` is a dependency as test utils are used and since `supertest` is in `devDependencies` it wouldn't be installed with the `--production` flag.

The simplest resolution would be to remove the `--production` flag from the production.md docs for installing/upgrading. Because the footprint of the `devDependencies` is not of much significance I think this would be best.

joshmorel:
Resolves #998

### DIFF
--- a/support/doc/tools.md
+++ b/support/doc/tools.md
@@ -13,7 +13,7 @@
 
 You need at least 512MB RAM to run the script.
 Scripts can be launched directly from a PeerTube server, or from a separate server, even a desktop PC.
-You need to follow all the following steps even if you are on a PeerTube server (including cloning the git repository in a different directory than your production installation).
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks @joshmorel!

Issue #765, opened 2018-07-03
Separate storage from staging directory for video encoding


sachaz:
Hi,

As we have fast and slow storage (for exemple: ssd and nfs mount) it will be helpfull to **separate the incoming directory and the storage video**.
 All video uploaded will go to the incoming directory, then after being processed it will me moved on the storage directory.
 Actualy, even if the cpu is the bootleneck for encoding video, an NFS mount could slow a lot the process.

 Finaly for the (near ?) future directories separation will be needed if we can imagine that Incoming directory could be an encoding farm separated of the storage using the web frontend.

Nutomic:
The "staging directory" could also be used as a cache for videos that are watched most often. That would avoid any extra reads from slow storage (like NFS). I know there are other ways to setup file caching, but they are more complicated to setup.

Extarys:
For the videos watched most often, should it be copied/moved to the faster directory on all instances or only on the instance that actually served that video more often?

If I have 3 instances of PeerTube for example, does all 3 instances move/copy the video to the faster storage? Should it be configurable? We should also think about setting a "cache" limit and the oldest most watched video could be deleted and replaced with the new one. If the fast storage is full of cached content, users will have issues uploading new videos.

Chocobozzz:
`tmp` directory added in https://github.com/Chocobozzz/PeerTube/commit/6040f87d143a5fa01db79867ece8197c3ce7be47

We don't transcode videos in this directory, because they are directly available. So we would have to copy the video in the `tmp/` directory, transcode it, and move it in the `videos/` directory. I don't think it's worth it but don't hesitate to create another issue if you do not agree.

PR #1054, opened 2018-09-12
Basic video redundancy implementation

Chocobozzz:
For now there is only 1 strategy: `most-views` that caches videos with the most views.


Nutomic:
Are those redundant videos stored in the normal video folder? Would be useful if we could put the cache on a faster disk.

Chocobozzz:
> Are those redundant videos stored in the normal video folder? Would be useful if we could put the cache on a faster disk.

The normal video folder should be in a fast disk.

> Edit: It would also be nice if local videos could be mirrored into the cache. That way we could use slow, cheap storage for all videos, and fast storage for popular videos.

Related to https://github.com/Chocobozzz/PeerTube/issues/765

Nutomic:
Well fast disks are expensive, and the majority of videos are barely ever watched. So it makes sense to store the few popular videos on a fast disk, and the rest on a slow disk (still fast enough for streaming). If all videos have to be stored on fast disks, that would make Peertube extremely expensive to host.

Chocobozzz:
When I talk about fast disk, I mean hard disk (and not a cold storage for example).

If you want a dedicated directory for old/unused videos (so you can put it in a slow and cheap storage), please create an issue.

Nutomic:
Okay done: #1086

PR #1074, opened 2018-09-16
Add possibility to manage trust_proxy list in docker image

LecygneNoir:
Hello!

This PR add the possibility for docker image to support trust_proxy
As docker is usually behind reverse proxy (at least the docker host), I think it would be a useful addition.
I have tried to follow customer-environment-variables guideline from nodejs config and js-yaml node packages, feel free to comment :hugs: 

Thanks!


rigelk:
Thanks @LecygneNoir :+1:

PR #1082, opened 2018-09-17
Import from local storage

m333w:
Hi, 

Is it possible to have the ability to import from local storage ? 

Best regards.


Chocobozzz:
Please create an issue, you created a PR.

> Is it possible to have the ability to import from local storage ?

I don't understand. Local storage cannot contain videos

PR #1099, opened 2018-09-21
Add dot into allowed actor username

lsde:
One should have an oportunity to include a dot into the username.
Currently, it breaks the flow if one has an SSO in front of PeerTube which creates users with "name.surname".

### DIFF
--- a/server/helpers/custom-validators/activitypub/actor.ts
+++ b/server/helpers/custom-validators/activitypub/actor.ts
@@ -27,7 +27,7 @@ function isActorPublicKeyValid (publicKey: string) {
     validator.isLength(publicKey, CONSTRAINTS_FIELDS.ACTORS.PUBLIC_KEY)
 }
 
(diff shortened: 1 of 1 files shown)


lsde:
Corresponding erroir:
"stack": "SequelizeValidationError: Validation error: \"name.surname\" is not a valid actor preferred username.

lsde:
The fact is that it was possible to create such usernames before 8a19bee1a1ee39f973bb37429e4f73c3f2873cdb.
Therefore i'd count is as a regression.

We are now stuck with having ~50 users with such usernames, but cannot create new ones without actually not dropping the ones who registered earlier (to compensate for the username change)..

Chocobozzz:
The fact is that before https://github.com/Chocobozzz/PeerTube/commit/8a19bee1a1ee39f973bb37429e4f73c3f2873cdb this regex was broken... But maybe adding a dot in actor username doesn't hurt

lsde:
OK, did not know that..anyway, dot in the username is quite common..
btw, thanks you for all the amazing work!

PR #1104, opened 2018-09-22
check old password before change

### LINKED ISSUE #1081: Require old password at password change
* **What is the expected behaviour?** When a user wants to change the account’s password, the current password needs to be also submitted, to prevent attackers from changing it, if they get access to the user’s session.
* **What do you see instead?** The current password is not needed to change it.

* **Reference:** [OWASP](https://www.owasp.org/index.php/Testing_for_weak_password_change_or_reset_functionalities_(OTG-AUTHN-009)#Test_Password_Change)

BO41:
disclaimer: first "big" PR to this project

please review my changes and test them. I am not sure if this suits the coding style of the project.
Also, it has been some time since I last used angular.

screenshot_2018-09-22 account settings - peertube
screenshot_2018-09-22 account settings - peertube 1

(please ignore the key icon)

fixes #1081


BO41:
@rigelk is it okay to name the functions like this?

should they be in this file or should the password check method be moved somewhere else, more global?

is it okay to use the `authService` and the `user` thingy?
I was thinking about using some other http method, but I don't know any documentation, so I though I just grab the username as it is already saved somewhere and then reuse the login method.

rigelk:
@BO41 the function naming is fine, and using `authService` is fine. My concern is more about using the `login` method and checking its errors. Any idea @Chocobozzz ?

Chocobozzz:
I think it would be better to move the old password verification on the server side. So the client would just have to send in addition the `current-password` field in the body.

Chocobozzz:
> I assume this is not implemented yet. Will you or someone else do this, or do you think I am able to do this myself?

I think you should be able to do this: it's not more complicated than the client side. You just need to add a parameter on the server side, and check its validity. If you don't manage to do this, I'll do it myself.

> Do you plan to merge this for now? I can change it once the server side is ready for it.

No I prefer to merge the final version.

BO41:
I found `server/middlewares/validators/users.ts` with the comment "TODO: Add old password verification". And I see that `body('')` holds submitted parameters.
But how can I check if the password is equal to the old one?

This method seems to just validate if the password is theoretical correct. Or is there a login method which I could reuse?

Chocobozzz:
> But how can I check if the password is equal to the old one?

See the `isPasswordMatch` method of `user` model: github.com/Chocobozzz/PeerTube/blob/develop/server/lib/oauth-model.ts#L71

BO41:
@Chocobozzz sorry, but I am unable to implement this. It is doing things I don't understand. I don't get any error messages and I don't know what is going on.
Either you try to explain it to me again or someone else has to implement it.

BO41:
@Chocobozzz nice! looks good.
Looking at your code, it seems like what I tried wasn't that much off the real solution.

**Thanks** for fixing and finishing this :smiley: :tada:

PR #1105, opened 2018-09-22
remove unused imports

BO41:
This should remove only unused imports and variables.

please review and correct me if I missed something

btw fyi: `app.module.ts` says:
>// FIXME: https://github.com/maxisam/ngx-clipboard/issues/133

and the issue seems to be fixed


Chocobozzz:
Thanks @BO41

PR #1106, opened 2018-09-22
WIP: add possibility to configure multiple trust_proxy

LecygneNoir:
Hello!

Following PR #1074, I have realized that my code does'nt allow to use multiple IPs as trust_proxies.

As it's not possible to pass array as env variables with docker, I have used a little trick through the entrypoint to parse the trust_proxy and add it directly in production.yaml.
I have also added comments in .env to explain how to pass corresponding multiple trust_proxies

After tests, it work as expected, and is very useful if you have a proxy chain below your docker :-)

Hope it helps!

Thanks


support/docker/production/.env
```diff
@@ -3,7 +3,10 @@ PEERTUBE_DB_PASSWORD=postgres_password
 PEERTUBE_WEBSERVER_HOSTNAME=domain.tld
 PEERTUBE_WEBSERVER_PORT=443
 PEERTUBE_WEBSERVER_HTTPS=true
+# If you need more than one IP as trust_proxy
+# pass them as an space separated list:
```
rigelk: "as _a_ space separated list"

support/docker/production/docker-entrypoint.sh
```diff
@@ -11,6 +11,18 @@ cp /app/config/default.yaml /config
 cp /app/support/docker/production/config/custom-environment-variables.yaml /config
 chown -R peertube:peertube /config
 
+# Parsing trust_proxy if available
+if [ ! -z "${PEERTUBE_TRUST_PROXY}" ]; then
+    # We need to replace space by comma
+    TRUST_PROXY="[${PEERTUBE_TRUST_PROXY// /,}]"
+    # Push the trust_proxy conf to production.yaml and remove old config
+    awk -v var="${TRUST_PROXY}" 's{if(/\s*-\s*.*/) next; else s=0}  /trust_proxy:/{print "trust_proxy: " var; s=1; next} 1' /config/production.yaml > /config/temp-production.yaml
```
rigelk: this seems a bit complex to maintain. I'm also unsure about using a temporary file…

support/docker/production/docker-entrypoint.sh
```diff
@@ -11,6 +11,18 @@ cp /app/config/default.yaml /config
...
+    # We use cat to override some filesystem problem with cp
+    cat /config/temp-production.yaml > /config/production.yaml
+    rm /config/temp-production.yaml
+fi
+chown -R peertube:peertube /config
```
rigelk: then could you remove the previous call to `chown`?

support/docker/production/.env
LecygneNoir: Woops, sorry for my poor English :'(
I'll fix that asap

support/docker/production/docker-entrypoint.sh
LecygneNoir: I hope too, but I get some strange error about rights when manipulating files inside the volume. I think it could be linked to existing files the entrypoint try to modify.

I have to admit it's mysterious to me this one, perhaps a docker security things about mapped linux UID, but I have to add both chown or sometimes the start failed :/

rigelk: :frowning: that's weird

support/docker/production/docker-entrypoint.sh
LecygneNoir: Yeah, it's indeed complicated, I have to twist my head to found a proper way to do that.
Th main problem is truxt_proxy is the only option in config using an array, and not a simple strin or true/false value.

It means my previous PR is only useful with one trust_proxy. I have tested a large variety of option, passing to node env variable with [ ], spaces, comma, json formated, but no one are corrects, and worse, node crash when parsing them, causing docker to restart indefinitely...

This is the only solution I found, parsing myself the env variable to create the array, and write it directly inside the configuration file. 

As we use bash, awk seems the most adapted tool, and as we have no modern awk in the docker image, we cannot directly modify the file in place, and need to store the result somewhere before writing again the production.yaml...

More simple bash commands like cat/EOF or cut/trim/echo are worse for modifying an entire file and dealing with all cases :'(

Others options would be to modify the main code of Peertube, like allowing configuring trust_proxy via the web interface, or parsing env variable at start with node, but only for docker needs it seemed even more complicated for me ^^"

Using the entrypoint for parsing env variables are still a good practice for docker, and I tried to be less complex and less space consuming possible. But I could understand it's not trivial.

I am fairly open to any advice about this problem, with gratitude! :hugs:

support/docker/production/docker-entrypoint.sh
LecygneNoir: Yes... However I suspect this is problem specific to kubernetes environment (That I used for tests beside docker run) because of its multiple security contexts mechanisms...

While odds, I think it acceptable to be sure right are corrects ^^

support/docker/production/docker-entrypoint.sh
rigelk: @LecygneNoir I don't have a better option.

Maybe explicit with comments that the said section is only here to support the PEERTUBE_TRUST_PROXY option. This way, if anyone ever bumps into maintainability issues, they can safely remove the incriminated code.

LecygneNoir: Good idea, I'll add more details about this choice :+1:

Chocobozzz:
I think you can set JSON in this env variable (an array)

See https://github.com/lorenwest/node-config/blob/76bebd85f5cbd1dcd3e6ea3ca3b075ab85d62e29/test/1-protected-test.js#L352

LecygneNoir:
Interesting, I have tried something similar with the _json during my tests, but Peertube crashed strangely.

More precisely, Peertube starts, functions correctly, but only when trying to play a video, it crashes with an error "IP address incorrect: IP1,IP2,IP3, ...", as if it does not parse the full array but consider it as a long IP address. But it was totally possible to browse the interface, upload videos, change parameters, etc.

It only crashes when playing video, so my guess was websocket or webtorrent has some difficulties with it and I let it aside.  
However, I haven't tried a full array with `key: value` format as I don't know which key to use (in the config, from yaml point of view, it's a simple array with only value).

I'll try to take some times to give a second chance to this one. Perhaps a format as `0: IP1, 1: IP2, ...` could do the job!

LecygneNoir:
@BO41 Yes sorry, I tried to rebase from develop into my branch, which lead to this... (pretty normal now I think about it)

I am very bad with git sorry, i'll craete a new one ^^'

PR #1110, opened 2018-09-23
add redirect after login

### LINKED ISSUE #294: Only redirected to the home after logged in
* **PeerTube version or commit:** 0cd4344f3cf529b15308fcf3eb7d7eb07726df56
* **Browser name/version:** Firefox 52

Steps to reproduce:
 1. Run Peertube, unsure you're not logged in
 2. Go on some page, such as the upload one
     → You're redirected to the login page
 3. Log in
     → You should be redirected to the upload page

You're redirected to the home. You have to click on "upload" again.

BO41:
this will redirect you back to the page you wanted to go before you were redirected to the login page.

I am not sure why I need `this.redirectUrl = null`.
If I remember correctly and looking at the [example](https://angular.io/guide/router#component-less-route-grouping-routes-without-a-component) this should not be needed. But maybe the `auth.service` is not destroyed at logout.

fixes #294

### DIFF
--- a/client/src/app/login/login.component.ts
+++ b/client/src/app/login/login.component.ts
@@ -72,9 +72,9 @@ export class LoginComponent extends FormReactive implements OnInit {
   }
 
   redirect () {
-    let redirect = this.authService.redirectUrl
+    const redirect = this.authService.redirectUrl
     if (redirect) {
-      this.router.navigate([redirect])
+      this.router.navigate([ redirect ])
     } else {
       this.redirectService.redirectToHomepage()
     }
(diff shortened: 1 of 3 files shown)


client/src/app/login/login.component.ts
```diff
@@ -69,6 +71,15 @@ export class LoginComponent extends FormReactive implements OnInit {
       )
   }
 
+  redirect () {
+    let redirect = this.authService.redirectUrl
```
rigelk: `const redirect`

rigelk:
Thanks @BO41 :slightly_smiling_face:

PR #1112, opened 2018-09-23
Add warning about RAM to contributing guide

am97:
I tried to run a dev instance of PeerTube on a VPS with Fedora. I tried with 1GB of RAM, then 2GB. In both cases, running `npm run dev:server` failed when the RAM got full. 

[See the screencast](https://framadrop.org/r/WPQds1jgVe#BuAlnYIxOU63BesiMhNRtKifPe5KbMFUCjEBW4F8TwU=)

I tried then with a VPS with 4GB of RAM, and it worked.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -44,7 +44,9 @@ before you start working on them :).
 
 ### Prerequisites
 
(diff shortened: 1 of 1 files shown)


rigelk:
Thanks @am97 !

PR #1118, opened 2018-09-25
Enable Simplified Chinese support

SerCom-KC:
Also slightly changed names for both zh languages:
`中文 (繁體, 台灣)` _Chinese (Traditional, Taiwan)_ → `繁體中文（台灣）` _Traditional Chinese (Taiwan)_

Let me know if anything goes wrong :P


Chocobozzz:
Perfect! Thanks a lot @SerCom-KC!

PR #1119, opened 2018-09-26
fix style dependency paths

FlatEarthTruther:
When installing node_modules with npm it's not necessary to put `../../node_modules` ... to include/import resources from installed npm modules when importing style dependencies.

### DIFF
--- a/client/src/sass/application.scss
+++ b/client/src/sass/application.scss
@@ -1,4 +1,4 @@
-$icon-font-path: '../../node_modules/@neos21/bootstrap3-glyphicons/assets/fonts/';
+$icon-font-path: '@neos21/bootstrap3-glyphicons/assets/fonts/';
 @import '_bootstrap';
 
 @import '_variables';
--- a/client/src/sass/include/_fonts.scss
+++ b/client/src/sass/include/_fonts.scss
@@ -1,4 +1,4 @@
-$FontPathSourceSansPro: '../../node_modules/npm-font-source-sans-pro/fonts';
+$FontPathSourceSansPro: 'npm-font-source-sans-pro/fonts';
 
 @font-face {
   font-family: 'Source Sans Pro';


FlatEarthTruther:
`npm install npm-font-source-sans-pro @neos21/bootstrap3-glyphicons`

Chocobozzz:
I don't understand and it does not work in dev mode

FlatEarthTruther:
The packager should know the location of the packages if you have the packages installed. If it doesn't work in dev mode, try to make sure the packages are installed in dev mode and it should work. The thing is that removing the `../../node_modules/` from the include paths make the app more stable in other development environments. This way it's not only for your machine where your dependencies are all installed a certain way. It's just cleaner this way. npm knows where the package is and so it figures that out for you, so you don't have to give it the relative path to the package. I hope that makes sense.

FlatEarthTruther:
Here is a stackoverflow discussion about this https://stackoverflow.com/questions/28283652/importing-sass-through-npm

BO41:
@FlatEarthTruther your stack discussion suggest using the tilde operator
`@import "~some-module/sass/app";`
https://stackoverflow.com/questions/28283652/importing-sass-through-npm/43182676#43182676

and not what you suggested.
or am I wrong?

FlatEarthTruther:
It's not necessary to use the `~`, but yeah I'm suggesting to remove the `../../node_modules/` from the imports

Chocobozzz:
Without the `~` I have 

```
WARNING in ./src/sass/application.scss (./node_modules/raw-loader!./node_modules/postcss-loader/lib??embedded!./node_modules/sass-loader/lib/loader.js??ref--15-3!./src/sass/application.scss)
[0] Module Warning (from ./node_modules/postcss-loader/lib/index.js):
[0] (Emitted value instead of an instance of Error) postcss-url: /home/florian/Coding/NodeJS/PeerTube/client/src/sass/include/_fonts.scss:8:2: Can't read file '/home/florian/Coding/NodeJS/PeerTube/client/src/sass/npm-font-source-sans-pro/fonts/WOFF2/TTF/SourceSansPro-Regular.ttf.woff2', ignoring
```

With the `~` it seems to work. Did you try with `npm run dev`? If yes, it's weird it works on your side and not on my side ^^

Chocobozzz:
So I put the `~` and it seems to work well. Thanks @FlatEarthTruther it's better now :)

https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/sass/application.scss#L1

FlatEarthTruther:
Sweet! Thank you! YouTube search "AGE OF DECEIT"

PR #1128, opened 2018-09-27
Add prompt to upgrade.sh to install pre-release version

Nutomic:
The upgrade.sh script doesn't let you install pre-release versions of Peertube. This change adds a prompt if a pre-release is available. Output looks like this:
```
$ ./upgrade.sh
Which version do you want to install?
[1] v1.0.0-beta.13 (stable)
[2] v1.0.0-beta.15 (pre-release)
```


Chocobozzz:
:+1: Thanks @Nutomic

PR #1135, opened 2018-09-28
WIP: Set bitrate limits for transcoding (fixes #638)

### LINKED ISSUE #638: Playback issues at 1080p on certain longer (>7 minute) videos
beta.7
* **Browser name/version:**
Firefox 60.0.1
* **NodeJS version:**
v8.11.2

Standard seeded playback without a "large" amount of buffering mid playback or reloading in HTTP.

Playing certain videos at 1080p is causing a lot of buffering until it falls back to HTTP (always before halfway, usually earlier). Allowing it to transcode down the resolution (only tested at 480p so far) lets it automatically drop down and stops this from happening.

Examples:
- Only 1080p: https://dialup.express/videos/watch/7b503a9b-36ec-4599-a6b7-441f0cbbe013
- Both 1080p and 480p: https://dialup.express/videos/watch/f9b569e1-e34b-4245-ba40-2b6efe504112

My server has 1891.32 down and 172.86 up, so bandwidth isn't an issue on that end. The speed on my end is 61.2 down and 5.32 up, so it's probably not that either.

The only reason this seems unusual is because I haven't seen another instance have problems like this, and I don't seen any reason for (unusual) playback issues.

* **Browser console log if useful (Gist/Pastebin...):**
https://pastebin.com/gzaZYYm7 (this is from before I started using the workaround, so it couldn't lower the resolution)

Nutomic:
I haven't tested this yet, because I already spent way too much time researching all the different recommendations for video bitrates. I haven't even tested this yet...


Nutomic:
I also found out about [Adaptive bitrate streaming](https://en.wikipedia.org/wiki/Adaptive_bitrate_streaming). Does Peertube implement this already?

rigelk:
We use our own implementation of adaptative bitrate streaming.

Nutomic:
Hmm, for some reason transcoding is not working on my local dev instance. ffmpeg version is n4.0.2, is that too new?

Also, I noticed that there aren't any tests for transcoding or other helpers yet. And I don't understand Typescript well enough to write that from scratch.

rigelk:
@Nutomic hopefully we have tests https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/videos/video-transcoder.ts (and we spent some decent time on them ^^)

Your version of ffmpeg should be fine. Newer versions of ffmpeg have never introduced problems, but rather improved performance.

server/helpers/ffmpeg-utils.ts
```diff
@@ -138,6 +138,43 @@ function transcode (options: TranscodeOptions) {
...
+    case VideoResolution.H_1080P:
+      // quality according to Google Live Encoder: 3000 - 6000 Kbps
+      // Quality according to YouTube Video Info: 2788 Kbps
+      command.outputOptions('-maxrate 3M -bufsize 6M')
+      break
+    }
+
+    // Force a keyframe every 2 seconds, and on scene cuts.
+    // https://superuser.com/a/1098329
+    command.outputOptions('-force_key_frames "expr:gte(t,n_forced*2)"')
```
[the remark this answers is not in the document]
rigelk: Reading the thread linked, it is apparently only the optional part of the keyframe fix. Including the first part is mandatory: `-g X -keyint_min X`.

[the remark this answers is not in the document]
rigelk: Actually running your code without that part passes the test. Since it's out of scope of the current issue at stake, I suggest you remove it so that we can merge the rest. We can always deal with that option in another PR once you have made it work.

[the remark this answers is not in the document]
Nutomic: Yeah I noticed that too. No idea why it is failing, as the same parameter works fine on the command line.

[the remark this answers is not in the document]
rigelk: If I might suggest using this diff: https://copycat.drycat.fr/?fc042d85910bda21#oPHiEhl9PSjfLLH3z2H6H31amvd12GGpA7kvX8pr6y8=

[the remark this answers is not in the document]
rigelk: It probably clashes with some other option in use in PeerTube ; but to properly debug it from within our library to ffmpeg (`fluent-ffmpeg`), we have yet to attach a logger.

Nutomic:
Wrote a test (didn't find that file before), added a script to optimize old videos, and moved the keyframe stuff into a new issue.

Transcoding is still not working if I upload a video into the dev instance, but it works fine from the script. Transcoding is enabled in Peertube settings. Maybe it will help if I reset the database etc.

package.json
```diff
@@ -50,6 +50,7 @@
     "generate-api-doc": "scripty",
     "parse-log": "node ./dist/scripts/parse-log.js",
     "prune-storage": "node ./dist/scripts/prune-storage.js",
+    "optimize-old-videos": "node ./dist/scripts/optimize-old-videos.js",
```
rigelk: You forgot to add the script source.

Nutomic: No idea where that went. Luckily I still had it open in the editor.

Nutomic:
Thanks :)

I'm not getting any upload error, Peertube just shows "waiting for transcoding", but doesn't start FFmpeg. It might just be a problem with my setup.

Nutomic:
I just realized that the optimize videos script still has some problems:
- redundant videos from other servers from other servers should be ignored
- torrents need to be regenerated after running the script

That means simply going over the files won't be enough, and we have to get the info from the database somehow.

rigelk:
Both very good points. The prune script makes the difference between "owned" and other videos that come from the redundancy, so at least you have an example of that. Torrents are generated in https://github.com/Chocobozzz/PeerTube/blob/8c2b9756f946354c09ff7b6ce5ce7fdae1d8b2b7/server/models/video/video.ts#L1355

Nutomic:
Thanks! It seems like I can get the VideoModel with `VideoModel.loadByUUIDWithFile(uuid)`, and then call `createTorrentAndSetInfoHash()` on it. But how do I get the VideoFileModel for that? It would be really good if there were some comments in the video.ts file (which has 1500 lines!).

scripts/optimize-old-videos.ts
```diff
@@ -0,0 +1,48 @@
...
+
+// TODO: copied from prune-storage.ts, should be moved to some common file
+function getUUIDFromFilename (filename: string) {
+  const regex = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/
+  const result = filename.match(regex)
+
+  if (!result || Array.isArray(result) === false) return null
+
+  return result[0]
+}
```
Nutomic: Any idea where I can put this function so that we don't need it both in prune-storage.ts and optimize-old-videos.ts?

rigelk:
Alright. From within a failing transcoding job's logger instance, I get the following stderr (stdout is empty):

```
Stream mapping:
  Stream #0:0 -> #0:0 (vp8 (native) -> h264 (libx264))  
  Stream #0:1 -> #0:1 (vorbis (native) -> aac (native))
[aac @ 0x562aaaf0dec0] [Eval @ 0x7ffc31f4a640] Undefined constant or missing '(' in 'undefined'
[aac @ 0x562aaaf0dec0] Unable to parse option value \"undefined\"
[aac @ 0x562aaaf0dec0] Error setting option maxrate to value undefined.
Error initializing output stream 0:1 -- Error while opening encoder for output stream #0:1 - maybe incorrect parameters such as bit_rate, rate, width or height
Conversion failed!"
```

The input file is a `.webm` from the tests. Most likely this is due to you getting `metadat.format.bit_rate`, which might return `undefined` for no apparent reason… :disappointed:

scripts/optimize-old-videos.ts
rigelk: Probably in `server/helpers/utils.ts`.

Nutomic:
Okay that one should be fixed. Not sure if tests are working now because they are timing out for me, but that might just be because of my slow laptop.

I still need help figuring out how to recreate the torrent (and update the file size?). Also, are the checks in `optimize-old-videos.ts` enough to make sure it doesn't touch any files that are currently being uploaded or transcoded?

rigelk:
@Nutomic a `VideoModel` holds an array of the files related to it (the different versions of a same video, each with a different resolution/fps pair). It's accessible as `VideoModel.VideoFiles: VideoFileModel[]`. I guess we don't document it because it follows the syntax defined by sequelize (which has a rich documentation), but that's still difficult for me ;)

Nutomic:
Thanks, I think I figured it out for the most part. This is all pretty complicated because I never touched Nodejs before. So I have to look up every little thing that I wouldn't even have to think about in Java...

There are still some problems:
- It looks like I really broke the transcoding somehow, that's why the tests are failing. But I have no clue what could be the problem, and I can't see any errors reported.
- optimize-old-videos.ts:48 throws an error, for which I couldn't find any solution online: `Error: Model not initialized: "ApplicationModel" needs to be added to a Sequelize instance before "findOne" can be called.`
- Repeating from my last comment in case you missed it:  are the checks in optimize-old-videos.ts enough to make sure it doesn't touch any files that are currently being uploaded or transcoded?

shared/models/videos/video-resolution.enum.ts
```diff
@@ -1,7 +1,58 @@
...
+  case VideoResolution.H_1080P:
+    if (fps === 60) {
+      // quality according to Google Live Encoder: 3000 - 6000 Kbps
+      // Quality according to YouTube Video Info: 4387 Kbps
+      return 4400 * 1000
+    } else {
+      // quality according to Google Live Encoder: 3000 - 6000 Kbps
+      // Quality according to YouTube Video Info: 3277 Kbps
+      return 3300 * 1000
+    }
```
Nutomic: I noticed that Peertube supports 60 fps video, so I set a higher bitrate for that. Unfortunately I can't use the `VIDEO_TRANSCODING_FPS` constant from `server/initializers` because that throws an error when I import it here (probably because this is in the `shared/` code).

rigelk:
@Nutomic I leave aside the questions I can't answer as of now, but don't worry I didn't forget about them. I'll let @Chocobozzz answer them next week.

server/tests/api/videos/video-transcoder.ts
```diff
@@ -281,6 +281,53 @@ describe('Test video transcoding', function () {
     }
   })
 
+  it('Should respect maximum bitrate values', async function () {
+    this.timeout(80000)
+
+    {
+      const videoAttributes = {
+        name: 'waiting video',
+        fixture: 'video_high_bitrate_1080p.mp4',
```
rigelk: You forgot to add the fixture in `server/tests/fixtures/video_high_bitrate_1080p.mp4`.

server/helpers/ffmpeg-utils.ts
```diff
@@ -106,6 +116,10 @@ type TranscodeOptions = {
 
 function transcode (options: TranscodeOptions) {
   return new Promise<void>(async (res, rej) => {
+    if (options.resolution === undefined) {
+      return rej('Invalid video file')
```
rigelk: rather than fail on undefined resolution (something which we didn't do so far), I'd rather keep the check for `options.resolution !== undefined` below and make 

```typescript
if (targetBitrate) command.outputOptions([`-maxrate ${ targetBitrate }`, `-bufsize ${ targetBitrate * 2 }`])
```

shared/models/videos/video-resolution.enum.ts
```diff
@@ -1,7 +1,58 @@
...
+    if (fps === 60) {
+      // quality according to Google Live Encoder: 2,250 - 6,000 Kbps
+      // Quality according to YouTube Video Info: 2634 Kbps
+      return 2600 * 1000
+    } else {
+      // quality according to Google Live Encoder: 1,500 - 4,000 Kbps
+      // Quality according to YouTube Video Info: 1752 Kbps
+      return 1750 * 1000
+    }
+  case VideoResolution.H_1080P:
```
rigelk: Why not make the highest case the default case when `resolution` is undefined?

server/tests/api/videos/video-transcoder.ts
Nutomic: Right, I keep forgetting to git add new files...

server/helpers/ffmpeg-utils.ts
Nutomic: I thought this would be fine, because there is probably no case where ffprobe can't discover the resolution, but still can transcode the video. But I will change it.

server/tests/api/videos/video-transcoder.ts
rigelk: The file is very large though. A few kB is the limit for fixtures, so that they don't clutter the repository. Hopefully we can make the video as short as we want.

Nutomic: Any idea where I can get a good file for that? The current file is from http://www.jell.yfish.us/, but there is no license, so it's probably proprietary. And even if I cut the video to one second, it would be very large.

Nutomic:
Okay I can probably just convert one of the existing files.

rigelk:
@Nutomic you could generate one on the fly with ffmpeg for the test, yes.

Nutomic:
This is a bit dirty because it creates the temp file in the fixtures folder, but I couldn't see an easy way to avoid that.

server/tests/api/videos/video-transcoder.ts
```diff
@@ -285,6 +286,13 @@ describe('Test video transcoding', function () {
     this.timeout(80000)
 
     {
+      // Generate high bitrate video on the fly.
+      // https://stackoverflow.com/a/18154439
+      ffmpeg(buildAbsoluteFixturePath('video_short1.webm'))
```
rigelk: You should wrap the call to ffmpeg in a Promise

Nutomic: Done, although I'm not sure what `res(imageName)` means. It gives me an error if I put that.

Nutomic:
@Chocobozzz I know you're busy, but can you have a look at this PR? Unfortunately I can't figure out the remaining problems on my own.

scripts/optimize-old-videos.ts
```diff
@@ -0,0 +1,55 @@
...
+    const isLocalVideo = await VideoRedundancyModel.isLocalByVideoUUIDExists(uuid)
+    const isMaxBitrateExceeded = videoBitrate > getMaxBitrate(videoFileResolution, fps)
+    if (uuid && isLocalVideo && isMaxBitrateExceeded) {
+      await optimizeVideo(uuid, inputPath, videoFileResolution, isPortraitMode)
+    }
+  }
+  console.log('Finished optimizing videos')
+}
+
+async function optimizeVideo (uuid: string, inputPath: string, resolution: number, isPortraitMode: boolean) {
```
Chocobozzz: You could use https://github.com/Chocobozzz/PeerTube/blob/develop/server/lib/video-transcoding.ts#L10

For example, add an optional parameter `videoInputPath?: string` and rename the function `optimizeVideoFile`

server/tests/api/videos/video-transcoder.ts
```diff
@@ -281,7 +282,67 @@ describe('Test video transcoding', function () {
     }
   })
 
+  it('Should respect maximum bitrate values', async function () {
+    this.timeout(80000)
+
+    {
+      // Generate a random, high bitrate video on the fly, so we don't have to include
+      // a large file in the repo.
+      // https://stackoverflow.com/a/15795112
```
Chocobozzz: nice work!

server/tests/api/videos/video-transcoder.ts
```diff
@@ -281,7 +282,67 @@ describe('Test video transcoding', function () {
...
+          .on('end', () => res)
+          .run()
+      })
+
+      const videoAttributes = {
+        name: 'waiting video',
+        fixture: 'video_high_bitrate_1080p.mp4',
+        waitTranscoding: false
+      }
+
```
Chocobozzz: I suggest to ensure your generated file has a higher bitrate than `getMaxBitrate`

server/tests/api/videos/video-transcoder.ts
```diff
@@ -281,7 +282,67 @@ describe('Test video transcoding', function () {
...
+      for (const server of servers) {
+        const res = await getVideosList(server.url)
+
+        const video = res.body.data.find(v => v.name === videoAttributes.name)
+
+        for (const resolution of ['240p', '360p', '480p', '720p', '1080p']) {
+          const path = join(root(), 'test2', 'videos', video.uuid + '-' + resolution + '.mp4')
+          const bitrate = await getVideoFileBitrate(path)
+          const fps = await getVideoFileFPS(path)
+          const resolution2 = await getVideoFileResolution(path)
```
Chocobozzz: Could you `expect(resolution2 + 'p').to.equal(resolution)` please?

server/tests/api/videos/video-transcoder.ts
```diff
@@ -281,7 +282,67 @@ describe('Test video transcoding', function () {
...
+      const videoDetails: VideoDetails = res2.body
+
+      expect(videoDetails.state.id).to.equal(VideoState.PUBLISHED)
+      expect(videoDetails.state.label).to.equal('Published')
+      expect(videoDetails.waitTranscoding).to.be.true
+    }
+  })
+
   after(async function () {
+    remove(buildAbsoluteFixturePath('video_high_bitrate_720p.mp4'))
```
Chocobozzz: put the generated name in a variable

shared/models/videos/video-resolution.enum.ts
```diff
@@ -1,7 +1,59 @@
...
+
+/**
+ * Sources for individual quality levels:
+ * Google Live Encoder: https://support.google.com/youtube/answer/2853702?hl=en
+ * YouTube Video Info (tested with random music video): https://www.h3xed.com/blogmedia/youtube-info.php
+ */
+export function getTargetBitrate (resolution: VideoResolution, fps: number) {
+  switch (resolution) {
+  case VideoResolution.H_240P:
+    // quality according to Google Live Encoder: 400 - 1,000 Kbps
```
Chocobozzz: 300 - 700

shared/models/videos/video-resolution.enum.ts
```diff
@@ -1,7 +1,59 @@
+import { VIDEO_TRANSCODING_FPS } from '../../../server/initializers'
```
Chocobozzz: You must not import constants in the models.

I would suggest to create a model representing `VIDEO_TRANSCODING_FPS`, in the constants use `const VIDEO_TRANSCODING_FPS: MyModel = { ... }` and use it `function getTargetBitrate (resolution: VideoResolution, fps: number, transcodingFPS: MyModel)`

shared/models/videos/video-resolution.enum.ts
```diff
@@ -1,7 +1,59 @@
...
+export function getTargetBitrate (resolution: VideoResolution, fps: number) {
+  switch (resolution) {
+  case VideoResolution.H_240P:
+    // quality according to Google Live Encoder: 400 - 1,000 Kbps
+    // Quality according to YouTube Video Info: 186 Kbps
+    return 250 * 1000
+  case VideoResolution.H_360P:
+    // quality according to Google Live Encoder: 400 - 1,000 Kbps
+    // Quality according to YouTube Video Info: 480 Kbps
+    return 500 * 1000
```
Chocobozzz: If this is the max rate, I think we should use the max youtube rate so 1000 Kbps no (even if we `*1.5` below)? (and it's higher on vimeo for example)

Nutomic: I really wasn't sure about the numbers, and I couldn't find any great sources either. I tried to keep close to the actual bitrates Youtube uses, because they spent a lot of time optimizing it. If the bitrates are too high, it won't work well for people with slow internet (and they will keep using youtube).

Chocobozzz: Yes but we talk about the max bitrate here, not the overall bitrate. Maybe we should use `-b:v` ffmpeg options instead of `maxrate`? I'm afraid with `maxrate`, we slaughter certain sequences of the video when it finally has an acceptable overall bitrate

Nutomic:
FYI I'm getting the following error when I compile, haven't changed anything in that file:
```
[0] ERROR in ../server/lib/activitypub/crawl.ts(4,1): error TS1202: Import assignment cannot be used when targeting ECMAScript modules. Consider using 'import * as ns from "mod"', 'import {a} from "mod"', 'import d from "mod"', or another module format instead.
```

The code in that line is `import Bluebird = require('bluebird')`. Changing that line to `import * as Bluebird from 'bluebird'` fixes the compile for me (but I didn't test any ActivityPub functionality.

Also, transcoding is working now. So you can ignore my previous post, just one question:  are the checks in optimize-old-videos.ts enough to make sure it doesn't touch any files that are currently being uploaded or transcoded?

Nutomic:
Okay here are some tests with the current code. I imported the videos into Peertube, with transcoding enabled for all resolutions. Bitrate is according to ffprobe output.

For the first test (high bitrate) I used the file `jellyfish-10-mbps-hd-h264.mkv` from http://www.jell.yfish.us/. After conversion to mp4, it has a bitrate of 9955 kb/s. Transcoded vide bitrates:
```
240p: bitrate: 258 kb/s
360p: bitrate: 515 kb/s
480p: bitrate: 925 kb/s
720p: bitrate: 1796 kb/s
1080p: bitrate: 3386 kb/s
```

For the second test (low bitrate), I generated a file with `ffmpeg -f lavfi -i testsrc -t 5 -vf scale=1920:1080 -pix_fmt yuv420p testsrc.mp4`. Bitrate is 223 kb/s. Transcoded vide bitrates:
```
240p: bitrate: 45 kb/s
360p: bitrate: 65 kb/s
480p: bitrate: 78 kb/s
720p: bitrate: 130 kb/s
1080p: bitrate: 258 kb/s
```

So the maximum bitrate is really close to Youtube values (even though it's a different video), and it scales down properly for low quality videos. Video quality also appears fine to me (that's subjective of course).

shared/models/videos/video-resolution.enum.ts
Nutomic: The site I linked in the code [1] specifically mentions using maxrate and bufsize for streaming, as does the FFmpeg documentation [2]. Unless you have a different source that recommends b:v for streaming, I think we should go with this.

[1] https://slhck.info/video/2017/03/01/rate-control.html
[2] https://trac.ffmpeg.org/wiki/EncodingForStreamingSites

shared/models/videos/video-resolution.enum.ts
Nutomic: I'm not sure how you want this to be implemented exactly. Right now I have this model, should I just return that from `ffmpeg-utils.getVideoFileFps()`? That means FPS would always be exactly 10/30/60, but nothing in between.
```
export enum VideoTranscodingFps {
    FPS_60 = 60,
    FPS_30 = 30,
    FPS_10 = 10
  }
```

Nutomic:
How can I run a single test for Peertube? Solutions I found online aren't working. I can't fix the test right now because some other test is failing right in the beginning (which seems entirely unrelated).

rigelk:
@Nutomic I'm usually running a single test via `npm run build:server;npm run mocha -- --exit --require ts-node/register/type-check --bail server/tests/api/videos/video-transcoder.ts` for instance.

Nutomic:
Wow that's extremely complicated, but it works. Would be good to have that in the documentation somewhere, and/or simplify it.

rigelk:
I guess we could simplify it, but meanwhile it is already in the documentation: https://github.com/Chocobozzz/PeerTube/blob/develop/.github/CONTRIBUTING.md#unit-tests

shared/models/videos/video-resolution.enum.ts
Nutomic: I see, then I replace `if (fps == 60)` with `if (fps == fpsTranscodingConstants.MAX)`. 

There's one problem related to this: if a video has 60 fps, it will be converted with 4400 Kbps, but if it has 59 fps, it only gets 3300 Kbps. But maybe this is something to be fixed later, this PR is already huge.

Chocobozzz: :+1: 

You will have to use KEEP_ORIGIN_FPS_RESOLUTION_MIN too :)

Nutomic: Hmm I'm not sure what bitrate should be used for 480p 60 fps video etc. How about we leave this for the future fps rework? I'd like to finish this PR at some point :P

Chocobozzz:
> The tests are now working, but I had to generate a rather long video (10 seconds), so the rate limits actually work with FFmpeg. This also makes the test slower.
Also, the test fixture is not deleted if the test fails. Is there any way to do that, or should I add it to .gitignore?

We will implement a cache, so put it in the gitignore.

shared/models/videos/video-resolution.enum.ts
Nutomic: Fixed :)

Nutomic:
Just rebased on develop. Tests are still failing, but that seems completely unrelated to my changes...

server/helpers/ffmpeg-utils.ts
```diff
@@ -138,6 +148,12 @@ function transcode (options: TranscodeOptions) {
       command = command.withFPS(fps)
     }
 
+    // Constrained Encoding (VBV)
+    // https://slhck.info/video/2017/03/01/rate-control.html
+    // https://trac.ffmpeg.org/wiki/Limiting%20the%20output%20bitrate
+    const targetBitrate = getTargetBitrate(options.resolution, fps, VIDEO_TRANSCODING_FPS)
+    command.outputOptions([`-b:v ${ targetBitrate }`, `-maxrate ${ targetBitrate }`, `-bufsize ${ targetBitrate * 2 }`])
```
Chocobozzz: One small question: what happens if the input file has an average bitrate < targetBitrate?

Nutomic:
Those are maximum values, so if possible, it transcodes with a lower bitrate. See my tests [here](https://github.com/Chocobozzz/PeerTube/pull/1135#issuecomment-426868150).

Thanks for all the help figuring this stuff out, @rigelk @Chocobozzz :)

server/helpers/ffmpeg-utils.ts
rigelk: @Nutomic won't the `-b:v ${ targetBitrate }` increase the bitrate of the video if it is lower than the targetBitrate? I feel like we should have just kept `maxrate` and `bufsize` to benefit from ffmpeg's automatic tuning.

Nutomic: Damn you're right. I just checked and it works fine if you remove the `-b:v` parameter (tested with low bitrate, high bitrate and very high bitrate like in the unit test). I was having some bug during development, where FFmpeg generated videos with a too high bitrate for no reason, but that is gone now.

Can one of you just make a commit to remove that parameter?

rigelk: Sure :) no worries, it's my fault for not double checking that

rigelk:
Should be fixed in e1d7b98bc762bf3a5e6590f827f67187713fb27e

Chocobozzz:
Perfect thanks @Nutomic & @rigelk

PR #1144, opened 2018-09-28
Allow specification of Channel ID in peertube-upload.js

### LINKED ISSUE #1134: peertube-upload.js to a specific channel
When uploading a folder of videos, I'd like to be able to tell the script to put them on a specific channel using a channel ID. The upload script does not seem to support this however. Is it supported on the upload endpoint?

If so, would appreciate script support for it :)

anoadragon453:
Closes #1134.

This PR creates a new command line flag for the `peertube-upload.js` script, `-C, --channel-id` that allows one to specify a channel to upload to rather than the default.

This was very helpful for my own scripts of uploading hundreds of videos to a custom channel, and I hope others will benefit from it too.

### DIFF
--- a/server/tools/peertube-upload.ts
+++ b/server/tools/peertube-upload.ts
@@ -15,6 +15,7 @@ program
   .option('-P, --privacy <privacy_number>', 'Privacy')
   .option('-N, --nsfw', 'Video is Not Safe For Work')
   .option('-c, --category <category_number>', 'Category number')
+  .option('-C, --channel-id <channel_id>', 'Channel ID')
   .option('-m, --comments-enabled', 'Enable comments')
   .option('-l, --licence <licence_number>', 'Licence number')
   .option('-L, --language <language_code>', 'Language ISO 639 code (fr or en...)')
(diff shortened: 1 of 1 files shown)


server/tools/peertube-upload.ts
```diff
@@ -105,7 +106,7 @@ async function run () {
 
   console.log('Uploading %s video...', program[ 'videoName' ])
 
-  const videoAttributes = {
+  let videoAttributes:any = {
```
rigelk: You don't need to remove the `const` nor the type inference. You could assign `program['channelId']` beforehand.

anoadragon453: But what if the user did not supply the channelId on the command line? What default value should we give?

rigelk: It should have value `undefined`.

anoadragon453: Oh I see, will fix thanks!

anoadragon453:
Updated with the fix :)

Chocobozzz:
Thanks @anoadragon453

PR #1148, opened 2018-09-29
show last commit hash alongside server version in footer

### LINKED ISSUE #835: Show last commit hash with server version
Currently we show the current version of Peertube but in some cases it is not precise enough. Users should be able to give the last commit hash when they report an issue so that we can identify what exact version they run.

rigelk:
resolves #835 and actually shows the tag name if different

### DIFF
--- a/server/controllers/api/config.ts
+++ b/server/controllers/api/config.ts
@@ -10,6 +10,7 @@ import { customConfigUpdateValidator } from '../../middlewares/validators/config
 import { ClientHtml } from '../../lib/client-html'
 import { auditLoggerFactory, CustomConfigAuditView, getAuditIdFromRes } from '../../helpers/audit-logger'
 import { remove, writeJSON } from 'fs-extra'
+import { version } from '../../tools/cli'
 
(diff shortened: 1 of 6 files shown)


Chocobozzz:
:+1:

server/controllers/api/config.ts
```diff
@@ -10,6 +10,7 @@ import { customConfigUpdateValidator } from '../../middlewares/validators/config
 import { ClientHtml } from '../../lib/client-html'
 import { auditLoggerFactory, CustomConfigAuditView, getAuditIdFromRes } from '../../helpers/audit-logger'
 import { remove, writeJSON } from 'fs-extra'
+import { version } from '../../tools/cli'
```
rigelk: @Chocobozzz argh that should have been `import { getVersion } from '../../helpers/utils'` (and then modify the call below)

Chocobozzz: https://github.com/Chocobozzz/PeerTube/commit/913f08f2bd5b03c121ee65751e938e4faa27e7f1

PR #1155, opened 2018-10-01
using "diff -u", for a clearer display

JohnXLivingston:
I think "diff -u" is more readable than "diff" to see changes in config files.


Chocobozzz:
Thanks @JohnXLivingston

Issue #1131, opened 2018-09-27
Angular configuration options (still missing noImplicitAny)


BO41:
The angular/ts allows some settings that might be good to change (for better performance):

- [`"noImplicitAny": true,`](https://angular.io/guide/typescript-configuration#noimplicitany-and-suppressimplicitanyindexerrors)
    >When the noImplicitAny flag is true and the TypeScript compiler cannot infer the type, it still generates the JavaScript files, but it also reports an error. Many seasoned developers prefer this stricter setting because type checking catches more **unintentional errors** at compile time.

- [`--build-optimizer`](https://angular.io/guide/deployment#optimize-for-production) (I guess this is not used; but may be enabled by default)
   >You may further **reduce bundle sizes** by adding the build-optimizer flag.

- [`strictInjectionParameters`](https://angular.io/guide/aot-compiler#strictinjectionparameters)
    >When set to true, this options tells the compiler to report an error for a parameter supplied whose injection type cannot be determined. When this option is not provided or is false, constructor parameters of classes marked with @Injectable whose type cannot be resolved will produce a warning.
Note: It is **recommended** to change this option explicitly to **true** as this option will default to true in the future.

- [`fullTemplateTypeCheck`](https://angular.io/guide/aot-compiler#fulltemplatetypecheck)
    >This option tells the compiler to enable the binding expression validation phase of the template compiler which uses TypeScript to validate binding expressions.
This option is false by default.
Note: It is **recommended** to set this to **true** because this option will default to true in the future.

please ignore everything that is already in use

Chocobozzz:
> --build-optimizer (I guess this is not used; but may be enabled by default)

https://github.com/Chocobozzz/PeerTube/blob/develop/client/angular.json#L44

I enabled the other options in https://github.com/Chocobozzz/PeerTube/commit/07524e229fa88a0d01af9d52ce124760f8ca5fa4

For the `noImplicitAny` it would be nice, but requires more work.

BO41:
@Chocobozzz okay. seems like either I didn't knew all the ways to activate options or angular changed since the last time I used it.
good to know that `--build-optimizer` was already in use.

I could open an PR for `noImplicitAny`. should be easy to fix, just take some time and work

Chocobozzz:
:+1:

PR #1137, opened 2018-09-28
noImplicitAny flag activated

BO41:
this enables the `noImplicitAny` flag

Imports are fixed following this [hint](https://stackoverflow.com/questions/44911538/angular-2-auth0-could-not-find-a-declaration-file-for-module-auth0-js/44911992#44911992)
I hope this won't break anything

>When the noImplicitAny flag is true and the TypeScript compiler cannot infer the type, it still generates the JavaScript files, but it also reports an error. Many seasoned developers prefer this stricter setting because type checking catches more unintentional errors at compile time.

closes: #1131

PR #1157, opened 2018-10-01
No implicit any flag true

BO41:
this enables the `noImplicitAny` flag

Imports are fixed following this hint
I hope this won't break anything

> When the noImplicitAny flag is true and the TypeScript compiler cannot infer the type, it still generates the JavaScript files, but it also reports an error. Many seasoned developers prefer this stricter setting because type checking catches more unintentional errors at compile time.

closes: #1131
replaces #1137


client/src/polyfills.ts
```diff
@@ -18,42 +18,45 @@
...
 
 /** IE10 and IE11 requires the following for the Reflect API. */
 
 // For Google Bot
-import 'core-js/es6/reflect';
+import 'core-js/es6/reflect'
 
-/** Evergreen browsers require these. **/
+/**
+ * Evergreen browsers require these.
```
Jorropo: Why using a multiple line comment here ?

BO41: thats the code style. but for some reason it wasn't used in this file, so because I came across it, I changed to fulfill the style rules

BO41: @Jorropo of course I could have replaced it with `// Evergreen browsers require these.`

BO41:
I am still having some trouble with imports

>Could not find a declaration file for module 'video.js'.

also `Window` has no index signature
for example here `window[ 'PeerTubePlayer' ] = PeerTubePlayer`

BO41:
down to just a few errors of those types:
```
src/assets/player/settings-menu-button.ts(249,24): error TS2503: Cannot find namespace 'videojs'.
src/standalone/player/player.ts(194,1): error TS7017: Element implicitly has an 'any' type because type 'Window' has no index signature.
```

using `const videojs = require('video.js')`  lets this fail `player: videojs.Player`

BO41:
damn it, broke something again

BO41:
problems solved with

>When the noImplicitAny flag is true, you may get implicit index errors as well. Most developers feel that this particular error is more annoying than helpful. You can suppress them with the following additional flag:

`"suppressImplicitAnyIndexErrors":true`

BO41:
I don't know how to solve the problem of using `videojs.Player`. Otherwise no problems occur anymore

McFlat:
`./node_modules/typescript/lib/lib.dom.d.ts` has `declare var window: Window;` while `./node_modules/@types/jschannel/index.d.ts` has `window: any;`

BO41:
@McFlat I am not 100% sure about the whole thing. But if you are, probably this would be the best option to make a PR at their end.

Since it isn't a problem for me and this PR anymore, _I_ would just leave it.

McFlat:
I've made a PR at their end to try to resolve this. I only made a simple change for now, but if they say to update anything else I'll do it, possibly still have to update the tests.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29711

McFlat:
I'm not sure we should have `"suppressImplicitAnyIndexErrors":true,` in the tsconfig.json, it could allow for many headaches later in the future

BO41:
@McFlat why is this flag bad?
It only suppresses some of the errors `noImplicitAny` throws. Nothing more and nothing less, if I know it correctly.

BO41:
@Chocobozzz @rigelk could you please review this PR? :)
I would prefer not to leave it open for long

McFlat:
The reason is because then it won't report errors with some other parts of the application and it will cause bugs, it makes the check much looser.

https://egghead.io/lessons/typescript-increase-typescript-s-type-safety-with-noimplicitany

McFlat:
The `types/jschannel/tsconfig.json` includes `noImplicitAny: true`

McFlat:
I think I just managed to fix the upstream, give me a minute

McFlat:
Okay for sure tests pass on upstream now with the changes I made, so lets wait until they approve it and your changes should work without the `"suppressImplicitAnyIndexErrors":true,`

McFlat:
screen shot 2018-10-14 at 6 32 58 pm

BO41:
@McFlat thanks, I will try to remove the flag, once your changes upstream got merged.
But I am still not sure if using this flag is that bad.

>When the noImplicitAny flag is true, you may get implicit index errors as well. Most developers feel that this particular error is more annoying than helpful. You can suppress them with the following additional flag:
"suppressImplicitAnyIndexErrors":true
The documentation setup sets this flag to true as well.

even [angular.io](https://angular.io/guide/typescript-configuration#noimplicitany-and-suppressimplicitanyindexerrors) "recommends" this

BO41:
@McFlat I assume the changes are live. For me it says `var window: Window`.

But the error is still there. And if I understood your changes correctly, this was clear.
Because you didn't change the index signature.
> ERROR in src/standalone/player/player.ts(194,1): error TS7017: Element implicitly has an 'any' type because type 'Window' has no index signature.

it is not about the type of window. it is about `window[ "someindex" ]`.

I fixed some of the errors like here https://github.com/Chocobozzz/PeerTube/pull/1157/files#diff-071d4fc0632db60808fbbbb4f6775966R19

McFlat:
@BO41 Yeah the fix was merged :)

screen shot 2018-10-15 at 11 37 46 pm

I found this stackoverflow discussion, maybe it can help you:
https://stackoverflow.com/questions/42193262/element-implicitly-has-an-any-type-because-type-window-has-no-index-signatur#42193905

McFlat:
Maybe I needed to add `[key: string]: any` to window? I need to get some sleep.

McFlat:
Yeah I'll make another pull request very soon.

McFlat:
Pull request created on the `Microsoft/TypeScript` repo https://github.com/Microsoft/TypeScript/pull/27924

McFlat:
All tests pass. Let's give it a few hours or days to wait for them to merge these changes.

screen shot 2018-10-16 at 11 12 30 am

rigelk:
Right now we are using Typescript 2.5.2. They would have to make a release and us adapt our code to Typescript's latest version for this merge to be actionable. All of this will take time, so in the meantime I suggest we merge this PR and open another one once both conditions are met.

McFlat:
If this was an issue in many places, more than one, then ok we could leave it, but since we are only adding that option to resolve a single issue it could creep up on us later and get out of hand, so better to just avoid it altogether if we can.

McFlat:
Here's what's happening with that pull request. I ask you guys is it valid what he says? or am I wrong? So far as I know what he says has nothing to do with what the pull request fixes.

screen shot 2018-10-16 at 4 53 05 pm

McFlat:
I'll keep trying to get them to fix the issue with window, seems to me they got too much pride over there in those parts

BO41:
@McFlat 
>Please trust that the repo maintainers know what they're talking about. We have dealt with tens of thousands of issues and PRs and understand what code changes do.

just let it be.
I am pretty sure they know what they are talking about.
You tried your best, and it seems that your suggested change isn't the right one.

McFlat:
@BO41 ok sounds good, I'll just leave it alone. It just seems wrong to me that they lock the window from adding index, when in reality the window can do that. If not then there would be an error, but according to MicroSoft windows have to be locked/closed, while in the real world they are open.

McFlat:
Let change `window[ 'PeerTubePlayer' ] = PeerTubePlayer` to `window.PeerTubePlayer = PeerTubePlayer` that will allow us to remove the `"suppressImplicitAnyIndexErrors":true,`

screen shot 2018-10-19 at 12 07 19 pm

PR #1162, opened 2018-10-01
add pid file in ExecStartPost to peertube systemd service

FlatEarthTruther:
There was a missing step in the systemd service, need to also create a pid file /var/run/peertube.pid

### DIFF
--- a/support/systemd/peertube.service
+++ b/support/systemd/peertube.service
@@ -9,6 +9,7 @@ Environment=NODE_CONFIG_DIR=/var/www/peertube/config
 User=peertube
 Group=peertube
 ExecStart=/usr/bin/npm start
+ExecStartPost=/bin/sh -c 'umask 022; pgrep peertube > /var/run/peertube.pid'
 WorkingDirectory=/var/www/peertube/peertube-latest
 StandardOutput=syslog
 StandardError=syslog


Chocobozzz:
Why do we need this pid file?

Jorropo:
Maybe this is better if peertube it self write this pid file, but that make an other problem, how lunch 2 instances at the same server, and what if we don't would it (like for Dev)

rigelk:
We are already setting a PID file in https://github.com/Chocobozzz/PeerTube/blob/develop/support/init.d/peertube

rhaamo:
I'm not sure that that ExecStartPost to create a PID is useful for Systemd as it auto-tracks PID by itself (and then allows to have multiple instances of PeerTube on the same server).
Itself this PID creation I don't see why and what is it purpose.

It is however required by init.d but that's another story.

FlatEarthTruther:
I added this because init.d sets a pid and systemd didn't set one, it would be useful to use from other tools to find what pid of peertube is, like from shell scripts that aren't part of peertube source code

rhaamo:
init.d set it because it needs it by design.

if you want to get the PID, systemd auto-track it and report it :
```
dashie@videaudio:~$ systemctl status peertube
● peertube.service - PeerTube daemon
   Loaded: loaded (/etc/systemd/system/peertube.service; enabled; vendor preset: enabled)
   Active: active (running) since Sat 2018-09-22 22:57:07 CEST; 1 weeks 2 days ago
 Main PID: 11007 (npm)
    Tasks: 21 (limit: 4915)
   CGroup: /system.slice/peertube.service
           ├─11007 npm
           ├─11017 sh -c node dist/server
           └─11018 peertube
```

I'm not sure it's a general use-case and needed thing, plus it make yet-another-thing-to-change if you want to run multiple peertube.

rigelk:
Like @rhaamo said - we wouldn't want to limit administrators to running one instance of PeerTube - plus you can easily get the PID of a service via `systemctl show --property="MainPID" <service>`, so this seems like a case covered by SystemD already.

Issue #951, opened 2018-08-22
Optionally allow multiple concurrent transcodes


scanlime:
On servers with many cores, it's possible for a single ffmpeg process to leave the machine idle even with a large number of threads.

I mentioned this in  #656. On the lower quality outputs especially, the x264 process has trouble keeping the 8 logical CPUs on my server occupied. For higher quality encodes, I can use all cores somewhat reliably but this requires a higher threads count around 12 or 16.

I think we'd be better off utilizing larger servers by capping the individual ffmpeg process's thread count at 4 or so, and running several concurrent ffmpeg processes.

scanlime:
I've tried modifying dist/server/initializers/constants.js to change JOB_CONCURRENCY['video-file'].

Testing with 3 concurrent encodes and 8 threads each. Server is still responsive and it's finally at 100% CPU.

scanlime:
As another data point: I've set up https request latency logging, and it seems like (on a physical server, at least) there's no noticeable latency penalty for overfilling the machine with transcode threads. There's close to 0% CPU idle and about 50% nice / 50% user in my system currently.

rigelk:
@scanlime maybe that's due to the niceness factor we apply to our ffmpeg process.

FlatEarthTruther:
I've created a pull request #1169 to allow admins to modify job attempts and job concurrency from the config file instead of having them hard coded.

FlatEarthTruther:
#1185

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/pull/3717

PR #1169, opened 2018-10-02
added job.attempts and jobs.concurrency to config file

FlatEarthTruther:


Jorropo:
I think that could be use full for tracker but I don't see why an admin will need to change job except for concurrent.video-file which is realeated to #951.
And that maybe could be changed  dynamically in the configuration tab.

FlatEarthTruther:
I've done as suggested #1185

PR #1172, opened 2018-10-03
fix menu to not overlap content on small screens

FlatEarthTruther:
remove margin-left: 0; from .main-col on max-width: 800px as it overrides a already good style setting which doesn't break the view on small screens. When toggle menu it makes the content shift over as intended for the menu to work, instead of overlap.


Chocobozzz:
Under 800px, the menu is absolute because we don't have enough width to display both the menu and the content. It's the reason why we don't need the `margin-lefŧ` anymore

PR #1185, opened 2018-10-03
add transcoding.concurrency to config

FlatEarthTruther:
Instead of hardcoding this value in constants would be better to allow for admins to change it from the config

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -120,6 +120,7 @@ user:
 transcoding:
   enabled: true
   threads: 1
+  concurrency: 1
   resolutions: # Only created if the original video has a higher resolution, uses more storage!
     240p: false
     360p: false
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -133,6 +133,7 @@ user:
 transcoding:
   enabled: true
   threads: 1
+  concurrency: 1
   resolutions: # Only created if the original video has a higher resolution, uses more storage!
     240p: false
     360p: false
--- a/server/initializers/constants.ts
+++ b/server/initializers/constants.ts
@@ -107,7 +107,7 @@ const JOB_CONCURRENCY: { [ id in JobType ]: number } = {
   'activitypub-http-fetcher': 1,
   'activitypub-follow': 3,
   'video-file-import': 1,
-  'video-file': 1,
+  'video-file': config.has('transcoding.concurrency') ? config.get<number>('transcoding.concurrency') : 1,
   'video-import': 1,
   'email': 5,
   'videos-views': 1


Chocobozzz:
Please update the client too: https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/%2Badmin/config/edit-custom-config/edit-custom-config.component.html#L200 :)

FlatEarthTruther:
Ok I'll have a look about how to do it and get it done, thanks

FlatEarthTruther:
Let me know it that is good, maybe there's something I missed. I've never done this before :) 🍻 Always learning something new and never stopping to seek the truth.

config/default.yaml
```diff
@@ -120,6 +120,9 @@ user:
 transcoding:
   enabled: true
   threads: 1
+  concurrency: 1
+  niceness: 15
+  ttl: 48 # 2 days, transcoding could be long
```
Chocobozzz: Use a string instead so users can specify '1 hour' or '2 days' (take example on `check_interval`)

FlatEarthTruther: ok sounds good

FlatEarthTruther: once it parses it on save it becomes an int and I can't seem to make it stay the duration string on save unless theres another function i don't know about yet which turns a int into a duration

FlatEarthTruther:
looks like we'll need another utility function to convert a int to a duration maybe call it like `toDuration(value: any)`

FlatEarthTruther:
I think parseDuration function is also missing case where if users give a string `true` or `false`, in that case we could use `1` or `0` respectively, also probably should strip any quotes out of the string in case they pass the string with quotes

FlatEarthTruther:
looks like [luxon](https://moment.github.io/luxon/) doesn't even have this, maybe with formatting it does idk
https://moment.github.io/luxon/docs/class/src/duration.js~Duration.html

client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
```diff
@@ -49,6 +52,31 @@ export class EditCustomConfigComponent extends FormReactive implements OnInit {
...
+      { value: 6, label: '6' },
+      { value: 7, label: '7' },
+      { value: 8, label: '8' },
+      { value: 9, label: '9' },
+      { value: 10, label: '10' },
+      { value: 11, label: '11' },
+      { value: 12, label: '12' },
+      { value: 13, label: '13' },
+      { value: 14, label: '14' },
+      { value: 15, label: '15' },
```
Jorropo: This from -20 to 19 (included), but user shouldn't place < 0 (cause if it does server will be stuck) so 1 to 19.
A text box with a form validator requiring a number > 0 and < 20 will be better. Because a 19 elements list is too big.

client/src/app/shared/forms/form-validators/custom-config-validators.service.ts
```diff
@@ -78,5 +81,29 @@ export class CustomConfigValidatorsService {
...
+        'min': this.i18n('Transcoding concurrency must be greater or equal to 1.')
+      }
+    }
+
+    this.TRANSCODING_NICENESS = {
+      VALIDATORS: [ Validators.required, Validators.min(1), Validators.max(15) ],
+      MESSAGES: {
+        'required': this.i18n('Transcoding niceness is required.'),
+        'min': this.i18n('Transcoding niceness must be greater or equal to 1.'),
+        'max': this.i18n('Transcoding niceness must be less than or equal to 15.')
```
Jorropo: Than 19 but very not recommended (it really slow down the process).

client/src/app/shared/forms/form-validators/custom-config-validators.service.ts
```diff
@@ -78,5 +81,29 @@ export class CustomConfigValidatorsService {
...
+      }
+    }
+
+    this.TRANSCODING_TTL = {
+      VALIDATORS: [ Validators.required ],
+      MESSAGES: {
+        'required': this.i18n('Transcoding ttl duration is required.')
```
Jorropo: Maybe an example like :
`2 days` or `12 hours`

PR #1187, opened 2018-10-03
Rewrite issue template

Nutomic:
The old order of items didn't really make sense to me.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -21,4 +21,4 @@
 * NodeJS version:
 
 * Link to browser console log if useful:
-* Link to server log if useful (journalctl or /var/www/peertube/storage/logs):
+* Link to server log if useful (journalctl or /var/www/peertube/storage/logs/):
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -2,15 +2,23 @@
...
+3.
+
+
+**Additional information**
+* PeerTube version or URL:
+* Browser name/version:
+* NodeJS version:
+
+* Link to browser console log if useful:
+* Link to server log if useful (journalctl or /var/www/peertube/storage/logs):
```
Jorropo: Maybe if /var/www/peertube/storage/logs finish with a / that will be more clear (for indicate that the logs are in the folder, not a single file).

Chocobozzz: :+1:

Nutomic: Makes sense, fixed :)

Chocobozzz:
> I would even add another section below the "bug report form", for "features requests" because they happen and should be a bit more structured too, if possible

I would prefer to keep the template as simple as possible, and we need to focus on bugs (that are the most difficult to reproduce and to fix).

Issue #1188, opened 2018-10-03
Feature: Allow to create usernames longer than 20 chars


FlatEarthTruther:
Some users have much longer usernames, I think would be better to limit at 32 chars.

<img width="379" alt="screen shot 2018-10-03 at 5 04 40 pm" src="https://user-images.githubusercontent.com/198260/46442445-48964c80-c72f-11e8-98d7-b5dc3c4f26b3.png">

Chocobozzz:
Why 32? Why not 50 for example?

FlatEarthTruther:
Because 32 is already pretty long enough to take into account a good amount of usernames that users may want to use, and it would fit into one line without doing some word-wrap to the next line, whereas 50 would have potential for word-wrap and make content shift down if some videos have shorter usernames and some longer things may be off balance in the frontend. 32 is a pretty good length, but giving the admins the ability to modify it from the config would be nice too

Jorropo:
@FlatEarthTruther I have just a question, how would you manage compatibility with old instances ?
Cause if you provide to an not already updated instances an 32 long username it will refuse it.

Nutomic:
FYI Youtube also limits the length of usernames to 20 characters, so I don't see why Peertube would need a higher limit.

https://stackoverflow.com/a/12738721

FlatEarthTruther:
Maybe that was with their V1 API, but with the V3 API things have changed, to be max 60 characters. https://stackoverflow.com/questions/35372320/youtube-and-twitch-channel-name-maximum-character-limit

PR #1189, opened 2018-10-03
add fields.users.username.max to config

FlatEarthTruther:
issue #1188 allow username to be longer than 20 chars, giving option to administrators to control this from config

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -163,3 +163,8 @@ services:
     # If false, we use an image link card that will redirect on your PeerTube instance
     # Change it to "true", and then test on https://cards-dev.twitter.com/validator to see if you are whitelisted
     whitelisted: false
+
+fields:
+  users:
+    username:
+      max: 20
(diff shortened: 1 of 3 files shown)


FlatEarthTruther:
The reason for the deep nesting is to allow for further to add options later in the future, when more fields and constraints are added in later versions.

I left 20 as the default as before, but we should ideal make the default to be 32 chars for usernames.
```
fields:
  users:
    username:
      max: 20
```

Chocobozzz:
Like @Jorropo said :+1:

FlatEarthTruther:
Maybe in the database set the limit to 128 or 64 and make that a hard limit, whereas the admin can set a soft-limit on the username max length.

FlatEarthTruther:
if 32 could be the new limit it would be ideal, since most usernames would be within that limit, I have downloaded many account videos and some accounts have long username, but for the most part they fit within the 32 char limit, so for it to be limited to 20 chars affects a lot of accounts

Issue #1171, opened 2018-10-02
Unable to build server from local dev box or from peertube server


FlatEarthTruther:
I've tried to build the server code both on my mac and on the debian 9 server that is currently running 1.0.0-beta-15 without success. It fails on the migrations part.

```
> peertube@1.0.0-beta.16 build /var/www/peertube/PeerTube
> SCRIPTY_PARALLEL=true scripty


> peertube@1.0.0-beta.16 tsc /var/www/peertube/PeerTube
> tsc


> peertube-client@1.0.0-beta.16 ng /var/www/peertube/PeerTube/client
> ng "build" "--output-path" "dist/en_US/" "--deploy-url" "/client/en_US/" "--prod" "--stats-json"

 10% building modules 3/3 modules 0 active(node:25411) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
 69% building modules 813/814 modules 1 active ...aption-add-modal.component.ngfactory.jsserver/controllers/api/videos/import.ts:219:3 - error TS2322: Type 'Bluebird<any>' is not assignable to type 'Bluebird<VideoImportModel>'.
  Types of property 'then' are incompatible.
    Type '{ <U>(onFulfill?: (value: any) => Resolvable<U>, onReject?: (error: any) => Resolvable<U>): Blueb...' is not assignable to type '{ <U>(onFulfill?: (value: VideoImportModel) => U | PromiseLike<U>, onReject?: (error: any) => U |...'.
      Type 'Bluebird<any>' is not assignable to type 'Bluebird<any>'. Two different types with this name exist, but they are unrelated.

219   return sequelizeTypescript.transaction(async t => {
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
220     const sequelizeOptions = { transaction: t }
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
243     return videoImport
    ~~~~~~~~~~~~~~~~~~~~~~
244   })
    ~~~~


server/initializers/migrations/0005-email-pod.ts:18:3 - error TS2719: Type 'Bluebird<void>' is not assignable to type 'Bluebird<void>'. Two different types with this name exist, but they are unrelated.
  Types of property 'then' are incompatible.
[...]

PR #1191, opened 2018-10-03
added installer command to package.json scripts; yarn run installer

FlatEarthTruther:
#1171

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "generate-api-doc": "scripty",
     "parse-log": "node ./dist/scripts/parse-log.js",
     "prune-storage": "node ./dist/scripts/prune-storage.js",
+    "installer": "rm -rf ./node_modules && rm -rf client/node_modules && yarn install --pure-lockfile",
     "postinstall": "cd client && yarn install --pure-lockfile",
     "tsc": "tsc",
     "spectacle-docs": "node_modules/spectacle-docs/bin/spectacle.js",


Chocobozzz:
I would prefer to keep the `yarn install...`, it's better for people to know what they're doing (in particular, using the `yarn` tool). The `rm` part has to be done on broken installations (because someone made the wrong command etc). Moreover, it does not cover the production case.

FlatEarthTruther:
It's just nicer to type less than to have to type more always, maybe we can have a dev-installer, prod-installer script instead if it doesn't cover the production use case, because for me even though I know what it does I don't always want to type it all the time and after going away for a month or two from the project I like to pop into the package json to see all the administration/development commands and go from there, so mostly I add all commands to the scripts for reference and to make things easier to remember after a while. I may go on vacation and come back to having to look through many docs to just find this command, whereas I could just find it in the package.json file instead. It anything we can multiple one for prod and one for dev, still it uses the install command originally but this just adds those hard to remember params in a single go.

FlatEarthTruther:
It's not like it's overwriting the install command, it just extends it. people can still use the yarn install all day long if they want, and part of the automation process who cares if it removes the node_modules and reinstalls them if broken install or not.

FlatEarthTruther:
This would've been a duplication of `yarn run danger:clean:modules` && `yarn install --pure-lockfile` so no worries, I guess after I spend more time on this project I'll get to know it more.

PR #1194, opened 2018-10-04
Update FAQ.md

Jorropo:


Nutomic:
Vuze also has webtorrent support: https://wiki.vuze.com/w/WebTorrent

rigelk:
Linking to https://github.com/webtorrent/webtorrent/issues/369 instead of quoting each and every client is probably better. But even then you are missing `webtorrent-hybrid`, which is useful to stream videos in the cli for instance.

Jorropo:
@rigelk I already thinked about telling of webtorrent hybrid but that seems too complicated for an FAQ.
And when you tell :
> Linking to webtorrent/webtorrent#369 instead of quoting each and every client is probably better.

You think that a sentence like that is better ?
You can find here a client list supporting webtorrent https://github.com/webtorrent/webtorrent/issues/369

Jorropo:
@zig-03 correct me if I'm wrong (cause I'm french) another limit to one and other say multiple. Cause multiple servers can do redundancy. If yes I think other is better.

Zig-03:
Hi @Jorropo 
if multiple servers can do the redundancy feature then the text would look something like this:
"It can also be helped by other servers using"

Jorropo:
@Zig-03 effectively I forgoted it.

Zig-03:
@Jorropo you missed one word "be"

> "It can also **be** helped by other servers using"

:)

Jorropo:
@Chocobozzz @Zig-03 made !
:smile:

Chocobozzz:
Thanks @Jorropo & @Zig-03

PR #1196, opened 2018-10-05
support/development/database.md

FlatEarthTruther:
Create a Development database document to help newcomers get setup in development, reasons for this file is because the setup differs a little bit from the production configuration.


Chocobozzz:
As @Jorropo said :)

PR #1198, opened 2018-10-05
WIP: Support more video type

### LINKED ISSUE #326: Allow all ffmpeg video format if transcoding is enabled
As PeerTube use ffmpeg to transcode videos, it should be possible to upload a video with any format supported by ffmpeg.

So, if transcoding is enable on the server, the file selector should not exclude file extensions like `.avi`.

Jorropo:
It closes #326.
If you think a format is important and I missed it, tell me here, I will add it.

But I have a questions, should I use `ffmpeg -formats | grep "^ D"` to get all supported and add it dynamically at the start of peertube ? (For me that is use less).
But I don't think that good cause user maybe will upload format that will require too many ram (like raw image flow) so I think using a small amount of format is better.

PS: currently peertube is far to supporting all popular format (I know that isn't the goal but we are for to far away of that).

TODO :
 - [X] Support more format if transcoding (server)
 - [ ] Force option : "wait transcoding" if the format isn't player compatible
 - [ ] Write test for server


rigelk:
There is a reason why we aren't accepting more formats :) It's basically because videos can be published before they are fully transcoded, and thus users can bump on the original video for a little while. If the video format is not supported by the browser, the user will understandably consider it a bug.

Hopefully now that we have an option to wait until transcoding, we can think of forcing non-streamable formats to wait until transcoding.

Jorropo:
@rigelk ok, I will do that.

Strit:
> Hopefully now that we have an option to wait until transcoding, we can think of forcing non-streamable formats to wait until transcoding.

I would love to see this, combined with the support of more formats.
Make it possible for me to upload more formats, but just don't publish them until they have been transcoded to web friendly formats.

McFlat:
@rigelk 

> There is a reason why we aren't accepting more formats :) It's basically because videos can be published before they are fully transcoded, and thus users can bump on the original video for a little while. If the video format is not supported by the browser, the user will understandably consider it a bug.

So if the video works in the browser, without even transcoding it, eg. `can bump on the original video`, then what's the point in even transcoding it? To get it in more resolutions? Does it make the quality any better or is the main reason just to get more resolutions of that video?

rigelk:
@McFlat because a video has the right format and right codec doesn't mean it has the right encoding options, the right bitrate, etc. Most h264 videos are not encoded using faststart for instance.

McFlat:
Yeah I can see how `faststart` is important, thanks for making it clearer for me.

rigelk:
@Jorropo do you plan to continue the PR? If yes, then I have a function to test if a file is streamable at https://framagit.org/rigelk/kitchensink/blob/master/src/ffmpeg-isStreamable/index.ts

Jorropo:
@rigelk yes, sorry, I have do a pause cause I havn't time but now I can work on, sorry for delay.

Yetangitu:
I'm using a somewhat modified version of this PR to import a host of video files produced by older digital compact camera's. In the process of getting the video's to play on mobile platforms I encountered a bunch of files which were encoded in pixel formats (e.g. yuvj422p) which are not supported on mobile (Android, probably not supported on iOS either). Getting these video's to play on mobile platforms requires a pixel format conversion along these lines:

https://github.com/Yetangitu/PeerTube/commit/a2a5711f4e0a136eb3b35dc2351055fe78d43f46

This forces transcoded material to use yuv420p which is supported on all platforms. Most modern input material already comes in this format and as such will not need conversion.

rigelk:
Closing due to inactivity, feel free to request a reopening.

Chocobozzz:
I'll implement this (in a few days/weeks). Assigned to myself https://github.com/Chocobozzz/PeerTube/issues/326

Issue #685, opened 2018-06-20
Give the possibility to opt-out the webtorrent


Insoleet:
Because of privacy reasons, some users could wan't not to share their public IP with the web torrent tracker. 

I think we should give the option for the user to opt-out of the web-torrent video sharing. If the server is overloaded, the user could watch the video with less quality or fluidity than web-torrent users, but at least his IP would remain private.

therahedwig:
I agree with this.

I think that informing people that their ip address is visible is waking them up to the fact bittorrent does this. This might be annoying right now, but it cannot be helped, and in general giving people more information on the topic is generally a good thing. However, to make a non-commercial youtube alternative, peer-to-peer is probably the only viable way. Therefore perhaps embrace that people the choice to do good by peering.

I think if people can choose to peer, a lot of them will feel empowered by the ability to do so, and I think that if channels can guide users through the weird YouTube subscription system, I don't think it will be hard for them to guide people to click the 'seed this video' button.

Similarly, even if bittorrent didn't have this showing-ip-problem, peertube channels would grow a need to encourage healthy seeding behaviour anyway, so I do not think that allowing opt-in/opt-out will have a huge negative effect in the long run.

RUSshy:
You own the platform, why should i need to protect my identity ?

If you can't do it while you own the platform, then your design sucks

**Don't share my IP, enf of the talk**

Imagine if everyone starts to use PeerTube, then everyone will be able to track me and my habits, and that just because your platform decide to share my identity with everyone else

On top of that i'm pretty sure this is not compatible with GPRD/RGPD

This should be opt-in with user agreement

thibaultamartin:
> If you can't do it while you own the platform, then your design sucks
> Don't share my IP, enf of the talk

@RUSshy, let's not forget PeerTube is developped by a single person hired by a non-profit. This is not even the v1, still a beta. Constructive criticism would help a lot :)

> On top of that i'm pretty sure this is not compatible with GPRD/RGPD

I do believe it is GDPR compliant, as these data are needed by the service to run. I need to dig a bit before giving a strong position though. PeerTube disclosing explicitly that your IP is shared publicly goes in the right direction: not hiding "flaws" from the user.

A more privacy-enabled mode could be thought of with ideas such as:
[...]

Issue #95, opened 2017-10-11
Should mobile user also seed ?


flyingrub:
Mobile users often don't have a lot of data plan so this could be a problem.

Some ideas : 
* Don't seed by default on 3g/4g
* Only seed on wifi ?
* Allow registered users to enable the seed on mobile network ?

Chocobozzz:
It's a very good idea. But I don't think we can detect if the user is on 3g/4g or on wifi, so just don't seed on mobile device.

NeodarZ:
Or just have an option like checkbox for seed ?

tcitworld:
> But I don't think we can detect if the user is on 3g/4g or on wifi

https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API

flyingrub:
After digging a little : https://github.com/webtorrent/webtorrent/issues/480 ...
I opened another issue : https://github.com/webtorrent/webtorrent/issues/1212

rigelk:
Since detecting if the user is on a mobile network has scarce support, we could make the assumption phones are using a mobile network and just detect is the client is on a phone (detecting that is way easier).

Of course that is not perfect, but it's patching the most critical use case, where phone contracts often have limited data volumes.

Speaking of a mobile detection library: https://github.com/KoderLabs/ngx-device-detector - but maybe there is something even more lightweight?

Chocobozzz:
Implemented in our new player in https://github.com/Chocobozzz/PeerTube/commit/39aad8cc856a8173e7a77c198b3968952d39f4b7

PR #1201, opened 2018-10-05
add webtorrent opt-out settings

rigelk:
- add a key in localstorage to remember the opt-out
- add a user setting

Visitors don't have an interface to modify the setting (yet) but that requires more work (modifying buttons in videojs is no small feat) and will be the subject of another PR. Better do things incrementally.

related to #685

### DIFF
--- a/client/src/app/+my-account/my-account-settings/my-account-video-settings/my-account-video-settings.component.html
+++ b/client/src/app/+my-account/my-account-settings/my-account-video-settings/my-account-video-settings.component.html
@@ -15,6 +15,18 @@
     </div>
   </div>
 
+  <div class="form-group">
+    <label i18n for="webTorrentPolicy">Policy regarding P2P technologies</label>
+
+    <div class="peertube-select-container">
+      <select id="webTorrentPolicy" formControlName="webTorrentPolicy">
+        <option i18n value="enable">Enable WebTorrent</option>
+        <option i18n value="disable">Disable WebTorrent globally</option>
+        <option i18n value="disable_on_mobile" disabled>Disable WebTorrent on mobile devices (not yet available)</option>
+      </select>
+    </div>
+  </div>
+
   <my-peertube-checkbox
     inputName="autoPlayVideo" formControlName="autoPlayVideo"
     i18n-labelText labelText="Automatically plays video"
--- a/client/src/app/+my-account/my-account-settings/my-account-video-settings/my-account-video-settings.component.html
+++ b/client/src/app/+my-account/my-account-settings/my-account-video-settings/my-account-video-settings.component.html
@@ -16,16 +16,20 @@
   </div>
 
   <div class="form-group">
-    <label i18n for="webTorrentPolicy">Policy regarding P2P technologies</label>
+    <label i18n for="webTorrentEnabled">Policy regarding P2P technologies</label>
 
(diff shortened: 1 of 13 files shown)


client/src/app/+my-account/my-account-settings/my-account-video-settings/my-account-video-settings.component.html
```diff
@@ -15,6 +15,18 @@
...
   </div>
 
+  <div class="form-group">
+    <label i18n for="webTorrentPolicy">Policy regarding P2P technologies</label>
+
+    <div class="peertube-select-container">
+      <select id="webTorrentPolicy" formControlName="webTorrentPolicy">
+        <option i18n value="enable">Enable WebTorrent</option>
+        <option i18n value="disable">Disable WebTorrent globally</option>
+        <option i18n value="disable_on_mobile" disabled>Disable WebTorrent on mobile devices (not yet available)</option>
```
Chocobozzz: I think you can remove this, since we will try to do it automatically (see https://github.com/Chocobozzz/PeerTube/issues/95#issuecomment-338431521)

So you would just have a boolean.

client/src/assets/player/peertube-player-local-storage.ts
```diff
@@ -10,6 +10,15 @@ function getStoredVolume () {
   return undefined
 }
 
+function getStoredWebTorrentPolicy () {
```
Chocobozzz: This function returns `undefined` or `true`?

client/src/assets/player/peertube-videojs-plugin.ts
```diff
@@ -288,7 +291,8 @@ class PeerTubePlugin extends Plugin {
         renderVideo(torrent.files[ 0 ], this.playerElement, renderVideoOptions, (err, renderer) => {
           this.renderer = renderer
 
-          if (err) return this.fallbackToHttp(done)
+          console.log('value this.playerRefusedP2P', this.playerRefusedP2P)
```
Chocobozzz: :p

client/src/assets/player/peertube-videojs-plugin.ts
```diff
@@ -288,7 +291,8 @@ class PeerTubePlugin extends Plugin {
...
+          if (err || this.playerRefusedP2P) return this.fallbackToHttp(done)
```
Chocobozzz: It's not here where you should fallback to http. Use it [here](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/assets/player/peertube-videojs-plugin.ts#L436)

client/src/assets/player/peertube-videojs-plugin.ts
rigelk: :sweat_smile:

client/src/assets/player/peertube-videojs-plugin.ts
rigelk: @Chocobozzz placing fallback anywhere in `initializePlayer()` doesn't work.

McFlat:
This is great. We can also mention that disabling webtorrent doesn't leak a users IP address and is an added security bonus feature, at the cost of bandwidth of course, still if only a single user is watching the video and not a group of users all at the same time there is not bandwidth saving, just a security benefit.

Issue #1147, opened 2018-09-29
Set Keyframe interval for FFmpeg


Nutomic:
The default for mp4 is a [keyframe every 250 frames](https://sites.google.com/site/linuxencoding/x264-ffmpeg-mapping) (so about every 10 seconds). This is bad for streaming, because it's harder to change resolutions. Youtube uses a keyframe interval of [2 seconds](https://streaminglearningcenter.com/blogs/whats-the-right-keyframe-interval.html).

According to [Stack Overflow](https://superuser.com/a/1098329), this should be the correct way to set a different keyframe interval. It works fine on the command line, but fails inside Peertube.
```
command.outputOptions('-force_key_frames "expr:gte(t,n_forced*2)"')
```

We could use the `keyint` parameter instead (described in the same link), but I don't understand how it works right now.

Edit: related to #1135.

Nutomic:
This is the error that FFmpeg returns with the option above:
```
Invalid duration specification for force_key_frames: "expr:gte(t
```
So it's only seeing the parameter up to the comma. To me that looks suspiciously like fluent-ffmpeg (or something else) is splitting the string on commas. I tried to add each part seperately with `command.addOutputOption()`, and to escape the comma with `\`, but that didn't change anything.

rigelk:
This does look like a `fluent-ffmpeg` problem. Maybe https://github.com/fluent-ffmpeg/node-fluent-ffmpeg/issues/682

Nutomic:
That sounds very similar, although it is about the outputs of fluent-ffmpeg, not the inputs. Also, it has been reported 1.5 years ago and nothing happened :/

rigelk:
Yes, as you have noticed, the [library is unmaintained (or low maintenance mode, rather)](https://github.com/fluent-ffmpeg/node-fluent-ffmpeg/wiki/Looking-for-a-new-maintainer)… :/

rigelk:
Looking at the keyframe interval itself, I doubt it will be of any use to us. Maybe I'm misunderstanding its use, but it's used by adaptative streaming server implementations. Our implementation is done client side, and doesn't rely on that much information.

rigelk:
I doubt seeking is useful every two frame, though. Usually a seek step of one second is used.

Nutomic:
I guess you misunderstood me? I proposed a keyframe interval of 2 seconds (same as youtube).

rigelk:
:open_mouth: I must be tired, sorry

PR #1205, opened 2018-10-05
Set keyframe interval

Nutomic:
Turns out that I was wrong in #1147, the `-g` paramater for keyframes is only deprecated for H265, but works fine for H264. 

Unfortunately fluent-ffmpeg doesn't return the keyframe interval, and doesn't have an option to set parameters. So I think the only way to get the keyframe interval of an existing video (for tests and `optimize-old-videos.ts`) is by running the command on a shell. But I'm not sure how that works, or if it's a good idea. This is the command to list all keyframe times ([source](https://stackoverflow.com/a/18088156)):
```
ffprobe -loglevel error -skip_frame nokey -select_streams v:0 -show_entries frame=pkt_pts_time -of csv=print_section=0
```
This PR depends on #1135.


Nutomic:
Here's a diff for easier reading:
```
diff --git a/server/helpers/ffmpeg-utils.ts b/server/helpers/ffmpeg-utils.ts
index 8e447117..26f99bc8 100644
--- a/server/helpers/ffmpeg-utils.ts
+++ b/server/helpers/ffmpeg-utils.ts
@@ -114,6 +114,10 @@ type TranscodeOptions = {
   isPortraitMode?: boolean
 }

+/**
+ * Transcodes a video to H264. Don't use it for other output formats, as
+ * parameters might be different.
+ */
     let command = ffmpeg(options.inputPath, { niceness: FFMPEG_NICE.TRANSCODING })
@@ -154,6 +158,11 @@ function transcode (options: TranscodeOptions) {
     const targetBitrate = getTargetBitrate(options.resolution, fps, VIDEO_TRANSCODING_FPS)
     command.outputOptions([`-b:v ${ targetBitrate }`, `-maxrate ${ targetBitrate }`, `-bufsize ${ targetBitrate * 2 }`])

+    // Keyframe interval of 2 seconds for faster seeking and resolution switching.
+    // https://streaminglearningcenter.com/blogs/whats-the-right-keyframe-interval.html
+    // https://superuser.com/a/908325
+    command.outputOption(`-g ${ fps * 2 }`)
+
     command
       .on('error', (err, stdout, stderr) => {
         logger.error('Error in transcoding job.', { stdout, stderr })
```

rigelk:
@Nutomic last time I refactored `ffmpeg-utils.ts`, I added presets which can be seen at the bottom of the file. It is meant to decouple the encoding profile and its options (which as you point out can be very specific) from the transcoding function.

So if you have a h264-specific encoding option, why not put it in `standard`, line 276? It might not be easy to pass the `fps`, though :thinking:

Nutomic:
Ah I wasn't sure what the preset was about exactly. How about renaming it to h264Profile or something like that? It would also make sense to move the code for video bitrate there.And then we could just pass fps and resolution as parameters?

rigelk:
@Nutomic sure! Right now only the standard preset is used, but the other two presets I added along were meant as the base for a later choice of a lighter transcoding profile by the admin (using veryfast for instance).

But it's just an idea of a later refactoring. Your PR is fine as is.

Nutomic:
I can certainly take care of that to save you some work. I'm just not sure what's the best way to solve it, because the presets function can't take any extra parameters. So either:
- change the function to `standard(ffmpeg, fps, resolution)`, and call it as a normal function from `transcode()`
- leave `standard()` as a preset function, and calculate fps/resolution again

rigelk:
@Nutomic could you rebase your PR now that the `ffmpeg-utils` have changed in develop?

PR #1206, opened 2018-10-06
WIP: config option transcoding concurrency

McFlat:


client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.html
```diff
@@ -221,6 +221,42 @@
             </div>
           </div>
 
+          <div class="form-group">
```
Jorropo: The transcoding concurrency included in the client is a bad idea cause it require a restart of the server to work :
https://github.com/Chocobozzz/PeerTube/blob/35d50b7dd26b3cf646b8845784927bb1ef18dfb3/server/lib/job-queue/job-queue.ts#L78 (as you can see transcoding concurrency is set at the init of the server).
Maybe we should keep it in the server but not in client :wink:

Jorropo:
You also must fix the test suite :)
https://travis-ci.org/Chocobozzz/PeerTube/jobs/437880126
https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/server/config.ts

McFlat:
to fix the test suite there needs to be a new utility function added by name of `toDuration`, there is currently a `parseDuration` to convert duration to integer number but not way right now to convert an integer number to a duration. I spoke about this before my github account got flagged. I was going to try to see if `luxon` would work out for this, instead of writing the function to just plug luxon in and possibly replace the current implementation of `parseDuration`

McFlat:
I wasn't aware that a server restart would be required to update the concurrency value, maybe you're right it's best to not expose this in the admin configurations and just keep it exposed in the config file instead. I did it that way before but then was told to make it available on the admin config interface too, but no worries it's not a waste at all, all this helped me to learn how the system is coded.

client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.html
Chocobozzz: Oh right nice catch @Jorropo (sorry @McFlat I forgot this)

config/default.yaml
```diff
@@ -120,6 +120,9 @@ user:
 transcoding:
   enabled: true
   threads: 1
+  concurrency: 1
+  niceness: 15 # from 1 to 18
+  ttl: '2 days' # 2 days, transcoding could be long. eg. '2 days' or '12 hours'
```
Chocobozzz: Maybe add a comment to explain what this `ttl` is (so admins understand this parameter is used to kill too long transcoding jobs).

config/default.yaml
```diff
@@ -120,6 +120,9 @@ user:
 transcoding:
   enabled: true
   threads: 1
+  concurrency: 1
```
Chocobozzz: Maybe add a comment explaining the difference between the `threads` configuration

server/initializers/constants.ts
```diff
@@ -118,7 +118,7 @@ const JOB_TTL: { [ id in JobType ]: number } = {
   'activitypub-http-fetcher': 60000 * 10, // 10 minutes
   'activitypub-follow': 60000 * 10, // 10 minutes
   'video-file-import': 1000 * 3600, // 1 hour
-  'video-file': 1000 * 3600 * 48, // 2 days, transcoding could be long
+  'video-file': parseDuration(config.get<string>('transcoding.ttl')), // 2 days, transcoding could be long
```
Chocobozzz: `CONFIG.TRANSCODING.TTL`

server/initializers/constants.ts
```diff
@@ -107,7 +107,7 @@ const JOB_CONCURRENCY: { [ id in JobType ]: number } = {
   'activitypub-http-fetcher': 1,
   'activitypub-follow': 3,
   'video-file-import': 1,
-  'video-file': 1,
+  'video-file': config.get<number>('transcoding.concurrency'),
```
Chocobozzz: `CONFIG.TRANSCODING.CONCURRENCY`

server/initializers/constants.ts
```diff
@@ -407,7 +410,7 @@ const VIDEO_RATE_TYPES: { [ id: string ]: VideoRateType } = {
 
 const FFMPEG_NICE: { [ id: string ]: number } = {
   THUMBNAIL: 2, // 2 just for don't blocking servers
-  TRANSCODING: 15
+  TRANSCODING: config.get<number>('transcoding.niceness')  // 15
```
Chocobozzz: `CONFIG.TRANSCODING.NICENESS` and remove `// 15``

client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.html
McFlat: No worries, I'll try to get all these things done today.

server/initializers/constants.ts
McFlat: I'll have to copy the whole CONFIG block up higher, because my editor is saying that when I do this(`CONFIG.TRANSCODING.CONCURRENCY`), it's being used before declaration.

McFlat: I'lll move the `CONFIG` block to be right after the `LAST_MIGRATION_VERSION`, seeing as this makes more sense for if we change or introduce more variables later into the `CONFIG` and use the definitions in the constants

McFlat:
@Chocobozzz I've made that changes you mentioned

McFlat:
Most of the tests pass, but there are two different builds giving some issues. I've tried to fix it but something to do with the way the transcoding object is shared between the client and server is not allowing me to fix the tests. Most builds pass, but two fail.

McFlat:
If we do decide to add the transcoding concurrency, niceness and ttl to the client later we can just let the admin know that, `If these options change a peertube restart is required.`

Chocobozzz:
> If we do decide to add the transcoding concurrency, niceness and ttl to the client later we can just let the admin know that, If these options change a peertube restart is required.

Good idea :+1: Please do it :)

> Looks like we need a parseBytes utility function. so instead of using integers in the config we can say '5GB' or whatever

I think it would be better if the server returns/get to/from the client the number of bytes (and only that).

McFlat:
For now I added a comment next to the options in the config that a peertube server restart is required if they change, and they shouldn't have to change all that often, so it's probably best to just keep then in the config file, and not expose them to the frontend/client, at least not yet. I've already tried to expose them to the client, but the tests fail because we are missing a utility function to parse integer into a duration, so until we add that utility function which takes a number and turns it into a duration then lets keep it only in the backend/server config and not expose it to the frontend/client.

Nutomic:
I don't see the point of that. You can already restart Peertube from the command line with systemd (or whatever you are using).

McFlat:
yeah, good point, it would be like adding a security risk if we added that capability

McFlat:
The only test that fails is the `misc` test. I'll try to fix that right now. All this shared-ness is kinda funky. Like those config options are only on the backend/server but they are seeping thru to the frontend client because of trying to keep the compiler happy.

```
ERROR in src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts(126,11): error TS2322: Type '{ instance: { name: any; shortDescription: any; description: any; terms: any; defaultClientRoute:...' is not assignable to type 'CustomConfig'.
  Types of property 'transcoding' are incompatible.
    Type '{ enabled: any; threads: any; resolutions: { '240p': any; '360p': any; '480p': any; '720p': any; ...' is not assignable to type '{ enabled: boolean; threads: number; concurrency: number; niceness: number; ttl: any; resolutions...'.
      Property 'concurrency' is missing in type '{ enabled: any; threads: any; resolutions: { '240p': any; '360p': any; '480p': any; '720p': any; ...'.
```

Chocobozzz:
Closing due to inactivity, feel free to request the reopening

PR #1213, opened 2018-10-06
WIP: change/ask for reset of actor's password as admin

### LINKED ISSUE #1052: Change/reset actor's password as admin
- Possibility for admin to change once's password and automaticaly send the new one to the user
    - Set it as random-no-see so that the admin don't see it.

rigelk:
resolves #1052 

update a user - peertube

TODO:

- [ ] fix direct password change

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -158,9 +158,7 @@
     "webpack-cli": "^3.0.8",
     "webtorrent": "https://github.com/webtorrent/webtorrent#e9b209c7970816fc29e0cc871157a4918d66001d",
     "whatwg-fetch": "^3.0.0",
-    "zone.js": "~0.8.5"
-  },
-  "dependencies": {
+    "zone.js": "~0.8.5",
     "generate-password-browser": "^1.0.2"
   }
 }
--- a/client/package.json
+++ b/client/package.json
@@ -159,5 +159,8 @@
     "webtorrent": "https://github.com/webtorrent/webtorrent#e9b209c7970816fc29e0cc871157a4918d66001d",
     "whatwg-fetch": "^3.0.0",
     "zone.js": "~0.8.5"
+  },
+  "dependencies": {
+    "generate-password-browser": "^1.0.2"
   }
 }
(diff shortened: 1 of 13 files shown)


client/package.json
```diff
@@ -160,5 +160,8 @@
     "webtorrent": "https://github.com/webtorrent/webtorrent#e9b209c7970816fc29e0cc871157a4918d66001d",
     "whatwg-fetch": "^3.0.0",
     "zone.js": "~0.8.5"
+  },
+  "dependencies": {
+    "generate-password-browser": "^1.0.2"
```
Chocobozzz: dev dep

Chocobozzz:
Can we have a screen from the danger section? :)

client/src/app/+admin/users/user-edit/user-password.component.html
```diff
@@ -0,0 +1,25 @@
...
+  <div class="form-group">
+
+    <div class="input-group mb-3">
+      <div class="input-group-prepend">
+        <div class="input-group-text">
+          <input type="checkbox" aria-label="Show password" (change)="togglePasswordVisibility()">
+        </div>
+      </div>
+      <input id="passwordField" #passwordField
+      [attr.type]="showPassword ? 'text' : 'password'" id="password"
```
Chocobozzz: indentation

client/src/app/+admin/users/user-edit/user-password.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  ) {
+    super()
+  }
+
+  ngOnInit () {
+    this.buildForm({
+      password: this.userValidatorsService.USER_PASSWORD
+    })
+
+    this.paramsSub = this.route.params.subscribe(routeParams => {
```
Chocobozzz: Why don't you use `Input()` from the parent component instead?

client/src/app/+admin/users/user-edit/user-update.component.ts
```diff
@@ -19,9 +19,12 @@ import { UserService } from '@app/shared'
 export class UserUpdateComponent extends UserEdit implements OnInit, OnDestroy {
   error: string
   userId: number
+  userEmail: string
   username: string
+  isAdministration = false
```
Chocobozzz: I don't understand why we need this, because we are in the `+admin/` directory so `isAdministration` should always be `true`

client/package.json
rigelk: I promise I did a `--save-dev` :p

Chocobozzz: it's `yarn install --dev ...`

Chocobozzz:
Merged manually

Issue #1186, opened 2018-10-03
Indicate that "Change video ownership" only works with local accounts


Nutomic:
* **PeerTube version or URL:** v1.0.0-rc.1

* **What do you see instead?**

After entering a username like @test@peertube.fr, and clicking submit, Peertube just shows a an error in the bottom right corner with "unknown error". 

* **What is the expected behaviour?**
The "change ownership" dialog should indicate that only local accounts are supported, for example with a text in the dialog. Entering a remote username should show a message that remote accounts are not supported yet.

* **Browser console log if useful:**
```
Backend returned code 400, errorMessage is: Unknown error. main.1053f9ccca20182cb383.js:1:421526
Huiz/o</t.prototype.handleError
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:421526
HvDw/s</t.prototype.changeOwnership/<
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:422369
9Z1F/l</e.prototype.error
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:157830
FFOo/c</e.prototype._error
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:407874
FFOo/c</e.prototype.error
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:407504
FFOo/c</e.prototype._error
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:407874
FFOo/c</e.prototype.error
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:407504
FFOo/c</e.prototype._error
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:407874
FFOo/c</e.prototype.error
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:407504
MGBS/i</e.prototype.notifyError
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:603462
rPjj/i</e.prototype._error
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:945311
FFOo/c</e.prototype.error
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:407504
MGBS/i</e.prototype.notifyError
https://peertube.social/client/en_US/main.1053f9ccca20182cb383.js:1:603462
rPjj/i</e.prototype._error
[...]

PR #1214, opened 2018-10-06
 Add explicit error message that changing video ownership only works with local accounts

lucas-dclrcq:
partialfixes #1186

### DIFF
--- a/server/middlewares/validators/videos/videos.ts
+++ b/server/middlewares/validators/videos/videos.ts
@@ -69,7 +69,6 @@ const videosAddValidator = getCommonVideoAttributes().concat([
     if (isAble === false) {
       res.status(403)
          .json({ error: 'The user video quota is exceeded with this video.' })
-         .end()
 
       return cleanUpReqFiles(req)
     }
@@ -82,7 +81,6 @@ const videosAddValidator = getCommonVideoAttributes().concat([
       logger.error('Invalid input file in videosAddValidator.', { err })
       res.status(400)
          .json({ error: 'Invalid input file.' })
-         .end()
 
       return cleanUpReqFiles(req)
     }
@@ -120,7 +118,6 @@ const videosUpdateValidator = getCommonVideoAttributes().concat([
       cleanUpReqFiles(req)
       return res.status(409)
         .json({ error: 'Cannot set "private" a video that was not private.' })
-        .end()
     }
 
     if (req.body.channelId && !await isVideoChannelOfAccountExist(req.body.channelId, user, res)) return cleanUpReqFiles(req)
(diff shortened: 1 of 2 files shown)


server/middlewares/validators/videos/videos.ts
```diff
@@ -238,8 +238,9 @@ const videosChangeOwnershipValidator = [
 
     const nextOwner = await AccountModel.loadLocalByName(req.body.username)
     if (!nextOwner) {
+      logger.debug('N')
```
rigelk: Please remove this line

lucas-dclrcq: All done. Sorry for that.

server/middlewares/validators/videos/videos.ts
```diff
@@ -239,7 +239,7 @@ const videosChangeOwnershipValidator = [
     const nextOwner = await AccountModel.loadLocalByName(req.body.username)
     if (!nextOwner) {
       res.status(400)
-        .type('json')
+        .json({ error: 'Changing video ownership to a remote account is not supported yet' })
         .end()
```
rigelk: Since you use `res.json()`, there is no need anymore for the `res.end()`, which is typically only used if we want to end the response without providing any data.

Chocobozzz:
Thanks for the PR @lucas-dclrcq :+1: 

But I think to resolve completely https://github.com/Chocobozzz/PeerTube/issues/1186, we should add a message on the client side too :)

lucas-dclrcq:
@Chocobozzz On the client side the error popup shows the error sent by the server. Is it not sufficient ?

server/middlewares/validators/videos/videos.ts
lucas-dclrcq: Hi, thanks for the info, I will change that. But in the video.ts file all the validators build their error response in that way. Should I refactor theses also ?

Chocobozzz:
@lucas-dclrcq it's better not to give people false hopes :) And server error messages are not translated on client side (yet)

server/middlewares/validators/videos/videos.ts
Chocobozzz: If you have time sure! :+1:

rigelk: If you don't mind that would be great :)

lucas-dclrcq:
So I'm thinking I can add a validator on the change ownership form that checks if the username contains the character '@' (since the username must contains only alphanumeric characters I think it can work)
And then show a localized form error saying that only local accounts are supported.

Would this be ok ?

Chocobozzz:
> So I'm thinking I can add a validator on the change ownership form that checks if the username field contains the character '@' (since the username contains only alphanumeric characters I think it can work)

:+1:

Chocobozzz:
Thanks @lucas-dclrcq!

PR #1231, opened 2018-10-09
Set keyframe interval for transcoding (fixes #1147)

### LINKED ISSUE #1147: Set Keyframe interval for FFmpeg

Nutomic:
The rebase for #1205 was quite messy, so I just applied the patch to develop instead. That broke the PR somehow, so here's the same changes again.

### DIFF
--- a/server/helpers/ffmpeg-utils.ts
+++ b/server/helpers/ffmpeg-utils.ts
@@ -147,12 +147,6 @@ function transcode (options: TranscodeOptions) {
       command = command.withFPS(fps)
     }
 
-    // Constrained Encoding (VBV)
-    // https://slhck.info/video/2017/03/01/rate-control.html
-    // https://trac.ffmpeg.org/wiki/Limiting%20the%20output%20bitrate
-    const targetBitrate = getTargetBitrate(options.resolution, fps, VIDEO_TRANSCODING_FPS)
-    command.outputOptions([`-maxrate ${ targetBitrate }`, `-bufsize ${ targetBitrate * 2 }`])
-
     command
       .on('error', (err, stdout, stderr) => {
         logger.error('Error in transcoding job.', { stdout, stderr })
@@ -327,7 +321,7 @@ async function presetH264 (ffmpeg: ffmpeg, resolution: VideoResolution, fps: num
   // https://slhck.info/video/2017/03/01/rate-control.html
   // https://trac.ffmpeg.org/wiki/Limiting%20the%20output%20bitrate
   const targetBitrate = getTargetBitrate(resolution, fps, VIDEO_TRANSCODING_FPS)
-  localFfmpeg.outputOptions([`-b:v ${ targetBitrate }`, `-maxrate ${ targetBitrate }`, `-bufsize ${ targetBitrate * 2 }`])
+  localFfmpeg.outputOptions([`-maxrate ${ targetBitrate }`, `-bufsize ${ targetBitrate * 2 }`])
 
   // Keyframe interval of 2 seconds for faster seeking and resolution switching.
   // https://streaminglearningcenter.com/blogs/whats-the-right-keyframe-interval.html
--- a/server/helpers/ffmpeg-utils.ts
+++ b/server/helpers/ffmpeg-utils.ts
@@ -126,7 +126,7 @@ function transcode (options: TranscodeOptions) {
 
     let command = ffmpeg(options.inputPath, { niceness: FFMPEG_NICE.TRANSCODING })
                     .output(options.outputPath)
-    presetH264(command, options.resolution, 0)
+    presetH264(command, options.resolution, fps)
 
     if (CONFIG.TRANSCODING.THREADS > 0) {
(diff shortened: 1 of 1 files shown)


server/helpers/ffmpeg-utils.ts
```diff
@@ -324,5 +323,16 @@ async function standard (_ffmpeg) {
 
   if (bitrate !== undefined) return localFfmpeg.audioBitrate(bitrate)
 
+  // Constrained Encoding (VBV)
+  // https://slhck.info/video/2017/03/01/rate-control.html
+  // https://trac.ffmpeg.org/wiki/Limiting%20the%20output%20bitrate
+  const targetBitrate = getTargetBitrate(resolution, fps, VIDEO_TRANSCODING_FPS)
+  localFfmpeg.outputOptions([`-b:v ${ targetBitrate }`, `-maxrate ${ targetBitrate }`, `-bufsize ${ targetBitrate * 2 }`])
```
rigelk: Please remove the fixed bitrate.

Nutomic: Oops, fixed.

server/helpers/ffmpeg-utils.ts
```diff
@@ -116,28 +116,27 @@ type TranscodeOptions = {
 
 function transcode (options: TranscodeOptions) {
   return new Promise<void>(async (res, rej) => {
+    let fps = await getVideoFileFPS(options.inputPath)
```
Chocobozzz: Why did you move this block? :)

server/helpers/ffmpeg-utils.ts
```diff
@@ -199,9 +192,9 @@ function getVideoFileStream (path: string) {
  * and quality. Superfast and ultrafast will give you better
  * performance, but then quality is noticeably worse.
  */
-function veryfast (_ffmpeg) {
-  _ffmpeg
-    .preset(standard)
+async function presetH264VeryFast (ffmpeg, resolution: VideoResolution, fps: number): ffmpeg {
```
Chocobozzz: Please type `ffmpeg`

server/helpers/ffmpeg-utils.ts
```diff
@@ -220,9 +213,9 @@ function veryfast (_ffmpeg) {
 /**
  * A preset optimised for a stillimage audio video
  */
-function audio (_ffmpeg) {
-  _ffmpeg
-    .preset(veryfast)
+async function presetStillImageWithAudio (ffmpeg, resolution: VideoResolution, fps: number): ffmpeg {
```
Chocobozzz: Please type `ffmpeg`

server/helpers/ffmpeg-utils.ts
Nutomic: Because fps is needed for the preset (just noticed that I forgot to set the parameter correctly).

server/helpers/ffmpeg-utils.ts
Nutomic: What do you mean? Not sure how this works in Typescript..

Nutomic: Oh just noticed it, I thought you mean something about the return type.

Chocobozzz: :+1:

server/helpers/ffmpeg-utils.ts
```diff
@@ -116,28 +116,27 @@ type TranscodeOptions = {
...
+    if (options.resolution !== undefined &&
+        options.resolution < VIDEO_TRANSCODING_FPS.KEEP_ORIGIN_FPS_RESOLUTION_MIN &&
+        fps > VIDEO_TRANSCODING_FPS.AVERAGE) {
+      fps = VIDEO_TRANSCODING_FPS.AVERAGE
+    }
+
     let command = ffmpeg(options.inputPath, { niceness: FFMPEG_NICE.TRANSCODING })
                     .output(options.outputPath)
-                    .preset(standard)
+    presetH264(command, options.resolution, fps)
```
rigelk: you need to assign the returned 'command' instance. Also `presetH264` is now `async function`, so you need to prefix its call with `await`.

Nutomic: Done. Although the return value isn't really needed afaik, because it's still the same object.

PR #1233, opened 2018-10-09
Scale bitrate linearly with FPS

Nutomic:
The current implementation after #1135 uses the same bitrate for any fps number, except for exactly 60 fps, where it uses a higher bitrate. As an extreme case, a video with 10 fps and a video with 59 fps will have the same bitrate (at the same resolution).

This is a suggestion to scale the bitrate linearly. Example bitrates (also as comment in the source code):
```
1080p10: 2420 kbps, 1080p30: 3300 kbps, 1080p60: 4620 kbps
 720p10: 1283 kbps,  720p30: 1750 kbps,  720p60: 2450 kbps
```
This approach will also work correctly with e.g. 480p60 video (if enabled). It should also work correctly with different `VIDEO_TRANSCODING_FPS.AVERAGE`, `VIDEO_TRANSCODING_FPS.MAX` etc values (untested).

You can visualize the formula here: https://www.wolframalpha.com/input/?i=3300+%2B+(x+-+30)+*+(1320%2F30);+0%3C%3Dx%3C%3D60&assumption=%7B%22C%22,+%220%22%7D+-%3E+%7B%22NumberMath%22%7D

If you agree with the idea, I will add some tests for `getTargetBitrate()`. Also feel free to suggest improvements to the documentation, as this is pretty complicated to explain.

Edit: Of course, the bitrate could also be scaled differently, e.g. quadratic scaling, linear scaling only below `VIDEO_TRANSCODING_FPS.AVERAGE`, etc.

### DIFF
--- a/shared/models/videos/video-resolution.enum.ts
+++ b/shared/models/videos/video-resolution.enum.ts
@@ -43,14 +43,20 @@ export function getBaseBitrate (resolution: VideoResolution) {
 
 /**
  * Calculate the target bitrate based on video resolution and FPS.
+ *
+ * The calculation is based on two values:
+ * Bitrate at VideoTranscodingFPS.AVERAGE is always the same as
+ * getBaseBitrate(). Bitrate at VideoTranscodingFPS.MAX is always
+ * getBaseBitrate() * 1.4. All other values are calculated linearly
+ * between these two points.
  */
 export function getTargetBitrate (resolution: VideoResolution, fps: number,
     fpsTranscodingConstants: VideoTranscodingFPS) {
   const baseBitrate = getBaseBitrate(resolution)
   // The maximum bitrate, used when fps === VideoTranscodingFPS.MAX
   // Based on numbers from Youtube, 60 fps bitrate divided by 30 fps bitrate:
-  //  720p: 2600 / 1750 = 1.48571428571
-  // 1080p: 4400 / 3300 = 1.33333333333
+  //  720p: 2600 / 1750 = 1.49
+  // 1080p: 4400 / 3300 = 1.33
   const maxBitrate = baseBitrate * 1.4
   const maxBitrateDifference = maxBitrate - baseBitrate
   const maxFpsDifference = fpsTranscodingConstants.MAX - fpsTranscodingConstants.AVERAGE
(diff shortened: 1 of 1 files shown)


scanlime:
Thanks, this definitely seems like an improvement!

Would it be easier to reason about the rate in terms of kbits per frame?

Nutomic:
Well I originally took the bitrate values for 30 fps and 60 fps from Youtube (see #1135), the other values are jsut scaled linearly from that. So you might ask Google about the reason. I'd say this is because at 60 fps, you have fewer changes between each frame compared to 30 fps, so you need less data for each frame.

If you know any good source for encoding bitrates, I'd love to see it. When I searched, I didn't find much general information (probably because I was mostly looking for info regarding youtube).

scanlime:
ok, thanks!

I suspect google did some kind of study that takes into account internet connection availability and the video quality, and this is what they came up with.

I’m having trouble finding good references. Here’s an old study that seems to indicate
there’s very little additional bandwidth necessary to maintain the same quality at 60fps vs 30fps, but this was on mostly lower res video and a specific problem domain, https://www.sri.com/sites/default/files/publications/3_07_h264_format_bitrate_quality_tradeoff_study.pdf

Nutomic:
@Chocobozzz @rigelk Any thoughts on this?

rigelk:
This sounds like a good idea, but I'm also having trouble finding references about it. The current PR is good as it is ; maybe we should be making another PR later once we have more references to figure it out.

Nutomic:
@Chocobozzz Maybe I should have mentioned this, but I would have prefered to get this into the next release together with #1135. Now any videos will have a low bitrate for fps between ~35 and 59, and we can't fix this later. And it would also be good to release #1231 together, so we have all the encoding changes together.

Is there anything else I need to do before merging?

Chocobozzz:
Sorry for the late @Nutomic 

> Now any videos will have a low bitrate for fps between ~35 and 59

It's not a big problem: we released an alpha and fps between 35/59 are not very common.

Nutomic:
Cool, thanks! I hope we get alpha 2 soon, so I can start optimizing all the old videos :)

PR #1239, opened 2018-10-11
add parseBytes utility function and tests

McFlat:
This is something we could use to parse values from config. Check the tests to see how it works

### DIFF
--- a/server/tests/helpers/core-utils.ts
+++ b/server/tests/helpers/core-utils.ts
@@ -15,7 +15,7 @@ describe('Parse Bytes', function () {
     expect(parseBytes(1048576)).to.be.eq(1048576)
     expect(parseBytes('1024')).to.be.eq(1024)
     expect(parseBytes('1048576')).to.be.eq(1048576)
-    
+
     // sizes
     expect(parseBytes('1B')).to.be.eq(1024)
     expect(parseBytes('1MB')).to.be.eq(1048576)
@@ -29,7 +29,7 @@ describe('Parse Bytes', function () {
     expect(parseBytes('1024MB')).to.be.eq(1073741824)
     expect(parseBytes('1024GB')).to.be.eq(1099511627776)
     expect(parseBytes('1024TB')).to.be.eq(1125899906842624)
-    
+
     // with whitespace
     expect(parseBytes('1 GB')).to.be.eq(1073741824)
     expect(parseBytes('1\tGB')).to.be.eq(1073741824)
@@ -40,7 +40,6 @@ describe('Parse Bytes', function () {
     expect(parseBytes('4TB 1024GB')).to.be.eq(5497558138880)
     expect(parseBytes('4TB 1024GB 0MB')).to.be.eq(5497558138880)
     expect(parseBytes('1024TB 1024GB 1024MB')).to.be.eq(1127000492212224)
-    
   })
 
   it('Should be invalid when given invalid value', async function () {
--- a/server/tests/helpers/core-utils.ts
+++ b/server/tests/helpers/core-utils.ts
@@ -1,0 +1,49 @@
+/* tslint:disable:no-unused-expression */
+
+import * as chai from 'chai'
+import 'mocha'
+import {
+  parseBytes
+} from '../../helpers/core-utils'
+
+const expect = chai.expect
+
+describe('Parse Bytes', function () {
+  it('Should pass when given valid value', async function () {
+    // just return it
+    expect(parseBytes(1024)).to.be.eq(1024)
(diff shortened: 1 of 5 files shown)


server/tests/helpers/core-utils.ts
```diff
@@ -0,0 +1,47 @@
...
+    // with whitespace
+    expect(parseBytes('1 GB')).to.be.eq(1048576)
+    expect(parseBytes('1\tGB')).to.be.eq(1048576)
+
+    // sum value
+    expect(parseBytes('1TB 1024MB')).to.be.eq(1074790400)
+    expect(parseBytes('4GB 1024MB')).to.be.eq(5242880)
+    expect(parseBytes('4TB 1024GB')).to.be.eq(5368709120)
+    expect(parseBytes('4TB 1024GB 0MB')).to.be.eq(5368709120)
+    expect(parseBytes('1024TB 1024GB 1024MB')).to.be.eq(1100586418176)
```
Nutomic: I never saw this kind of format used anywhere. Is it really necessary to support?

Nutomic: If you remove this functionality, you could use a simple switch for `parseBytes()`, and get rid of the complicated regex stuff.

McFlat: We are using integers all over the place, and in the config some tuff is commented out where  it uses the strings but they aren't parsed. So this functionality is pretty much needed the way things are headed

McFlat:
```
redundancy:
  videos:
    check_interval: '1 hour' # How often you want to check new videos to cache
    strategies:
#      -
#        size: '10GB'
#        # Minimum time the video must remain in the cache. Only accept values > 10 hours (to not overload remote instances)
#        min_lifetime: '48 hours'
#        strategy: 'most-views' # Cache videos that have the most views
#      -
#        size: '10GB'
#        # Minimum time the video must remain in the cache. Only accept values > 10 hours (to not overload remote instances)
#        min_lifetime: '48 hours'
#        strategy: 'trending' # Cache trending videos
#      -
#        size: '10GB'
#        # Minimum time the video must remain in the cache. Only accept values > 10 hours (to not overload remote instances)
#        min_lifetime: '48 hours'
#        strategy: 'recently-added' # Cache recently added videos
#        min_views: 10 # Having at least x views
```

McFlat: also to specify the quotas, instead of using bytes be can just give it a string value

```
user:
  # Default value of maximum video BYTES the user can upload (does not take into account transcoded files).
  # -1 == unlimited
  video_quota: -1
  video_quota_daily: -1
```

McFlat: If you can come up with a better way to do it without regex, go for it

Nutomic: I meant the part with having TB and GB both in the same value. If you leave that out, you can just do something like '10 GB'.split(' ')` and do a switch on the second part.

Chocobozzz:
Thanks for the PR @McFlat, but I don't think we need it. See my comment: https://github.com/Chocobozzz/PeerTube/pull/1206#issuecomment-428863963

The API REST should try to reduce the input data, and it's not really difficult for the client to parse a number of bytes :) (We gave the `bytes` pipe in Angular)

McFlat:
The Angular pipe is only for the frontend/client, but this parseBytes is a backend/sever function. It's not to be used on the frontend. I noticed that pipe on the frontend, but we are missing a function like this on the backend/server.

Chocobozzz:
> The Angular pipe is only for the frontend/client, but this parseBytes is a backend/sever function. It's not to be used on the frontend. I noticed that pipe on the frontend, but we are missing a function like this on the backend/server.

Yep, but if the backend only returns bytes instead of strings, and only get bytes from the client we don't need this any more. It's an api rest, it's not intended to get human readable formats

rigelk:
@Chocobozzz I guess the main use case would be the configuration file where the admin has to enter the values directly

Chocobozzz:
oh :man_facepalming: 

Sorry.

server/tests/helpers/index.ts
```diff
@@ -0,0 +1 @@
+import './core-utils'
```
Chocobozzz: Add this test to `travis.sh` (in `misc` for example)

McFlat: Ok done, let's cross our fingers that it fixed the misc test

McFlat:
I've squashed the commits in this pull request

McFlat:
This is ready to be merged if you guys approve of this function

Chocobozzz:
Thanks @McFlat

PR #1241, opened 2018-10-11
Use DB informations from config/production.yaml in upgrade script

ldidry:
Avoid hardcoded values in upgrade script.
Avoid asking for DB password.
Uses python (usually installed on your system, even with minimal installations) and some of its standard lib modules.

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -23,8 +23,13 @@ fi
 
 # Backup database
 SQL_BACKUP_PATH="$PEERTUBE_PATH/backup/sql-peertube_prod-$(date +"%Y%m%d-%H%M").bak" 
+DB_USER=$(node -e "console.log(require('js-yaml').safeLoad(fs.readFileSync('$PEERTUBE_PATH/config/production.yaml', 'utf8'))['database']['username'])")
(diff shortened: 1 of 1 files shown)


scripts/upgrade.sh
```diff
@@ -23,8 +23,13 @@ fi
 
 # Backup database
 SQL_BACKUP_PATH="$PEERTUBE_PATH/backup/sql-peertube_prod-$(date +"%Y%m%d-%H%M").bak" 
+DB_USER=$($PEERTUBE_PATH/peertube-latest/node_modules/.bin/js-yaml $PEERTUBE_PATH/config/production.yaml | python -c 'import json,sys;obj=json.load(sys.stdin);print obj["'database'"]["'username'"]';)
```
Chocobozzz: May I suggest:

```
$ cd "$PEERTUBE_PATH/peertube-latest"
$ node -e "console.log(require('js-yaml').safeLoad(fs.readFileSync('$PEERTUBE_PATH/config/production.yaml', 'utf8'))['database']['username'])"
$ ...
$ ...
```

To avoid python calls?

ldidry: Sure. I didn't knew how to do that in NodeJS, that's why I did it in Python.

ldidry: Done.

Chocobozzz:
Thanks @ldidry

PR #1252, opened 2018-10-12
add Content Security Policy

### LINKED ISSUE #1190: Implement Content Security Policy (CSP)
This would improve the security, and prevent XSS attacks. Here's a good overview:

https://www.html5rocks.com/en/tutorials/security/content-security-policy/

Also check out these links:
https://infosec.mozilla.org/guidelines/web_security#content-security-policy
https://observatory.mozilla.org/analyze.html?host=peertube.social

rigelk:
resolves #1190


Nutomic:
If you paste the raw header, I can test it again on my instance.

rigelk:
@Nutomic `Content-Security-Policy: default-src 'none'; connect-src *; media-src 'self'; font-src 'self' data:; img-src 'self' data:; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; form-action 'self'; frame-ancestors 'none'; base-uri 'self'; plugin-types 'none'; manifest-src 'self'; frame-src 'self'; worker-src 'self'; upgrade-insecure-requests`

beware though, on the embed route `frame-ancestors` is set to `*`.

Nutomic:
Okay I set the different headers for / and /videos/embed in Nginx with report-only, so far no reports. Will let you know if something comes up.

By the way we still need to get rid of inline CSS and Javascript.

rigelk:
@Nutomic there is no way to get rid of `unsafe-inline`, but browsers supporting CSP 2 can ignore it if we specify a hash or a nonce. That's why I've added SRI, but I still need to dynamically add the hash to the CSP, and make sure it changes for each language.

Nutomic:
Looks like we have to change `media-src` to `'self' blob`: https://stackoverflow.com/questions/28467789/content-security-policy-object-src-blob#31258285

```
[2018-10-14 03:38:39.258] [WARN] csp - violation of [media-src] from [default-src 'none'; connect-src *; media-src 'self'; font-src 'self' data:; img-src 'self' data:; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; form-action 'self'; frame-ancestors 'none'; base-uri 'self'; plugin-types 'none'; manifest-src 'self'; frame-src 'self'; worker-src 'self'; upgrade-insecure-requests; report-uri /csp-reporting;]
  on https://peertube.social/videos/watch/bf1db223-a61b-4ab6-a44c-609a82569327 by blob
  agent: Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/69.0.3497.100 Safari/537.36
  referrer: https://www.youtube.com/
```

And it would probably be a good idea to release this as report-only at first. Or at least test the actual code on a live instance.

rigelk:
Tests are failing because of the `subresourceIntegrity` option added in the production angular configuration. I'm on this.

rigelk:
Okay, adding SRI/nonces to remove the `unsafe-inline` is probably better in a second PR.

@Nutomic if you don't see any other problem, I'll merge it ASAP :)

Nutomic:
Chromium is still showing this error for me, even though it should be fixed (https://github.com/Chocobozzz/PeerTube/issues/1190#issuecomment-429066791):
```
Invalid plugin type in 'plugin-types' Content Security Policy directive: ''none''. Did you mean to set the object-src directive to 'none'?
```
Looks like we have to set `connect-src * data:` ([link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src)):
```
[2018-10-16 18:37:45.128] [WARN] csp - violation of [connect-src *] from [default-src 'none'; connect-src *; media-src 'self' blob:; font-src 'self' data:; img-src 'self' data:; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; form-action 'self'; frame-ancestors 'none'; base-uri 'self'; plugin-types 'none'; manifest-src 'self'; frame-src 'self'; worker-src 'self'; upgrade-insecure-requests; report-uri /csp-reporting;]
  on https://peertube.social/videos/watch/fd3c1a4f-4a77-403a-aa07-f9a855482fee by data line:1
  agent: Mozilla/5.0 (iPad; CPU OS 10_3_1 like Mac OS X) AppleWebKit/603.1.30 (KHTML, like Gecko) Version/10.0 Mobile/14E304 Safari/602.1
  referrer: https://www.reddit.com/r/nsfwasmr/
```

There are still a couple of violations of media-src, but they are all from the same user, and I can't reproduce it in Firefox or Chromium:
```
[2018-10-16 18:27:47.001] [WARN] csp - violation of [media-src 'self' blob:] from [default-src 'none'; connect-src *; media-src 'self' blob:; font-src 'self' data:; img-src 'self' data:; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; form-action 'self'; frame-ancestors 'none'; base-uri 'self'; plugin-types 'none'; manifest-src 'self'; frame-src 'self'; worker-src 'self'; upgrade-insecure-requests; report-uri /csp-reporting;]
  on https://peertube.social/videos/watch/72aa9b2c-8225-4dbf-b1e8-3abc478d32e4 by https://video.pyrignis.fr
  agent: Mozilla/5.0 (iPhone; CPU iPhone OS 11_4_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/11.0 Mobile/15E148 Safari/604.1
  referrer: https://joinpeertube.org/en/
```

This one might just be because I misconfigured the embed policy in Nginx:
```
[2018-10-16 18:50:46.478] [WARN] csp - violation of [frame-ancestors] from [default-src 'none'; connect-src *; media-src 'self' blob:; font-src 'self' data:; img-src 'self' data:; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; form-action 'self'; frame-ancestors 'none'; base-uri 'self'; plugin-types 'none'; manifest-src 'self'; frame-src 'self'; worker-src 'self'; upgrade-insecure-requests; report-uri /csp-reporting;]
  on https://peertube.social/videos/embed/07dc0814-b769-4206-b984-384cdddafdfc?autoplay=1&auto_play=1 by https://peertube.social/videos/embed/07dc0814-b769-4206-b984-384cdddafdfc?autoplay=1&auto_play=1
  agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/69.0.3497.100 Safari/537.36
```

I still don't think we should release it in this form. Either include the CSP as report-only, or run it on a test instance for a while.

server/middlewares/csp.ts
```diff
@@ -0,0 +1,42 @@
...
+  browserSniff: false,
+  reportOnly: true
+})
+
+const embedCSP = helmet.contentSecurityPolicy({
+  directives: Object.assign(baseDirectives, {
+    frameAncestors: ['*']
+  }),
+  browserSniff: false, // assumes a modern browser, but allows CDN in front
+  reportOnly: true
```
Nutomic: With this configuration, violations will only be reported in the browser console. So it would be really hard to tell if it works, without asking all the users. It would be better if you either hardcode a report url, so admins can collect data with csp-logger, or if you included csp-logger which puts the reports in a known location.

server/middlewares/csp.ts
```diff
@@ -0,0 +1,42 @@
...
+  // objectSrc: ["'none'"], // only define to allow plugins, else let defaultSrc 'none' block it
+  formAction: ["'self'"],
+  frameAncestors: ["'none'"],
+  baseUri: ["'self'"],
+  pluginTypes: ["'none'"],
+  manifestSrc: ["'self'"],
+  frameSrc: ["'self'"], // instead of deprecated child-src / self because of test-embed
+  workerSrc: ["'self'"], // instead of deprecated child-src
+  upgradeInsecureRequests: true,
+  blockAllMixedContent: true
```
mkoppmann: This line is not needed, because `upgrade-insecure-requests` already blocks HTTP requests.

rigelk: Indeed, even the (mozilla) documentation says "The `upgrade-insecure-requests` directive is evaluated before `block-all-mixed-content` and if it is set, the latter is effectively a no-op."

Correcting. Thanks!

server/middlewares/csp.ts
rigelk: @Nutomic added in https://github.com/Chocobozzz/PeerTube/pull/1252/commits/1b92e937ff5aeac7366fa8004cf02681f17ed25c

rigelk:
PR is merge-ready @Chocobozzz

server/controllers/client.ts
```diff
@@ -22,6 +22,7 @@ clientsRouter.use('/videos/watch/:id',
 
 clientsRouter.use('' +
   '/videos/embed',
+  embedCSP,
```
Chocobozzz: It should be already done with https://github.com/Chocobozzz/PeerTube/pull/1252/files#diff-5c092da67283b0b787b618509fa7b96fR58

server/middlewares/csp.ts
```diff
@@ -0,0 +1,45 @@
+import * as helmet from 'helmet'
+import { CONFIG } from '../initializers/constants'
+
+const baseDirectives = Object.assign({},
+  {
+    defaultSrc: ["'none'"], // by default, not specifying default-src = '*'
+    connectSrc: ['*', 'data:'],
+    mediaSrc: ["'self'", 'blob:'],
```
Chocobozzz: I think we could specify a remote HTTP URL too (on http fallback of a remote video)

support/docker/production/config/custom-environment-variables.yaml
```diff
@@ -101,9 +101,13 @@ transcoding:
...
 
 instance:
   name: "PEERTUBE_INSTANCE_NAME"
   description: "PEERTUBE_INSTANCE_DESCRIPTION"
   terms: "PEERTUBE_INSTANCE_TERMS"
+
+services:
+  csp-logger:
+    __name: "PEERTUBE_SERVICES_CSPLOGGER"
+    __format: "json"
```
Chocobozzz: isn't a string?

server/controllers/client.ts
rigelk: `embedCSP` is purposedly different from `baseCSP`.

Chocobozzz: oups :+1:

support/docker/production/config/custom-environment-variables.yaml
rigelk: I don't know, the syntax isn't documented anywhere.

Nutomic: Is this the endpoint for the [csp-logger](https://github.com/mozilla/csp-logger)? In that case it should be an URL, no?

Nutomic: Maybe I'm misunderstanding how this file works, but it seems like json is the wrong type for an URL.

Chocobozzz: https://github.com/lorenwest/node-config/wiki/Environment-Variables#custom-environment-variables

Nutomic:
This PR has been merged for half a year now. When is it going to be ready for production use? Right now it is doing absolutely nothing.

On a related note, Mozilla suggests also setting the `X-Frame-Options` for older browsers: https://observatory.mozilla.org/analyze/peertube.social

Chocobozzz:
@Nutomic I don't understand. It's in production as opt-in config option: https://github.com/Chocobozzz/PeerTube/blob/develop/config/production.yaml.example#L110. It's up to you to enable it or not.

Nutomic:
Yes exactly, it says there that csp is still being tested, so that sounds like it is not ready for production use yet. If it is ready, it should be enabled by default for everyone.

PR #1260, opened 2018-10-13
fix migrations to not use config constant values as it can introduce …

### LINKED ISSUE #1259: Migrations should not use config constants as the constant values can change later
**What happened?**
Some of the migrations have field max values that come from config constants, this will introduce bugs later when the values change. We should instead use static string values.

What I see in [0120-video-null.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0120-video-null.ts) migration:
type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEOS.DESCRIPTION.max),

What I see in [0195-support.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0195-support.ts) migration:
```
type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEOS.SUPPORT.max),
type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEO_CHANNELS.SUPPORT.max),
type: Sequelize.STRING(CONSTRAINTS_FIELDS.USERS.DESCRIPTION.max),
type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEOS.DESCRIPTION.max),
```

What I see in [0245-user-blocked.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0245-user-blocked.ts) migration:
```
type: Sequelize.STRING(CONSTRAINTS_FIELDS.USERS.BLOCKED_REASON.max),
```

What I see in [0250-video-abuse-state.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0250-video-abuse-state.ts) migration:
```
type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEO_ABUSES.MODERATION_COMMENT.max),
```

What I see in [0255-video-blacklist-reason.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0255-video-blacklist-reason.ts) migration:
```
type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEO_BLACKLIST.REASON.max),
```

I expected to see in [0120-video-null.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0120-video-null.ts) migration: 
```
type: Sequelize.STRING(10000),
```

I expected to see in [0195-support.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0195-support.ts) migration:
```
type: Sequelize.STRING(500),
type: Sequelize.STRING(500),
type: Sequelize.STRING(250),
type: Sequelize.STRING(10000),
```

I expected to see in [0245-user-blocked.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0245-user-blocked.ts) migration:
```
type: Sequelize.STRING(250),
```

I expected to see in [0250-video-abuse-state.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0250-video-abuse-state.ts) migration:
```
type: Sequelize.STRING(300),
```

I expected to see in [0250-video-abuse-state.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0255-video-blacklist-reason.ts) migration:
```
type: Sequelize.STRING(300),
```

**Steps to reproduce:**

**Additional information**
In the [0195-support.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/migrations/0195-support.ts) migration we are using constants to set the field max values and this can go wrong if we change the max values in a subsequent migration. We should change any and all max values in migrations to be a string value set in that migration and not a constant that can later change.

```
{
    const data = {
      type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEOS.SUPPORT.max),
      allowNull: true,
      defaultValue: null
    }
    await utils.queryInterface.addColumn('video', 'support', data)
  }

  {
    const data = {
      type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEO_CHANNELS.SUPPORT.max),
      allowNull: true,
      defaultValue: null
    }
    await utils.queryInterface.addColumn('videoChannel', 'support', data)
  }

  {
    const data = {
      type: Sequelize.STRING(CONSTRAINTS_FIELDS.USERS.DESCRIPTION.max),
      allowNull: true,
      defaultValue: null
    }
    await utils.queryInterface.addColumn('account', 'description', data)
  }

  {
    const data = {
      type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEOS.DESCRIPTION.max),
      allowNull: true,
      defaultValue: null
    }
    await utils.queryInterface.changeColumn('video', 'description', data)
  }
```

* PeerTube version or URL:
PeerTube [1.0.0] (0e5ff97f)>

* Browser name/version:
Not applicable

* NodeJS version:
v9.11.2

McFlat:
…bugs later when they change; (fixes #1259)


McFlat:
The API-2 test is failing the first and third time because of `server/tests/api/server/redundancy.ts`:
```
AssertionError: Webseed http://localhost:9001/static/webseed/3e624fde-e5b5-4eab-b6c5-b26bd5b3e84a not found in magnet:?xs=http%3A%2F%2Flocalhost%3A9002%2Fstatic%2Ftorrents%2F3e624fde-e5b5-4eab-b6c5-b26bd5b3e84a-720.torrent&xt=urn:btih:dc10df0fa3b6fc0b2a511c5f8a58d25de70efcc3&dn=video+2+server+2&tr=ws%3A%2F%2Flocalhost%3A9002%2Ftracker%2Fsocket&tr=http%3A%2F%2Flocalhost%3A9002%2Ftracker%2Fannounce&ws=http%3A%2F%2Flocalhost%3A9002%2Fstatic%2Fwebseed%2F3e624fde-e5b5-4eab-b6c5-b26bd5b3e84a-720.mp4 on server http://localhost:9001: expected undefined not to be undefined
```
and failing the second attempt because of the last number in the port is different (`2`), maybe we can remove that last number `1` so it can match 1 or 2 or 3 or whatever the test server port starts it on for subsequent attempts.

rigelk:
I don't understand why you want to rewrite the previous migrations. Why not write a new migration that does database modifications for existing instances and not just new instances?

McFlat:
The reason is, because when a person starts a new install is will use the modified constant values right away, instead of how they started out and then a migration was created to modify them to be the new max value. This can introduce bugs for new installs when a migration runs that updates a field max value which came from a constant, possibly won't introduce bugs with installs that have a migration version, but with new installs it will be affected once we add migrations that modify field max values, it will just set the same value as in the constants, basically not modifying it at all etc.

McFlat:
I'm only fixing the existing migrations by setting the values statically to what they would already be, then when we modify the values later it will avoid any possible bugs. 

In migrations we should never use a variable like this, we should always use a static value which will always be the same expected value for that migration, otherwise things can behave unexpectedly in the future.

McFlat:
Current instance are already up to date to the current values, and so these migration fixes don't affect current instances, they only affect new instances for when we add new migrations that modify these values.

rigelk:
Ah okay I see, that makes sense now that I can read the commit (was on mobile) :')

McFlat:
Cool, yeah this is fixes some major bugs that could happen in the future. We should always use static values in migrations and not from variables that can change later and cause havoc. Going forward we should avoid using constants or variables in migrations that could be one value one moment and another value in the future.

This is ready to be merged once you guys approve of these fixes.

Issue #1258, opened 2018-10-13
My account settings "description cannot be more than 250 characters long"


McFlat:
<img width="323" alt="screen shot 2018-10-13 at 10 59 55 am" src="https://user-images.githubusercontent.com/43896676/46907393-46728180-ced7-11e8-9dae-09bfedae99e7.png">

We need to make the description to be much longer on the `/my-account/settings` page. I understand that the channels a user has can have much longer descriptions, but there's no real good reason to limit this section to be so short, not like there is all that much stuff on that page anyway. 

Also it doesn't seem to parse markdown, we should also enable markdown for this user description field, perhaps make it like the ones the users channels have. Also could be nice to add a support section just like the user channels have. Just to keep things consistent, in case visitors go to one place and not to another they will see it.

While we're at it we should probably increase the description and support field lengths from 500 to 1000 characters.

McFlat:
In the `0195-support.ts` migration we are using constants to set the field max values and this can go wrong if we change the max values in a subsequent migration. We should change any and all max values in migrations to be a string value set in that migration and not a constant that can later change.

```
{
    const data = {
      type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEOS.SUPPORT.max),
      allowNull: true,
      defaultValue: null
    }
    await utils.queryInterface.addColumn('video', 'support', data)
  }

  {
    const data = {
      type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEO_CHANNELS.SUPPORT.max),
      allowNull: true,
      defaultValue: null
    }
    await utils.queryInterface.addColumn('videoChannel', 'support', data)
  }

  {
    const data = {
      type: Sequelize.STRING(CONSTRAINTS_FIELDS.USERS.DESCRIPTION.max),
      allowNull: true,
      defaultValue: null
    }
    await utils.queryInterface.addColumn('account', 'description', data)
  }

  {
    const data = {
      type: Sequelize.STRING(CONSTRAINTS_FIELDS.VIDEOS.DESCRIPTION.max),
      allowNull: true,
      defaultValue: null
    }
    await utils.queryInterface.changeColumn('video', 'description', data)
  }
```

McFlat:
okay, well it seems that it does parse markdown, just not heading tags, ones starting with `#` char, but no worries, I guess main thing is links

McFlat:
#1262

McFlat:
[...]

PR #1262, opened 2018-10-13
Feature/migration description & support fields length 1000

McFlat:
Make Description field max length 1000 chars, 500 is good but 1000 in much better. 

Created 2 migrations in total; 1 migration to modify the field lengths of videoChannel and account, and 1 migration to add new support field to account.

Also added `support` field for account, in account settings, but it's still not showing up for some reason.


McFlat:
would be nice if can fix the 0285 migration to addColumn after some other column instead to the end of table after the createAt and updateAt fields before merging this pull request

McFlat:
I had to remove the support field I tried to add to the account table, because there's no way to make that field to be added after the description field on postgres, description field being 1000 max length should be enough now though.

McFlat:
pull request #1267 is the client/server part to resolve the issue, while this one 1262 is only the migration; all of it fixes the issue #1258

PR #1265, opened 2018-10-14
username/display_name/video_channel_name min length 1 and max length 50

### LINKED ISSUE #1263: Allow for username max length to be 50 chars instead of 20, 20 is too limiting
We need to allow longer usernames. Youtube api version 1 had shorter usernames, I think it was 20 chars, then in v3 they made the usernames to be 52 or 54 chars(not sure right now, I'll have to find that stackoverflow discussion again if you insist).

Maximum length of a Facebook username is currently 50 characters.

I reported this issue before, but it disappeared into thin air, so I'm bringing it up again because it's still an issue. I will make a pull request sometime today to make the username max length 50 chars.

However 64 chars would be better, just because Google and Facebook set a limit of 50, doesn't mean we have to set their exact limit. We are in fact creating this project because youtube is too limiting, if I'm not mistaken.

McFlat:
! still some bug on the frontend complains but if you remove the disabled property it creates the account just fine (fixes #1263)


McFlat:
yes, not sure if it's something on my end only, but it still complains about username being less than 3 or more than 20 chars, but can "Inspect" the elements and remove the `disabled` field from the submit button and it works fine.

McFlat:
We have this in the client code now:
```
    this.USER_USERNAME = {
      VALIDATORS: [
        Validators.required,
        Validators.minLength(1),
        Validators.maxLength(50),
        Validators.pattern(/^[a-zA-Z0-9]+[a-zA-Z0-9-._]+$/)
      ],
      MESSAGES: {
        'required': this.i18n('Username is required.'),
        'minlength': this.i18n('Username must be at least 1 character long.'),
        'maxlength': this.i18n('Username cannot be more than 50 characters long.'),
        'pattern': this.i18n('Username should start with a letter or number; dots, dashes and underscores are allowed.')
      }
    }
```
But in when I run it I see this validation error:
screen shot 2018-10-15 at 8 09 35 pm
screen shot 2018-10-15 at 8 10 48 pm

However I can remove the `disabled` attribute added to the submit/signup button and it all works fine, it creates the account.
screen shot 2018-10-15 at 8 11 36 pm

McFlat:
Could possibly be a bug in the validator

McFlat:
We should probably add some better tests for this buggy issue.

client/src/app/shared/forms/form-validators/user-validators.service.ts
```diff
@@ -23,15 +23,15 @@ export class UserValidatorsService {
     this.USER_USERNAME = {
       VALIDATORS: [
         Validators.required,
-        Validators.minLength(3),
-        Validators.maxLength(20),
-        Validators.pattern(/^[a-z0-9._]+$/)
+        Validators.minLength(1),
+        Validators.maxLength(50),
+        Validators.pattern(/^[a-zA-Z0-9]+[a-zA-Z0-9-._]+$/)
```
rigelk: Your regular expression is preventing `minLength` from being useful: the `pattern` test fails before the `minLength`.

Shouldn't the regex be `^[a-zA-Z0-9-._]+$` ?

McFlat: The reason I made it that way is to not allow users to have a username that starts with a `.`, `-`, `_`, so if they should want a single letter username it would have to be alphanumeric

McFlat:
Wait so you can create a username with a single letter or a username with more than 20 chars long without seeing the validation errors? if so then for sure it's only with my environment

McFlat:
I've been using `node 9.11.2` while working on this, I'll have to check with node version 8.x to see if it behaves differently.

client/src/app/shared/forms/form-validators/user-validators.service.ts
McFlat: that regular expression is actually buggy, it allows minimum 2 chars, I had to fix it

client/src/app/shared/forms/form-validators/user-validators.service.ts
```diff
@@ -23,15 +23,15 @@ export class UserValidatorsService {
     this.USER_USERNAME = {
       VALIDATORS: [
         Validators.required,
-        Validators.minLength(3),
-        Validators.maxLength(20),
-        Validators.pattern(/^[a-z0-9._]+$/)
+        Validators.minLength(1),
+        Validators.maxLength(50),
+        Validators.pattern(/^[a-zA-Z0-9][a-zA-Z0-9-._]*$/)
```
Chocobozzz: Please keep lowercases. With uppercases, it needs more work on backend

client/src/app/shared/forms/form-validators/video-channel-validators.service.ts
```diff
@@ -14,28 +14,28 @@ export class VideoChannelValidatorsService {
     this.VIDEO_CHANNEL_NAME = {
       VALIDATORS: [
         Validators.required,
-        Validators.minLength(3),
-        Validators.maxLength(20),
-        Validators.pattern(/^[a-z0-9._]+$/)
+        Validators.minLength(1),
+        Validators.maxLength(50),
+        Validators.pattern(/^[a-zA-Z0-9][a-zA-Z0-9-._]*$/)
```
Chocobozzz: same here (lowercase)

client/src/app/shared/forms/form-validators/user-validators.service.ts
McFlat: Ok I'll change it to lowercase, no worries

client/src/app/shared/forms/form-validators/video-channel-validators.service.ts
McFlat: ok sounds good

McFlat: I'll work on the `server/tests/api/check-params/video-channels.ts` right now

McFlat: The video channel name tests seem to be fine. We could add more test cases to check username `...` or channel name `...` etc that don't meet the criteria of the regex, but the one check with a space in the name 'super name' already covers one use case that should fail, maybe we should add more failing use cases?

McFlat:
All my tests seem to work all good, now that I figured out port 3000 not port 9000. Everything works great, I think this is ready to be merged unless you guys want to add something else like more tests that check the different variations of username or video channel names that can be created.

Chocobozzz:
Should be fixed if you rebase develop

McFlat:
I'll fix the merge conflict right now

client/src/app/shared/forms/form-validators/video-channel-validators.service.ts
```diff
@@ -14,28 +14,28 @@ export class VideoChannelValidatorsService {
     this.VIDEO_CHANNEL_NAME = {
       VALIDATORS: [
         Validators.required,
-        Validators.minLength(3),
-        Validators.maxLength(20),
-        Validators.pattern(/^[a-z0-9._]+$/)
+        Validators.minLength(1),
+        Validators.maxLength(50),
+        Validators.pattern(/^[a-z0-9][a-z0-9-._]*$/)
```
Chocobozzz: `-` char is not accepted by mastodon i think

McFlat: I'll have to fix that, or maybe we can have mastadon accept it?

rhaamo: There is two issues at least:
- https://github.com/tootsuite/mastodon/issues/9309
- https://github.com/tootsuite/mastodon/issues/9066

but that doesn't seems to be planned for the moment..

McFlat: To me it seems that mastodon accepts everything accept the `.` char, which it removes if it sees it.

https://github.com/tootsuite/mastodon/blob/master/app/validators/unique_username_validator.rb#L7

McFlat: screen shot 2018-11-25 at 6 04 44 am

rhaamo: I'm not sure this is the only thing.
According to Gargron here: https://github.com/tootsuite/mastodon/issues/9323#issuecomment-441005348 dashes are still refused from remote usernames.

McFlat: screen shot 2018-11-25 at 6 06 07 am

McFlat: I'd say it's safe to merge the pull request without any other changes to it

McFlat:
This pull request is ready for merge, the buggy issues I was having was due to my environment, specifically with accessing the app on a unrecommended port, other than that everything looks good. I also checked the username with mastodon and it seems to work fine without issues, we are compatible with mastodon.

client/src/app/shared/forms/form-validators/video-channel-validators.service.ts
rhaamo: This is still not safe, see my previous comment.
You are only showing basic HTML5 validation (field filled or not).
Validate the form and it will refuse it:
image

There is two issues still opened, and Gargron stated three days ago that dashes are **not accepted** from remote usernames.

rhaamo:
> I also checked the username with mastodon and it seems to work fine without issues, we are compatible with mastodon. Mastodon actually accepts spaces in their usernames too, but I don't think we should support spaces, just a longer length and dots, dashes, underscores is good enough.

Please see my previous comments, mastodon accepts **only** : letters, numbers and underscores

Validate the register form and you will see it yourself:
image

client/src/app/shared/forms/form-validators/video-channel-validators.service.ts
McFlat: Here's a scenario where a user can have spaces in their very long username, even beyond the 50 chars limit we have with this pull request. `very-long username-a user likes just because he likes it`
screen shot 2018-11-25 at 6 14 14 am

rhaamo: No, you didn't **submitted** the form, the HTML5 shows a green border only because the field is **filled** and not empty (because of the "required" html attribute).

PR #1267, opened 2018-10-14
Feature/description support fields length 1000

McFlat:
This is the client/server part of pull request #1262 which was only the migration, but the migration is included in this pull request so I closed the other pull request(1262)


Jorropo:
What is the performance impact ?
(How many it slow down peertube) ?

server/initializers/migrations/0280-description-support.ts
```diff
@@ -0,0 +1,53 @@
...
+      defaultValue: null
+    }
+    await utils.queryInterface.changeColumn('videoChannel', 'description', data)
+  }
+
+  {
+    const data = {
+      type: Sequelize.STRING(1000),
+      allowNull: true,
+      defaultValue: false
```
Chocobozzz: null

McFlat: ah yes, you're right

McFlat: I've also increased the video support side on the client side, thanks for pointing that out

client/src/app/shared/forms/form-validators/video-validators.service.ts
```diff
@@ -79,7 +79,7 @@ export class VideoValidatorsService {
     }
 
     this.VIDEO_SUPPORT = {
-      VALIDATORS: [ Validators.minLength(3), Validators.maxLength(500) ],
+      VALIDATORS: [ Validators.minLength(3), Validators.maxLength(1000) ],
       MESSAGES: {
         'minlength': this.i18n('Video support must be at least 3 characters long.'),
         'maxlength': this.i18n('Video support cannot be more than 500 characters long.')
```
Chocobozzz: 1000 :)

rigelk:
```suggestion
         'maxlength': this.i18n('Video support cannot be more than 1000 characters long.')
```

McFlat: I've made the fix, not sure where my focus went haha

Chocobozzz:
Thanks @McFlat

PR #1272, opened 2018-10-14
add cli option to run without client

rigelk:


McFlat:
can we rename this npm script to be `start:server` instead of `start:noClient`? 

Perhaps if we have a server only script we should also add a client only script, `start:client`

rigelk:
@McFlat mhm yes that's probably more consistent

McFlat:
@rigelk great work on this so far, I can see how some admins will need this feature

PR #1276, opened 2018-10-15
ability to enable/disable "_channel" suffix for created accounts

### LINKED ISSUE #1275: Allow admins to set the created channel suffix, so can be '' or '_channel'; a boolean
On my instance after creating a channel I always go into the database manually and remove the suffix from the channel name. It would be cool if we can make this configurable in the config, to add a suffix to the created user username when creating a channel or to not add the suffix. We can just keep the setting like that, as a boolean.

```
let channelName = userCreated.username + '_channel'
```

all works fine after I just go into the database manually and remove the suffix so I figure that it won't hurt if we allow admins to modify it, it's only a visual part of the url, but the url can determine if it's a channel or a user by other means instead of the '_channel' suffix.

McFlat:
fixes #1275

### DIFF
--- a/client/src/app/shared/video/video-miniature.component.ts
+++ b/client/src/app/shared/video/video-miniature.component.ts
@@ -34,6 +34,7 @@ export class VideoMiniatureComponent implements OnInit {
     // If the video channel name an UUID (not really displayable, we changed this behaviour in v1.0.0-beta.12)
     // -> Use the account name
     if (
+      this.video.channel.name === this.video.account.name ||
       this.video.channel.name === `${this.video.account.name}_channel` ||
       this.video.channel.name.match(/^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/)
     ) {
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -103,6 +103,7 @@ signup:
   enabled: false
   limit: 10 # When the limit is reached, registrations are disabled. -1 == unlimited
   requires_email_verification: false
+  channel_suffix: false # when this is true, adds '_channel' to the end of username for channel when user created.
   filters:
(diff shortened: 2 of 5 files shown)


McFlat:
This is not meant to be turned on and off from the frontend client, so let's just leave this in the config for admins to set it and forget it

Chocobozzz:
Thanks for the issue @McFlat but see my comment: https://github.com/Chocobozzz/PeerTube/issues/1275#issuecomment-429759952

McFlat:
@Chocobozzz they aren't duplicates, one is of type `Person` and the other is of type `Group`

McFlat:
@Chocobozzz I don't mean to keep bugging you about this but after a user creates an account and a channel is created for the user with the '_channel' suffix, the user can create a channel on their own that doesn't even have the suffix. Here's a screenshot of the use case, really this pull request could be merged without any problems. So either we fix the bug where it doesn't add the suffix to any created channel after that or we allow admins to either add or remove the suffix altogether and fix the bug so it will add the suffix for any channels created after as well depending on if true or false to use the suffix. My preference would be to allow admins to disable the suffix and fix the bug. I can fix it if you allow me to do it.

screen shot 2018-10-17 at 2 26 09 pm

McFlat:
But webfinger discovers users, not channels right? so if a user comments on another instance then it's that user that's making the comment not the channel doing it.

> WebFinger is a protocol specified by the Internet Engineering Task Force IETF that allows for discovery of information about people and things identified by a URI.[1] Information about a person might be discovered via an "acct:" URI, for example, which is a URI that looks like an email address.

https://en.wikipedia.org/wiki/WebFinger

If anything we could add `channel:` to the discovery to extend the webfinger, but as it stands right now, it functions by the actor `User`, not the actor `Channel` aka `Group`. Even if there is a duplicate name `Group` then it doesn't affect the webfinger, because it uses the actor `User` for the federation not the actor `Group`.

Chocobozzz:
There is no `channel:` or `chan:` URI scheme spec in webfinger

PR #1279, opened 2018-10-15
fix tslint config

BO41:
both `tslint.json`s show warnings about a wrong configuration

https://palantir.github.io/tslint/rules/member-ordering/

https://palantir.github.io/tslint/rules/indent/

### DIFF
--- a/client/tslint.json
+++ b/client/tslint.json
@@ -9,9 +9,15 @@
     "no-unused-variable": false, // Memory issues
     "member-ordering": [true, {
       "order": [
-        "public-before-private",
-        "static-before-instance",
-        "variables-before-functions"
+        "public-static-field",
+        "private-static-field",
+        "public-instance-field",
+        "private-instance-field",
+        "public-constructor",
+        "private-constructor",
+        "public-instance-method",
+        "protected-instance-method",
+        "private-instance-method"
         ]}
     ],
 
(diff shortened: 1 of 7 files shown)


rigelk:
I'm still getting a warning `Warning: member-ordering - Bad member kind: public-before-private` with your modifications.

client/tslint.json
```diff
@@ -7,10 +7,12 @@
...
-    "member-ordering": [true,
-      "public-before-private",
-      "static-before-instance",
-      "variables-before-functions"
+    "member-ordering": [true, {
+      "order": [
+        "public-before-private",
+        "static-before-instance",
+        "variables-before-functions"
+        ]}
```
rigelk: A similar syntax by the tslint docs should be:

```
"public-static-field",
"public-instance-field",
"public-constructor",
"private-static-field",
"private-instance-field",
"private-constructor",
"public-instance-method",
"protected-instance-method",
"private-instance-method"
```

BO41: I can't reproduce the warnings you get. my ide seems not to show them.
But you are right, according to the docs `"public-before-private"` and the others are not allowed, yet they still seem to work.

Should I change it to your proposal or to something of the examples?

rigelk: Try with a terminal and `npm run travis -- lint`.

Yes, I've not tried this proposal but it should be close enough :crossed_fingers:

rigelk:
Thanks @BO41

P.S.: I'll let @Chocobozzz answer on that one :sweat_smile:

Chocobozzz:
> quick question btw:

It's because we share the same template between `user-create` and `user-update`

Issue #1277, opened 2018-10-15
Optimize Website Load Speed


Nutomic:
Out of curiosity I just run some loading speed tests on peertube.social, and it seems there is some room for improvement. Here are the tests I tried:

https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fpeertube.social%2F&tab=desktop
https://gtmetrix.com/reports/peertube.social/fcff8bfY
https://www.uptrends.com/tools/website-speed-test
https://developers.google.com/web/tools/lighthouse/

The main suggestions are:
- [ ] get rid of render-blocking Javascript and CSS
- [x] specify browser caching for static files
- [ ] <strike>enable compression for json (seems like this would be a risk for [BREACH](http://breachattack.com/), and anyway would only save 26 KB out of 912 KB)</strike>

Chocobozzz:
> get rid of render-blocking Javascript and CSS

PeerTube HTML is built by JavaScript so I'm not sure we can do anything here (or use server side rendering but it's quite difficult).

> specify browser caching for static files

Do you have the URLs?

> enable compression for json (seems like this would be a risk for BREACH, and anyway would only save 26 KB out of 912 KB)

Yep, we do not enabled it on JSON endpoints for this reason

rigelk:
Maybe we can improve the lazy-loading of the JS and CSS?

BO41:
one other thing I was wondering about:
using `'use strict';` in JavaScript can make the code [run faster](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode)(2. Point).
So shouldn't the transpiled typescript contain this string?

I am not sure if there is already an option in typescript or angular. But if I understand it correctly, there should be one (so probably needs a fix upstream).

any thoughts on this?

rigelk:
@BO41 Typescript has a `strict` option, which enables `noImplicitAny` and `strictNullChecks` and generates the `'use strict'` you refer to. So a prerequisite would be to make our code compatible with these two options.

Nutomic:
@Chocobozzz Here are the files listed by gtmetrix.com:
```
/api/v1/config/ (expiration not specified)
/api/v1/oauth-clients/local (expiration not specified)
/api/v1/videos/categories (expiration not specified)
/api/v1/videos/languages (expiration not specified)
/api/v1/videos/licences (expiration not specified)
/api/v1/videos/privacies (expiration not specified)
[...]

PR #1280, opened 2018-10-15
add useStrict flag

BO41:
as described in https://github.com/Chocobozzz/PeerTube/issues/1277#issuecomment-429799326, using `'use strict';` would be a good decision.

No problems seem to appear.

### DIFF
--- a/client/tsconfig.json
+++ b/client/tsconfig.json
@@ -8,6 +8,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "noImplicitAny": false,
+    "alwaysStrict": true,
     "target": "es5",
     "typeRoots": [
       "node_modules/@types"


BO41:
>@BO41 Typescript has a strict option, which enables noImplicitAny and strictNullChecks and generates the 'use strict' you refer to. So a prerequisite would be to make our code compatible with these two options.

@rigelk It seems that `noImplicitAny` and `strictNullChecks` are not needed to be active.
The code seem compatible already

rigelk:
Ah indeed! Typescript make a difference between strict and alwaysStrict options apparently:

```
As of TypeScript 2.3 in early June 2017, the --strict flag enables the following four compiler options:

    --strictNullChecks
    --noImplicitAny
    --noImplicitThis
    --alwaysStrict
```

We only need alwaysStrict to benefit from the `'use strict'`.

BO41:
@rigelk thanks for clarifying :+1: 
today, it would be even more

--strict | boolean | false | Enable all strict type checking options. Enabling --strict enables --noImplicitAny, --noImplicitThis, --alwaysStrict, --strictNullChecks, --strictFunctionTypes and --strictPropertyInitialization.
-- | -- | -- | --

rigelk:
Thanks @BO41!

PR #1282, opened 2018-10-15
increase timeout in redundancy tests

McFlat:
Increase timeout so to avoid the test failing because `expect undefined not to be undefined`

```
Should cache video 2 webseed on the first video
```

Been having to push the code over and over again because this test would fail https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/server/redundancy.ts#L454

### DIFF
--- a/server/tests/api/server/redundancy.ts
+++ b/server/tests/api/server/redundancy.ts
@@ -456,19 +456,19 @@ describe('Test videos redundancy', function () {
 
       await waitJobs(servers)
 
-      await wait(7000)
+      await wait(10000)
 
       try {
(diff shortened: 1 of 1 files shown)


rigelk:
You're right. It's been annoying me lately as I had to frantically restart the `api-2` test for all PRs… :persevere:

rigelk:
Since the error is not due to a timeout, the PR can't fix anything. The issue is real however - any idea how this could be fixed @Chocobozzz?

Chocobozzz:
@rigelk By removing the test :p

I'll try to fix it

McFlat:
ok sounds good, thanks guys

Issue #946, opened 2018-08-22
Video metadata should include created/published dates


scanlime:
It would be really useful for video metadata to optionally include not just the date the content was uploaded, but the date it was originally produced or published.

Lots of people are using PeerTube to host archival content, but often I'll open a video and have no idea when it was from. Diligent uploaders could use the description to indicate this, but the current importer script doesn't do this. Without manual intervention, videos imported from YouTube for example have no information about when the content was produced.

It would be really useful to include date fields in the UI, defaulting to the current day. Later on, it would be great to search for archival content according to its production date rather than its upload date.

The metadata used by Archive.org is an inspiration to me, and might provide some other clues about data we may want to start collecting as early as we can.

Chocobozzz:
Some students are working on this issue. :)

am97:
I guess we should discuss the details about the implementation here :

At the moment, according to this [file](https://github.com/Chocobozzz/PeerTube/blob/4293e362af90fd61f22d5b88d926165e970e53db/shared/models/videos/video.model.ts#L40) a video has three dates : 
- `createdAt` 
- `updatedAt`
- `publishedAt`
    - When privacy changes from private to public/unlisted, this field is set to NOW ([more info](https://github.com/Chocobozzz/PeerTube/commit/2922e048de95738b3319054ce0778f873a34a0ee))

We want to add a field `producedAt`, which would be the date when a video was originally produced (e.g. 05/02/1936 for Charlie Chaplin - Modern Times)

am97:
First question: should we add also a `uploadedAt` field? This would be the date when the video was uploaded on another platform, like YouTube. But we may end up having too much dates for a video.

am97:
Second: what would be the right way to display the dates in the web UI? Both side by side? Or should the uploader choose the date to be displayed?

Displaying the dates side by side would be a bit ugly:
![screenshot_2018-10-22](https://user-images.githubusercontent.com/18473412/47303788-2c4a4900-d625-11e8-884e-449795b973cf.png)

scanlime:
BO41, the point of having this in metadata other than description is we can use it as a sort key. I’d like to be able to sort channels and playlists by the original publication date. Otherwise episodes of a series will be out of order if they didn’t all transcode in order.

BO41:
@scanlime ah I see. makes sense. :+1: 
Then @am97 I would add the information at the bottom, to the language and the license.

clementbrizard:
[...]

PR #1285, opened 2018-10-15
WIP : add production date for videos

clementbrizard:
This PR is related to the following issue #946.
I'm working on it with @auberanger and @am97, we are all CS students.

This first commit is the very beginning of our work, a kind of "Hello World" commit to PeerTube's code after a few documentation contributions.

Until now, we called the field `producedAt`. It can be changed of course.

### DIFF
--- a/client/src/app/shared/forms/form-validators/video-validators.service.ts
+++ b/client/src/app/shared/forms/form-validators/video-validators.service.ts
@@ -16,6 +16,7 @@ export class VideoValidatorsService {
   readonly VIDEO_TAGS: BuildFormValidator
   readonly VIDEO_SUPPORT: BuildFormValidator
   readonly VIDEO_SCHEDULE_PUBLICATION_AT: BuildFormValidator
+  readonly VIDEO_ORIGINALLY_PUBLISHED_AT: BuildFormValidator
 
   constructor (private i18n: I18n) {
 
@@ -91,6 +92,11 @@ export class VideoValidatorsService {
       MESSAGES: {
         'required': this.i18n('A date is required to schedule video update.')
       }
+    }
+
+    this.VIDEO_ORIGINALLY_PUBLISHED_AT = {
+      VALIDATORS: [ ],
+      MESSAGES: {}
     }
   }
 }
--- a/client/src/app/videos/+video-edit/shared/video-edit.component.html
+++ b/client/src/app/videos/+video-edit/shared/video-edit.component.html
@@ -114,6 +114,20 @@
               </div>
             </div>
 
+            <div class="form-group">
+              <label i18n for="originallyPublishedAt">Original publication date</label>
+              <my-help i18n-preHtml preHtml="This is the date when the content was originally published (e.g. the release date for a film)"></my-help>
+              <p-calendar
+                id="originallyPublishedAt" formControlName="originallyPublishedAt" [dateFormat]="calendarDateFormat"
+                [locale]="calendarLocale" [showTime]="true" [hideOnDateTimeSelect]="true" [monthNavigator]="true" [yearNavigator]="true" [yearRange]="myYearRange"
+              >
+              </p-calendar>
+
+              <div *ngIf="formErrors.originallyPublishedAt" class="form-error">
+                {{ formErrors.originallyPublishedAt }}
+              </div>
+            </div>
+
             <my-peertube-checkbox
               inputName="nsfw" formControlName="nsfw"
               i18n-labelText labelText="This video contains mature or explicit content"
--- a/client/src/app/videos/+video-edit/shared/video-edit.component.ts
+++ b/client/src/app/videos/+video-edit/shared/video-edit.component.ts
@@ -45,6 +45,7 @@ export class VideoEditComponent implements OnInit, OnDestroy {
 
   calendarLocale: any = {}
   minScheduledDate = new Date()
+  myYearRange = '1880:' + (new Date()).getFullYear()
 
   calendarTimezone: string
   calendarDateFormat: string
@@ -99,7 +100,8 @@ export class VideoEditComponent implements OnInit, OnDestroy {
       thumbnailfile: null,
       previewfile: null,
       support: this.videoValidatorsService.VIDEO_SUPPORT,
-      schedulePublicationAt: this.videoValidatorsService.VIDEO_SCHEDULE_PUBLICATION_AT
+      schedulePublicationAt: this.videoValidatorsService.VIDEO_SCHEDULE_PUBLICATION_AT,
+      originallyPublishedAt: this.videoValidatorsService.VIDEO_ORIGINALLY_PUBLISHED_AT
     }
 
     this.formValidatorService.updateForm(
--- a/client/src/app/videos/+video-watch/video-watch.component.html
+++ b/client/src/app/videos/+video-watch/video-watch.component.html
@@ -33,7 +33,6 @@
           <div>
             <div class="d-block d-sm-none"> <!-- only shown on small devices, has its conterpart for larger viewports below -->
               <h1 class="video-info-name">{{ video.name }}</h1>
-
               <div i18n class="video-info-date-views">
                 Published {{ video.publishedAt | myFromNow }} - {{ video.views | myNumberFormatter }} views
               </div>
@@ -158,6 +157,14 @@
           <div class="video-attribute">
             <span i18n class="video-attribute-label">Privacy</span>
             <span class="video-attribute-value">{{ video.privacy.label }}</span>
+          </div>
+
+          <div class="video-attribute">
+            <span i18n class="video-attribute-label">Originally published on</span>
+            <span *ngIf="!video.originallyPublishedAt" class="video-attribute-value">Unknown</span>
+            <span *ngIf="video.originallyPublishedAt" class="video-attribute-value">
+		    {{ video.originallyPublishedAt | date: 'dd MMMM yyyy' }}
+	    </span>
           </div>
 
           <div class="video-attribute">
--- a/server/initializers/migrations/0295-add-originally-published-at.ts
+++ b/server/initializers/migrations/0295-add-originally-published-at.ts
@@ -1,0 +1,42 @@
+import * as Sequelize from 'sequelize'
+
+async function up (utils: {
+  transaction: Sequelize.Transaction,
+  queryInterface: Sequelize.QueryInterface,
+  sequelize: Sequelize.Sequelize
+}): Promise<void> {
+
+  {
+    const data = {
+      type: Sequelize.DATE,
+      allowNull: true,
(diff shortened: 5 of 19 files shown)


client/src/app/shared/forms/form-validators/video-validators.service.ts
```diff
@@ -92,5 +93,10 @@ export class VideoValidatorsService {
         'required': this.i18n('A date is required to schedule video update.')
       }
     }
+
+    this.VIDEO_PRODUCED_AT = {
+      VALIDATORS: [ ],
```
clementbrizard: Nothing put here at the moment but we may add something.

client/src/app/videos/+video-edit/shared/video-edit.component.html
```diff
@@ -114,6 +114,11 @@
               </div>
             </div>
 
+            <div class="form-group">
+              <label i18n for="producedAt">Production date</label>
+              <input type="text" id="producedAt" formControlName="producedAt" />
```
clementbrizard: Here we may put the same type as the `schedulePublicationAt` field.

client/src/app/videos/+video-edit/shared/video-edit.component.ts
```diff
@@ -98,7 +98,8 @@ export class VideoEditComponent implements OnInit, OnDestroy {
       thumbnailfile: null,
       previewfile: null,
       support: this.videoValidatorsService.VIDEO_SUPPORT,
-      schedulePublicationAt: this.videoValidatorsService.VIDEO_SCHEDULE_PUBLICATION_AT
+      schedulePublicationAt: this.videoValidatorsService.VIDEO_SCHEDULE_PUBLICATION_AT,
+      producedAt: this.videoValidatorsService.VIDEO_PRODUCED_AT
```
clementbrizard: We may add in this file an `onProducedAtAdded()` function to check if the given date is prior to the creation date.

client/src/app/videos/+video-edit/shared/video-edit.component.html
McFlat: instead of `schedulePublicationAt` may I suggest `schedulePublishedAt`

McFlat: or perhaps instead of `schedulePublishedAt` just `publishedAt`, since the other field is `producedAt`

client/src/app/videos/+video-edit/shared/video-edit.component.ts
McFlat: no doubt it will always be produced before published, unless there was an error on the part of the uploader or some kind of bug, so lets not make it fail in that case, maybe throw up a warning, to notify about it but to stop it.

client/src/app/videos/+video-edit/shared/video-edit.component.html
McFlat: Unless this will be used to schedule when the video will be viewable by visitors, in that case it's fine I think, maybe in that case we'd have both `schedulePublicationAt` and `publishedAt`, one for when it is viewable and the other as a datetime stamp of when it was published at the original place it was uploaded, like youtube for instance.

client/src/app/shared/forms/form-validators/video-validators.service.ts
McFlat: when downloading/importing a video it uses youtube-dl to do it from many different providers. From youtube the video info json have these date field `"upload_date": "20100503"`, some other providers like bitchute don't give an `upload_date` or they may give it in a different format other than `Ymd`, so really I'm not sure we should put any validators, we should just try to parse it in the many different formats it can come in, if it exists and it not then oh well

McFlat: maybe on top of having the `producedAt` field we could also have an `uploadedAt` field in that case.

client/src/app/videos/+video-edit/shared/video-edit.component.ts
clementbrizard: We were indeed thinking about some error the user could make when setting this field. OK for displaying a warning.

client/src/app/videos/+video-edit/shared/video-edit.component.html
clementbrizard: > instead of `schedulePublicationAt` may I suggest `schedulePublishedAt`

`schedulePublicationAt` is already in production if I'm not wrong.

clementbrizard: > or perhaps instead of `schedulePublishedAt` just `publishedAt`, since the other field is `producedAt`

`publishedAt` is already used for when you make the video public : https://github.com/Chocobozzz/PeerTube/commit/2922e048de95738b3319054ce0778f873a34a0ee

client/src/app/shared/forms/form-validators/video-validators.service.ts
clementbrizard: Thanks for the review. This is more about specification so I suggest we go to the issue's discussion : https://github.com/Chocobozzz/PeerTube/issues/946

clementbrizard: > when downloading/importing a video it uses youtube-dl to do it from many different providers. From youtube the video info json have these date field `"upload_date": "20100503"`, some other providers like bitchute don't give an `upload_date` or they may give it in a different format other than `Ymd`, so really I'm not sure we should put any validators, we should just try to parse it in the many different formats it can come in, if it exists and it not then oh well

We may do it if it is decided to save the date when a video was uploaded on another platform.

McFlat: If the date is available from the other platform, then lets use it, if not then just ignore it. I'm thinking it will be useful for users to know when the video was uploaded so they can know if they've already watched it before, in case the same video is on multiple instances they follow they can use that as a metric to determine if it's a new upload or not.

am97:
We made some advancements recently, we are not far from finishing. To Do:
* some client side code to insert the original publication date in the database
* add original publication date to video description
We should finish this in the next days.

client/src/app/shared/forms/form-validators/video-validators.service.ts
```diff
@@ -16,6 +16,7 @@ export class VideoValidatorsService {
   readonly VIDEO_TAGS: BuildFormValidator
   readonly VIDEO_SUPPORT: BuildFormValidator
   readonly VIDEO_SCHEDULE_PUBLICATION_AT: BuildFormValidator
+  readonly VIDEO_ORIGINAL_PUBLISHED_AT: BuildFormValidator
```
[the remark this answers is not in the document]
rigelk: it's really similar. Make little difference to me.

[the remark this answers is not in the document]
Nutomic: The current version sounds wrong grammatically, `originally` is definitely better.

client/src/app/videos/+video-watch/video-watch.component.html
```diff
@@ -35,7 +35,7 @@
               <h1 class="video-info-name">{{ video.name }}</h1>
 
               <div i18n class="video-info-date-views">
-                Published {{ video.publishedAt | myFromNow }} - {{ video.views | myNumberFormatter }} views
+                Published on {{ video.originalPublishedAt | date: 'dd MMMM yyyy' }} - {{ video.views | myNumberFormatter }} views
```
Chocobozzz: Do we really want to display the `originalPublishedAt` field by default? @Nutomic @rigelk

rigelk: We had settled on some variant of https://github.com/Chocobozzz/PeerTube/issues/946#issuecomment-432253009

Also please note the date format varies: https://en.wikipedia.org/wiki/Date_format_by_country | Do we already deal with date formats by language, or is it fine to leave that aside for now (probably?).

clementbrizard:
Hello there,
Just to say this PR is ready for review, we will try to be as reactive as possible to the feedbacks you could give us.

## To sum-up :
- the user can give the original publication date of his video when uploading it. This new variable was named `originallyPublishedAt`
- this date is displayed first when watching a video while the publication date is in the metadata

### Details of the behaviour depending on the settings when uploading a video
6 possibilities :
  - the user uploads a video, gives `originallyPublishedAt` and sets video to public : simplest case
  - same but video set to private : when video changed to public, `originallyPublishedAt` is not changed obviously
  - same but now the user doesn't give `originallyPublishedAt` : this receives the `createdAt` value.
  - same but `originallyPublishedAt` is not given. When video is set to public,  `originallyPublishedAt` follows the behaviour of `publishedAt` : updated with the current date (`NOW`)
  - now we consider schedule publications. First case : `originallyPublishedAt` is given. When video is published, `originallyPublishedAt` doesn't change obviously.
  - Final case : same as previous one but `originallyPublishedAt` is not given. When video is published, `originallyPublishedAt` is updated with current date, as `publishedAt`

See [the commit](https://github.com/Chocobozzz/PeerTube/pull/1285/commits/56dbe1e7932c44208cea92bd4883ab934d5e171c) for more details.

client/src/app/shared/forms/form-validators/video-validators.service.ts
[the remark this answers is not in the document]
clementbrizard: Done now

Chocobozzz:
We should continue to display the `publishedAt` attribute near the views (for the sake of consistency with the video miniature date), and display the `originallyPublishedAt` field in the `video-attribute` block.

Moreover, if `originallyPublishedAt` is not specified just set it to `null`, it will be simpler.

@rigelk What do you think?

am97:
> We should continue to display the `publishedAt` attribute near the views (for the sake of consistency with the video miniature date), and display the `originallyPublishedAt` field in the `video-attribute` block.
> 
> Moreover, if `originallyPublishedAt` is not specified just set it to `null`, it will be simpler.
> 
> @rigelk What do you think?

Ping @rigelk 
Should we keep the code as is, or do as @Chocobozzz suggested ?

rigelk:
We should do as suggested by @Chocobozzz. Consistency is indeed important, and changing miniatures to display the `originallyPublishedAt` date would disturb visitors who expect to see the date of publication.

server/initializers/migrations/0295-add-originally-published-at.ts
```diff
@@ -0,0 +1,42 @@
+import * as Sequelize from 'sequelize'
```
Chocobozzz: It's the migration number 325 now, and you need to update the LAST_MIGRATION_VERSION: https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/constants.ts#L19

Chocobozzz:
Merged manually. Thanks @clementbrizard

PR #1286, opened 2018-10-16
user/instance muting

Chocobozzz:
* user mutes an account
 * user mutes an instance
 * instance mutes an account  
 * instance mutes another instance 

With this code, it should not be hard to add user/instance blocking in the future (same tables, same endpoint with an additional parameter etc).


.travis.yml
```diff
@@ -42,6 +42,7 @@ matrix:
   - env: TEST_SUITE=api-1
   - env: TEST_SUITE=api-2
   - env: TEST_SUITE=api-3
+  - env: TEST_SUITE=api-4
```
rigelk: :cold_sweat:

client/src/app/+accounts/accounts.component.html
```diff
@@ -10,8 +10,15 @@
           <div class="actor-name">{{ account.nameWithHost }}</div>
 
           <span *ngIf="user?.blocked" [ngbTooltip]="user.blockedReason" class="badge badge-danger" i18n>Banned</span>
-
-          <my-user-moderation-dropdown buttonSize="small" [user]="user" (userChanged)="onUserChanged()" (userDeleted)="onUserDeleted()">
+          <span *ngIf="account.mutedByUser" class="badge badge-danger" i18n>Muted</span>
+          <span *ngIf="account.mutedServerByUser" class="badge badge-danger" i18n>Muted by your instance</span>
+          <span *ngIf="account.mutedByInstance" class="badge badge-danger" i18n>Instance muted</span>
+          <span *ngIf="account.mutedServerByInstance" class="badge badge-danger" i18n>Instance muted by your instance</span>
```
rigelk: We could use [dropdown headers](https://getbootstrap.com/docs/4.0/components/dropdowns/#menu-headers) to separate actions specific to the user from moderation actions that apply instance-wide.

Chocobozzz: https://github.com/Chocobozzz/PeerTube/issues/1290

PR #1293, opened 2018-10-16
Remove hard-coded 8GB upload limit in client

scanlime:
Ideally we'd know what the specific server's configured upload limit
is before starting, but this 8GB limit is not useful if an administrator
has changed the nginx post limit on the server.

### DIFF
--- a/client/src/app/videos/+video-edit/video-add-components/video-upload.component.ts
+++ b/client/src/app/videos/+video-edit/video-add-components/video-upload.component.ts
@@ -113,12 +113,6 @@ export class VideoUploadComponent extends VideoSend implements OnInit, OnDestroy
     const videofile = this.videofileInput.nativeElement.files[0] as File
     if (!videofile) return
 
-    // Cannot upload videos > 8GB for now
-    if (videofile.size > 8 * 1024 * 1024 * 1024) {
-      this.notificationsService.error(this.i18n('Error'), this.i18n('We are sorry but PeerTube cannot handle videos > 8GB'))
-      return
-    }
-
     const bytePipes = new BytesPipe()
     const videoQuota = this.authService.getUser().videoQuota
     if (videoQuota !== -1 && (this.userVideoQuotaUsed + videofile.size) > videoQuota) {
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -96,8 +96,18 @@ server {
     proxy_set_header Host $host;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 
-    # Hard limit, PeerTube does not support videos > 8GB
+    # This is the maximum upload size, which roughly matches the maximum size of a video file
+    # you can send via the API or the web interface. By default this is 8GB, but administrators
+    # can increase or decrease the limit. Currently there's no way to communicate this limit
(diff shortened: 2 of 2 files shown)


McFlat:
I think most video being uploaded won't even get close to that limit, even full length movies won't reach that limit. I personally lowered the limit to 4GB on my instance.

scanlime:
The limit was only in two places, the nginx config and the client. Client limit is going to be incorrect. I agree something easier to config would be better, but this patch fixes the problem for me today.

Just uploaded a 35GB file with this patch and all is well. It’s okay if not everyone needs this, but i find it useful and it doesn’t seem like it makes the client less correct or less useful for others.

Nutomic:
You should probably adjust the comment in the Nginx config too: `Hard limit, PeerTube does not support videos > 8GB`.

Chocobozzz:
Thanks for the PR @scanlime.

May I suggest to add a notification information if the video > 8GB? Like "Your video is very heavy, the upload could fail" or something like that.

Nutomic:
@Nerothos you can adjust it in the nginx config (client_max_body_size).

scanlime:
@chocobozzz, at what point would we decide to show that message, and what point would it serve? what is so special about 8gb?

McFlat:
What if say we give the users an option to sync the video to dropbox or some file sharing service and then allow them to import it from that service? so it wouldn't be a web browser doing the uploading of the video it can happen from a desktop sync software which could be more dependable in some use cases

scanlime:
I think you can use video import from URL to download from dropbox already, someone correct  me if this doesn’t work but it seems like it should be supported using existing tools.

In case of imports from url, torrent, or the command line tool, this limit never would have applied. This is why i’ve only noticed recently, i had been importing 16GB videos just fine from the command line :)

McFlat:
Yeah so in that case we can have a message state that if a user has large videos to import they can input multiple urls of the videos to import and they will be imported from a background job, then when it completes they will receive an email and go to update any info of the video before it goes public. 

Maybe we can allow users to import a list of videos from any url and they will just import and then email them to notify them when they are ready to be published that way?

Chocobozzz:
> and what point would it serve? what is so special about 8gb?

Because it's the default value of our nginx template. But if you want to set this value dynamically in admin configuration go for it

scanlime:
I've just increased the limit from 40GB to 200GB in nginx and I'm still using this patch; uploaded a 48GB video yesterday with no problem.

McFlat:
scanlime, what if there are say 10 users uploading a 40GB file? does it scale? I'm betting not so much

scanlime:
mcflat, i would then need 400GB of temp space, which i’m aware of and i have. I don’t understand what this argument is about; the higher upload limit works fine for me. are you arguing that no instance should have an upload limit higher than the default or what?

McFlat:
I'm suggesting that there is possibility of a denial-of-service

scanlime:
This PR is for a client side change, it has no bearing on the DoS resilience of the server or lack of.

scanlime:
no good reason i picked 200GB instead of unlimited; seemed slightly better to reject totally bogus sizes instead of not. anyway, i don’t think that has anything to do with this client side pull request

PR #1296, opened 2018-10-18
Improve some of the wording around abuse reports UI

### LINKED ISSUE #1295: Confusing term for Delete abuse
Hello !

As an administrator, I receive a report abuse.
I find confusing the term "Delete" 

image

Apparently it deletes the abuse, but we can think that it deletes the video.
Maybe juste change the term to "Delete the abuse" ?

barbeque:
To resolve #1295, I fixed these two strings to make it more obvious that it is referring to the _report_, and not the _content_, that is going to be deleted.

In order to update the i18n files, I ran `npm run build`, which gave me a lot of internationalization warnings for missing translations but only seemed to touch the `en-US` file. I hope this is OK for now, and that the translators to other languages are still able to translate these new strings without much difficulty.

PeerTube is a great project!


Chocobozzz:
I think @barbeque used the extractor :) 

Thanks!

rigelk:
Oh right, I overlooked the update id and track, sorry.

Issue #1298, opened 2018-10-18
optimize-old-videos throws Error: Model not initialized


Nutomic:
I just updated my instance to develop (commit bcf21a376f1e26cb3e74236e4cc41909310d4c32), and tried to run the `optimize-old-videos` script. It looks like we missed something in the PR, because it immediately throws an error:

```
$ sudo -u peertube NODE_CONFIG_DIR=/var/www/peertube/config NODE_ENV=production npm run optimize-old-videos

> peertube@1.1.0-alpha.1 optimize-old-videos /var/www/peertube/versions/peertube-develop
> node ./dist/scripts/optimize-old-videos.js

{ Error: Model not initialized: "VideoModel" needs to be added to a Sequelize instance before "scope" can be called.
    at new ModelNotInitializedError (/var/www/peertube/versions/peertube-develop/node_modules/sequelize-typescript/lib/models/errors/ModelNotInitializedError.js:18:16)
    at checkInitialization (/var/www/peertube/versions/peertube-develop/node_modules/sequelize-typescript/lib/models/BaseModel.js:64:23)
    at Function._target.(anonymous function) [as scope] (/var/www/peertube/versions/peertube-develop/node_modules/sequelize-typescript/lib/models/BaseModel.js:34:21)
    at Function.listLocal (/var/www/peertube/versions/peertube-develop/dist/server/models/video/video.js:146:29)
    at /var/www/peertube/versions/peertube-develop/dist/scripts/optimize-old-videos.js:24:54
    at Generator.next (<anonymous>)
    at /var/www/peertube/versions/peertube-develop/dist/scripts/optimize-old-videos.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (/var/www/peertube/versions/peertube-develop/dist/scripts/optimize-old-videos.js:3:12)
    at run (/var/www/peertube/versions/peertube-develop/dist/scripts/optimize-old-videos.js:23:12)
    at Object.<anonymous> (/var/www/peertube/versions/peertube-develop/dist/scripts/optimize-old-videos.js:16:1)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Function.Module.runMain (module.js:694:10)
    at startup (bootstrap_node.js:204:16)
    at bootstrap_node.js:625:3
[...]

PR #1304, opened 2018-10-18
Make backups of files in optimize-old-videos.ts

Nutomic:
As mentioned by @Chocobozzz (https://github.com/Chocobozzz/PeerTube/issues/1298#issuecomment-431040970), it makes sense to make backups of the video files before optimizing them. And it would be even better if that would be handled automatically by the script, so this is what I implemented. I also added some output to tell what the script is doing currently.

It currently compares the duration as a check, not sure if that's the best option.

### DIFF
--- a/scripts/optimize-old-videos.ts
+++ b/scripts/optimize-old-videos.ts
@@ -1,10 +1,11 @@
 import { CONFIG, VIDEO_TRANSCODING_FPS } from '../server/initializers/constants'
-import { getVideoFileBitrate, getVideoFileFPS, getVideoFileResolution } from '../server/helpers/ffmpeg-utils'
+import { getVideoFileBitrate, getVideoFileFPS, getVideoFileResolution, getDurationFromVideoFile } from '../server/helpers/ffmpeg-utils'
 import { getMaxBitrate } from '../shared/models/videos'
 import { VideoModel } from '../server/models/video/video'
 import { optimizeVideofile } from '../server/lib/video-transcoding'
 import { initDatabaseModels } from '../server/initializers'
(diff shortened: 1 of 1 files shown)


Chocobozzz:
LGTM

rigelk:
LGTM, but since we are doing potentially very long-running operations, we should try to ensure two things:

- make the script auto-heal (restore the backup) when interrupted (for instance if the user presses <kbd>Ctrl</kbd>+<kbd>C</kbd>.
- make the script go through all files to know which will need optimization, so that we can show the user 
  - how many files to transcode are remaining
  - how many GB are left to transcode

Nutomic:
Sorry meant to type "update". I thought it creates a temp file for transcoding. So we don't have to do anything (except delete the temp file).

rigelk:
Since we are also making a backup, we have to restore it too.

Nutomic:
We don't have to restore it, because the original file is still in place, unchanged. But yes, we should delete the backup file.

McFlat:
@rigelk No it's not, it's actually a separate issue. But I'm willing to bet it's related to this issue, because transcoding/importing a video shaves a second off of the original length of the video.

rigelk:
@McFlat please open a dedicated issue for that problem then.

Nutomic:
@rigelk Is anything else needed here? Pretty sure the test failures are not related to my changes.

rigelk:
@Nutomic no that should be ready to merge

Nutomic:
I ran this on my server now, it took about 2 days for 500 GB of videos. Looks like everything went fine, I didn't notice any problems.

PR #1306, opened 2018-10-18
WIP: update angular version

### LINKED ISSUE #1305: update to angular 7

BO41:
>Package "angular2-hotkeys" has an incompatible peer dependency to "rxjs" (requires "^5.5.0", would install "6.3.3").
Incompatible peer dependencies found. See above.

TODO:
- [ ] update core `ng update @angular/core`

closes #1305


BO41:
@rigelk should I take your changes from https://github.com/Chocobozzz/PeerTube/pull/1252/commits/51136cfed06dc8fd6c9ec51fd28f73d49f126fc1#diff-ddbfe834d8d4aa567d1a15674b8d343aR54 or keep what I did, and remove the include and its comment?
Regarding https://github.com/Chocobozzz/PeerTube/issues/1277#issuecomment-432773782

BO41:
>$ ng update @angular/cli @angular/core
An unexpected error happened; package webtorrent has no version null.

I am not sure what is going on, and why this error is thrown

PR #1313, opened 2018-10-20
(doc) add manpage to peertube cli

rigelk:
This mostly concerns distribution packagers.


server/tools/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+`import-videos|import`: import a video from a streaming platform to a remote instance
+
+    $ peertube import \
+        -u "PEERTUBE_URL" \
+        -U "PEERTUBE_USER" \
+        --password "PEERTUBE_PASSWORD" \
+        -t "TARGET_URL"
+
+    The target URL can be directly the video file, or any of the supported sites of youtube-dl. Already downloaded videos will not be uploaded twice, so you can run and re-run the script in case of crash, disconnection…
```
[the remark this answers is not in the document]
rigelk: @Nutomic the script downloads the video locally and then uploads it like the upload script (more or less). So we don't care whether the option is enabled by the server or not. It's just a normal upload.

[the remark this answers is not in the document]
Nutomic: Ah okay, I assumed it uses the server side upload. You might want to add a note about that then (because of data transfer etc).

package.json
```diff
@@ -119,6 +120,7 @@
     "jsonld-signatures": "https://github.com/Chocobozzz/jsonld-signatures#rsa2017",
     "lodash": "^4.17.10",
     "magnet-uri": "^5.1.4",
+    "marked-man": "^0.2.1",
```
Chocobozzz: dev dep no?

Chocobozzz: dev dep?

scripts/generate-cli-doc.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -eu
+
+node_modules/marked-man/bin/marked-man server/tools/README.md > dist/server/tools/peertube.8
```
Chocobozzz: :+1:

PR #1334, opened 2018-10-22
allow peertube-import-videos.ts CLI script to run concurrently

McFlat:
The way tool video imports works right now doesn't scale if you run multiple imports at the same time, it deletes another processes file, to fix the issue we should sha256 the url of the video being downloaded and the thumbnail as well because it can happen with thumbnails since currently it uses the same path, not allowing for multiple import jobs to run at once.

All tests pass, so that's good.

I tried to make this pull request before, but my previous account got flagged


rigelk:
Thanks @McFlat!

PR #1338, opened 2018-10-22
add possibility to configure multiple trust_proxy (was: #1106)

LecygneNoir:
Hello,

I have broken the PR #1106 when trying to rebase peertube/develop in the branch, so I create this new clean PR :-)

Following suggestion in the PR, I made tests with json instead of Yaml, and indeed found a way for Peertube to use trust_proxy directly from env variables without external compilation anymore :heart: 

We need to pass trust_proxy as a full array such as `["IP1", "IP2"]`  
If using a external deployment tool such as docker-compose or helm for kubernetes we need to surround it with simple quote: `'["IP1", "IP2"]'`

But with that, Peertube is able to parse the list without problem, tested interface, upload, play a video, update a video, etc. I do not have problems anymore where it crashed before.

I am currently using it from two days in my dev env without visible problem :pray: 

It's clearly more simple than the first PR with parsing yaml!

Do you need more tests or modification or do it seems acceptable?

Thanks :hugs:


Chocobozzz:
Thanks @LecygneNoir!

PR #1339, opened 2018-10-22
search UUID and description fixes #1330

### LINKED ISSUE #1330: Search for video UUID should return that video
**What happened?**

Pasted a video UUID into search (verified that the UUID worked when constructing a URL with it), no results returned.

Expected a search for any exact identifier for that video to return the video.

**Steps to reproduce:**

1. Paste UUID into search

McFlat:
Haven't tested it but it should work, will have to test it sometime today once I get some videos imported locally.

fixes #1330 

- make video search to search uuid instead of only full uri to the video, include description field in search
- attributeInclude uuid, description just in case

### DIFF
--- a/server/models/video/video.ts
+++ b/server/models/video/video.ts
@@ -1082,8 +1082,11 @@ export class VideoModel extends Model<VideoModel> {
               '(' +
               'SELECT "video"."id" FROM "video" ' +
               'WHERE ' +
+              'lower(immutable_unaccent("video"."uuid")) % lower(immutable_unaccent(' + escapedSearch + ')) OR ' +
               'lower(immutable_unaccent("video"."name")) % lower(immutable_unaccent(' + escapedSearch + ')) OR ' +
-              'lower(immutable_unaccent("video"."name")) LIKE lower(immutable_unaccent(' + escapedLikeSearch + '))' +
+              'lower(immutable_unaccent("video"."name")) LIKE lower(immutable_unaccent(' + escapedLikeSearch + ')) OR ' +
+              'lower(immutable_unaccent("video"."description")) % lower(immutable_unaccent(' + escapedSearch + ')) OR ' +
+              'lower(immutable_unaccent("video"."description")) LIKE lower(immutable_unaccent(' + escapedLikeSearch + ')) ' +
               'UNION ALL ' +
               'SELECT "video"."id" FROM "video" LEFT JOIN "videoTag" ON "videoTag"."videoId" = "video"."id" ' +
               'INNER JOIN "tag" ON "tag"."id" = "videoTag"."tagId" ' +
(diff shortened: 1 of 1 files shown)


server/models/video/video.ts
```diff
@@ -1082,8 +1082,11 @@ export class VideoModel extends Model<VideoModel> {
               '(' +
               'SELECT "video"."id" FROM "video" ' +
               'WHERE ' +
+              'lower(immutable_unaccent("video"."uuid")) % lower(immutable_unaccent(' + escapedSearch + ')) OR ' +
```
Chocobozzz: We want an exact search for the uuid, we don't need the similarity

server/models/video/video.ts
```diff
@@ -1082,8 +1082,11 @@ export class VideoModel extends Model<VideoModel> {
...
               'lower(immutable_unaccent("video"."name")) % lower(immutable_unaccent(' + escapedSearch + ')) OR ' +
-              'lower(immutable_unaccent("video"."name")) LIKE lower(immutable_unaccent(' + escapedLikeSearch + '))' +
+              'lower(immutable_unaccent("video"."name")) LIKE lower(immutable_unaccent(' + escapedLikeSearch + ')) OR ' +
+              'lower(immutable_unaccent("video"."description")) % lower(immutable_unaccent(' + escapedSearch + ')) OR ' +
+              'lower(immutable_unaccent("video"."description")) LIKE lower(immutable_unaccent(' + escapedLikeSearch + ')) ' +
```
Chocobozzz: I don't think a like on description is a good idea

server/models/video/video.ts
```diff
@@ -1082,8 +1082,11 @@ export class VideoModel extends Model<VideoModel> {
               '(' +
               'SELECT "video"."id" FROM "video" ' +
               'WHERE ' +
+              'lower(immutable_unaccent("video"."uuid")) % lower(immutable_unaccent(' + escapedSearch + ')) OR ' +
               'lower(immutable_unaccent("video"."name")) % lower(immutable_unaccent(' + escapedSearch + ')) OR ' +
-              'lower(immutable_unaccent("video"."name")) LIKE lower(immutable_unaccent(' + escapedLikeSearch + '))' +
+              'lower(immutable_unaccent("video"."name")) LIKE lower(immutable_unaccent(' + escapedLikeSearch + ')) OR ' +
+              'lower(immutable_unaccent("video"."description")) % lower(immutable_unaccent(' + escapedSearch + ')) OR ' +
```
Chocobozzz: We need to create an index of the x first description words/characters first

server/models/video/video.ts
Nutomic: It might be useful if you don't want to type the entire uuid string.

Chocobozzz: I don't think it would be useful, as we use random uuid v4 so a substring/similar uuid has no meaning

Chocobozzz:
Closing due to inactivity, feel free to request the reopening

Issue #1211, opened 2018-10-06
add PiP support


BO41:
I think it would be nice to support picture in picture, just like youtube now, or even better, like twitch.tv

**What happened?**

The video stops playing when I switch to a different site, or look for another video.
And I am unable to see the video when I scroll down to read a comment.

**What do you expect to happen instead?**

A small player should pop up and show the video.

parts:
- [ ] button to pop out pip for changing site https://github.com/Chocobozzz/PeerTube/issues/1211#issuecomment-427708393
- [x] open pip on scroll down
- [x] add video controls to pip player
- [ ] add options to control pip behavior

rigelk:
I imagine it like the video miniature player that goes to the corner in YouTube's mobile app. Is it how you envisioned it?

However it would only work while you stay on the same site, not beyond when you switch to a different site.

BO41:
@rigelk yes, exactly. Like the youtube app, youtube.com, newpipe, twitch.tv, ...
And of course only on PeerTube and the Angular app.

![screenshot_2018-10-06 twitch](https://user-images.githubusercontent.com/29450609/46570493-5208fa00-c965-11e8-9e2e-165817a750d8.png)
twitch.tv if you scroll or click somewhere else
![screenshot_2018-10-06 youtube](https://user-images.githubusercontent.com/29450609/46570497-59300800-c965-11e8-869e-e5b6b8902f91.png)
youtube if you click on the PiP button

rigelk:
Thanks for the screenshot! That's what I was missing. I didn't know they had a corner player (PiP sounds more like putting a video in a corner of the video itself), but it looks great.

McFlat:
+1

BO41:
@XenonFiber I see. Good point :+1: 
And having the small window when scrolling? This shouldn't bother, right?

XenonFiber:
@BO41 That would probably be fine since you're still on the same video page. It feels like something you should still have to choice to disable, though.

rigelk:
Following PR #1345, PR #1463 takes over. It's still needs to be disabled by default, and expose the option to enable it to non-authenticated users.

BenLubar:
You can right click a video, then right click again without moving the mouse, and then select "Picture in Picture" from the browser menu in browsers that support it.

This method also works on YouTube.

BO41:
Most major browsers now support PiP, so no need to implement it

Chocobozzz:
[...]

PR #1345, opened 2018-10-24
popup player

BO41:
what it looks at the moment
screenshot_2018-10-24 test
I could need some help by removing the controls, or just the download stats.

fixes part of #1211

### DIFF
--- a/client/src/app/videos/+video-watch/video-watch.component.scss
+++ b/client/src/app/videos/+video-watch/video-watch.component.scss
@@ -64,11 +64,21 @@ $other-videos-width: 260px;
   }
 }
 
+@keyframes fade {
+  from {
+    opacity: 0;
+  }
+  to {
+    opacity: 1;
+  }
+}
+
 .video-element-wrapper-popup {
   position: fixed;
   bottom: 20px;
   right: 20px;
   z-index: 30;
+  animation: fade 1.5s;
   box-shadow: 0 5px 10px 0 rgba(0,0,0,.3),0 1px 8px -3px rgba(0,0,0,.26),0 2px 16px 0 rgba(0,0,0,.28);
 
   /deep/ .video-js {
--- a/client/src/app/videos/+video-watch/video-watch.component.scss
+++ b/client/src/app/videos/+video-watch/video-watch.component.scss
@@ -64,11 +64,11 @@ $other-videos-width: 260px;
   }
 }
 
-@keyframes fade {
-  from {
+@include keyframes (fade) {
+  0% {
     opacity: 0;
   }
-  to {
+  100% {
     opacity: 1;
   }
 }
--- a/client/src/app/videos/+video-watch/video-watch.component.scss
+++ b/client/src/app/videos/+video-watch/video-watch.component.scss
@@ -69,10 +69,16 @@ $other-videos-width: 260px;
   bottom: 20px;
   right: 20px;
   z-index: 30;
+  box-shadow: 0 5px 10px 0 rgba(0,0,0,.3),0 1px 8px -3px rgba(0,0,0,.26),0 2px 16px 0 rgba(0,0,0,.28);
 
   /deep/ .video-js {
     width: 280px !important;
     height: 160px !important;
+  }
+
+  /deep/ .vjs-peertube-displayed,
+  /deep/ .vjs-theater-control {
+    display: none !important;
   }
 }
 
--- a/client/src/app/videos/+video-watch/video-watch.component.scss
+++ b/client/src/app/videos/+video-watch/video-watch.component.scss
@@ -64,6 +64,18 @@ $other-videos-width: 260px;
   }
 }
 
+.video-element-wrapper-popup {
+  position: fixed;
+  bottom: 20px;
+  right: 20px;
+  z-index: 30;
+
+  /deep/ .video-js {
+    width: 280px !important;
+    height: 160px !important;
+  }
+}
+
 .alert {
   text-align: center;
 }
--- a/client/src/app/videos/+video-watch/video-watch.component.ts
+++ b/client/src/app/videos/+video-watch/video-watch.component.ts
@@ -1,4 +1,4 @@
-import { catchError } from 'rxjs/operators'
+import { catchError, debounceTime } from 'rxjs/operators'
 import { ChangeDetectorRef, Component, ElementRef, Inject, LOCALE_ID, NgZone, OnDestroy, OnInit, ViewChild } from '@angular/core'
 import { ActivatedRoute, Router } from '@angular/router'
 import { RedirectService } from '@app/core/routing/redirect.service'
@@ -6,7 +6,7 @@ import { peertubeLocalStorage } from '@app/shared/misc/peertube-local-storage'
 import { VideoSupportComponent } from '@app/videos/+video-watch/modal/video-support.component'
 import { MetaService } from '@ngx-meta/core'
 import { NotificationsService } from 'angular2-notifications'
-import { forkJoin, Subscription } from 'rxjs'
+import { forkJoin, Subscription, fromEvent } from 'rxjs'
 // FIXME: something weird with our path definition in tsconfig and typings
 // @ts-ignore
 import videojs from 'video.js'
@@ -61,6 +61,7 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
   hasAlreadyAcceptedPrivacyConcern = false
   remoteServerDown = false
   hotkeys: Hotkey[]
+  playerPopedOut = false
 
   private videojsLocaleLoaded = false
   private paramsSub: Subscription
@@ -140,6 +141,10 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
       }, undefined, this.i18n('Subscribe to the account'))
     ]
     if (this.isUserLoggedIn()) this.hotkeysService.add(this.hotkeys)
+
+    fromEvent(window, 'scroll')
+      .pipe(debounceTime(250))
+      .subscribe(() => this.onScroll())
   }
 
   ngOnDestroy () {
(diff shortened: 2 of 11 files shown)


rigelk:
If I recall well, adding `vjs-is-mobile` to the player element hides the stats plugin :thinking:

BO41:
@rigelk did not work the way I tried. But now it is fine too :wink: 

with controls:
screenshot_2018-10-24 video_2018-10-13_17-45-34 1
-----
after controls hide:
screenshot_2018-10-24 video_2018-10-13_17-45-34 2

rigelk:
@BO41 Nice! Have you tried it on multiple resolutions? How does it behave on mobile?

BO41:
@rigelk works on every resolution. It may be a little bit big on mobile, but I am not able to test it on my phone, and the browser emulator isn't the same.

I would like to add a transition between the two states, but I don't know how. `transition` does not seem to work.

ezgif com-video-to-gif

rigelk:
@BO41 that's already a nice POC! But indeed a transition would really ease the sight of it :thinking:

BO41:
seems like the `position` change isn't animatable.
this is the best I can make
ezgif com-video-to-gif 1

BO41:
@rigelk yeah, sure. Is there a real settings page with such settings? Than I will add it :wink: 
What do you think about the rest of the code?
Should I keep the `playerPopedOut` variable and is a `debounceTime` needed, or rather none?

BO41:
`my-account-video-settings.component` is probably the right place for such an option, right?
Will add an option there

rigelk:
@BO41 you need to add the column to the database, via a database migration

rigelk:
@BO41 there are quite a few examples in https://github.com/Chocobozzz/PeerTube/tree/develop/server/initializers/migrations

BO41:
@rigelk thanks! Worked fine.
I think I am done now, **please review** as I might have missed a place to add my property

client/src/app/videos/+video-watch/video-watch.component.ts
```diff
@@ -140,6 +141,11 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
       }, undefined, this.i18n('Subscribe to the account'))
     ]
     if (this.isUserLoggedIn()) this.hotkeysService.add(this.hotkeys)
+
+    if (this.user.pipPlayer) {
```
rigelk: `this.user` is not defined at that point.

BO41: I can test my code again, but before pushing the last commit, my code worked fine and the setting was respected :confused:

rigelk: You mean commit https://github.com/Chocobozzz/PeerTube/pull/1345/commits/e58cd73f914c8da31d5597942510d9d40113083d is at fault?

rigelk: What you said was ambiguous. I thought "before pushing the last commit" = "the commit before the last one".

Ping me whenever :slightly_smiling_face:

BO41: the code works fine, tested it again.
The setting is respected and no error is logged to the console.

rigelk: Ok, I'll try again later tonight then.

What is your browser/node version?

BO41:
@rigelk feel free to pull my changes/branch and test it yourself. Better safe than sorry

rigelk:
@BO41 that's already what I had done - will do it more thoroughly this time

client/src/app/videos/+video-watch/video-watch.component.scss
```diff
@@ -64,6 +64,34 @@ $other-videos-width: 260px;
   }
 }
 
+@keyframes fade {
```
Chocobozzz: Maybe you can put this in the _mixins file

client/src/app/videos/+video-watch/video-watch.component.scss
```diff
@@ -64,6 +64,34 @@ $other-videos-width: 260px;
...
+  box-shadow: 0 5px 10px 0 rgba(0,0,0,.3),0 1px 8px -3px rgba(0,0,0,.26),0 2px 16px 0 rgba(0,0,0,.28);
+
+  /deep/ .video-js {
+    width: 280px !important;
+    height: 160px !important;
+  }
+
+  /deep/ .vjs-peertube-displayed,
+  /deep/ .vjs-theater-control {
+    display: none !important;
```
Chocobozzz: theater icon should already be hidden https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/sass/player/peertube-skin.scss#L471

client/src/app/videos/+video-watch/video-watch.component.ts
Chocobozzz: Should be placed in the `if (this.isUserLoggedIn())` block, because you might be unlogged (and thus `this.user` is undefined)

client/src/app/videos/+video-watch/video-watch.component.ts
```diff
@@ -140,6 +141,11 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
...
+      fromEvent(window, 'scroll')
```
Chocobozzz: add debounce or throttle (you have an example in the infinite scroll directive)

client/src/app/videos/+video-watch/video-watch.component.ts
```diff
@@ -534,4 +540,17 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
       this.player = undefined
     }
   }
+
+  private onScroll () {
+    const playerElementWrapper = this.elementRef.nativeElement.querySelector('#video-element-wrapper')
+    const position = playerElementWrapper.getBoundingClientRect()
+
+    if (position.bottom <= 150 && !this.playerPopedOut) {
```
Chocobozzz: Could you add a comment to explain the reasoning?

client/src/app/videos/+video-watch/video-watch.component.ts
BO41: do you think a debounce would feel nice?
I think this might feel strange if it takes a time until the change appears, therefor I removed the debounce time.

client/src/app/videos/+video-watch/video-watch.component.ts
BO41: at the moment, the popup player is enabled by default. Should I enable it for not logged in users?

client/src/app/videos/+video-watch/video-watch.component.scss
BO41: it is not.
Why should it be? the option to go into theater mode is available, but in the popup player, it doesn't make sense and only takes space

client/src/app/videos/+video-watch/video-watch.component.scss
BO41: >/PeerTube/node_modules/sass-lint/lib/rules/property-sort-order.js:108
              name = prop.first('ident');                      
TypeError: Cannot read property 'first' of null

rigelk:
@BO41 Testing the endpoint in

client/src/app/videos/+video-watch/video-watch.component.scss
Chocobozzz: Rofl'

BO41: > Rofl'

what? I can't commit with those errors, although the css works and looks right. Or is there a way to force commit?

rigelk: You can skip the pre-commit lint-stage with `--no-verify`.

BO41: so the errors which appear are save to be ignored?

rigelk: It looks more like an error of `sass-lint`, but I don't know much about SASS. If you are sure it is valid, you can bypass the linting.

BO41:
2018-11-14-141433_843x676_scrot
shouldn't those files be ignored by `.gitignore` as they appear after running `$ npm run build`

client/src/app/videos/+video-watch/video-watch.component.scss
Chocobozzz: I often have issues with this pre hook and sass-lint, maybe we should remove the hook at least

Chocobozzz:
> shouldn't those files be ignored by .gitignore as they appear after running $ npm run build

These files are created during the build process, and then removed

client/src/app/videos/+video-watch/video-watch.component.scss
BO41: @rigelk I have absolutely no idea about sass/scss, but others say this code is fine and it works, so I assume it must be right.

@Chocobozzz does it have any pros? Generally I am a fan of many tests and lints, as it prevents errors and a bad coding style. But if it only produces errors...

BO41:
@Chocobozzz 
>and then removed

they are probably not removed at my system as the command fails because of those errors https://github.com/Chocobozzz/PeerTube/pull/1354#issuecomment-437586130

BO41:
```
(node:15183) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
npm ERR! code ELIFECYCLE
```
I think it doesn't want me to write tests.

I'll push my last changes. Please take the code and write the tests yourself, I am not good at this anyway :)

BO41:
I still have no idea why it ended with an error. And searching for an error with something I don't understand isn't fun. Especially not, if I didn't even make a change to the existing code

Chocobozzz:
Closing for https://github.com/Chocobozzz/PeerTube/pull/1463

PR #1351, opened 2018-10-26
update to angular 7

### LINKED ISSUE #1305: update to angular 7

BO41:
I am not sure why `import 'core-js/es7/reflect'` wasn't removed from `client/src/polyfills.ts`.
Should I remove it manually, since the [update notes](https://blog.angular.io/version-7-of-angular-cli-prompts-virtual-scroll-drag-and-drop-and-more-c594e22e7b8c) suggest that this should happen?
>the update to v7 will automatically remove this from your polyfills.ts

closes #1305


BO41:
perhaps it would be good to way a bit longer. But I am not sure where the problem is, so maybe there won't be a "fix".
Can someone try this on their machine and if it works better open a new PR?

PR #1354, opened 2018-10-28
update angular to v7

### LINKED ISSUE #1305: update to angular 7

BO41:
this time it looks right.
Everything seems to work, still

closes #1305


BO41:
I am not sure, but it looks to me, that the travis fail has nothing to do with my changes.
Or am I wrong?

>src/app/+admin/users/user-list/user-list.component.html(73,81): : Property 'onUserChanged' does not exist on type 'UserListComponent'.
for example

rigelk:
@BO41 it might be due to your removal of `import 'core-js/es7/reflect'`, since it's the part responsible for the decorator, which is what's giving the error.

BO41:
but since the official angular update did remove this import and they say nothing should break, does this still make sense?
I assume that the error is within a dependency which isn't ready for v7 yet, but that doesn't make too much sense either 
>ERROR in node_modules/@ng-bootstrap

Do you know anything to fix this issue, or should I try to update the problematic dependency?

BO41:
https://www.npmjs.com/package/@ng-bootstrap/ng-bootstrap:

ng-bootstrap | Angular | Bootstrap CSS
-- | -- | --
1.x.x | 5.0.2 | 4.0.0
2.x.x | 6.0.0 | 4.0.0
3.x.x | 6.1.0 | 4.0.0
4.x.x | 7.0.0 | 4.0.0

according to the release of ng-bootstrap version 4 is angular 7 compatible https://github.com/ng-bootstrap/ng-bootstrap/issues/2764#issuecomment-433338370

so I would suggest updating. But there might be other dependencies that need to be updated as well.
Still go for an update to angular 7 (and with this ng-bootstrap 4)?

rigelk:
If you manage to update ng-bootstrap, LGTM. You can still continue the full update to v7 but you might bump into a library that is still incompatible.

BO41:
updated ng-bootstrap, I hope the `TEST_SUITE=misc` now passes

BO41:
@rigelk do you have a clue why misc still fails?

BO41:
could somebody please take a look at the error log of travis?
All those errors seem **critical**, as if someone deleted methods. But not references.

I assume angular 7 is just looking closer at the code and that the mistakes were there for a long time

@rigelk @Chocobozzz

Chocobozzz:
> All those errors seem important, as methods get called, which do not exist.

I'll fix this and merge manually at the end of the week :+1:

PR #1363, opened 2018-10-30
Tiniest proof reading

jagannathBhat:


rigelk:
That's a fitting commit message :sweat_smile: 

Thanks anyway @jagannathBhat

PR #1368, opened 2018-10-31
Add Introduction to README (short explanation of what is PeerTube)

thomaskuntzz:
Provide small explanation of what is PeerTube and point to resources to learn more (combined with 'Want to see in action?' links)

I had a hard time coming up with a short and still decent explanation of what PeerTube is and how it works, feel free to improve it as you see fit 😉😊

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,23 @@ directly in the web browser with <a href="https://github.com/feross/webtorrent">
   </a>
 </p>
 
+Introduction
+----------------------------------------------------------------
+
+PeerTube is a free, decentralized and federated video platform developed as an alternative to other platforms that centralize our data and attention, such as YouTube, Dailymotion or Vimeo. :clapper:
+
+PeerTube is a software that enables the creation of a network of multiple small interconnected video hosting providers, called instances. :movie_camera: :globe_with_meridians:
+
+To learn more, see:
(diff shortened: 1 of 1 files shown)


rigelk:
The introduction is halfway between the short text below the title and the motivation paragraph. We should be moving the motivation paragraph to the FAQ, referring to it in your introduction, and trying to not say twice what is said in the short text below the title.

thomaskuntzz:
Oh, yeah, that's right, we could sort of combine these two. However I see that there are two FAQs, one in FAQ.md on GitHub and the other on joinpeertube.org. I feel it could be redundant, even though the one one GitHub seems more about technical topics than the other. What do you think?

rigelk:
Indeed, but that redundancy of the FAQs is acceptable IMHO, since the one in GitHub is more technical, and the one on joinpeertube.org can be translated.

thomaskuntzz:
Okay but then explaining the motivation behind the project would be more relevant on joinpeertube.org, wouldn't it?

rigelk:
The motivation is relevant for both FAQs.

rigelk:
In the end I'll reorganise the "Motivation" based on your changes. Thanks @thomaskuntzz !

PR #1370, opened 2018-11-03
autoplay videos

### LINKED ISSUE #1355: Autoplay next video
A nice feature would be an autoplay button, that automatically play the next video listed in "other videos" (as on youtube).

BO41:
I need to get the recommended videos, but at the moment not sure how

closes #1355


BO41:
>I need to get the recommended videos, but at the moment not sure how

help appreciated :)

BO41:
`this.nextVideoUrl` has the correct value in `onRecommendations`. But ` console.log(this.nextVideoUrl)` in `autoplayNext` prints `undefined`.
I don't understand why

rigelk:
@BO41 `onRecommendation` is never called.

BO41:
it is called at https://github.com/Chocobozzz/PeerTube/pull/1370/files#diff-98e5043904e920080dfb51d322af61c7R202. Once the `@output` is emitted.
I tested it with a console.log and it prints the correct value.

BO41:
> @BO41 it's probably because of the context change (self =/= this).

what do you mean with that? from line 329 to line 547 the context should be the same. and `this.nextVideoUrl` shouldn't be undefined as I initialize it with `''`.

`onRecommendations` is only called once, so the variable should only be changed once

BO41:
why do I still get `stack": "server_error: column User.autoPlayNextVideo does not exist\`?
didn't I make a correct db migration?

rigelk:
@BO41 maybe your database is in an incorrect state - try with a new one.

client/src/app/videos/+video-watch/video-watch.component.ts
```diff
@@ -323,6 +325,10 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
     return this.video && this.video.scheduledUpdate !== undefined
   }
 
+  onRecommendations (videos: Video[]) {
+    this.nextVideoUrl = videos[0].uuid
```
Chocobozzz: check the length of `videos` before assigning the first element + I fear that we'll have an infinite autoplay loop 

If think we should keep in memory the previous videos that we played to not play them again

client/src/app/videos/+video-watch/video-watch.component.ts
```diff
@@ -438,6 +444,9 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
       videojs(this.playerElement, videojsOptions, function (this: videojs.Player) {
         self.player = this
         this.on('customError', ({ err }: { err: any }) => self.handleError(err))
+        if (self.user.autoPlayNextVideo) {
```
Chocobozzz: Check the user if logged in first

client/src/app/videos/+video-watch/video-watch.component.ts
BO41: >an infinite autoplay loop

if the recommendation service returns always the same video, it will be an endless loop.
But how could I prevent this? Just remembering the last could resolve in a 3 way loop.

And shouldn't this rather be solved in the recommendation service?

BO41:
>Could you add update and check params test too please?

could you explain this to me, as I don't understand it in this wording @Chocobozzz

client/src/app/videos/+video-watch/video-watch.component.ts
Chocobozzz: > if the recommendation service returns always the same video

Not only. If the service returns recently added videos, and you are watching the latest upload video you'll have (in the `[0]` index) the second last one uploaded video. Then you autowatch this video, and the service will return the latest uploaded video that you'll re-watch etc etc

> And shouldn't this rather be solved in the recommendation service?

It would be a solution :+1:

Chocobozzz:
> could you explain this to me, as I don't understand it in this wording @Chocobozzz

Add tests for this attributes (in this example we check the nsfw param, that is a boolean too):
 * To check your validator middleware works: https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/check-params/users.ts#L291
 * To check your controller works: https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/users/users.ts#L371

rigelk:
Closing due to inactivity. Feel free to reopen.

LoveIsGrief:
I'm interested in picking this up. How could I do so since the branch has been deleted :cry:

Findus23:
@LoveIsGrief The code is still here. You can e.g. download [`https://github.com/Chocobozzz/PeerTube/pull/1370`](https://github.com/Chocobozzz/PeerTube/pull/1370) and apply it with `git apply thepatchfile`.

PR #1386, opened 2018-11-09
Rename context stats to state

tcitworld:
I guess it refers to the VideoState enum used here as `state` instead.


tcitworld:
Spotted through an AP call to a video, we have 
```json
{
  "@context": {
    "stats": "sc:Number"
  },
  "state": 1,
}
```
`state` isn't defined anywhere and `stats` doesn't seem to be used anyway.

Chocobozzz:
Thanks @tcitworld

Issue #1080, opened 2018-09-17
SEO improvements


Zig-03:
_SEO - Search Engine Optimization_

The thing is that almost none of the instances/video titles/content is discoverable by SEs (Search Engines). 
For example, there is an instance called **peertube.social**, and let's say we need to find a channel called "chriswere" so if we search for:
["chriswere peertube social"](https://duckduckgo.com/?q=chriswere+peertube+social) - we get no results. 

Let's be more exact and add 2 more keywords from [this](https://peertube.social/videos/watch/13ce5139-06f5-4c4b-9289-426d588d2836) video of ()chriswere:
["chriswere peertube social Chromium BSU"](https://duckduckgo.com/?q=chriswere+peertube+social+Chromium+BSU) - zero results.

To fix this (at least for now), this article should do: https://backlinko.com/on-page-seo (this is a general purpose article, not everything applies to peertube, but most of it I'd say)

rigelk:
We're essentially missing custom video urls, if I get the gist of this page.

Zig-03:
@rigelk a ctr+u on a video page reveals that there are no header like H1, H2, H3. So there could be more missing steps.

Also this file is weird to me https://peertube.social/robots.txt
I don't know how crawlers interpret this:

> User-agent: *
> **Disallow: ' '**

After Disallow it's usually a link or a path or nothing https://moz.com/learn/seo/robotstxt 

So perhaps bots aren't even allowed to enter and crawl these website?

rigelk:
I have changed the html tag of the video title to `<h1>` on the video view, and attributes of miniatures now include the title in the `alt` attribute.

Chocobozzz:
I'm reopening, because we still have SEO issues. For example, we need to index only local videos/accounts/video channels. 

Moreover, google does not index our videos because "Google chose different canonical than user".

BO41:
non video pages are missing opengraph/twitter tags and a canonical URL.
and not found urls are **not** returning a `404` status code

BO41:
a nice thing to add would be another opengraph tag
>`video:tag` - string array - Tag words associated with this movie.

Zig-03:
We need H tags as well. <H1> for the each video title and <H2> for each video description. 

BTW right now comments aren't displayed when viewing the page source, was this intended to be like this?

fflorent:
[...]

PR #1396, opened 2018-11-13
change og video type

BO41:
`og:video:type` should be ["MIME type of the video. Either `application/x-shockwave-flash` or `video/mp4`."](https://developers.facebook.com/docs/sharing/webmasters#markup)

But it is
```html
<meta property="og:video:type" content="text/html" />
```

this may help #1080

### DIFF
--- a/server/lib/client-html.ts
+++ b/server/lib/client-html.ts
@@ -116,7 +116,7 @@ export class ClientHtml {
 
       'og:video:url': embedUrl,
       'og:video:secure_url': embedUrl,
-      'og:video:type': 'text/html',
+      'og:video:type': 'video/mp4',
       'og:video:width': EMBED_SIZE.width,
       'og:video:height': EMBED_SIZE.height,
 


tcitworld:
I'm not sure about this, because `og:video:url` points to the embed URL (which is `text/html`, not the video URL itself).

Did you test this somewhere to see what it does ? cc @Chocobozzz

Chocobozzz:
@tcitworld Yes on https://peertube2.cpy.re/videos/watch/217eefeb-883d-45be-b7fc-a788ad8507d3

But it does not really change anything on facebook side

BO41:
vimeo on the other hand uses `text/html`

tcitworld:
In the case of YouTube, the `og:video:url` directly links to the video (with the right `application/x-shockwave-flash` Content-Type header).

My concern is with the HTTP Content-Type header of whatever's at `og:video:url` not being the same as what's inside `og:video:type`.

BO41:
I am just trying to improve things here, not trying to break anything or force my option on this.

So thank you @tcitworld for starting this discussion; better safe than sorry.
I suggest either you or @Chocobozzz search for this (hopefully finding a real doc) and either leave it or revert it

Chocobozzz:
Okay thanks @tcitworld, I reverted this PR in https://github.com/Chocobozzz/PeerTube/commit/fb651cf2d426e7429a7c571ab7d93becd63ad116

PR #1400, opened 2018-11-14
add sudo to commands

BO41:
```
createdb -O peertube peertube_test1   
createdb: database creation failed: ERROR:  must be member of role "peertube"
```

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -140,7 +140,7 @@ Then, we can create the databases (if they don't already exist):
 
 ```
 $ sudo -u postgres createuser you_username --createdb
-$ createdb -O peertube peertube_test{1,2,3}
+$ sudo -u postgres createdb -O peertube peertube_test{1,2,3}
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Did you execute `sudo -u postgres createuser you_username --createdb --superuser`?

BO41:
probably without `--superuser` as I accidentally created this user already.
Would this fix the error? Then sorry for this PR, please close :)

Chocobozzz:
> Would this fix the error?

yes

.github/CONTRIBUTING.md
```diff
@@ -140,7 +140,7 @@ Then, we can create the databases (if they don't already exist):
 
 ```
 $ sudo -u postgres createuser you_username --createdb
```
BO41: @Chocobozzz at first I ran this command, thats the reason the user already existed. And here is no `--superuser`.
So theoretically, here `sudo` would be needed, right?

PR #1415, opened 2018-11-20
Rewrite Docker image with Linux Alpine for smaller size

Nutomic:
original image: 1.35GB
new image: 409MB

This will make it much easier to install the image (smaller download), and to store images (less disk space used). Most of the savings are from switching to the Alpine image as a base, and generating less intermediate images. It might be possible to save more space by deleting build dependencies or source files, but I'm not sure what else can be removed.

 I haven't tested it yet, but I will probably set up a test server in the next days.

### DIFF
--- a/support/docker/production/Dockerfile.stretch
+++ b/support/docker/production/Dockerfile.stretch
@@ -1,66 +1,50 @@
-FROM node:8-stretch
-
-RUN set -ex; \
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Hello,

I'm sorry but I don't want to maintain an Alpine image. We had some weird issues with it regarding our native dependencies. But feel free to maintain it and add a link in our doc!

PR #1420, opened 2018-11-21
Improve FreeBSD setup instructions

0mp:
Use install(1) instead of cp(1) and chmod(1) to install
the service file. Also, use sysrc(8) to enable the PeerTube
service in rc.conf(5).


rigelk:
Thanks @0mp !

PR #1422, opened 2018-11-21
Various improvements for docker-compose

Nutomic:
- automatically redirect http requests to https
- include postfix so that email sending works without any extra configuration
- don't expose the api publicly because this can result in security issues ([documentation](https://docs.traefik.io/configuration/api/#security), [previous security issue](https://twitter.com/EdOverflow/status/1031267058615508994))
- remove unneeded `docker-compose down` in upgrade instructions ([see my comment here](https://github.com/Chocobozzz/PeerTube/commit/9007daff8262bcb539eb5fa74919828f425b5f69#commitcomment-31373736))

I used this configuration on a test server and everything seems to work fine.


Chocobozzz:
Thanks @Nutomic

PR #1425, opened 2018-11-22
Check free storage before upgrading

Nutomic:
I just had my upgrade fail because most of my server's disk space was taken up by the video cache. So I added this quick check to the upgrade script.

The minimum free space of 1 GB is somewhat arbitary, because each Peertube installation only takes around 270 MB on my server. But it's better to have too much than too little.

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -20,6 +20,14 @@ if [ ! -e "$PEERTUBE_PATH/versions" -o ! -e "$PEERTUBE_PATH/config/production.ya
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


scripts/upgrade.sh
```diff
@@ -20,6 +20,14 @@ if [ ! -e "$PEERTUBE_PATH/versions" -o ! -e "$PEERTUBE_PATH/config/production.ya
   exit 1
 fi
 
+REMAINING=$(df -k $PEERTUBE_PATH | awk '{ print $4}' | sed -n 2p)
```
Chocobozzz: Check `awk` and `sed` are installed first

Chocobozzz:
Thanks @Nutomic

Issue #1341, opened 2018-10-22
Indicate to users how "trending" works


Nutomic:
The trending list sorts videos by number of views in the last 7 days (default value). But it is not indicated to users that this is the case, the ordering appears quite random. This can be especially confusing when a one year old video is listed next to one from a few days ago.

In order to clarify this, why don't we add a line like "123 views in 7 days" to each video on the trending page? That would make it immediately obvious to users why each video is listed there.

This would also be a great way to differentiate Peertube from Youtube, where the sort order is completely intransparent. And it would bring us closer to the philosophy of Mastodon, namely to avoid "magic" algorithms influence the content users see.

Related to #1207.

BO41:
good thing you point this out.
But I guess it would be even clearer to add a simple text box, or popup or whatever and write 2-3 sentences. Then it is clear for sure

auberanger:
I'm currently working on this issue and on the #1409 (which are related).
I'm actually nor satisfied by the idea of a popup or by the use of the help component here. At least, it would be better to make it appear on hover of the Trending title. Otherwise, a simple text box appearing from "Trending" when hovered or clicked could be a good solution. But in this case, is there a preexisting component I could use ?

Issue #1207, opened 2018-10-06
How does the "Trending" algorithm work?


Nutomic:
Looking at the [trending page of my instance](https://peertube.social/videos/trending), it is not clear to me how this is generated. For example, the trending videos interval in the config is set to 7, but I see videos that are more than half a year old. I also couldn't find any info in the Github repo how this works exactly. It would be good if this could be added to the user documentation or somewhere else.

Chocobozzz:
Ping @rigelk 

The trending algorithm takes into account views from the last *x* days (specified in the admin configuration), regardless of the videos published date.

McFlat:
Would it be possible to expose some kind of settings for the trending algorithm in the admin?

Chocobozzz:
@McFlat What kind of settings?

McFlat:
@Chocobozzz To be able to:
1. specify a list of video to exclude from trending
2. update trending interval days in the frontend "Administration" for admins
3. have a list of videos to include in trending or possibly a new section for "Featured Videos" would be nicer, on the trending page. So that the page can have some featured videos at the top, which would be selected out of a list of videos by the admin

buoyantair:
I have an idea of sorts, can users actually implement their own trending algorithms? 
We can actually extract the component that implements the current trending algorithm into it's own package of sorts that can be plugged into Peertube. This means that later on anybody can change their trending algorithms by creating a new plugin and then connecting it to Peertube

Chocobozzz:
Closing this issue, the algorithm is now explained in a tooltip.

PR #1458, opened 2018-12-05
WIP : Indicate to users how "trending" works

auberanger:
This PR is related to the following issue #1341 
I've just give a try to a simple help icon with a tooltip in order to explain how trending works. It looks like this : 
screenshot_2018-12-05 trending videos - peertube
screenshot_2018-12-05 trending videos - peertube-clicked

In my view, this solution is not really satisfying. I was thinking about something subtler, as I'm going to explain in the issue comments.

### DIFF
--- a/client/src/app/shared/video/abstract-video-list.html
+++ b/client/src/app/shared/video/abstract-video-list.html
@@ -1,8 +1,11 @@
 <div [ngClass]="{ 'margin-content': marginContent }">
   <div class="videos-header">
     <div *ngIf="titlePage" class="title-page title-page-single">
-      {{ titlePage }}
+      <div placement="bottom" [ngbTooltip]="titleTooltip" container="body">
+        {{ titlePage }}
+      </div>
     </div>
+
     <my-feed [syndicationItems]="syndicationItems"></my-feed>
 
     <div class="moderation-block" *ngIf="displayModerationBlock">
--- a/client/src/app/core/server/server.service.ts
+++ b/client/src/app/core/server/server.service.ts
@@ -83,6 +83,11 @@ export class ServerService {
           enabled: false
         }
       }
+    },
+    trending: {
+      videos: {
+        intervalDays: 0
+      }
     }
   }
   private videoCategories: Array<VideoConstant<number>> = []
--- a/client/src/app/shared/video/abstract-video-list.scss
+++ b/client/src/app/shared/video/abstract-video-list.scss
@@ -19,8 +19,8 @@
 
   my-feed {
     display: inline-block;
-    position: relative;
     top: 1px;
+    min-width: 60px;
   }
 
   .moderation-block {
--- a/client/src/app/shared/video/abstract-video-list.ts
+++ b/client/src/app/shared/video/abstract-video-list.ts
@@ -39,6 +39,7 @@ export abstract class AbstractVideoList implements OnInit, OnDestroy {
   ownerDisplayType: OwnerDisplayType = 'account'
   firstLoadedPage: number
   displayModerationBlock = false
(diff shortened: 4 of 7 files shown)


Chocobozzz:
I would prefer to use a tooltip as you said in https://github.com/Chocobozzz/PeerTube/issues/1341#issuecomment-444502177

https://ng-bootstrap.github.io/#/components/tooltip/examples

You could change the title to "Trending videos in the last x days", too

auberanger:
Following your advice, I've changed the help component by a ng-tooltip and the title of the trending page. However, I don't manage to find where exactly is defined the Trending algorithm. It seems to be related to the videoService.getVideos() method and to the RestService class but I don't understand precisely how. I've also looked into the admin configuration as suggested in this [comment](https://github.com/Chocobozzz/PeerTube/issues/1207#issuecomment-427561081), without success.

Also, I don't manage to find any mention of a parameter for the last "X" days.

Would you have a tip for helping me about this ?

Chocobozzz:
Sorry indeed I have not been very clear

> Also, I don't manage to find any mention of a parameter for the last "X" days.

You'll have to add it to [the config endpoint](https://github.com/Chocobozzz/PeerTube/blob/develop/server/controllers/api/config.ts#L54) from [the constants](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/constants.ts#L220)

The trending algorithm is very simple: it sum the views made in the last x days, and sort them. (see the [query builder](https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/video/video.ts#L1345) and [the sort](https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/utils.ts#L22)

auberanger:
Hello @Chocobozzz, you can find a first draft of the resolution of the issue in the previous commits.

client/src/app/shared/video/abstract-video-list.html
```diff
@@ -1,8 +1,18 @@
 <div [ngClass]="{ 'margin-content': marginContent }">
   <div class="videos-header">
     <div *ngIf="titlePage" class="title-page title-page-single">
-      {{ titlePage }}
+      <ng-template i18n #tipContent>
+        The <b>Trending algorithm</b> sums the views made in the last {{ trendingDays }} days for the videos of the related instances, and sort them.
```
Chocobozzz: I think we should remove the `<b>` tag, so we can put this text in a `titleTooltip` variable in `abstract-video-list.ts` (the children choose to fill it or not). This way, you would not have this ngIf in the template:

```
<div placement="bottom" [ngbTooltip]="titleTooltip" container="body">
  {{ titlePage }}
</div>
```

In the trending videos component, `titlePage` should be `titlePage = this.i18n('{{titlePage}} for the last {{trendingDays}} days', ...)`

auberanger: Thanks for this much more elegant solution !

Aluriak:
IMHO,

    The Trending algorithm sums the views made in the last {{ trendingDays }} days for the videos of the related instances, and sort them

is quite technical and would be simpler for most people if worded as

    trending videos are those totalizing the greatest number of views during the last {{ trendingDays }} days

And also, maybe use *during the last day* instead for when `trendingDays` is 1.

auberanger:
I agree with you @Aluriak. Howewer, would it not be even clearer to use _during the last 24 hours_ instead of _during the last day_ instead for when `trendingDays` is 1 ?

Aluriak:
> would it not be even clearer to use *during the last 24 hours instead of during the last day*

yes, it would.

Chocobozzz:
Thanks @auberanger

PR #1464, opened 2018-12-07
add change-password well-known url

rigelk:
spying on @tcitworld's stars, I found yet another well-known url to implement :sweat_smile:

### DIFF
--- a/server/controllers/static.ts
+++ b/server/controllers/static.ts
@@ -136,6 +136,12 @@ staticRouter.use('/.well-known/dnt/',
   }
 )
 
+staticRouter.use('/.well-known/change-password',
+  (_, res: express.Response) => {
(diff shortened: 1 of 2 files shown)


tcitworld:
And there's some discussions going on here https://news.ycombinator.com/item?id=18618193

server/controllers/static.ts
```diff
@@ -136,6 +136,13 @@ staticRouter.use('/.well-known/dnt/',
   }
 )
 
+staticRouter.use('/.well-known/change-password',
+  (_, res: express.Response) => {
+    res.location('/my-account/settings')
+    res.redirect('/my-account/settings')
```
Chocobozzz: https://stackoverflow.com/a/22678011

Issue #993, opened 2018-08-31
Warn in case of error in email middleware


joshmorel:
As discussed here #977 - the user should be notified if an email isn't sent because of some system error Currently for asking password reset or sending email verification link - success is indicated even if it failed on back-end (e.g. SMTP not configured properly).

PR #1470, opened 2018-12-08
rename ask-email link to ask-send-email

joshmorel:
I encountered this bug when working on #993 . The link is likely rarely needed (only if verification string is wrong) which is why it has not been encountered.


rigelk:
Thanks @joshmorel !

PR #1471, opened 2018-12-08
WIP: Warn if email is not enabled for email verification

joshmorel:
Work in progress related to #993 

I originally opened the issue when implementing the email verification feature. This is likely an edge case as it would require the site admin to explicitly enable "Signup requires email verification" without smtp properly setup. 

In the work in progress I implemented a similar error on the client side to that used for ask password reset.

request-new-email

I'm looking for feedback on which other related items to implement (or not): 

1. Client-side error if admin enables "Signup requires email verification" config when email is not enabled
2. Client-side error if user tries to signup when email verification is required and email is not enabled
3. Server-side error (500 response?) via middleware for requests dependent on email (ask password reset, ask verification email)

### DIFF
--- a/client/src/app/+verify-account/verify-account-ask-send-email/verify-account-ask-send-email.component.html
+++ b/client/src/app/+verify-account/verify-account-ask-send-email/verify-account-ask-send-email.component.html
@@ -2,21 +2,27 @@
   <div i18n class="title-page title-page-single">
     Request email for account verification
   </div>
-
-  <form *ngIf="requiresEmailVerification; else emailVerificationNotRequired" role="form" (ngSubmit)="askSendVerifyEmail()" [formGroup]="form">
-    <div class="form-group">
-      <label i18n for="verify-email-email">Email</label>
-      <input
-        type="email" id="verify-email-email" i18n-placeholder placeholder="Email address" required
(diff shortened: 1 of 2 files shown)


Chocobozzz:
I think we should prevent peertube from starting if it is incorrectly configured. For example, in the `checkConfig` function: https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/checker-after-init.ts#L34

Nutomic:
But you can also change these settings in the web interface so there needs to be a warning/error as well.

joshmorel:
Agreed. I will implement both checks. 

If both are implemented is it still worth keeping the user warning above? (this should be a real slim edge case if the two other checks are implemented).

Chocobozzz:
> If both are implemented is it still worth keeping the user warning above? (this should be a real slim edge case if the two other checks are implemented).

If both are implemented yep remove the warning above :+1:

Chocobozzz:
Any news on this PR @joshmorel?

joshmorel:
@Chocobozzz  - put this on the backburner as I was working on another issue (almost done!).

I think I can finish by this weekend so we can close this soon.

joshmorel:
Started again with #1642

PR #1474, opened 2018-12-09
Update FreeBSD startup script

gegeweb:
Adding procname='node:' to find pid for stopping the service launched with node.
Previous version doesn't work without this since the upgrade to FreeBSD 11.2.


rigelk:
Thanks @gegeweb!
Could you explain your changes in the PR or commit description?

gegeweb:
I'v just add to line, since the upgrade to FreeBSD 11.2 stopping the service with `service stop pertube`doesn't works. Processus pid not find.
So I've add 'procname='node:' to deal with.

It works now on my system, I can do restart or stop succefully.

Chocobozzz:
Thanks @gegeweb

gegeweb:
Well... it doesn't work as expected.
If there are other process nodes, stopping the service will kill all process nodes, even if it is not Peertube.
I can't figure out how to get the pid number... I'll figure out how to do it!

rigelk:
@gegeweb then please open an issue tracking the problem so that we don't forget it exists.

gegeweb:
@rigelk  Done! #1477

Issue #1477, opened 2018-12-11
FreeBSD startup script stopping all node's process


gegeweb:
My merge request #1474 doesn't works as expected.
That stop all the node's process, not only Peertube.

If someone have an idea how to deal with pid number, they are most welcome! ;)

PR #1479, opened 2018-12-12
Use daemon(8) in the FreeBSD startup service

0mp:
This commit aims to solve the problem existing in #1477. Thanks to using
the daemon(8) utility, the service is able to always stop the correct
process as it is going to check both PID and the process name before
killing the PeerTube service.

----

I am not running PeerTube so I couldn't test this patch. I just saw [a call for help on Mastodon](https://stoneartprod.xyz/@gegeweb/101226782080800619) and decided to contribute. I'm happy to help you improve this solution if it happens to be broken.

### DIFF
--- a/support/freebsd/peertube
+++ b/support/freebsd/peertube
@@ -24,8 +24,9 @@ NODE_CONFIG_DIR=/var/www/peertube/config \
 USER=peertube"
 peertube_user=peertube
 procname="node:"
+pidfile="/var/run/peertube.pid"
 
-command="/usr/local/bin/npm"
-command_args="start >> /var/log/peertube/${name}.log 2>&1 &"
+command="/usr/sbin/daemon"
+command_args="-p ${pidfile} -o /var/log/peertube/${name}.log -t ${name} -u ${peertube_user} /usr/local/bin/npm start"
 
 run_rc_command "$1"


Chocobozzz:
Ping @gegeweb to test the patch

gegeweb:
I'm going to look at this solution, it doesn't work at this time.
Starting fail with this error `daemon: failed to set user environment`.

I'll check how to others have dealing with node and daemon…

Thanks!

gegeweb:
And there is another error, the peertube user can't write the pid in `/var/run`, it's nedeed to create a `peertube` folder owned by the peertube user to keep the file with the pid number.

gegeweb:
As I was saying on Mastodon, it is not necessary to use daemon.
We have to go back to the version before my merge request and just change the start command : `/usr/local/bin/node dist/server` instead of `npm start`.

I just tested it right now. 

I'll send a new merge request.

Thank you again!

0mp:
I am going to close this PR then. Feel free to reopen it if needed. :+1:

gegeweb:
Hi @0mp, I have a question about using daemon instead of launching the node process directly.
Do you think that will restart the server after a « crash » caused by an unmanaged exception on peertube side? (Like pm2 do)

If yes I'll probably look how to fix this error `daemon: failed to set user environment.` for use it for peertube, and the Mastodon stream too.

0mp:
daemon(8) has -r and -R options for restarting its child. In that case you need to store the PID of the daemon (with -P) instead of the child's one (with -p) as you need to kill the parent process (the daemon) in order to stop the service.

gegeweb:
OK, thanks.

PR #1488, opened 2018-12-15
More systemd service hardening

mkoppmann:
* I set the `PrivateDevices=` directive to true, because I doubt that Peertube is being used on Raspberry Pies currently.
* Because the installation guide recommends `/var/www/peertube` as the home directory for the Peertube user, I think it’s safe to remove the ability for Peertube to access `/home`, `/root`, and `/run/user`.
* The service still has sys admin capabilities, and in theory can circumvent the sandbox restrictions. In the [manual](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#ReadWritePaths=) it is recommend to drop this capability for an effective sandbox.

### DIFF
--- a/support/systemd/peertube.service
+++ b/support/systemd/peertube.service
@@ -22,12 +22,17 @@ PrivateTmp=true
 ; Mount /usr, /boot, and /etc as read-only for processes invoked by this service.
 ProtectSystem=full
 ; Sets up a new /dev mount for the process and only adds API pseudo devices
-; like /dev/null, /dev/zero or /dev/random but not physical devices. Disabled
-; by default because it may not work on devices like the Raspberry Pi.
-PrivateDevices=false
+; like /dev/null, /dev/zero or /dev/random but not physical devices. May not work
+; on devices like the Raspberry Pi.
+PrivateDevices=true
 ; Ensures that the service process and all its children can never gain new
 ; privileges through execve().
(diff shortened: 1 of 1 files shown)


Chocobozzz:
> I set the PrivateDevices= directive to true, because I doubt that Peertube is being used on Raspberry Pies currently.

I think some people do, or will.

mkoppmann:
@rigelk Great, thank you :). I would write something like this; you can adapt it to your liking:

A systemd unit template is provided at `support/systemd/peertube.service`.

### `PrivateDevices`

<div class="alert alert-warning" role="alert">
  <strong>Warning:</strong> this won't work on Raspberry Pi. That's
  why we don't enable it by default.
</div>

`PrivateDevices=true` sets up a new `/dev` mount for the Peertube process and only adds API pseudo devices like `/dev/null`, `/dev/zero`, or `/dev/random` but not physical devices.

### `ProtectHome`

`ProtectHome=true` sandboxes Peertube such that the service can not access the `/home`, `/root`, and `/run/user` folders. If your local Peertube user has its home folder in one of the restricted places, either change the home directory of the user or set this option to `false`.

rigelk:
Thanks @mkoppmann - your changes to the documentation are live too!

mkoppmann:
Thank you

PR #1490, opened 2018-12-15
Throw error if MailDev doesn't run

### LINKED ISSUE #1482: tests succeed even though maildev silently fails & quits
**What happened?** Test suite succeeds even though email-related tests don't run due to `MailDev` silent failure.

**What do you expect to happen instead?** Tests should run or at least fail if they can't run for some reason - e.g. SMTP already using port.

**Steps to reproduce:**

This actually demonstrated in this Travis Job. `"Test users account verification"` which uses `mockSmtpServer` doesn't actually run the suite of tests but the job succeeds overall. This is because  `new MailDev` silently fails because one was already running on that port from `"Test emails"`.

travis-job-succeeds
https://travis-ci.org/Chocobozzz/PeerTube/jobs/467150770#L969

Mea culpa, I introduced this when adding email account verification feature. Previously `mockSmtpServer` was only used in one test suite so this issue wouldn't be encountered. I didn't notice it because I was first running only `"Test users account verification"` with all tests passing. When I ran all the tests together locally, well, it succeeded so I figured all was good!

I can implement a fix and also add some kind of expect logic to fail the tests if the server can't run (e.g.  developer has mail server already running 1025, or something unforeseen).

joshmorel:
Resolves #1482 

I discovered that if MailDev instantiation fails the process is terminated with a 0 exit code. The only way to report this back as an error is to run as forked child process.

Besides the approach an alternate option is to use a different library - which I'm open to.

Now - it fails when expected (e.g. postfix listening on port 1025):

peertube-smtp-test-fails-as-expected

And also keeps running when expected when used in multiple test files:

peertube-smtp-test-continues-as-expected

I'm open to criticisms on approach with singleton class.

### DIFF
--- a/server/tests/api/server/email.ts
+++ b/server/tests/api/server/email.ts
@@ -20,7 +20,7 @@ import {
   ServerInfo,
   setAccessTokensToServers
 } from '../../../../shared/utils'
-import { mockSmtpServer } from '../../../../shared/utils/miscs/email'
+import { MockSmtpServer } from '../../../../shared/utils/miscs/email'
 import { waitJobs } from '../../../../shared/utils/server/jobs'
 
 const expect = chai.expect
@@ -41,7 +41,7 @@ describe('Test emails', function () {
   before(async function () {
     this.timeout(30000)
 
-    await mockSmtpServer(emails)
+    await MockSmtpServer.Instance.collectEmails(emails)
 
     await flushTests()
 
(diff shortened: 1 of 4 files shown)


Chocobozzz:
Thanks @joshmorel 

If you find another library, that is simpler to use don't hesitate to make a PR.

Issue #1359, opened 2018-10-29
Error 499 on large uploads (with nginx)


JohnXLivingston:
**What happened?**

When I upload a large file (4Go), the progress bar says "processing...", then I got a 499 http error code on server side (Client Closed Request).

**Additional information**
* PeerTube version or URL: https://www.yiny.org (version 1.0.1, on nginx)
* Browser name/version: Firefox 63

* browser console log:
Backend returned code 0, errorMessage is: Unknown error. main.1572864908cc797acead.js:1:423429

![image](https://user-images.githubusercontent.com/38844060/47677297-02fa6180-dbbf-11e8-9f6c-765feca453f4.png)

* nginx access logs:
[29/Oct/2018:20:00:05 +0000] "POST /api/v1/videos/upload HTTP/2.0" 499 0 "https://www.yiny.org/videos/upload" "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:63.0) Gecko/20100101 Firefox/63.0"

* peertube logs:
Oct 29 20:00:05 yiny peertube[15813]: [www.yiny.org:443] 2018-10-29 20:00:05.875 info: 82.124.0.0 - - [29/Oct/2018:20:00:05 +0000] "POST /api/v1/videos/upload HTTP/1.0" - - "https://www.yiny.org/videos/upload" "Firefox"

After that, the following file is in my storage/videos folder:
-rw-r--r--  1 peertube peertube 2.2G Oct 29 20:00 f8edb67880a043bcf6cf077418168572.mp4
(Note that it is only 2.2G, but the browser uploaded 3.71G)

The prune-storage script want to remove this file. 

There is no failed job.

The server is a little VM. I notice that during the upload, the processor is running hard.

JohnXLivingston:
If I use the "import from an url" feature, and paste a direct link to this video file, it works. It's only the upload process who fails.

poka-IT:
You have to change in your nginx configuration (/etc/nginx/sites-enabled/peertube) the client_max_body_size : 
```
  location / {
#    proxy_pass http://localhost:9000;
    proxy_pass http://tubep2p;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header Host $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;

    # For the video upload
    client_max_body_size 2G;
    proxy_connect_timeout       600;
    proxy_send_timeout          600;
    proxy_read_timeout          600;
[...]

PR #1494, opened 2018-12-16
 This fix #1460 and #1359 

### LINKED ISSUE #1460: Video upload fails with "Unknown Error"
When I try to upload a specific video file to my instance, it fails every time. I tried to force reload the page and even restart the server, but nothing helped. The browser console only shows:
```
Backend returned code 0, errorMessage is: Unknown error.
```
I didn't have time to check the server logs yet.

I could upload the video without problems after running `ffmpeg -i old.mp4 new.mp4`.

Video file: https://nextcloud.nutomic.com/index.php/s/qrCW4m6cREL7faB

* PeerTube version or URL: https://peertube.social
* Browser name/version: Firefox 63

gegeweb:
This PR fix the error on upload for large video (see #1460 and #1359).


rigelk:
> The [proxy_next_upstream](http://nginx.org/r/proxy_next_upstream) directive is a configuration directive to control re-request from a [group of upstream servers](http://nginx.org/r/upstream) by a `proxy_pass` if a request to one of them fails. It doesn't make sense without `proxy_pass` and an `upstream` block defined.

Since the template doesn't contain any `upstream` block, it looks like the fix you are suggesting is specific to your setup.

gegeweb:
OK, I close this PR.
I just had that error (Unknown error and client side, and the '499' in nginx log) again with this directive.

Do I reopen an issue?

Issue #1436, opened 2018-11-29
close Download Video popup after validation


roipoussiere:
When we click on *Download Video*, a popup appears with several options. After clicking *Download*, the download starts but the popup remains here. This is a little bit disturbing because we may think that the action failed. Maybe this button should close this popup.

juliencrepin62:
I will edit the file to close the pop-up when the the user clicks on the download button.

buoyantair:
@juliencrepin62 Hey, just wanted to checkup on this issue

rokamoto30:
Is this issue still active? I'm a new programmer looking into open source projects and I was wondering how I could start contributing to this project.

Chocobozzz:
@rokamoto30 Yes you can take it. Or any "good first issue": https://github.com/Chocobozzz/PeerTube/labels/good%20first%20issue%20%3Abeginner%3A

Chocobozzz:
Implemented by @LeoMouyna in https://github.com/Chocobozzz/PeerTube/commit/11b3f14c758e57f9cb5f32d8140fa043094bb3d7

PR #1496, opened 2018-12-16
Close pop-up after dowload (see #1436)

juliencrepin62:
Add in the download function "this.hide()".


client/src/app/videos/+video-watch/modal/video-download.component.ts
```diff
@@ -33,6 +33,7 @@ export class VideoDownloadComponent implements OnInit {
 
   download () {
     window.location.assign(this.getLink())
+    this.hide()
```
Chocobozzz: `this.hide` is not defined

PR #1498, opened 2018-12-18
Add WebFinger support

acid-chicken:
Some of federation projects don't support `/authorize_interaction?acct={uri}` endpoint (e.g. [Misskey](https://github.com/syuilo/misskey)). So I propose to support WebFinger on the workflow of the remote subscribe button.

### DIFF
--- a/client/src/app/shared/user-subscription/remote-subscribe.component.ts
+++ b/client/src/app/shared/user-subscription/remote-subscribe.component.ts
@@ -37,7 +37,21 @@ export class RemoteSubscribeComponent extends FormReactive implements OnInit {
 
   formValidated () {
     const address = this.form.value['text']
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks @acid-chicken

rigelk:
As a follow-up, I suggest thinking again about the text written. So far the UI showed "follow with a Mastodon account" as this was both the accurate depiction of the front capabilities *and* the best course for the λ user.

Should we switch to a more platform-neutral description of what the account handle should be?

c.f. https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/shared/user-subscription/subscribe-button.component.html#L43

acid-chicken:
@rigelk I think that it's a good suggestion.

How about "follow with an ActivityPub-compliant service account"?

PR #1535, opened 2019-01-06
increase waitJobs pendingJobs timeout to 2000

### LINKED ISSUE #1492: Test should have correct follows count 3 fails locally
**What happened?**

```
  1) Test follows
       Should propagate data on a new following
         Should have the correct follows counts 3:

      AssertionError: peertube@localhost:9001 on http://localhost:9003: expected 1 to equal 2
      + expected - actual

      -1
      +2
```

Test succeeds.

In `./shared/utils/server/jobs.ts` if I change from:

```javascript
    // Retry, in case of new jobs were created
    if (pendingRequests === false) {
      await wait(1000)

      await Promise.all(tasks)
    }
```
to:
 
```javascript
    // Retry, in case of new jobs were created
    if (pendingRequests === false) {
      await wait(2000)

      await Promise.all(tasks)
    }
```

It works. Can we just increase the wait time it or is there a better way?

**Steps to reproduce:**

```bash
git pull upstream develop
yarn run test
```

**Additional information**
* PeerTube version or URL: n/a
* Browser name/version: n/a
* NodeJS version: 8.12.0

4 cores/8 vCPUs with 16GB RAM

joshmorel:
Resolves #1492

### DIFF
--- a/shared/utils/server/jobs.ts
+++ b/shared/utils/server/jobs.ts
@@ -58,7 +58,7 @@ async function waitJobs (serversArg: ServerInfo[] | ServerInfo) {
 
     // Retry, in case of new jobs were created
     if (pendingRequests === false) {
-      await wait(1000)
+      await wait(2000)
 
       await Promise.all(tasks)
     }


Chocobozzz:
Thanks @joshmorel

Issue #1536, opened 2019-01-07
postgresql container not named


adrian-amaglio:
**What happened?**
The postgresql container is not named and therefore not found wen doing `docker-compose up`. The command line displays : ` "err": "SequelizeHostNotFoundError: getaddrinfo ENOTFOUND db db:5432`

**What do you expect to happen instead?**
The postgresql container should be found

**Steps to reproduce:**
Following the docker compose installation documentation

**Fix**
Add a container_name field in docker-compose.yml
```
  postgres:
    image: postgres:10-alpine
    container_name: db
```

adrian-amaglio:
I opened a pull request #1537

rigelk:
Are you running with the default support files?

What container specifically uses the "db" name to reach the `postgres` container? In https://github.com/Chocobozzz/PeerTube/blob/develop/support/docker/production/config/production.yaml#L21 you can see the hostname used is the default name, not `db`.

adrian-amaglio:
That’s wierd, I’m using this file [https://raw.githubusercontent.com/chocobozzz/PeerTube/master/support/docker/production/docker-compose.yml](https://raw.githubusercontent.com/chocobozzz/PeerTube/master/support/docker/production/docker-compose.yml)
found in the instructions here [https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/docker.md](https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/docker.md)
Is this document outdated ?

Chocobozzz:
> Is this document outdated ?

No, @rigelk linked the PeerTube **configuration** file, not the docker compose file. 

Try to re pull/re build your docker image first, we are in v1.1.0 now. Then, paste the content of your`.env` file if it still does not work.

PR #1537, opened 2019-01-07
Answer to peertube issue 1536. Add a name to postgresql container

adrian-amaglio:
The default configuration miss a name in the postgresql container declaration (needed by the other containers).


Chocobozzz:
Closing (see https://github.com/Chocobozzz/PeerTube/issues/1536#issuecomment-452111634)

PR #1551, opened 2019-01-13
Patch docker entrypoint to speed up the chown at startup

LecygneNoir:
Hello!

Here is a little patch to speed up the docker start on large instance.
Indeed the `chown -R` is executed on all files, even if the right are already ok, which cost IO and is very long with lots of files.

The patch uses find to read (instead of write) each files and apply the chown only if necessary, which speed up the start.

From benchmark on my instance, it reduces the start from 6 minutes to less than 1 minute :)

Datalove <3

### DIFF
--- a/support/docker/production/docker-entrypoint.sh
+++ b/support/docker/production/docker-entrypoint.sh
@@ -9,7 +9,7 @@ fi
 # Always copy default and custom env configuration file, in cases where new keys were added
 cp /app/config/default.yaml /config
 cp /app/support/docker/production/config/custom-environment-variables.yaml /config
-chown -R peertube:peertube /config
+find /config ! -user peertube -exec chown peertube:peertube {} \;
 
 # first arg is `-f` or `--some-option`
 # or first arg is `something.conf`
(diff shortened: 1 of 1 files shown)


Nutomic:
How about throwing an error instead? It should be really uncommon to have the wrong permissions, so I don't see the point in slowing down every startup for that.

LecygneNoir:
I agree that you should not have files created with wrong permission inside, it's very unlikely.

However I choose to keep the chown because we speak about docker here. Volume are exposed on the host, we do not know what the admin or user could do with it, possibly he launch Peertube tools manually, or copy files, it's difficult to take the assertion that rights are always correct :-/

Throwing an error is something possible indeed, but it'll break the docker start. 

With kubernetes, it means it will restart indefinitely causing downtime, until the admin could catch the error which could be difficult as kubernetes tends to delete/recreate pods very quickly.

With compose, I am unsure. The compose will fail, but is it possible to display the error for the specific container that fails? Or compose fail silently?

We can display an error without stopping the start, but very likely the admin will miss it (as node write lots of things too) and having incorrect rights could cause bigger damages after.

Perhaps we could add an option to check rights? In fully autonomous system, nothing is done manually, the docker could start quickly, but if there is a migration, videos added manually, passing a ENV variable to the docker to trigger the chown checks?

Issue #1237, opened 2018-10-10
Discuss segmented video architecture to support live streaming and long videos


scanlime:
Hi, I'm quite interested in solutions to #970 and #151. Both of these issues have at their root that PeerTube currently requires each video to be a single media file.

For background: There are multiple possible solutions to this problem, including very common ones (HTTP Live Streaming or DASH) and more experimental (github/Novage/p2p-media-loader). The most typical solution would be to use a growing .m3u8 playlist full of very short mp4 files which optionally grows during a broadcast. This same playlist format can be used for non-streaming (video on demand) content to provide fast loading and seeking even on very long streams.

I'm opening this issue to ask the PeerTube developers whether it's realistic to expect a solution to these issues in the current architecture. If so, this may be a good place to discuss a migration plan, which may involve a way for the protocol and servers to support both segmented and whole video files. If not, it would be good to know why not, and we can perhaps move on to other projects that still try to integrate at the ActivityPub level, and this may be a good place to plan for protocol interoperability.

scanlime:
In #1253, Chocobozzz says the video format is stable. I assume this means Peertube wouldn't consider replacing either current format (mp4 over HTTP, and the same but wrapped in WebTorrent) with an alternative segmented format (HLS, DASH, etc).

Replacing the format entirely would be a compelling option in some ways, since it leads to a straightforward server without any duplication of the videos. It burdens the client with remuxing in the event users want to download the video as a single file.

Another option might be to include HLS (.m3u8) as well as playlist-of-torrent formats alongside the current monolithic video files. This would either require duplicate storage space, or on-the-fly remuxing.

Another alternative is that videos could have multiple types of backend storage, with the missing formats generated on-the-fly until they've been requested frequently enough to cache.

Would it be at all possible to switch entirely to a new segmented format, or are we definitely stuck supporting multiple formats?

scanlime:
Another format option: Segmented MP4 wrapped in DASH. We can use a DASH MPD file to describe all available resolutions, and we can use multiple files if necessary during a live-stream. After live-streaming, or for single-file uploads, we'd remux into a segmented MP4 that can be used as-is or the individual segments can be referenced from the DASH MPD.

Mp4box is a tool to remux mp4 files using segments,
https://gpac.wp.imt.fr/2011/02/02/mp4box-fragmentation-segmentation-splitting-and-interleaving/

scanlime:
[...]

PR #1566, opened 2019-01-24
HLS experimental support

Chocobozzz:
https://github.com/Chocobozzz/PeerTube/issues/1237#issuecomment-457220443


mrlika:
Why do you use videojs? Is it used all over PeerTube?

P2P Media Loader support is not perfect in case of videojs because it does not use its build in [HTTP streaming](https://github.com/videojs/http-streaming) plugin but relies on a custom plugin that provides hlsjs engine support in the player. And the plugin is not actively developed.

mrlika:
Ok. I will try to add videojs HTTP streaming support for P2P Media Loader in nearest future.

rigelk:
@mrlika :astonished: that would be more than welcome ! Thanks

tcitworld:
:tada:

Chocobozzz:
Demo: https://peertube2.cpy.re/videos/watch/4b997fc0-e106-42d9-bff9-ae9d64902bbb?mode=p2p-media-loader

PR #1580, opened 2019-01-28
fix typo in test accounts api validators

joshmorel:
Noticed this when searching for "Test users API validators" - both in `check-params/users.ts` & `check-params/accounts.ts` so this one must be a typo.


Chocobozzz:
Thanks @joshmorel

PR #1619, opened 2019-02-10
Nginx config file: remove text/html from gzip_types

bnjbvr:
As stated by https://nginx.org/en/docs/http/ngx_http_gzip_module.html, text/html is always part of the gzip_types. This removes a warning when checking the Nginx configuration files.


rigelk:
Thanks @bnjbvr !

PR #1623, opened 2019-02-11
Update year to 2019

C0derLint:
Was reading through the README when I noticed the year on it was still 2018. What else could I do but open a PR? ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ Quonfucius, IP Solution, \_Laure\_, @lex666, 0x010C, 3dsman, 3rw4n-G3D, aallrd, 
 
 ## License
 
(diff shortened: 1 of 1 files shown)


Nutomic:
I think this should be something like 2018-2019. But I dont know much about copyright.

Serkan-devel:
I think peertube has existed even earlier

C0derLint:
@Nutomic The first commit to the project is in 2015 :D

Chocobozzz:
From https://www.gnu.org/licenses/gpl-howto.en.html (**The copyright notice** section) I think we should write 2015-2019

Chocobozzz:
Thanks @C0derLint

Issue #1445, opened 2018-12-02
Comment posted negative seconds ago


scanlime:
I assume there may be a small time desynchronization between server and client, but in any case i don't think this is what we want to be displaying:

![screen shot 2018-12-02 at 11 11 24 am](https://user-images.githubusercontent.com/923812/49343693-351a3b80-f623-11e8-8867-4ad3af43f76c.png)

I'd expect we could solve this by rounding any times more recent than a couple seconds to "just now" or something, ignoring negative values effectively.

Poslovitch:
I may work on this one this weekend

Poslovitch:
It seems it happens when publishing videos as well...
![image](https://user-images.githubusercontent.com/20014332/49675209-f64a1280-fa74-11e8-97db-6673976e01d1.png)

Chocobozzz:
@Poslovitch any news on this? :)

Poslovitch:
@Chocobozz hi, sorry, I'll have to dismiss my contribution. I don't feel confident enough with this language. Sorry again :(

zacharystenger:
Hi, I'd like to fix this issue. I believe I've got a proper fix [here](https://github.com/zacharystenger/PeerTube/tree/negative-seconds). It uses Math.max to make a negative seconds result equal to zero instead. It also gets rid of a redundant Math.floor operation.

rigelk:
@zacharystenger thanks ! Please make a pull request.

Chocobozzz:
Fixed by @zacharystenger in https://github.com/Chocobozzz/PeerTube/commit/ec24796ea83dbe15863c4f074844abbb4cdd5e08

PR #1625, opened 2019-02-12
Fix negative seconds by displaying 0 instead (#1445)

zacharystenger:
Use Math.max to make negative seconds equal to zero instead. Remove redundant Math.floor operation.


Chocobozzz:
Thanks @zacharystenger

Issue #371, opened 2018-03-21
Add LDAP user backend


tigre-bleu:
Clearly not a top priority, but it would be nice to support LDAP integration as Mastodon does.

I would be happy to provide peertube service to my users when this will be implemented.

thegcat:
Same here.

anmol26s:
Hi
This would be priority feature for YunoHost too.

rigelk:
Looking rapidly at what's available I found [vesse/passport-ldapauth](https://github.com/vesse/passport-ldapauth). Maybe we could use that for the implementation.

meoso:
this is what's keeping us from implementing.  we would need to auth with Active Directory and integrate with Moodle somehow.

rigelk:
@meoso what do you mean by "integrate with Moodle"?

meoso:
sorry, our Moodle authenticates with Acitve Dir (LDAP), that's all.  Maybe we would just embed URL links into moodle. (not really my job function, i just know we need "something" to serv video much like a youtube clone and i really like the p2p offload of PeerTube)

ShenZhouHong:
I think integration with LDAP or some other form of Single Sign On authentication is more important than we think. Although it's a bit of an "enterprise-y" feature, some users are hosting decentralised "cyber-collectives" where federation services like Mastodon, Nextcloud, and Matrix are all offered under one URL and umbrella. All 3 of these services support LDAP, which makes it trivial to have a central registration page, so users automatically get accounts on all 3 of the services without any hassle. If Peertube supports an SSO backend, it would make it even easier to integrate :)

Thatoo:
I'm also waiting for LDAP support to install Peertube (and mastodon) on my familly yunohost [CHATONS](http://chatons.org/) server.

WingsLikeEagles:
How about CAS authentication, that way people can use whatever auth they want to implement (ldap, x.509, 2 factor, etc).
https://www.apereo.org/projects/cas

ldidry:
> How about CAS authentication, that way people can use whatever auth they want to implement (ldap, x.509, 2 factor, etc).

True, but putting up a working CAS is not so easy. Not as easy as a setting LDAP configuration in peertube conf.

I think that adding CAS authentication would be a good thing, but not for that reason.

meoso:
I know the dev-ops team here is implementing CAS with our moodle, so here is a +1.

immae:
Hey there,
[...]

PR #1626, opened 2019-02-12
Add LDAP authentication

immae:
This PR implements LDAP authentication. The behavior is the expected standard behavior for LDAP authentication: a master LDAP account can search for the user that tries to log-in, then if found we try to bind as that user and if successful, the user is logged in.

I tried to add some tests, but I couldn’t find any existing tests for models and libs (the parts of the code that I touched)...

Fixes https://github.com/Chocobozzz/PeerTube/issues/371

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -43,6 +43,19 @@ redis:
   auth: null
   db: 0
 
+auth:
+  local:
+    enabled: true
+  ldap:
+    enabled: true
+    url: ldap://localhost:389/dc=example,dc=com
+    insecure_tls: false
+    bind_dn: cn=admin,dc=example,dc=com
+    bind_password: adminPass
+    base: dc=example,dc=com
+    mail_entry: "mail"
+    user_filter: "(|(email=%username%)(uid=%username%))"
+
 smtp:
   hostname: null
   port: 465
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -43,6 +43,19 @@ redis:
   auth: null
   db: 0
 
+auth:
+  local:
+    enabled: true
+  ldap:
+    enabled: true
+    url: ldap://localhost:389/dc=example,dc=com
+    insecure_tls: false
+    bind_dn: cn=admin,dc=example,dc=com
+    bind_password: adminPass
+    base: dc=example,dc=com
+    mail_entry: "mail"
+    user_filter: "(|(email=%username%)(uid=%username%))"
+
 # SMTP server to send emails
 smtp:
   hostname: null
--- a/server/initializers/constants.ts
+++ b/server/initializers/constants.ts
@@ -14,7 +14,7 @@ import { CONFIG, registerConfigChangedHandler } from './config'
 
 // ---------------------------------------------------------------------------
 
-const LAST_MIGRATION_VERSION = 370
+const LAST_MIGRATION_VERSION = 375
 
 // ---------------------------------------------------------------------------
 
--- a/server/lib/oauth-model.ts
+++ b/server/lib/oauth-model.ts
@@ -75,7 +75,13 @@ function getRefreshToken (refreshToken: string) {
 async function getUser (usernameOrEmail: string, password: string) {
   logger.debug('Getting User (username/email: ' + usernameOrEmail + ', password: ******).')
 
-  const user = await UserModel.loadByUsernameOrEmail(usernameOrEmail)
+  let user : UserModel
+  if (CONFIG.AUTH.LDAP.ENABLED) {
+    user = await UserModel.findOrCreateLDAPUser(usernameOrEmail)
+  }
+  if (!user && CONFIG.AUTH.LOCAL.ENABLED) {
+    user = await UserModel.loadByUsernameOrEmail(usernameOrEmail)
+  }
   if (!user) return null
 
   const passwordMatch = await user.isPasswordMatch(password)
--- a/server/models/account/user.ts
+++ b/server/models/account/user.ts
@@ -1,4 +1,5 @@
 import { FindOptions, literal, Op, QueryTypes } from 'sequelize'
+import { Ldap } from '../../lib/ldap'
 import {
   AfterDestroy,
   AfterUpdate,
@@ -44,7 +45,9 @@ import { AccountModel } from './account'
 import { NSFWPolicyType } from '../../../shared/models/videos/nsfw-policy.type'
 import { values } from 'lodash'
 import { NSFW_POLICY_TYPES } from '../../initializers/constants'
+import { CONFIG } from '../../initializers/config'
 import { clearCacheByUserId } from '../../lib/oauth-model'
+import { createUserAccountAndChannelAndPlaylist } from '../../lib/user'
 import { UserNotificationSettingModel } from './user-notification-setting'
 import { VideoModel } from '../video/video'
 import { ActorModel } from '../activitypub/actor'
@@ -112,6 +115,11 @@ export class UserModel extends Model<UserModel> {
   @IsEmail
   @Column(DataType.STRING(400))
   email: string
+
+  @AllowNull(true)
+  @Default(null)
+  @Column
+  ldapDn: string
 
   @AllowNull(true)
   @Default(null)
@@ -362,6 +370,48 @@ export class UserModel extends Model<UserModel> {
     return UserModel.findOne(query)
   }
 
+  static loadByLdapDn (ldapDn: string) {
+    const query = {
+      where: {
+        ldapDn
+      }
+    }
+
+    return UserModel.findOne(query)
+  }
+
+  static async findOrCreateLDAPUser (username: string) {
+    try {
+      const userInfos = await Ldap.Instance.findUser(username)
+      const user = await UserModel.loadByLdapDn(userInfos['dn'])
+      if (user) {
+        return user
+      } else {
+        return await UserModel.createLDAPUser(username, userInfos)
+      }
+    } catch (e) {
+      return null
+    }
+  }
+
+  static async createLDAPUser (username: string, userInfos: {}) {
+    const userToCreate = new UserModel({
+      username,
+      password: 'SomeInvalidPassword',
+      email: userInfos[CONFIG.AUTH.LDAP.MAIL_ENTRY],
+      ldapDn: userInfos['dn'],
+      nsfwPolicy: CONFIG.INSTANCE.DEFAULT_NSFW_POLICY,
+      autoPlayVideo: true,
+      role: UserRole.USER,
+      videoQuota: CONFIG.USER.VIDEO_QUOTA,
+      videoQuotaDaily: CONFIG.USER.VIDEO_QUOTA_DAILY,
+      emailVerified: true,
+      adminFlags: UserAdminFlag.NONE
+    })
+    const { user } = await createUserAccountAndChannelAndPlaylist(userToCreate)
+    return user
+  }
+
   static loadByUsernameAndPopulateChannels (username: string) {
     const query = {
       where: {
@@ -529,7 +579,11 @@ export class UserModel extends Model<UserModel> {
   }
 
   isPasswordMatch (password: string) {
-    return comparePassword(password, this.password)
+    if (this.ldapDn === null) {
+      return comparePassword(password, this.password)
+    } else {
+      return Ldap.Instance.checkUser(this.ldapDn, password)
+    }
   }
 
   toFormattedJSON (parameters: { withAdminFlags?: boolean } = {}): User {
--- a/package.json
+++ b/package.json
@@ -125,6 +125,7 @@
     "js-yaml": "^3.5.4",
     "jsonld": "~1.1.0",
     "jsonld-signatures": "https://github.com/Chocobozzz/jsonld-signatures#rsa2017",
+    "ldapjs": "^1.0.2",
     "lodash": "^4.17.10",
     "magnet-uri": "^5.1.4",
     "memoizee": "^0.4.14",
--- a/server/initializers/config.ts
+++ b/server/initializers/config.ts
@@ -34,6 +34,21 @@ const CONFIG = {
     AUTH: config.has('redis.auth') ? config.get<string>('redis.auth') : null,
     DB: config.has('redis.db') ? config.get<number>('redis.db') : null
   },
+  AUTH: {
+    LOCAL: {
+      ENABLED: config.has('auth.local.enabled') ? config.get<boolean>('auth.local.enabled') : true,
+    },
+    LDAP: {
+      ENABLED: config.has('auth.ldap.enabled') ? config.get<boolean>('auth.ldap.enabled') : false,
+      URL: config.has('auth.ldap.url') ? config.get<string>('auth.ldap.url') : null,
+      INSECURE_TLS: config.has('auth.ldap.insecure_tls') ? config.get<boolean>('auth.ldap.insecure_tls') : false,
+      BIND_DN: config.has('auth.ldap.bind_dn') ? config.get<string>('auth.ldap.bind_dn') : null,
+      BIND_PASSWORD: config.has('auth.ldap.bind_password') ? config.get<string>('auth.ldap.bind_password') : null,
+      BASE: config.has('auth.ldap.base') ? config.get<string>('auth.ldap.base') : null,
+      MAIL_ENTRY: config.has('auth.ldap.mail_entry') ? config.get<string>('auth.ldap.mail_entry') : 'mail',
+      USER_FILTER: config.has('auth.ldap.user_filter') ? config.get<string>('auth.ldap.user_filter') : null
+    },
+  },
   SMTP: {
     HOSTNAME: config.get<string>('smtp.hostname'),
     PORT: config.get<number>('smtp.port'),
--- a/server/initializers/migrations/0375-user-ldap-dn.ts
+++ b/server/initializers/migrations/0375-user-ldap-dn.ts
@@ -1,0 +1,26 @@
+import * as Sequelize from 'sequelize'
+
+async function up (utils: {
+  transaction: Sequelize.Transaction,
+  queryInterface: Sequelize.QueryInterface,
+  sequelize: Sequelize.Sequelize
+}): Promise<void> {
+
+  {
+    const data = {
+      type: Sequelize.STRING,
+      allowNull: true,
+      defaultValue: null
+    }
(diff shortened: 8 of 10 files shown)


config/production.yaml.example
```diff
@@ -33,6 +33,17 @@ redis:
...
+  enable: true
+  # Disallow non-ldap users (it also disables root!)
+  ldap_only: true
+  url: ldap://localhost:389/dc=example,dc=com
+  bind_dn: cn=admin,dc=example,dc=com
+  bind_password: adminPass
+  base: dc=example,dc=com
+  mail_entry: "mail"
+  user_filter: "(|(email=%username%)(uid=%username%))"
+
```
ldidry: You may want to add a setting to prevent connection failure when you use ldaps and the LDAP server has a self-signed certificate (for ex.).

I didn't use the same module than you to connect to LDAP, but on [ep_mypads](https://framagit.org/framasoft/Etherpad/ep_mypads/), I had to add this optional setting in the LDAP settings:

```
    "tlsOptions": {
        "rejectUnauthorized": false
    },
```

If you want to easily test your code against a self-signed certificated ldaps, you can use this docker image: https://github.com/rroemhild/docker-test-openldap

immae: Good catch @ldidry I added that option.

I used inspiration from etherpad-lite to write this PR, so it’s definitely the same node module ldapjs that is used in both cases :)

(And thanks for pointing me to ep_mypads module, it seems great and I didn’t know about!)

config/default.yaml
```diff
@@ -33,6 +33,9 @@ redis:
   auth: null
   db: 0
 
+ldap:
+  enable: false
```
Chocobozzz: `enabled` for consistency

And add all ldap keys please

config/production.yaml.example
```diff
@@ -33,6 +33,18 @@ redis:
   auth: null
   db: 0
 
+ldap:
+  enable: true
```
Chocobozzz: `enabled`

config/production.yaml.example
```diff
@@ -33,6 +33,18 @@ redis:
...
+  # Disallow non-ldap users (it also disables root!)
+  ldap_only: true
```
Chocobozzz: I'm not sure it's a good idea. What if instead we have something like:

```
auth:
  local:
    enabled: boolean
  ldap:
    enabled: boolean
    ...
```

server/initializers/constants.ts
```diff
@@ -16,7 +16,7 @@ let config: IConfig = require('config')
 
 // ---------------------------------------------------------------------------
 
-const LAST_MIGRATION_VERSION = 340
+const LAST_MIGRATION_VERSION = 341
```
Chocobozzz: Use 345 so we can insert other migrations in patch versions

server/initializers/constants.ts
```diff
@@ -177,6 +177,17 @@ const CONFIG = {
...
+  LDAP: {
+    ENABLE: config.has('ldap.enable') ? config.get<boolean>('ldap.enable') : false,
+    LDAP_ONLY: config.has('ldap.ldap_only') ? config.get<boolean>('ldap.ldap_only') : false,
+    URL: config.has('ldap.url') ? config.get<string>('ldap.url') : null,
+    INSECURE_TLS: config.has('ldap.insecure_tls') ? config.get<boolean>('ldap.insecure_tls') : false,
+    BIND_DN: config.has('ldap.bind_dn') ? config.get<string>('ldap.bind_dn') : null,
+    BIND_PASSWORD: config.has('ldap.bind_password') ? config.get<string>('ldap.bind_password') : null,
+    BASE: config.has('ldap.base') ? config.get<string>('ldap.base') : null,
+    MAIL_ENTRY: config.has('ldap.mail_entry') ? config.get<string>('ldap.mail_entry') : 'mail',
+    USER_FILTER: config.has('ldap.user_filter') ? config.get<string>('ldap.user_filter') : '(|(email=%username%)(uid=%username%))'
```
Chocobozzz: Prefer `null` instead of `'(|(email=%username%)(uid=%username%))'`

server/initializers/constants.ts
```diff
@@ -177,6 +177,17 @@ const CONFIG = {
...
   },
+  LDAP: {
+    ENABLE: config.has('ldap.enable') ? config.get<boolean>('ldap.enable') : false,
+    LDAP_ONLY: config.has('ldap.ldap_only') ? config.get<boolean>('ldap.ldap_only') : false,
+    URL: config.has('ldap.url') ? config.get<string>('ldap.url') : null,
+    INSECURE_TLS: config.has('ldap.insecure_tls') ? config.get<boolean>('ldap.insecure_tls') : false,
+    BIND_DN: config.has('ldap.bind_dn') ? config.get<string>('ldap.bind_dn') : null,
+    BIND_PASSWORD: config.has('ldap.bind_password') ? config.get<string>('ldap.bind_password') : null,
+    BASE: config.has('ldap.base') ? config.get<string>('ldap.base') : null,
+    MAIL_ENTRY: config.has('ldap.mail_entry') ? config.get<string>('ldap.mail_entry') : 'mail',
```
Chocobozzz: Prefer `null` instead of `'mail'`

server/lib/oauth-model.ts
```diff
@@ -66,7 +66,13 @@ function getRefreshToken (refreshToken: string) {
 async function getUser (usernameOrEmail: string, password: string) {
   logger.debug('Getting User (username/email: ' + usernameOrEmail + ', password: ******).')
 
-  const user = await UserModel.loadByUsernameOrEmail(usernameOrEmail)
+  let user
```
Chocobozzz: `let user: UserModel`

server/models/account/user.ts
```diff
@@ -354,6 +361,91 @@ export class UserModel extends Model<UserModel> {
...
+        if (user) {
+          return user
+        } else {
+          return UserModel.createLDAPUser(username, userInfos)
+        }
+      })
+      .catch(() => { return null })
+  }
+
+  static findLDAPUser (username: string) {
```
Chocobozzz: Put it in a `lib/ldap.ts` module, so you create the client only once

server/models/account/user.ts
```diff
@@ -354,6 +361,91 @@ export class UserModel extends Model<UserModel> {
...
+      }
+    }
+
+    return UserModel.findOne(query)
+  }
+
+  static findOrCreateLDAPUser (username: string) {
+    let userInfos
+
+    return Promise.resolve(UserModel.findLDAPUser(username))
```
Chocobozzz: Use async/await instead

Example: `const { dn ] = await UserModel.findLDAPUser(username)` etc

You can use try/catch, but please log the error

server/models/account/user.ts
```diff
@@ -354,6 +361,91 @@ export class UserModel extends Model<UserModel> {
...
+      scope: 'sub',
+      attributes: [ CONFIG.LDAP.MAIL_ENTRY, 'dn' ]
+    }
+
+    return new Promise(function (resolve, reject) {
+      client.bind(CONFIG.LDAP.BIND_DN, CONFIG.LDAP.BIND_PASSWORD, function (err) {
+        if (err) reject()
+        let entries = []
+        client.search(CONFIG.LDAP.BASE, opts, function (err, search) {
+          if (err) reject()
```
Chocobozzz: `reject(err)`

server/models/account/user.ts
```diff
@@ -354,6 +361,91 @@ export class UserModel extends Model<UserModel> {
...
+    })
+    const opts = {
+      filter,
+      scope: 'sub',
+      attributes: [ CONFIG.LDAP.MAIL_ENTRY, 'dn' ]
+    }
+
+    return new Promise(function (resolve, reject) {
+      client.bind(CONFIG.LDAP.BIND_DN, CONFIG.LDAP.BIND_PASSWORD, function (err) {
+        if (err) reject()
```
Chocobozzz: `reject(err)`

server/models/account/user.ts
```diff
@@ -354,6 +361,91 @@ export class UserModel extends Model<UserModel> {
...
+        client.search(CONFIG.LDAP.BASE, opts, function (err, search) {
+          if (err) reject()
+          search.on('searchEntry', function (entry) {
+            entries.push(entry.object)
+          })
+          search.on('end', function (result) {
+            if (entries.length === 1) {
+              resolve(entries[0])
+            } else {
+              reject()
```
Chocobozzz: Reject an error please

server/models/account/user.ts
```diff
@@ -354,6 +361,91 @@ export class UserModel extends Model<UserModel> {
...
+            }
+          })
+        })
+      })
+    })
+  }
+
+  static createLDAPUser (username: string, userInfos: {}) {
+    return Promise.resolve(userInfos)
```
Chocobozzz: Use async/await

You don't need `Promise.resolve`

server/models/account/user.ts
```diff
@@ -354,6 +361,91 @@ export class UserModel extends Model<UserModel> {
...
+      .then((userInfos) => {
+        const userToCreate = new UserModel({
+          username: username,
+          password: 'SomeInvalidPassword',
```
Chocobozzz: `null` no?

server/models/account/user.ts
```diff
@@ -354,6 +361,91 @@ export class UserModel extends Model<UserModel> {
...
+        })
+      })
+    })
+  }
+
+  static createLDAPUser (username: string, userInfos: {}) {
+    return Promise.resolve(userInfos)
+      .then((userInfos) => {
+        const userToCreate = new UserModel({
+          username: username,
```
Chocobozzz: `username: username,` -> `username,`

immae:
> What if the user username or email is changed in LDAP? We can say we don't support username change, but we should support email change.

Right, we should update the e-mail when it changes on LDAP side

> What do we do when someone registers on the instance with a username that already exists in LDAP, but not in our local database (because the user has not yet connected for example)

It is created automatically

server/models/account/user.ts
immae: password field cannot be null, and I didn’t want to remove this constraint as I don’t know the consequences in the rest of the code. If you think it’s safe, we can make it nullable

config/production.yaml.example
immae: After some thought, I don’t know if it is a good idea to support both, apart from a transitionning period: supporting both local and "remote" (be it ldap or something else) user will most probably provoke clashes in usernames at some point.

So my thought would be to have ldap_only to true, and warn the user that ldap_only: false should be used only during transition to ldap

server/initializers/constants.ts
immae: Ok

server/initializers/constants.ts
immae: `mail` is a standard field name for e-mail in ldap, so I think we should keep it

config/production.yaml.example
Chocobozzz: Yes, it was my second concern:

"What do we do when someone registers on the instance with a username that already exists in LDAP, but not in our local database (because the user has not yet connected for example)"

When someone registers, we could check in LDAP if the username already exists. Because even if there is not much sense to have LDAP and local auth enabled, we'll have this kind of use case with other authentication methods.

Anyway, please use the `auth:` parent key. We can then throw errors/warnings at startup if we decide we don't support LDAP & local authentication.

server/models/account/user.ts
Chocobozzz: Yep I think we can allow null on this column

immae:
@Chocobozzz : I finally had the time to address your requests.

I didn’t manage to make the password field nullable : I changed the column in the database to make it nullable, changed in the UserModel too, and made the isUserPasswordValid validator more flexible, but it keeps failing. I may have missed something but I couldn’t find what.

I changed some of the Promise as required, but in lib/ldap.ts there are still some: the ldapjs library works with callbacks, and I don’t know how to turn that into an async/await process.

immae:
(about password field nullable: it fails in `lib/user.ts` `createUserAccountAndChannelAndPlaylist`, at  line `userToCreate.save(userOptions)`, regardless of `validate` value)

server/initializers/constants.ts
immae: (changed for 375)

Chocobozzz:
> I didn’t manage to make the password field nullable : I changed the column in the database to make it nullable, changed in the UserModel too, and made the isUserPasswordValid validator more flexible, but it keeps failing. I may have missed something but I couldn’t find what.

No problem, I'll merge your PR in a separate branch (next week or the week after) and will update the code myself. 

Please could you add unit tests?

server/lib/oauth-model.ts
```diff
@@ -75,7 +75,13 @@ function getRefreshToken (refreshToken: string) {
 async function getUser (usernameOrEmail: string, password: string) {
   logger.debug('Getting User (username/email: ' + usernameOrEmail + ', password: ******).')
 
-  const user = await UserModel.loadByUsernameOrEmail(usernameOrEmail)
+  let user : UserModel
+  if (CONFIG.AUTH.LDAP.ENABLED) {
+    user = await UserModel.findOrCreateLDAPUser(usernameOrEmail)
+  }
+  if (!user && CONFIG.AUTH.LOCAL.ENABLED) {
```
Chocobozzz: I think we should put the local check first. Because users that were created before the LDAP activation should be able to log in with their "old" account.

Chocobozzz:
> >  What do we do when someone registers on the instance with a username that already exists in LDAP, but not in our local database (because the user has not yet connected for example)

> It is created automatically

IMHO you should add a check in the registration controller/middleware, so users cannot register a username or email that already exists in LDAP. What do you think?

immae:
[IRC.log](https://github.com/Chocobozzz/PeerTube/files/3171667/IRC.log)

immae:
[IRC.log](https://github.com/Chocobozzz/PeerTube/files/3171672/IRC.log)

battosai30:
Hi guys,

Any update on this feature ? I'm thinking about implement PeerTube in severall structure that I managed but I really need LDAP for my users as I already use other services (like RocketChat, Nextcloud etc ...) which needs authentification and already support LDAP and allowed my users to only have a single login set on all platforms.

Regards and thanks for your great job

Chocobozzz:
Hi,

I plan to improve the plugin system that would allow to register an `auth` method. Then, (I or @immae if he wants) will externalize this PR in a plugin.

The algo would be:
 * Every auth plugin would have a weight, defined by the administrator
 * On login, we check if the username exists in database. If yes, try to login against this user (it could be an external auth method)
 * If the user does not exist, try to login with every auth plugin ordered by `weight DESC`. On success with a particular auth plugin, create the user in the database with a column that specify the auth method (`LDAP`, `local` etc)

@immae What do you think?

immae:
@Chocobozzz yes that seems to be perfect :) Maybe some details would need to be adapted (for auth methods that require more than just a password), but the idea seems good

server/lib/oauth-model.ts
battosai30: Usually it's not the good strategy : 
1) If LDAP is enabled, you first  check if login and password are ok in LDAP server.
2) If not, and if fallback login option is activated, you check in local app database if login/password are OK.

Usually LDAP server is the authentification security reference, so it's normal that's if activated it's checked first.

jadjay:
Thanks for this !
I hope it will be soon released...

immae:
@Chocobozzz : should we maybe close this stale PR? Please note that I’m  following the project (with a little delay, I’m currently at latest 2.X.X), so I’m always able to provide an up to date PR whenever the topic comes back on the table

immae:
NB: I didn’t follow the project closely enough to be able to tell wether you implemented that plugin system you talk about in above comment, if so please tell me and I can have a look at it...

Chocobozzz:
@immae Nope, it's a reminder for me. If all goes well, I should develop an LDAP plugin this year.

Chocobozzz:
Plugin done in https://framagit.org/framasoft/peertube/official-plugins/-/tree/master/peertube-plugin-auth-ldap

Plugin auth support WIP on https://github.com/Chocobozzz/PeerTube/tree/feature/auth-plugins

UnlimitedCookies:
Thank you so much for creating the highly requested ldap auth plugin! ❤️ 

I am still wondering: 
- How and when can I install this? (Because I couldn't find it in the plugin search yet)
- And will there be documentation (for general OpenLDAP usage)?
- Where can I find the auth weight setting then?

rigelk:
@UnlimitedCookies the plugin specifies compatibility only starting with the next release, as can be read in its `package.json`, so you will have to wait for version 2.2 :wink:

Chocobozzz:
> And will there be documentation (for general OpenLDAP usage)?
Where can I find the auth weight setting then?

I'm not sure to understand these questions

willrimmer:
Looking forward to LDAP support! One login to rule them all...

UnlimitedCookies:
@willrimmer And the amazing thing is, it's already implemented ;)

jdaviescoates:
@UnlimitedCookies yes, but how to actually implement/ use it? Needs documenting.

willrimmer:
Thats awesome! Is there guidelines on implementation?

July 23, 2020 6:49 PM, "UnlimitedCookies" <notifications@github.com (mailto:notifications@github.com?to=%22UnlimitedCookies%22%20<notifications@github.com>)> wrote:
	@willrimmer (http://url2781.famfam.us/ls/click?upn=eLZImTJF5xLTGfznWHzZtpmbFRbx0RofvM7Beoi2fQ12xmdQ-2FmozCUj987QfjUCc3931_U31jt5HgHBh7k3Grc5s4l7xuQERo1tl9uwqaRQ-2FWpkyKdwlIl44SYpn94CEvcAa1yKzGkaGD3jvkcCK4MiQYhKiXcNHIKEQY7nNDHNBSjQpQYyxq3D7NzgPRhkpSsvhtpYw01LZB9Mf4HC-2F-2FPfJHUPJpmh52jdCc8xpRbSkhQwmb7R-2FX-2BMNMZFaXrx-2FLomAZsQAIT9GGWfZ-2BJJ4kgDz-2BZbJ7dvzuhJmLOKkBtEKIC-2BGRtgCnNlDCGfTaG7GY-2BOpVd8jU7vh8VRkYnND0rN13YNv3Ab5GOML6Co-2FphaByHbk-3D And the amazing thing is, it's already implemented ;)

Chocobozzz:
@jdaviescoates https://docs.joinpeertube.org/#/admin-managing-users?id=external-auths-support

PR #1637, opened 2019-02-15
add quarantine videos feature

### LINKED ISSUE #138: Add manual video validation by a moderator
Hello
After two bad upload today (ru serie without copyright >< ), I thing is good to validate video on upload ... (And add option "validate all video or check each video" ?). After, it's cool to etablish a reputation for each user.
For exemple, if user upload 10 good vidéos, after it doesn't not need validate. :)

I think, my english is very bad tonight :(

Regard

joshmorel:
Resolves #138 

@Chocobozzz - had to add a lot so I'll make a few points on design decisions:

* Considered creating separate VideoState at first but decided `video.quarantined` boolean would be best. First, I estimated it would mean less code changes now and in the future. Second, a video can be released from quarantine by moderator while in either `TO_TRANSCODE` or `PUBLISHED` state so a separate state didn't make sense.
* For the quarantined video list UI I used the video list similar to my-videos instead of other moderation lists. My rationale was that this would facilitate better workflow for the moderator as seeing the thumbnail would be important and could support quickly releasing videos for more trusted users.
* I was running into issues with using `notifyOnPendingVideoPublished` in multiple contexts so decided to create 3 distinct public functions (which call the same private function) making everything more explicit and easier to reason about

Do not hesitate with any questions or suggestions.


server/tests/api/videos/video-quarantine.ts
```diff
@@ -0,0 +1,189 @@
...
+      videoId = res.body.video.id
+      await waitJobs([ server, followingServer ])
+
+      const followingRes = await getVideosList(followingServer.url)
+      const videos = followingRes.body.data
+      expect(videos).to.be.an('array')
+      expect(videos.length).to.equal(releasedVideosCount)
+    })
+
+    it('Should propogate the released video to following server', async function () {
```
Chocobozzz: propagate

server/tests/client.ts
```diff
@@ -50,7 +50,7 @@ describe('Test a client controllers', function () {
     server.video = videos[0]
   })
 
-  it('Should have valid Open Graph tags on the watch page with video id', async function () {
+  it.skip('Should have valid Open Graph tags on the watch page with video id', async function () {
```
Chocobozzz: why?

Chocobozzz:
Beautiful :heart_eyes: 

Thanks a lot @joshmorel 

>  Considered creating separate VideoState at first but decided `video.quarantined` boolean would be best. First, I estimated it would mean less code changes now and in the future. Second, a video can be released from quarantine by moderator while in either `TO_TRANSCODE` or `PUBLISHED` state so a separate state didn't make sense.

Did you consider using the videos blacklist system instead of the `quarantined` boolean? We could imagine adding `type` column with `blacklist` or `quarantine`as values. Because the blacklist and the quarantine features are very similar.

I don't think using the videos comment system to communicate with the uploader is the best way to do it. I think we should create separate table/endpoints to store this type of communication. But I can do it :)

server/tests/client.ts
joshmorel: Good catch thanks

joshmorel:
> Did you consider using the videos blacklist system instead of the `quarantined` boolean? We could imagine adding a `type` column with `blacklist` or `quarantine`as values. Because the blacklist and the quarantine features are very similar.

Good idea but I missed it. I had discussed using the blacklist system to communicate issues related to videos which the moderator is hesitant to release, however, I did not think of using an "automatic blacklist" for new videos on instances wanting this feature. 

I feel that would have been a better first stab at resolving the original issue as less effort would have been required and more existing code could be leveraged. The main advantage in not using the blacklist is that the use cases, although sharing similarities, are different and you'd want to facilitate and optimize the workflows in different ways while also not causing confusion. However, I can see doing more of this workflow customization on the front-end while leveraging the existing blacklist object and related logic (video not visible to other users, video not propagated etc) on the back-end. 

The up-front effort is expended but there's definitely a consideration for maintainability of the code base to better extend the blacklist. Let me know if you'd like me to try it. I'm not against taking a stab at seeing how keeping the front-end _mostly_ similar (so it is more custom to the workflow) while leveraging a _typed_ blacklist for the back-end might work.

Chocobozzz:
> The up-front effort is expended but there's definitely a consideration for maintainability of the code base to better extend the blacklist. Let me know if you'd like me to try it. I'm not against taking a stab at seeing how keeping the front-end _mostly_ similar (so it is more custom to the workflow) while leveraging a _typed blacklist_ on the back-end might work.

If you have time to do this, go for it! Otherwise, I can do it. But I completely agree: keep the same front that you realized and that perfectly meets the moderator need, and try to use the blacklist system on the backend.

joshmorel:
@Chocobozzz I'm pretty close to getting this done (https://github.com/joshmorel/PeerTube/tree/feature/quarantine-videos-use-blacklist)  but looking for general feedback and I also need input on one specific thing.

I'm proposing using "blacklist" sub-typed into manual and automatic. It does add some complexity to existing blacklist related functionality but overall I would say it seems like the better option from a code reuse & complexity perspective.

For instances using the feature moderators will see two lists:

peertube-blacklist-sub-typed

For instances not using the feature (it might be useful to find out how many actually use it eventually) everything will look mostly like before:

peertube-blacklist-manual-only

The one thing I need and am looking for suggestions on the best place and way to implement is a list of blacklisted video (with type filter) so I can implement the same `AbstractVideoList` as I did with the quarantine list component on the front-end. The existing `VideoBlacklist` model is extremely far from the necessary `Video` format. Where/how would be the best place to implement this? Is it a new route within the video controller, or the video blacklist controller, or a parameter in an existing route? Or do I just not implement the `AbstractVideoList` component and do something custom instead while adding a few necessary required attributes added to the `VideoBlacklist` model (e.g. thumbnail etc). Or do you have another better idea?

Chocobozzz:
@joshmorel the new branch seems very good to me. 

> The existing VideoBlacklist model is extremely far from the necessary Video format.

If I understand well, you're missing some attributes in the `video` field of https://github.com/Chocobozzz/PeerTube/blob/develop/shared/models/videos/blacklist/video-blacklist.model.ts#L8?
If so, you can replace the custom `video` field by the [Video object](https://github.com/Chocobozzz/PeerTube/blob/develop/shared/models/videos/video.model.ts#L41). You can then use the [FOR_API scope](https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/video/video.ts#L201) (that includes all attributes your will need in the abstract videos component) in [video blacklist -> listForApi](https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/video/video-blacklist.ts#L53).

The auto blacklist feature is very interesting, because we could imagine having different rules in the future. For example, disabling auto blacklist per user. So I'm thinking something like:

```
auto_blacklist:
  # New videos automatically blacklisted so moderators can review before publishing
  videos:
    new_users:
      enabled: false
```

in the configuration file would be better (or something like that, if you have a better formulation don't hesitate).

joshmorel:
What would we mean by `new_users`? Right now if the feature is enabled it's applied to all videos of all users except admins/mods (who have `BYPASS_AUTO_BLACKLIST_RIGHT`). @Nutomic requested future improvement where a threshold could be set - if user had 1,2,3,x videos reviewed, removed from blacklist and published they would be exempt (quote below). So by `new_user` related to this, meaning perhaps someone who had uploaded no videos yet? Another idea could be to have a setting on users to explicitly give them the bypass right without making them moderators.

> One option I would like to see is an option to validate videos automatically once one (or two or three) videos by a user have been approved. So only the first videos by a user would need manual validation. But of course that can be added later.

Chocobozzz:
My configuration example means "automatically blacklist videos of new users". But the use case of @Nutomic (with the threshold) is better. I just want to have an evolutive configuration key, so we can add in the future other parameters like "automatically blacklist videos of new users who registered < 30 days ago" or "automatically blacklist videos of new users who uploaded < 3 videos" but also "apply auto blacklist to new videos containing a specific term" etc.

> Right now if the feature is enabled it's applied to all videos of all users except admins/mods (who have BYPASS_AUTO_BLACKLIST_RIGHT

I think it's better to not have this `BYPASS_AUTO_BLACKLIST_RIGHT` right. Just skip auto blacklist to users that have the `MANAGE_VIDEO_BLACKLIST` right. And in the future, we may add a `skipAutoVideoBlacklist` boolean in the `user` table so admins can choose who can skip the auto blacklisting.

So another proposal could be

```
auto_blacklist:
  # New videos automatically blacklisted so moderators can review before publishing
  videos:
    of_users:
      enabled: false
      registration_lt: '3 days' # auto blacklist videos of users who registered less than 3 days ago
      uploaded_videos_lt: 5 # auto blacklist videos of users who uploaded less than 5 videos
    with_name:
       enabled: false
       containing:  # auto blacklist videos that have a name that contains 'toto'
         - 'toto'
```

Of course the `registration_lt` and `uploaded_videos_lt` keys could be implemented in a long time, but we must anticipate this kind of addition.

joshmorel:
Okay using blacklist now and I incorporated the recommended suggestions.

A few comments/pictures:

If enabled there are two moderation lists related to blacklist - "Manually blacklisted" and "Auto-blacklist". The former is as before but includes only the `MANUAL` type:

PeerTube-blacklist-list-manual-only

The later implements the abstract video list component based on my assumption that that kind of a UI would better facilitate the moderator's workflow for assessing videos with the name, user, length and thumbnail immediately displayed. It includes only the type AUTO_BEFORE_PUBLISHED:

Peertube-blacklist-list-auto

If the feature is not enabled then the blacklist list and moderation item title is as before. The filter is removed, my assumption being if the feature is enabled then disabled if there are any of the later type they would still want to be seen in the list.

 Peertube-blacklist-list-combined

There were two UI items I was unsure regarding but my opinion is they are okay for now and can be tweaked if needed once feedback can be gathered.

First, when I was using quarantine the video-watch component should should a release button beneath the videos for mods. I decided to leave this out for now. The mod can unblacklist like normal when in that view.

Second, I'm allowing both unblacklist and publish notifications to be sent to user in the case where auto-blacklist is removed and the video is actually published - i.e. not wait for transcoding, private or scheduled update. If it is not actually published for one of those reasons then only the unblacklist notification is sent. I don't think the duplicate notifications would be an issue but if they are it can be changed later.

Peertube-auto-blacklist-unblacklist-notifications

The user-notification comment tests were failing in both develop and this branch. So I just skipped them for my testing but all other user-notification tests passed.

Chocobozzz:
Thank you very much @joshmorel. Merging and I'll try to refactor videos list components.

PR #1642, opened 2019-02-17
check email enabled for requiresEmailVer config

### LINKED ISSUE #993: Warn in case of error in email middleware

joshmorel:
Instead of #1471

Resolves #993

> Chocobozzz: I think we should prevent peertube from starting if it is incorrectly configured. For example, in the `checkConfig` function: https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/checker-after-init.ts#L34
> Nutomic: But you can also change these settings in the web interface so there needs to be a warning/error as well.
> Chocobozzz: If both are implemented yep remove the warning above

@Chocobozzz you already implemented the 1st so this is just for the 2nd. I only check if email is enabled for `requiresEmailVerification` since you had `contactForm` as warning only.

I was considering some logic on the client to disable the checkbox if email is not enabled but I feel this is good enough. Let me know if you'd like me to.

peertube-email-config-check

### DIFF
--- a/server/middlewares/validators/config.ts
+++ b/server/middlewares/validators/config.ts
@@ -2,6 +2,8 @@ import * as express from 'express'
 import { body } from 'express-validator/check'
 import { isUserNSFWPolicyValid, isUserVideoQuotaValid, isUserVideoQuotaDailyValid } from '../../helpers/custom-validators/users'
 import { logger } from '../../helpers/logger'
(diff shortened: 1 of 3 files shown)


Chocobozzz:
Thanks @joshmorel

PR #1697, opened 2019-02-26
make the server tools a yarn workspace

rigelk:
It allows to define extra production dependencies necessary to make CLI tools work in `--production` while keeping a separation of concerns.

Long term, it allows to publish a standalone script on npmjs.com.

Note that workspaces can be disabled by configuring yarn with `yarn config set workspaces-experimental false`, thus reducing further the production dependencies by limiting them to the production dependencies of the root `package.json`.

### DIFF
--- a/server/tools/package.json
+++ b/server/tools/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "@peertube/remote",
+  "version": "1.0.0",
+  "private": true,
+  "dependencies": {
+    "application-config": "^1.0.1",
+    "bluebird": "^3.5.3",
+    "commander": "^2.19.0",
+    "fs-extra": "^7.0.1",
+    "lodash": "^4.17.11",
+    "prompt": "^1.0.0",
+    "request": "^2.88.0",
+    "summon-install": "^0.4.6",
+    "supertest": "^3.4.2"
+  },
+  "summon": {
+    "silent": true
+  }
+}
--- a/support/doc/tools.md
+++ b/support/doc/tools.md
@@ -30,7 +30,6 @@
 
 You need at least 512MB RAM to run the script.
 Scripts can be launched directly from a PeerTube server, or from a separate server, even a desktop PC.
-You need to follow all the following steps even if you are on a PeerTube server (including cloning the git repository in a different directory than your production installation because the scripts utilize non-production dependencies).
 
 ### Dependencies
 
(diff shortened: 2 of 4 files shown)


support/doc/tools.md
```diff
@@ -42,20 +41,12 @@ Clone the PeerTube repo to get the latest version (even if you are on your PeerT
...
-$ CLONE="$(pwd)/PeerTube"
+$ CLONE="$(pwd)/PeerTube && cd $CLONE"
 ```
 
-Run ``yarn install --pure-lockfile``
-```
-$ cd ${CLONE}
-$ yarn install --pure-lockfile
-```
+Run `yarn install --production --pure-lockfile` or `NOCLIENT=1 yarn install --production --pure-lockfile` if you wish to save space by not having dependencies for the web interface of PeerTube.
```
Chocobozzz: If you build the server you need the dev dependencies, therefore you cannot use `--production`

And do you have a documentation link regarding the `NOCLIENT` env variable?

rigelk: Ah right, I put the `--production` from the point of view of an install using a release. Let me correct that.

As for `NOCLIENT`, I introduced it in https://github.com/Chocobozzz/PeerTube/commit/590924284c571d30a6b1852296b2043f3bfa5400.

support/doc/tools.md
```diff
@@ -42,20 +41,12 @@ Clone the PeerTube repo to get the latest version (even if you are on your PeerT
...
-$ CLONE="$(pwd)/PeerTube"
+$ CLONE="$(pwd)/PeerTube && cd $CLONE"
 ```
 
-Run ``yarn install --pure-lockfile``
-```
-$ cd ${CLONE}
-$ yarn install --pure-lockfile
-```
+Run `yarn install --pure-lockfile` or `NOCLIENT=1 yarn install --pure-lockfile` if you wish to save space by not having dependencies for the web interface of PeerTube.
```
Chocobozzz: I think it's better when users do not have a choice to make and when the command takes the full width so:

 ####

Install dependencies:

```
$ NOCLIENT=1 yarn install --pure-lockfile
```

Build server:

```
$ npm run build:server
```

server/tools/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@peertube/remote",
```
Chocobozzz: @peertube/cli no?

server/tools/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "version": "1.2.1",
```
Chocobozzz: CLI has its own version (1.0.0 for example)

support/doc/tools.md
```diff
@@ -42,20 +41,12 @@ Clone the PeerTube repo to get the latest version (even if you are on your PeerT
...
-$ CLONE="$(pwd)/PeerTube"
+$ CLONE="$(pwd)/PeerTube && cd $CLONE"
 ```
 
-Run ``yarn install --pure-lockfile``
-```
-$ cd ${CLONE}
-$ yarn install --pure-lockfile
-```
+Run `yarn install --pure-lockfile` or `NOCLIENT=1 yarn install --pure-lockfile` if you wish to save space by not having dependencies for the web interface of PeerTube. Likewise if you are on a production installation that has already its server compiled, you may add the `--production` flag and skip the next command.
```
Chocobozzz: Remove `yarn install --pure-lockfile` since it's better to use the `NOCLIENT` version

rigelk: Okay.

server/tools/package.json
rigelk: Since remote tools are going to get published and not really the local tools, I felt like there needed some disambiguation here. Anyway it's not like we are publishing the tool right now, we can still change the name :slightly_smiling_face:

server/tools/package.json
rigelk: Okay.

Chocobozzz:
Closing in flavour of https://github.com/Chocobozzz/PeerTube/pull/1840

PR #1719, opened 2019-03-17
Typo for yarn version

quentinDupont:
I think you might want to say that ;)


tcitworld:
One year ago it was the latest. :)

rigelk:
Latest version is actually [1.15.2](https://github.com/yarnpkg/yarn/releases/tag/v1.15.2). We might want to reformulate so that we don't put the version but a link to the latest.

tcitworld:
That would be https://github.com/yarnpkg/yarn/releases/latest

PR #1745, opened 2019-03-29
Add the video tags restrictions to the API docs

### LINKED ISSUE #1744: Video tags restrictions aren't in the API docs
**What happened?**

When using the https://docs.joinpeertube.org/api.html#tag/Video/paths/~1videos~1upload/post enpoint I had `400` return code with a message that isn't very detailed.

I expected the limitations to be detailed in the API docs.

**Steps to reproduce:**

Try uploading a video with a tag that is longer than 30 chars, or a list of more than 5 tags, and you'll received a `400` with a message looking like the following:

```json
{"errors":
  {"tags":
    {"location": "body",
     "param": "tags",
     "value": ["un très long tag qui ne passera pas parce qu'il fait plus de 30 caractères"],
     "msg": "Should have correct tags"
    }
  }
}
```

**Additional information**
* PeerTube version: 1.2.1

magopian:
Fixes #1744

### DIFF
--- a/support/doc/api/openapi.yaml
+++ b/support/doc/api/openapi.yaml
@@ -842,7 +842,7 @@ paths:
                   description: Video name
                   type: string
                 tags:
-                  description: Video tags
+                  description: Video tags (maximum 5 tags each between 2 and 30 characters)
                   type: array
                   items:
                     type: string


support/doc/api/openapi.yaml
```diff
@@ -842,7 +842,7 @@ paths:
                   description: Video name
                   type: string
                 tags:
-                  description: Video tags
+                  description: Video tags (maximum 5 tags each between 2 and 30 characaters)
```
yohanboniface: charac**a**ters? ;)

Chocobozzz:
Thanks @magopian & @yohanboniface

Issue #1737, opened 2019-03-27
Client Server Service loadConfig Timing Issue when Using Specific URL on First Page Load


own3mall:
1.  Clear browser cache.
2.  Visit /signup page (without visiting any other page) using direct URL (https://myinstance/signup)
3.  loadConfig API XHR is made, but it doesn't complete in time before checks such as `return this.serverService.getConfig().signup.requiresEmailVerification` are made.

Default config options are returned on first page load because loadConfig hasn't completed, and the settings have not yet been cached in localStorage.  Thus, the page doesn't respect your settings since the database configuration settings weren't loaded in time.

Not sure how to fix this, but you can't assume that every user is going to hit the home page of https://myinstance/ to save the config settings in local storage.  This could throw off functionality on many pages.

own3mall:
The way I handle this in PHP is that global settings are loaded from a MySQL database in a blocking synchronous call which get stored in a session variable as cache.  The settings are then loaded from the session cache (server side) for a while, but every 5th request or so, I reload settings from the database and override what is stored in the session cache.  This helps prevent users from continuing to use invalid setting values if they are changed by an admin while their session is active. 

In PeerTube, the `loadConfig` call needs to complete before the application should continue doing anything since many pages rely on these settings and can render differently depending on the setting.  However, I can't find a way to make the http.get request synchronous in angular because it goes against what angular stands for, and this is why frameworks like angular suck.  They're not flexible for every scenario.

Global server settings really need to be handled on the server side... not the client side.

own3mall:
Ok, I fixed this in my branch:

Key commits:

https://github.com/own3mall/PeerTube/commit/8834eb3a5999e94efabeacd23867335ad0de7763
https://github.com/own3mall/PeerTube/commit/6de02a05bea600e5da2a10f7f708e1c6e54c5da4

Range of commits for 3/29/2019:

https://github.com/own3mall/PeerTube/commits/develop

I think this piece should be integrated in your branch ASAP and will prevent future problems like this.  Besides, in my branch, it is imperative for my implementation of Recaptcha.

Chocobozzz:
Fixed by https://github.com/Chocobozzz/PeerTube/commit/f3ae606caf74c8a173ce44bf3f913f1488f97d7f

PR #1749, opened 2019-03-31
Load Config Before Doing Anything Else

own3mall:
Fix for https://github.com/Chocobozzz/PeerTube/issues/1737

### DIFF
--- a/client/src/app/app.module.ts
+++ b/client/src/app/app.module.ts
@@ -1,4 +1,4 @@
-import { LOCALE_ID, NgModule, TRANSLATIONS, TRANSLATIONS_FORMAT } from '@angular/core'
+import { LOCALE_ID, NgModule, TRANSLATIONS, TRANSLATIONS_FORMAT, APP_INITIALIZER } from '@angular/core'
 import { BrowserModule } from '@angular/platform-browser'
 import { ServerService } from '@app/core'
 import { ResetPasswordModule } from '@app/reset-password'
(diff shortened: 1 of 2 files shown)


Chocobozzz:
It's weird `loadConfig` is not finished in the `signup` function :thinking: It should be fast.

own3mall:
You can't ensure timing on async operations, and if you look at the link to stackoverflow I included in the source code, you'll see it's a common problem and that the way I did it is the correct way to handle loading the config.  

https://stackoverflow.com/questions/39033835/angularjs2-preload-server-configuration-before-the-application-starts

Chocobozzz:
Then prefer using `switchMap` [here](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/signup/signup.component.ts#L53).

This way you don't block the app until the configuration is loaded.

own3mall:
It's not just the signup page though.  It could be any endpoint or page.  When it comes to loading the settings, it really should load the config first before doing anything else.  Plus, you said the call should be quick, and it is.  And, after this call is made, it is just referenced from localStorage, so I really don't see it being a bad thing.

Chocobozzz:
We'll wait the config in these pages then. I don't want the app to wait 1 endpoint if this is not required for most pages.

own3mall:
That's the whole point.  It could happen on any page, and there's no guarantee the load config call will be completed in time with the way it's currently setup in the main branch.  Up to you if you want to leave it completely broken in the main branch :-1: .  It's fixed in mine though.  :+1:

Chocobozzz:
Fixed in https://github.com/Chocobozzz/PeerTube/commit/f3ae606caf74c8a173ce44bf3f913f1488f97d7f

PR #1756, opened 2019-04-04
Add /accounts/:username/ratings endpoint

yohanboniface:
Sorry, I'm not a TypeScript user (and far from being a JS expert), so I hope I've done it correctly :s

Thanks for reviewing :)


server/controllers/api/users/me.ts
```diff
@@ -72,6 +72,11 @@ meRouter.get('/me/videos/:videoId/rating',
   asyncMiddleware(getUserVideoRating)
 )
 
+meRouter.get('/me/videos/ratings',
```
Chocobozzz: Prefer to use the account endpoint instead https://github.com/Chocobozzz/PeerTube/blob/develop/server/controllers/api/accounts.ts#L37, because in the future some users may want to show what they liked.

server/controllers/api/users/me.ts
```diff
@@ -72,6 +72,11 @@ meRouter.get('/me/videos/:videoId/rating',
...
+  authenticate,
+  asyncMiddleware(getUserVideoRatings)
```
Chocobozzz: Please add pagination, sort and a rate type filter

server/models/account/account-video-rate.ts
```diff
@@ -88,6 +89,16 @@ export class AccountVideoRateModel extends Model<AccountVideoRateModel> {
     return AccountVideoRateModel.findOne(options)
   }
 
+  static listForAccount (accountId: number) {
+    const query = {
+      where: {
+        accountId
+      }
```
Chocobozzz: Please join on videos like https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/video/video-blacklist.ts#L70

server/models/account/account-video-rate.ts
```diff
@@ -88,6 +89,16 @@ export class AccountVideoRateModel extends Model<AccountVideoRateModel> {
     return AccountVideoRateModel.findOne(options)
   }
 
+  static listForAccount (accountId: number) {
```
Chocobozzz: `listByAccountForApi`

server/models/account/account-video-rate.ts
```diff
@@ -185,4 +196,11 @@ export class AccountVideoRateModel extends Model<AccountVideoRateModel> {
       else if (type === 'dislike') await VideoModel.increment({ dislikes: -deleted }, options)
     })
   }
+
+  toFormattedJSON (): UserVideoRate {
+    return {
+      videoId: this.videoId,
```
Chocobozzz: `this.Video.toFormattedJSON()` (you'll have to create an `AccountVideoRate` type)

server/controllers/api/users/me.ts
yohanboniface: I'm OK to make this change, but just to be sure: isn't playlists the way to share a set of videos to others? And isn't exposing my likes in the API a bit tricky on the privacy topic?

Chocobozzz: > I'm OK to make this change, but just to be sure: isn't playlists the way to share a set of videos to others?

Yes you're right. But I still prefer to use the account endpoint, just in case.

> And isn't exposing my likes in the API a bit tricky on the privacy topic?

By default this endpoint should be only accessible by the authenticated user (add a validator middleware)

yohanboniface: >  by the authenticated user

Just to be sure: you mean the account owner, not any authenticated user?

Chocobozzz: Yep!

buoyantair:
Wait, what does ratings for username mean?

server/middlewares/validators/users.ts
```diff
@@ -317,6 +318,21 @@ const userAutocompleteValidator = [
   param('search').isString().not().isEmpty().withMessage('Should have a search parameter')
 ]
 
+const ensureUserAccountValidator = [
```
Chocobozzz: `ensureAuthUserOwnsAccountValidator`

server/middlewares/validators/users.ts
```diff
@@ -317,6 +318,21 @@ const userAutocompleteValidator = [
...
+  async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    const user = res.locals.oauth.token.User
+    const userAccount = await AccountModel.load(user.Account.id)
```
Chocobozzz: You already have the account in `res.locals.account` so just check `res.locals.account.id === user.Account.id` :)

Chocobozzz:
Thanks @yohanboniface

PR #1759, opened 2019-04-05
Add the 'email' entry in the default.yaml config file

### LINKED ISSUE #1758: Setting the SMTP server in the configuration isn't enough to enabled the emails
**What happened?**

I was having the message "We are sorry, you cannot recover you password because your instance administrator did not configure the PeerTube email system." when asking for a password reset.

Setting up the `SMTP` in the configuration (and having the confirmation from the logs that I was `Successfully connected to SMTP server.`) didn't do the trick.

I thought setting the `SMTP` configuration was enough, and didn't find any reference to the following configuration that is needed:

```
email:
  enabled: true
```

**Steps to reproduce:**

Do not set the above configuration, and ask for a password reset.

**Additional information**
* PeerTube version: v1.2.1...3a0fb65c

magopian:
Fixes #1758

I'm not sure where this could be documented, but at least having it in the `config/default.yaml` file may give a hint that this configuration is needed to enable the emails.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -43,6 +43,9 @@ smtp:
   ca_file: null # Used for self signed certificates
   from_address: 'admin@example.com'
 
+email:
+  enabled: false # Set to `true` if you want to enable emails (to reset a password for example)
+
 # From the project root directory
 storage:
   tmp: 'storage/tmp/' # Used to download data (imports etc), store uploaded files before processing...


magopian:
I have no idea why the travis build is failing :/

Chocobozzz:
We don't use this configuration in peertube. https://github.com/Chocobozzz/PeerTube/issues/1758 bug comes from something else.

magopian:
I don't understand: I grepped on the `We are sorry, you cannot recover you password because your instance administrator did not configure the PeerTube email system.
` error message and found https://github.com/Chocobozzz/PeerTube/blob/32157c0892ed000d15ad8fbc89e90fd005a5f4e8/client/src/app/login/login.component.html#L64-L66

Grepping on `isEmailDisabled` led me to https://github.com/Chocobozzz/PeerTube/blob/32157c0892ed000d15ad8fbc89e90fd005a5f4e8/client/src/app/login/login.component.ts#L46-L48, which led me to think this was the missing configuration.

Did I miss something?

Adding this configuration indeed does remove this message.

Chocobozzz:
Oh okay. This key is generated on server side: https://github.com/Chocobozzz/PeerTube/blob/develop/server/controllers/api/config.ts#L69

magopian:
ah, i understand! Thanks a lot for the pointer!

PR #1783, opened 2019-04-24
Add argument to dockerfile to pass options to npm run build

NaPs:
Allow to pass extra options to the npm run build eg: --light or --light-fr to not build all client languages (speed up build time if i18n is not required)


Chocobozzz:
Thanks @NaPs

PR #1785, opened 2019-04-24
Dont transcode videos when it is not required

Nutomic:
Haven't tested it yet, but this is the idea.

### DIFF
--- a/server/helpers/ffmpeg-utils.ts
+++ b/server/helpers/ffmpeg-utils.ts
@@ -173,7 +173,7 @@ function transcode (options: TranscodeOptions) {
   })
 }
 
-async function canDoQuickTranscode (path: string) {
+async function canDoQuickTranscode (path: string): Promise<boolean> {
   // NOTE: This could be optimized by running ffprobe only once (but it runs fast anyway)
   const videoStream = await getVideoStreamFromFile(path)
   const parsedAudio = await audio.get(path)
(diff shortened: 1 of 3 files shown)


Nutomic:
Okay it is working now in my tests. Please have a look at it. I will add tests for the canDoQuickTranscode() function if the existing code makes sense.

@Chocobozzz I am a bit confused what the purpose of optimizeVideofile() and transcodeOriginalVideofile() is. Apparently one is used when the resolution already fits, and the other is used if the video needs to resized to another resolution. But I dont see why you cant handle that in the same function. And the function names are not clear at all.

Chocobozzz:
`optimizeVideofile` is about optimize the original file (replace it), and `transcodeOriginalVideofile` create smaller resolutions. You can change their name/refactor if you want.

Nutomic:
I couldnt really think of any better function names, but I added comments to explain this. Is there anything else needed before merging?

Chocobozzz:
Thanks @Nutomic

raphgilles:
I don't know how you have do that, but I have a similar suggestion.

Don't transcode videos when it's not required, of course, and when we use yt-download to download all formats used by the instance and don't transcode the videos downloaded.

Nutomic:
@raphgilles you should open an issue for this (or maybe there already is one)

PR #1787, opened 2019-04-25
Allow to control RATES_LIMIT from configuration

yohanboniface:
My scenario is that I'm running a migration from a previous server to PeerTube, and I'm creating users and uploading video for those users, so I need to increase a lot this limit while doing the migration.

Thanks for reviewing :)

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -8,17 +8,16 @@ webserver:
   https: false
   hostname: 'localhost'
   port: 9000
-  rates_limit:
-    login:
-      # For 5 minutes (in ms).
-      window_ms: 300000
-      # 15 attemps
-      max: 150
-    ask_send_email:
-      # For 5 minutes (in ms).
-      window_ms: 300000
-      # 3 attemps
-      max: 3
+
+rates_limit:
+  login:
+    # 15 attempts in 5 min
+    window: 5 minutes
+    max: 15
+  ask_send_email:
+    # 3 attempts in 5 min
+    window: 5 minutes
+    max: 3
 
 # Proxies to trust to get real client IP
 # If you run PeerTube just behind a local proxy (nginx), keep 'loopback'
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -8,6 +8,17 @@ webserver:
   https: false
   hostname: 'localhost'
   port: 9000
+  rates_limit:
+    login:
+      # For 5 minutes (in ms).
+      window_ms: 300000
+      # 15 attemps
+      max: 150
+    ask_send_email:
+      # For 5 minutes (in ms).
+      window_ms: 300000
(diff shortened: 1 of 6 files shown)


config/default.yaml
```diff
@@ -8,6 +8,17 @@ webserver:
   https: false
   hostname: 'localhost'
   port: 9000
+  rates_limit:
```
Chocobozzz: Create a dedicated section for `rates_limit` instead of adding it to the `webserver` section

server/initializers/constants.ts
```diff
@@ -441,12 +441,12 @@ let CONSTRAINTS_FIELDS = {
 
 const RATES_LIMIT = {
   LOGIN: {
-    WINDOW_MS: 5 * 60 * 1000, // 5 minutes
-    MAX: 15 // 15 attempts
+    WINDOW_MS: config.get<number>('webserver.rates_limit.login.window_ms'),
```
Chocobozzz: use `parseDuration` so you can set `5 minutes` or `1 hour` in configuration

config/default.yaml
```diff
@@ -8,6 +8,17 @@ webserver:
...
+    login:
+      # For 5 minutes (in ms).
+      window_ms: 300000
+      # 15 attemps
```
Chocobozzz: 15 or 150?

server/initializers/constants.ts
yohanboniface: Should I then rename the config property in `window`, given the value will not be expressed in `ms` anymore?

config/default.yaml
yohanboniface: Oops, forgot my own setting used while testing :(

server/initializers/constants.ts
Chocobozzz: yep

config/default.yaml
```diff
@@ -9,6 +9,16 @@ webserver:
   hostname: 'localhost'
   port: 9000
 
+rates_limit:
+  login:
+    window: 5 min
```
Chocobozzz: Does `min` works? :thinking:

yohanboniface:
Hold on, I've been hit by https://github.com/Chocobozzz/PeerTube/commit/6dd9de95dfa39bd5c1faed00d1dbd52cd112bae0#diff-2c0718e420302bfd05b677790eb9eb57 anyway

config/default.yaml
yohanboniface: Humm, it did not complain… but given it was complaining for other things I'm unsure :p

yohanboniface:
@Chocobozzz I think I'm back on Earth (after many push --force :p )

Chocobozzz:
Thanks

Issue #912, opened 2018-08-08
Translate emails


Chocobozzz:
For now, emails are only in english.

Would be nice to translate them.

yohanboniface:
@Chocobozzz can you elaborate a bit on what needs to be done?

We may join the effort, as we want custom emails (at least in French and with our own UI URLs and a custom instance name in the signature).

Thanks :)

cc @vinyll @magopian

yohanboniface:
I guess we first need #1732 ?

lucacalcaterra:
At this time is this implemented on current relase ? I can translate the pug template, but lost all in next updates.

rigelk:
@lucacalcaterra the issue is not closed so no, it's not

Chocobozzz:
Implemented by https://github.com/Chocobozzz/PeerTube/commit/d6e4dac0322314888f4308892666d40139fc4051

We create a new `server-internal` component in weblate so contributors can help us to translate emails and other potential server strings: https://weblate.framasoft.org/projects/peertube/server-internal/#translations

PR #1789, opened 2019-04-26
Allow to control emails prefix and signature with config

yohanboniface:
Relates to #912 

Thanks for reviewing :)

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -53,9 +53,11 @@ smtp:
   ca_file: null # Used for self signed certificates
   from_address: 'admin@example.com'
 
-emails:
-  signature: "PeerTube"
-  prefix: "[PeerTube]"
+email:
+  body:
+    signature: "PeerTube"
+  object:
+    prefix: "[PeerTube]"
 
 # From the project root directory
 storage:
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -53,6 +53,10 @@ smtp:
   ca_file: null # Used for self signed certificates
   from_address: 'admin@example.com'
 
+emails:
+  signature: "PeerTube"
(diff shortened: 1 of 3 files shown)


Nutomic:
It would probably make more sense to allow custom email templates. If you go down this path of single variables for elements in the emails, you will end up with dozens of variables. That will be completely unmaintainable.

yohanboniface:
I'd agree in general, but those are two elements that are used in each and every email, so imho it makes sense to have them in a DRY configuration place anyway.

Plus, allowing templates is another story than this one, here we have something done that allows more branded emails, vs something that is to be done, and it's not that simple.

Chocobozzz:
I agree with both opinions. Having email templating would be interesting, but it's not incompatible with signature and object prefix variables.

config/default.yaml
```diff
@@ -53,6 +53,10 @@ smtp:
   ca_file: null # Used for self signed certificates
   from_address: 'admin@example.com'
 
+emails:
+  signature: "PeerTube"
```
Chocobozzz: Maybe

```
body:
  signature:
```

config/default.yaml
```diff
@@ -53,6 +53,10 @@ smtp:
...
+  prefix: "[PeerTube]"
```
Chocobozzz: And 

```
object:
  prefix: 
```

config/default.yaml
```diff
@@ -53,6 +53,10 @@ smtp:
   ca_file: null # Used for self signed certificates
   from_address: 'admin@example.com'
 
+emails:
```
Chocobozzz: `email` for consistency

Chocobozzz:
Thanks @yohanboniface

PR #1791, opened 2019-04-30
Fix sort order for Greek language

gunchleoc:
I could not find the Greek language wile uploading some subtitles and I only found out that it existed when I translated the language list.

I suggest that we drop the "Modern" prefix - we don't call English "Modern English" to distinguish it from "Middle English" either ;)

### DIFF
--- a/client/src/locale/source/iso639_en_US.xml
+++ b/client/src/locale/source/iso639_en_US.xml
@@ -122,7 +122,7 @@
         <source>Dzongkha</source>
       </trans-unit>
       <trans-unit id="Modern Greek (1453-)">
-        <source>Modern Greek (1453-)</source>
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks @gunchleoc 

I fixed this myself because we needed to update the script that generates this file: https://github.com/Chocobozzz/PeerTube/commit/d8cfccfbc4895c3664aad7174d1570c309982857

gunchleoc:
Great, thanks for taking care of it!

I also have some nds (Low German) captions that I can't upload - what would it take to get the language in the list? Just edit the 2 files, or is there some other place in the code that needs updating?

Chocobozzz:
You can add it to the `additionalLanguages` object

PR #1799, opened 2019-05-04
Allow to use PEERTUBE_NEW_PASSWORD env var to change user password

ldidry:
It makes the reset-password suitable for automated commands (like in an installation script)


Chocobozzz:
`echo $PASSWORD | npm run reset-password -- -u my_user` does not address the need?

ldidry:
Mmh, yes, it should work, indeed. I proposed that since Salt allow to set env vars when using cmd.run, and I find that writing prettier than using a pipe. You can close the PR if you want.

Issue #147, opened 2017-12-11
Remote storage mounted via FUSE, Amazon S3, SFTP and more


ghost:
UPDATED by @rigelk:

There are two ways to see remote storage:

1. the instance admin adds remote storage via the admin interface. Arguably he can already do that by hand if he has access to the hosted machine, but that's another story.
2. the registered user adds their credentials to remote storage via their user interface. This way they can increase their storage capacity independently of their own quota.

For those who have followed the discussions about IPFS, the problem is not exactly the same. Here it is simpler instance, since we do not need to consider each video/group of videos nor its integration with the client to really benefit from it. Here we just consider the remote storage as a bucket sitting at the disposal of the PeerTube server.

---

Libraries we should look at:
- [rclone](https://github.com/ncw/rclone/) (Amazon Drive, Amazon S3 / Dreamhost / Ceph / Minio / Wasabi, Backblaze B2, Box, Dropbox, FTP, Google Cloud Storage, Google Drive, HTTP, Hubic, Mega, Microsoft Azure Blob Storage, Microsoft OneDrive, OpenDrive, Openstack Swift / Rackspace cloud files / Memset Memstore / OVH / Oracle, Cloud Storage, pCloud, QingStor, SFTP, Webdav / Owncloud / Nextcloud, Yandex Disk, local filesystem)
- [goofys](https://github.com/kahing/goofys/) (Amazon S3, Amplidata, DreamObjects (Ceph), EMC Atmos, Google Cloud Storage, OpenStack Swift, S3Proxy, Minio (limited))
- [pkgcloud](http://github.com/pkgcloud/pkgcloud) (Amazon, Azure, Google, HP, Openstack, Rackspace)
- [node-ftps](https://github.com/Atinux/node-ftps) (FTP, FTPS, SFTP)

We should also consider adding a caching system, since most of these backends are slow by nature (a.k.a. cold storage):
- [catfs](https://github.com/kahing/catfs)
- [pcachefs](https://github.com/ibizaman/pcachefs)

Now, I would be cautious about pricing models of some services supported. Disabling them by default, making a huge banner saying "here be dragons" before activating them, would be the least we can do.

rigelk:
Looking for libraries that do this, I have seen [pkgcloud](http://github.com/pkgcloud/pkgcloud) that seems to provide that kind of abstraction − but it doesn't feature as much backends supported as [rclone](https://github.com/ncw/rclone/) nor as thriving a community.

[...]

PR #1810, opened 2019-05-13
Transcode in STORAGE.TMP_DIR for s3fs compatibility (#147)

libertysoft3:
This PR allows trancoding to succeed when `storage.videos` is a [s3fs-fuse](https://github.com/s3fs-fuse/s3fs-fuse) mounted directory. It changes the transcode procedure(s) to: 

1. Transcode video to `storage.tmp_dir` (local disk)
1. Read/compute metadata like fps
1. Move video to `storage.videos` (local disk or s3fs) as expected

By transcoding to local disk and reading metadata from there, object storage related transcoding errors like 'moov atom not found' and 'ffprobe exited with code 1' can be avoided. These issues seem to exist because object storage is 'eventually consistent'- metadata cannot be read immediately after creating a video. This commit assumes that `storage.tmp_dir` will remain a local disk directory.

Here's how I mounted my `storage.videos` directory:

    $ s3fs your-space-name /var/www/peertube/storage/videos -o url=https://region.digitaloceanspaces.com -o allow_other -o use_path_request_style -o uid=1000 -o gid=1000

### DIFF
--- a/server/lib/video-transcoding.ts
+++ b/server/lib/video-transcoding.ts
@@ -13,11 +13,12 @@ import { CONFIG } from '../initializers/config'
 
 async function optimizeVideofile (video: VideoModel, inputVideoFileArg?: VideoFileModel) {
   const videosDirectory = CONFIG.STORAGE.VIDEOS_DIR
+  const transcodeDirectory = CONFIG.STORAGE.TMP_DIR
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks @libertysoft3

PR #1813, opened 2019-05-14
fix: Close modal after clicking on download

LeoMouyna:
Use the NgActiveModal method to close it on a click by adding a specific
modalActive object and use it on download function.

See issue: Chocobozzz#1436


Chocobozzz:
Thanks @LeoMouyna

PR #1830, opened 2019-05-21
Fix docker-compose setup so Peertube sees correct IP

Nutomic:
Peertube was not getting the correct client IP on my instance, so here is what I did to fix it.

According to [this comment](https://github.com/docker/compose/issues/4336#issuecomment-273341619), docker-compose essentially uses a random subnet by default, so we have to specify that in docker-compose.yml. The same subnet has to be added to .env as trust_proxy.

Slightly unrelated, but can you tell me a good place to link my [peertube-compose](https://radical.town/@felix/102106907762520601) project? @rigelk suggested I add it to the wiki here, but then I have to add a new page or completely rewrite the "Ansible playbook" page, and I dont know how that should look.

### DIFF
--- a/support/docker/production/.env
+++ b/support/docker/production/.env
@@ -5,8 +5,7 @@ PEERTUBE_WEBSERVER_PORT=443
 PEERTUBE_WEBSERVER_HTTPS=true
 # If you need more than one IP as trust_proxy
 # pass them as a comma separated array:
-PEERTUBE_TRUST_PROXY=["127.0.0.1"]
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Thanks @Nutomic 

> Slightly unrelated, but can you tell me a good place to link my peertube-compose project?

Yes please create a new page in the wiki. Something like "Peertube setup with Ansible and Docker-Compose".

Nutomic:
Alright, I added it here: https://github.com/Chocobozzz/PeerTube/wiki/Peertube-setup-with-Ansible-and-Docker-Compose

PR #1839, opened 2019-05-24
fix: fixing build failure.

LeoMouyna:
Rename reload to unload function.

No issue


Chocobozzz:
Sorry @LeoMouyna, should be fixed in develop

PR #1840, opened 2019-05-24
Move CLI dependencies in their own package.json

Chocobozzz:
Supersedes https://github.com/Chocobozzz/PeerTube/pull/1697

We just don't use workspaces, so the CLI tools dependencies are not installed by default. Instead, the user has to run a special command to install CLI dependencies and build the TS files.


Chocobozzz:
@rigelk Good idea :+1:

Issue #1850, opened 2019-05-27
Features list shows transcoding 'on' while it is 'off' in config


LiPeK:
In the transcoding configuration section I have disabled all the available resolutions and the 'Transcoding enabled' checkbox.
Now the 'Features found on this instance' section on about page shows the 'Transcode your videos in multiple resolutions' as enabled.

It seems that the problem is located here:
https://github.com/Chocobozzz/PeerTube/blob/46b2cec73c3ee10fe9f751a301e4afff0a56b371/server/controllers/api/config.ts#L54

The workaround:
Enable all resolutions and disable transcoding.

PR #1851, opened 2019-05-27
Update config.ts - issue 1850

LiPeK:
This should fix https://github.com/Chocobozzz/PeerTube/issues/1850


Chocobozzz:
Thanks @LiPeK!

Issue #1775, opened 2019-04-18
Logs rotation


Grayswandir:
Hello,
Peertube version: v1.2.1

Logs rotation seems to be integrated to the application PT with this rules:
- If the logs file reach 12 Mo -> start new file.
- New logs filename is peertubeXX+1.log

`-rw-r--r-- 1 peertube peertube 12582987 avril 16 14:31 peertube.log
-rw-r--r-- 1 peertube peertube 12582990 avril 17 16:34 peertube1.log
-rw-r--r-- 1 peertube peertube 12583125 avril 18 11:34 peertube2.log
-rw-r--r-- 1 peertube peertube  1965436 avril 18 14:23 peertube3.log
`
Do you think that it's possible to have a config parameter to let logrotate doing the jobs instead please?

NassimBounouas:
Hello,

If no one is working on this issue, I'm interested to do it.

I looked on the way Peertube is logging and I found that the winston library is used and configured in [this file](https://github.com/Chocobozzz/PeerTube/blob/76062d9f96e06a23a2efc8a727ea9c5394d21466/server/helpers/logger.ts#L55).

The log limitations seems not to be mandatory according to [this documentation](https://github.com/winstonjs/winston/blob/master/docs/transports.md#file-transport) and [this example](https://github.com/winstonjs/winston/blob/master/examples/quick-start.js#L19).

The log files destination is configured using the `config` package and the [default configuration](https://github.com/Chocobozzz/PeerTube/blob/develop/config/default.yaml#L69) is pointing to a dedicated folder in the project root directory.

I reached the conclusion that adding a "log rotation" parameter in the configuration, to remove log rotations options from the winston file transport could be a solution. This parameter would be set as true by default letting the user able to set it to false and deactivate the log rotation. Letting the user deactivating this mecanism makes him responsible of the logrotate configuration [(for example this way)](https://www.digitalocean.com/community/tutorials/how-to-manage-logfiles-with-logrotate-on-ubuntu-16-04) and of the peer tube log folder configuration [there](https://github.com/Chocobozzz/PeerTube/blob/develop/config/default.yaml#L69).

Please, can someone confirm me that I understood the issue and the viability of my solution (@Chocobozzz / @rigelk )?

Thank you !

rigelk:
[...]

PR #1881, opened 2019-06-08
Feature/logrotation

NassimBounouas:
This pull concerns #1775 and let the user deactivate winston log library rotation.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -75,6 +75,7 @@ storage:
 
 log:
   level: 'info' # debug/info/warning/error
+  rotation: false
 
 search:
   # Add ability to fetch remote videos/actors by their URI, that may not be federated with your instance
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -75,7 +75,8 @@ storage:
 
 log:
   level: 'info' # debug/info/warning/error
-  rotation: true # Enabled by default, if disabled make sure that 'storage.logs' is pointing to a folder handled by logrotate
+  rotation:
+    enabled : true # Enabled by default, if disabled make sure that 'storage.logs' is pointing to a folder handled by logrotate
 
 search:
   # Add ability to fetch remote videos/actors by their URI, that may not be federated with your instance
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -75,7 +75,7 @@ storage:
 
 log:
   level: 'info' # debug/info/warning/error
-  rotation: false
+  rotation: true # Enabled by default, if disabled make sure that 'storage.logs' is pointing to a folder handled by logrotate
 
 search:
   # Add ability to fetch remote videos/actors by their URI, that may not be federated with your instance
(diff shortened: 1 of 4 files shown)


server/helpers/logger.ts
```diff
@@ -45,23 +45,28 @@ const labelFormatter = winston.format.label({
   label
 })
 
+const fileLoggerOptions = {
```
Chocobozzz: `const fileLoggerOptions: FileTransportOptions = {`

server/helpers/logger.ts
```diff
@@ -45,23 +45,28 @@ const labelFormatter = winston.format.label({
...
+  filename: path.join(CONFIG.STORAGE.LOG_DIR, 'peertube.log'),
+  handleExceptions: true,
+  format: winston.format.combine(
+    winston.format.timestamp(),
+    jsonLoggerFormat
+  )
+}
+
+if (CONFIG.LOG.ROTATION) {
+  fileLoggerOptions['maxsize'] = 1024 * 1024 * 12
```
Chocobozzz: With the change above, now you can use

```fileLoggerOptions.maxsize```

File: config/default.yaml [hunk 3]
```diff
@@ -75,6 +75,7 @@ storage:
 
 log:
   level: 'info' # debug/info/warning/error
+  rotation: true # Enabled by default, if disabled make sure that 'storage.logs' is pointing to a folder handled by logrotate
```
Chocobozzz: Prefer:

```
  rotation:
    enabled: true
```

So we can add other parameters in the future (file max size etc)

Please update https://github.com/Chocobozzz/PeerTube/blob/develop/config/production.yaml.example too

server/helpers/logger.ts
NassimBounouas: Thank you @Chocobozzz, I searched a way of setting fields more properly and I didn't find this type. It's really better.

config/default.yaml
NassimBounouas: Thank you, I did it ! By doing this, it will be easier to complete [this request](https://github.com/Chocobozzz/PeerTube/issues/1775#issuecomment-498587347).

server/initializers/config.ts
```diff
@@ -83,7 +83,8 @@ const CONFIG = {
   },
   TRUST_PROXY: config.get<string[]>('trust_proxy'),
   LOG: {
-    LEVEL: config.get<string>('log.level')
+    LEVEL: config.get<string>('log.level'),
+    ROTATION: config.get<boolean>('log.rotation')
```
Chocobozzz: `log.rotation.enabled`

NassimBounouas: Indeed, I completely forgot this modification, thank you for your look-out. I apologize.

Chocobozzz:
Thanks @NassimBounouas

PR #1901, opened 2019-06-13
Fix #1870

### LINKED ISSUE #1870: Change wording of "Video uploads require manual validation by moderators	"
All other fields are X when users are restricted and checked when users are not.  However, having manual approval is a restriction.

Currently:
image

Ideally (with same config):
image

Jorropo:
I think placing a red cross here isn't justified, that a judgment on how an instance should works and changing the sentence is probably more tolerant.
Should fix #1870

### DIFF
--- a/client/src/app/shared/instance/instance-features-table.component.ts
+++ b/client/src/app/shared/instance/instance-features-table.component.ts
@@ -50,7 +50,9 @@ export class InstanceFeaturesTableComponent implements OnInit {
         value: config.signup.allowed
       },
       {
-        label: config.autoBlacklist.videos.ofUsers.enabled ? this.i18n('Video uploads require manual validation by moderators') : this.i18n('Uploads are published without moderator verification'),
+        label: config.autoBlacklist.videos.ofUsers.enabled ?
+          this.i18n('Video uploads require manual validation by moderators') :
+          this.i18n('Uploads are published without moderator verification'),
         value: true
       },
       {
--- a/client/src/app/shared/instance/instance-features-table.component.ts
+++ b/client/src/app/shared/instance/instance-features-table.component.ts
@@ -50,8 +50,8 @@ export class InstanceFeaturesTableComponent implements OnInit {
         value: config.signup.allowed
       },
       {
-        label: this.i18n('Video uploads require manual validation by moderators'),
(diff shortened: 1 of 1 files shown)


client/src/app/shared/instance/instance-features-table.component.ts
```diff
@@ -50,8 +50,10 @@ export class InstanceFeaturesTableComponent implements OnInit {
         value: config.signup.allowed
       },
       {
-        label: this.i18n('Video uploads require manual validation by moderators'),
-        value: config.autoBlacklist.videos.ofUsers.enabled
+        label: config.autoBlacklist.videos.ofUsers.enabled ?
+          this.i18n('Video uploads require manual validation by moderators') :
+          this.i18n('Uploads are published without moderator verification'),
+        value: true
```
rigelk: please use the variable instead

Jorropo: ? the whole goal of this pr is to not do any judgment about how you manage your instances, so `value:` must all ways stay at `true` and the ternary operator switch the message by a message or its opposite (and if I invert the sense of the message I must also invert the value wich finish by always false are always true, I choosed always true because that more friendly).
I'm not realy sure about what you are talking, so wich variable do you talk ?

Chocobozzz: Could we imagine something like 

```
label: 'Video upload'
value: 'Requires manual validation by moderators' or 'Automatically published'
```

?

Jorropo: @Chocobozzz I've tried but, I find that quite uggly :
Capture
The space in the bottom of Video upload is way too big and the 4 wrap line for a small sentance isn't very pretty.
So I've tried to get a use of this big open blank space in the middle of the screen but that create some problems with reponsiveness.

Since my view is very subjective, are you sure you want this ?

Chocobozzz: Yes I think this is the clearest to make things understood. Just try to increase the `min-width` value in https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/shared/instance/instance-features-table.component.scss#L10

Chocobozzz:
Closing in favor of https://github.com/Chocobozzz/PeerTube/pull/2091

PR #1902, opened 2019-06-13
Update contributing to use multiple remote.

Jorropo:
Using multiple remote simplify a lot updating and pulling.


Chocobozzz:
Thanks

PR #1903, opened 2019-06-13
WIP: Fix #498

### LINKED ISSUE #498: Basic markdown formatting in comments
Would it be possible to allow some basic text formatting in comments (bold, italic, underlined) ?
I think bold would be especially useful. I don't like writing some words in capitals, making the readers think I'm screaming at them. As you know, Youtube comments are full of capital letters because it doesn't have text formatting. This could be another way to make Peertube more interesting.

Jorropo:
Fix #498 

I cant actualy make it works.
For an unknown reason `<br />` got been escaped while the same code doesn't do this in :
```
client/src/app/videos/+video-watch/video-watch.component.ts
```
It seems to be done by `avoidTruncatedTags` in :
```
client/src/app/shared/renderer/markdown.service.ts
```
but again in :
```
client/src/app/videos/+video-watch/video-watch.component.ts
```
this doesn't happend and I cant get why.

### DIFF
--- a/client/src/app/videos/+video-watch/comment/video-comment.component.ts
+++ b/client/src/app/videos/+video-watch/comment/video-comment.component.ts
@@ -4,7 +4,7 @@ import { VideoCommentThreadTree } from '../../../../../../shared/models/videos/v
 import { AuthService } from '../../../core/auth'
 import { Video } from '../../../shared/video/video.model'
 import { VideoComment } from './video-comment.model'
-import { HtmlRendererService } from '@app/shared/renderer'
+import { MarkdownService } from '@app/shared/renderer'
 
 @Component({
   selector: 'my-video-comment',
@@ -28,7 +28,7 @@ export class VideoCommentComponent implements OnInit, OnChanges {
   newParentComments: VideoComment[] = []
 
   constructor (
-    private htmlRenderer: HtmlRendererService,
+    private markdownService: MarkdownService,
     private authService: AuthService
   ) {}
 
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Because in video watch we inject the HTML:

```
<div class="video-info-description-html" [innerHTML]="videoHTMLDescription"></div>
```

If you add basic markdown in comments, please add a preview bloc when writing the comment.

Jorropo:
@Chocobozzz this is the same things here :
```html
<div class="comment-html" [innerHTML]="sanitizedCommentHTML"></div>
```
And the problem is `<br />` transform to `&lt;br /&gt;` after been returned from `await this.markdownService.textMarkdownToHTML(this.comment.text)`.
I've thinked to first make the text in `await this.htmlRenderer.toSafeHtml(this.comment.text)` but nothing change (logical because that not linked).

> If you add basic markdown in comments, please add a preview bloc when writing the comment.

Yes, that a good idea, I didn't even think of that. Would you the preview in the bottom of the writing box (the preview box should open only when you are writing), asside of the write box (not disponible for mobile), or in a preview tab (like github does) ?

Chocobozzz:
> And the problem is <br /> transform to &lt;br /&gt; after been returned from await this.markdownService.textMarkdownToHTML(this.comment.text).
I've thinked to first make the text in await this.htmlRenderer.toSafeHtml(this.comment.text) but nothing change (logical because that not linked).

Ah yes sorry. Try to create another markdown renderer that accepts html (because we already sanitized it): https://github.com/markdown-it/markdown-it#init-with-presets-and-options

> Yes, that a good idea, I didn't even think of that. Would you the preview in the bottom of the writing box (the preview box should open only when you are writing), asside of the write box (not disponible for mobile), or in a preview tab (like github does) ? (I think first one is the best)

I have no preference :)

rigelk:
@Jorropo do you plan on finishing this PR, or do you need help with it?

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/commit/1aa7543403f278ad88d64e368e77bcb3efa58dd7

PR #1907, opened 2019-06-15
Add a Dockerfile to share Peertube on ARM device with docker.

jacen92:
PeerTube become available on ARM device like rock64 (ARM64 chip).
I didn't test on Raspberry PI (not enough RAM I guess).
The build is really slow (due to multiple languages) but works pretty well on Rock64 with other containers running along it.
I will be able to share the docker image via dockerhub soon if needed with some optimizations.

Merci

### DIFF
--- a/support/docker/production/Dockerfile.arm
+++ b/support/docker/production/Dockerfile.arm
@@ -1,0 +1,65 @@
+FROM arm64v8/debian:stretch-slim
+
+RUN set -ex; \
+    if ! command -v gpg > /dev/null; then \
+      apt-get update; \
+      apt-get install -y --no-install-recommends \
+        gnupg \
+        dirmngr \
+      ; \
+      rm -rf /var/lib/apt/lists/*; \
+fi
+
+# install nodejs and npm
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get update \
+&& apt-get install -y apt-utils nano htop curl git build-essential gcc g++ make \
+&& curl -sL https://deb.nodesource.com/setup_8.x | bash - \
+&& apt-get install -yq nodejs \
+&& rm -rf /var/lib/apt/lists/ \
+&& npm install npm@latest -g && npm install yarn@latest -g
+
+# Install dependencies
+RUN apt-get update \
+    && apt-get -y install ffmpeg \
+    && rm /var/lib/apt/lists/* -fR
+
+# Add peertube user
+RUN groupadd -r peertube \
+    && useradd -r -g peertube -m peertube
+
+# grab gosu for easy step-down from root
+RUN set -eux; \
+	apt-get update; \
+	apt-get install -y gosu; \
+	rm -rf /var/lib/apt/lists/*; \
+	gosu nobody true
+
+# Install PeerTube
+WORKDIR /app
+COPY . ./
+RUN chown -R peertube:peertube /app
+
+USER peertube
+# from: https://github.com/yarnpkg/yarn/issues/5259#issuecomment-421067226
+RUN for attempt in $(seq 1 100); do yarn install --pure-lockfile --verbose && break; done
+RUN npm run build && rm -r ./node_modules ./client/node_modules
+RUN for attempt in $(seq 1 100); do yarn install --pure-lockfile --verbose --production && break; done && yarn cache clean
+
+USER root
+
+RUN mkdir /data /config
+RUN chown -R peertube:peertube /data /config
+
+ENV NODE_ENV production
+ENV NODE_CONFIG_DIR /config
+
+VOLUME /data
+VOLUME /config
+
+COPY ./support/docker/production/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
+
+# Run the application
+CMD ["npm", "start"]
+EXPOSE 9000
--- a/support/docker/production/Dockerfile.arm
+++ b/support/docker/production/Dockerfile.arm
@@ -1,5 +1,10 @@
 FROM arm64v8/debian:stretch-slim
 
+# Allow to pass extra options to the npm run build
+# eg: --light --light-fr to not build all client languages
+#     (speed up build time if i18n is not required)
+ARG NPM_RUN_BUILD_OPTS
+
 RUN set -ex; \
     if ! command -v gpg > /dev/null; then \
       apt-get update; \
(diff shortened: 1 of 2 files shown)


support/docker/production/Dockerfile.arm
```diff
@@ -0,0 +1,65 @@
...
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get update \
+&& apt-get install -y apt-utils nano htop curl git build-essential gcc g++ make \
+&& curl -sL https://deb.nodesource.com/setup_8.x | bash - \
+&& apt-get install -yq nodejs \
+&& rm -rf /var/lib/apt/lists/ \
+&& npm install npm@latest -g && npm install yarn@latest -g
+
+# Install dependencies
+RUN apt-get update \
```
Jorropo: apt-get start to be more and more depreacated, you should consider switching to aptitude (will probbably change nothing since most of apt-get are now symbolink to aptitude)
For your script that means replacing each apt-get to apt

support/docker/production/Dockerfile.arm
```diff
@@ -0,0 +1,65 @@
...
+
+# Install PeerTube
+WORKDIR /app
+COPY . ./
+RUN chown -R peertube:peertube /app
+
+USER peertube
+# from: https://github.com/yarnpkg/yarn/issues/5259#issuecomment-421067226
+RUN for attempt in $(seq 1 100); do yarn install --pure-lockfile --verbose && break; done
+RUN npm run build && rm -r ./node_modules ./client/node_modules
```
Jorropo: yarn run build ?

support/docker/production/Dockerfile.arm
```diff
@@ -0,0 +1,65 @@
...
+ENV NODE_CONFIG_DIR /config
+
+VOLUME /data
+VOLUME /config
+
+COPY ./support/docker/production/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
+
+# Run the application
+CMD ["npm", "start"]
```
Jorropo: yarn ?

support/docker/production/Dockerfile.arm
```diff
@@ -0,0 +1,65 @@
...
+fi
+
+# install nodejs and npm
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get update \
+&& apt-get install -y apt-utils nano htop curl git build-essential gcc g++ make \
+&& curl -sL https://deb.nodesource.com/setup_8.x | bash - \
+&& apt-get install -yq nodejs \
+&& rm -rf /var/lib/apt/lists/ \
+&& npm install npm@latest -g && npm install yarn@latest -g
```
Jorropo: npm ins't usefull for peertube, you can do evrythings with yarn.
It comes with nodejs package but I think its beter to follow https://yarnpkg.com/lang/fr/docs/install/#debian-stable
And only update yarn. Upgrading npm is a waste of time and space.
(that not a full change request, more a comment)

support/docker/production/Dockerfile.arm
jacen92: Should I update the Dockerfile.strech in the same time?

support/docker/production/Dockerfile.arm
jacen92: It seems I was not on the correct branch but there is this command used in the original dockerfile.

Replaced by `&& npm run build -- $NPM_RUN_BUILD_OPTS \` later.

support/docker/production/Dockerfile.arm
jacen92: Also the same in the current Dockerfile, I do not changed this part.

support/docker/production/Dockerfile.arm
jacen92: Keep that in mind for future image optimization.
This one is not perfect but I will improve it later.

Thanks

support/docker/production/Dockerfile.arm
Jorropo: I havn't tested it for your docker but aptitude is in debian way beffore stretch, so a replacement of `apt-get` to `apt` should be complete.

support/docker/production/Dockerfile.arm
Jorropo: that doesn't realy change somethings, -- means pass this to the script and yarn also handle that.

support/docker/production/Dockerfile.arm
Jorropo: Yes, that here gaining somethings like 30sec at best and probably 10mb at best, so no need to press :) .

support/docker/production/Dockerfile.arm
Jorropo: so it should be fixed in the current Dockerfile.
yarn run start and npm start does the same things but npm takes more seconds to load (that best in latest update but still not at the level of yarn (ok that only 3s to 5s but on small device like RPI it could counts way more)).

Jorropo: Edit, yarn start works, no need of run

jacen92: Ok I will test that.
Just to clarify I am working on rock64 and not raspberry.

support/docker/production/Dockerfile.arm
jacen92: Anyway I will test each of this improvement on both ARM and x86 in the coming days.

support/docker/production/Dockerfile.arm
jacen92: Done in all dockerfiles

support/docker/production/Dockerfile.arm
jacen92: Done in all dockerfiles (I keep the argument name for other automation tools)

support/docker/production/Dockerfile.arm
jacen92: Done in all dockerfiles

support/docker/production/Dockerfile.arm
jacen92: In this case npm is installed with nodejs so we will just make sure we got the latest version

jacen92:
images are available here: https://hub.docker.com/r/ngargaud/peertube/tags

Chocobozzz:
Hi @jacen92 

I'm really sorry but I don't want/have time to maintain another official Docker image for PeerTube. But please don't hesitate to create your own repository and add the installation steps in the documentation website: https://docs.joinpeertube.org/#/install-unofficial

jacen92:
Hi

I understand but it is just the same image for an other platform.
How should I add the link to my repository? I don't know how to proceed.

Nutomic:
Can we at least merge the apt-get -> apt and npm -> yarn switches for the x86 docker image? Faster startup and build time would be very useful.

jacen92:
In this case and if Chocobozzz would not maintain it we could add the dockerfile without release the image for arm, just for who are interested and with a disclaimer for example.

Nutomic:
@jacen92 @Jorropo Could either of you make a PR with only the apt-get -> apt and npm -> yarn changes in the existing Docker file? I would definitely like to see those performance improvements on my instance. Otherwise I will make the PR myself.

jacen92:
Done here: https://github.com/Chocobozzz/PeerTube/pull/1952

forzagreen:
I built versions v2.4.0 and v3.0.0 in an ARM machine, and push them to [forzagreen/peertube](https://hub.docker.com/r/forzagreen/peertube) in Docker Hub:

- `docker pull forzagreen/peertube:arm64-v3.0.0`
- `docker pull forzagreen/peertube:arm64-v2.4.0`

I had to do the same thing for postfix-relay : cf. https://github.com/wader/postfix-relay/issues/21

GitHub Actions already supports multi-architecture Docker builds: https://github.com/marketplace/actions/build-and-push-docker-images#multi-platform-image

But I think PeerTube is building Docker images in GitLab CI with kaniko, which is not _yet_ supporting multi-architecture builds. When the PR https://github.com/GoogleContainerTools/kaniko/pull/1531 gets merged and released, I'll ask ask to open this issue again.

Issue #1843, opened 2019-05-24
Add complete username on account page


valvin1:
**What happened?**
When I want to add an account on my friendica instance I need to have a complete AP username `@username@server.tld` but I can't find it on account page. I can only see "username" whitout server.tld.

**What do you expect to happen instead?**
Having near the user (in small) a complete username `@username@server.tld` or `username@server.tld`

**Steps to reproduce:**
In my case I was watching a video on [Touhoppai instance](https://peertube.touhoppai.moe/videos/watch/8a9a054a-b1ac-4658-b363-e3212bc52054)
I click on user who posted the video and i'm redirected on its [account / video page](https://peertube.touhoppai.moe/accounts/touhoppai/videos)
I can only copy username : `touhoppai` but not its complete username `@touhoppai@peertube.touhoppai.moe`

**Additional information**
* PeerTube version or URL: v1.3.0-rc2
* Browser name/version:Firefox 67

NassimBounouas:
Hello,

I guess that the complete username is not displayed when the account is displayed from its original instance. In this case, the name is truncated.

**What I found :**
- The account complete name is displayed [there](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/%2Baccounts/accounts.component.html#L10).
- This field is created [there](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/shared/account/account.model.ts#L21) and the function called is [this one](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/shared/actor/actor.model.ts#L31).

This is, I guess, the explanation for this difference between [Touhoppai profile on its instance](https://peertube.touhoppai.moe/accounts/touhoppai/videos) and [this profile on p2p legal instance](https://tube.p2p.legal/accounts/linux@peertube.mastodon.host/videos).

Is this behavior mandatory @rigelk  or can I do a PR with a modification ?

Thank you !

Chocobozzz:
Hi,

This behaviour is intended to easily differentiate between a local account and a remote account. I think we should keep this behaviour, except on the account page. What do you think?

NassimBounouas:
Hi @Chocobozzz !

[...]

PR #1913, opened 2019-06-18
Feature/completeUsernameCopy

NassimBounouas:
This pull concerns #1843 and add a button on account page to copy a complete username.


client/src/app/+accounts/accounts.component.ts
```diff
@@ -70,4 +70,18 @@ export class AccountsComponent implements OnInit, OnDestroy {
           )
     }
   }
+
+  copyText(val: string){
```
NassimBounouas: Declaration of public instance method not allowed after declaration of private instance method.

NassimBounouas: Fixed

Chocobozzz:
Thanks for the PR @NassimBounouas 

Could you use [ngx-clipboard](https://github.com/maxisam/ngx-clipboard)? Like we do in the [share modal](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/videos/%2Bvideo-watch/modal/video-share.component.html#L17)

client/src/assets/images/global/copy.html
```diff
@@ -0,0 +1,6 @@
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
Chocobozzz: This is a [softies](https://robbiepearce.com/softies/) icon?

NassimBounouas: No it's not, I will change it !

NassimBounouas:
As you suggested @Chocobozzz, I replaced the icon by the `glyphicon-copy` and the copy function has been replaced by ngx-clipboard component. If you have any other suggestion, I'll do it with pleasure.

Chocobozzz:
Thanks @NassimBounouas!

Issue #1916, opened 2019-06-19
Connection impossible with email


bristow:
**What happened?**

It seems that in version 1.3.1, you can no longer connect with an email with capital letters.
If, as an administrator, I modify this email and put it in lowercase, then we can use this email again to connect.
No problem with the ID of these accounts

**What do you expect to happen instead?**

In a previous version of PT, we could connect with an email with capital letters

**Additional information**
* PeerTube version or URL: 1.3.1
* Browser name/version: Firefox
* NodeJS version: ?

NassimBounouas:
Hi,

I have been able to reproduce this bug. I'll investigate to suggest a fix.

NassimBounouas:
On the client side, the email is converted to lower case before sending the request [(implemented there)](https://github.com/Chocobozzz/PeerTube/commit/c1521ca3d757bee91f7dfbb15b3717162bf4997d).

On the server side, sequelize-typescript is used as ORM to query the postgresql database. There I need a confirmation (maybe from @Chocobozzz ?). As I understood, the API request is received by [this function](https://github.com/Chocobozzz/PeerTube/blob/develop/server/controllers/api/users/index.ts#L158). And from there, the token generation is handled by oauth2-server [(there)](https://github.com/Chocobozzz/PeerTube/blob/develop/server/middlewares/oauth.ts#L76). The credentials verification is based on the user retrieve [there](https://github.com/Chocobozzz/PeerTube/blob/develop/server/lib/oauth-model.ts#L75).

So I guess that the problem is coming from [this query](https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/account/user.ts#L402), the lowercase email sent by the client is tested against the saved email without converting it to lowercase so the query is case sensitive.

My first suggestion is to modify the sequelize query [like this](https://github.com/NassimBounouas/PeerTube/commit/4f68f648a41464b9fa3221c8b3fa9ec7b67465f2).

I tested it locally and it seems to solve the problem but I'm really not sure. If @rigelk or @Chocobozzz can review my suggestion before I make a PR, It would be kind.

Thank you

Chocobozzz:
Unfortunately emails are (in theory) case sensitive. So I think you should revert my commit, and use `iLike` for the username on the server.

NassimBounouas:
[...]

PR #1917, opened 2019-06-19
Fix/connection with email

NassimBounouas:
This pull concerns #1916 and make login case sensitive for email / case insensitive for username.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -181,13 +181,13 @@ Then, we can create the databases (if they don't already exist):
 
 ```
 $ sudo -u postgres createuser you_username --createdb --superuser
-$ createdb -O peertube peertube_test{1,2,3,4,5,6}
+$ npm run clean:server:test
 ```
 
 Build the application and run the unit/integration tests:
 
 ```
-$ npm run build
+$ npm run build -- --light
 $ npm test
 ```
 
--- a/client/src/app/core/auth/auth.service.ts
+++ b/client/src/app/core/auth/auth.service.ts
@@ -153,7 +153,7 @@ export class AuthService {
       response_type: 'code',
       grant_type: 'password',
       scope: 'upload',
-      username: username.toLowerCase(),
+      username,
       password
     }
 
--- a/server/models/account/user.ts
+++ b/server/models/account/user.ts
@@ -380,7 +380,7 @@ export class UserModel extends Model<UserModel> {
   static loadByEmail (email: string) {
     const query = {
       where: {
-        email
+        email: { [ Op.iLike ]: email}
       }
     }
 
@@ -392,7 +392,7 @@ export class UserModel extends Model<UserModel> {
 
     const query = {
       where: {
-        [ Op.or ]: [ { username }, { email } ]
+        [ Op.or ]: [ { username }, { email: { [ Op.iLike ]: email } } ]
       }
     }
 
--- a/server/models/account/user.ts
+++ b/server/models/account/user.ts
@@ -360,7 +360,7 @@ export class UserModel extends Model<UserModel> {
   static loadByUsername (username: string) {
     const query = {
       where: {
-        username
+        username: { [ Op.iLike ]: username }
       }
     }
 
@@ -370,7 +370,7 @@ export class UserModel extends Model<UserModel> {
   static loadByUsernameAndPopulateChannels (username: string) {
     const query = {
       where: {
-        username
+        username: { [ Op.iLike ]: username }
       }
     }
 
@@ -380,7 +380,7 @@ export class UserModel extends Model<UserModel> {
   static loadByEmail (email: string) {
     const query = {
       where: {
-        email: { [ Op.iLike ]: email}
+        email
       }
     }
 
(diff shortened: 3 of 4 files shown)


Chocobozzz:
Could you add a unit test? 

Here: https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/users/users.ts#L114
See the documentation: https://github.com/Chocobozzz/PeerTube/blob/develop/.github/CONTRIBUTING.md#unit-tests

NassimBounouas:
Hi @Chocobozzz,

Thank you for the pointer. I thought about adding  a unit test but I didn't know where to do it. I'll work on this.

NassimBounouas:
@Chocobozzz I tried to run the test suite following the doc before adding mine and I encountered this problem :

```
  Test a client controllers
    1) "before all" hook for "Should have valid Open Graph tags on the watch page with video id"
    2) "after all" hook for "Should have valid index html updated tags (title, description...)"


  0 passing (2m)
  2 failing

  1) Test a client controllers
       "before all" hook for "Should have valid Open Graph tags on the watch page with video id":
     Error: Timeout of 120000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/user/dev/PeerTube/server/tests/index.ts)
```

Do you any idea of the root cause ? Thank you !

Chocobozzz:
@NassimBounouas Just run the test you updated, because running all the test suite is quite long. Travis can do this for you.

To run the client controllers test, you need to `npm run build -- --light` (to build the english client). Could you update the contributing guide and replace `npm run build` with `npm run build -- --light`?

NassimBounouas:
@Chocobozzz I'll update the contrib guide at the same time I will push a test. Thank you for the advice. 

While I'm trying to just run one controller test I'm encountering a timeout. What can be the reason ?

`
$ npm run mocha -- --exit --require ts-node/register/type-check --bail server/tests/api/index.ts

> peertube@1.3.1 mocha /home/user/dev/PeerTube
> mocha "--exit" "--require" "ts-node/register/type-check" "--bail" "server/tests/api/index.ts"



  Test activitypub
    1) "before all" hook for "Should return the account object"


  0 passing (30s)
  1 failing

  1) Test activitypub
       "before all" hook for "Should return the account object":
     Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/user/dev/PeerTube/server/tests/api/index.ts)
`

Chocobozzz:
Try to run `npm run clean:server:test && NODE_APP_INSTANCE=1 NODE_ENV=test npm start` to see what's wrong

NassimBounouas:
I'll recreate a clean environment later in the day. I'm using a postgresql db and a redis db that are both run into a docker container so the cleaning script can't be directly used.

NassimBounouas:
Hi @Chocobozzz,

I finally successed to run the test suite with databases directly installed on my computer and I have few questions. In the documentation [there](https://docs.joinpeertube.org/#/contribute-getting-started?id=unit-tests) there is those two commands :

```
$ sudo -u postgres createuser you_username --createdb --superuser
$ createdb -O peertube peertube_test{1,2,3,4,5,6}
```

1) Does the part  `-O peertube` on the second command should not be the same as `you_username` ?
2) Does the second command needs a specific version of postgresql because it returns me the following error : 

```
$ sudo -u postgres createdb -O peertube peertube_test{1,2,3,4,5,6}
createdb: too many command-line arguments (first is "peertube_test3")
Try "createdb --help" for more information.
```
I created 6 databases prefixed with `peertube_test` and it seems to work.

According your answers I will update the contribution guide.

Thank you

Chocobozzz:
> Does the part -O peertube on the second command should not be the same as you_username ?

No it should be `peertube`, because it's the postgresql user we declared in the default configuration file that test instances use. `sudo -u postgres createuser you_username --createdb --superuser` is used by PeerTube test scripts (`npm run clean:server:test`) to create and drop databases without using the `postgres` user. 

> Does the second command needs a specific version of postgresql because it returns me the following error 

I think the documentation is wrong. We could replace this with `npm run clean:server:test` that will create these databases.

NassimBounouas:
Hi @Chocobozzz !

After some experimentation I think that I understood the peertube unit tests structure and add a test about the username case insensitivity during the login. Can you tell me if my test fits  to your expectation ? 

Just to be sure about what I understood : the credentials used in the test suite are set [there](https://github.com/Chocobozzz/PeerTube/blob/develop/shared/extra-utils/server/servers.ts#L211) from an extract (using regex) of the test server standard output [there](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/installer.ts#L150), am I right ?

I also updated the documentation for the build part and the test databases clean process.

Subsidiary question : I saw few unit tests that are in pending state, is there any issue related to this in order to contribute ?

Thank you

Chocobozzz:
> Just to be sure about what I understood : the credentials used in the test suite are set there from an extract (using regex) of the test server standard output there, am I right ?

Yep!

> Subsidiary question : I saw few unit tests that are in pending state, is there any issue related to this in order to contribute ?

I don't remember all of them, but I think you can try to implement some of them :)

PR #1935, opened 2019-06-22
Update postgresql version for Gentoo

Wirebrass:
Update version of postgresql in rc-update and rc-service commands.


rigelk:
Thanks @Wirebrass!

PR #1937, opened 2019-06-23
For Gentoo : update @peertube set (delete vips)

Wirebrass:
The media-libs/vips package does not exist on Gentoo (and all works fine on my PeerTube POC without vips).


Chocobozzz:
Thanks @Wirebrass :)

PR #1938, opened 2019-06-23
For Gentoo : set testing keyword for yarn

Wirebrass:
Gentoo has not "stable" keyword for yarn (cf source, next line), you need to add "testing" keyword for yarn.
Source : https://packages.gentoo.org/packages/sys-apps/yarn


Chocobozzz:
Thanks @Wirebrass

PR #1940, opened 2019-06-23
Make video files immutable (fixes #1777)

### LINKED ISSUE #1777: Video files are not cacheable
I am not using the default Peertube installation where nginx has direct access to video files on the local disk. Instead, I am requesting all files through Peertube. In this setup, I noticed that fragments of video files were never cached. Turns out this is because Peertube is sending the following header:
```
cache-control: public, max-age=0
```

To verify, you can run the following request (you can use network tools in Firefox to copy the request from your own instance):
```
curl 'https://peertube.social/static/webseed/610a67fe-df2a-482d-ad9a-aeb57433187b-720.mp4' -H 'range: bytes=2031616-2097151' -I
``` 

I would expect that the cache-control header is set to `immutable`, or to a very long `max-age`.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
https://stackoverflow.com/a/1383359

Edit: To be clear, I only checked headers for the webseed folder, but this problem might also affect files in the redundancy folder.

Nutomic:
This PR puts newly uploaded videos into the folder `storage/videos/transcoding`, and moves them to `storage/videos/` once transcoding is finished. This means that video files are never changed/replaced on disk, and they can easily be cached.

Side note: it would also be good to use to use cache-control: immutable for other static resources like css files, js files and so on to reduce the number of requests ([Improving Performance with Cache-Control: Immutable](https://www.keycdn.com/blog/cache-control-immutable)).

I havent tested this PR yet.

### DIFF
--- a/server/controllers/api/videos/index.ts
+++ b/server/controllers/api/videos/index.ts
@@ -212,7 +212,7 @@ async function addVideo (req: express.Request, res: express.Response) {
   }
 
   // Move physical file
-  const videoDir = CONFIG.STORAGE.VIDEOS_DIR
+  const videoDir = CONFIG.STORAGE.VIDEOS_TRANSCODING_DIR
   const destination = join(videoDir, video.getVideoFilename(videoFile))
   await move(videoPhysicalFile.path, destination)
   // This is important in case if there is another attempt in the retry process
--- a/server/controllers/static.ts
+++ b/server/controllers/static.ts
@@ -5,6 +5,7 @@ import {
   ROUTE_CACHE_LIFETIME,
   STATIC_DOWNLOAD_PATHS,
   STATIC_MAX_AGE,
+  VIDEO_MAX_AGE,
   STATIC_PATHS,
   WEBSERVER
 } from '../initializers/constants'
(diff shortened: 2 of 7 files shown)


Nutomic:
@Chocobozzz Could you have a look at this? I tested it, and there are some problems with getting the correct video path in `server/models/video/video.ts`. I tried to fix that with my last commit, but apparently its not working. And I cant really test it myself because after a few minutes of testing, npm always fills up my RAM until the entire OS freezes.

Chocobozzz:
Hi,

I'm sorry but I don't think it will work. Video files are not immutable, because they can be re-encoded (they will probably be in the future with new available video codecs). The only way I see to use this header is to update the video file names by adding an identifier like a sha sum. But it's a big work.

Nutomic:
Hmm but re-encoding a video file would be really rare, and probably initiated by the admin, right? So in that case they could clear the cache on the the server manually. And instead of making the videos immutable, we could just make them cacheable, at least for 2 hours like we do with thumbnails? 

The problem is that Peertube uses a lot of CPU to serve video files directly. This screenshot is from 1 MByte/s of traffic at most. When I used nginx serving directly from the filesystem, I never saw more than 1-2% of CPU usage. Sure I could serve videos from disk again, but I think caching would be good to have.

Screenshot_20190702_130202

Chocobozzz:
Sure I understand your issue. But I don't think it's the good solution.

The nginx cache is a very good idea, but maybe you should cache the videos for ~2 hours directly in nginx. Or directly serve videos from the disk, which is the most efficient solution.

When we'll use the HLS player by default in the future, you won't have this high CPU usage anymore since HLS segments are much bigger than torrent chunks (and so require much fewer HTTP requests)

Nutomic:
Yes I could just force cache it for 2 hours now, but you are right that this will be problematic with transcoding. So this PR would at least make the 2 hour caching possible (even if videos are not immutable).

Chocobozzz:
I'm sorry but I don't want to add this another complexity in peertube just for this specific usecase, since it will be fixed by our new player & you can serve the files with nginx from the disk

Nutomic:
Alright. Though I still think that making videos cacheable would be an improvement.

Issue #1741, opened 2019-03-28
Column quota sortable in user list ?


bristow:
It would be very convenient to be able to sort users by quota used, like 'created' and 'username' columns.

I think it's an issue for the label "Good First Issue" :)

darnuria:
Hi, @rigelk would you mentor me on this issue ? I can start working on it tomorrow.

rigelk:
@darnuria of course!

darnuria:
@rigelk hi! Is that the right componnent to edit: `client/src/app/+admin/users/user-list/user-list.component.html` ? I am new to AngularJS.

Chocobozzz:
@darnuria yes, but you'll also have to update the [backend](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/constants.ts#L42) (just add `videoQuota` to the array)

darnuria:
Shall we continue the discussion inside the merge request or here?

Chocobozzz:
in the mr

Chocobozzz:
Implemented by @darnuria in https://github.com/Chocobozzz/PeerTube/commit/4511437e19d2ce521e28f8417512d47773207da4 & https://github.com/Chocobozzz/PeerTube/commit/aa3796bda5a6776e0d21048239f25d385a8e7545

PR #1982, opened 2019-07-22
[WIP] Add sortable for videoQuotaUsed in user-list component.

darnuria:


darnuria:
With @rigelk I found the component to change in the client, but it need changes in the server side:

```[1] [localhost:9000] 2019-07-23 01:28:20.753 warn: Incorrect request parameters {
[1]   "path": "/api/v1/users/?start=0&count=10&sort=videoQuotaUsed",
[1]   "err": {
[1]     "sort": {
[1]       "location": "query",
[1]       "param": "sort",
[1]       "value": "videoQuotaUsed",
[1]       "msg": "Should have correct sortable column"
[1]     }
[1]   }
[1] }
```
It's because videoQuotaUsed is a function not a SQL column.

(mostly notes to myself and @rigelk to continue the PR in the week)

There is some bits of solution in `server/models/account/user.ts` and  `server/models/utils.ts`.

Chocobozzz:
See https://github.com/Chocobozzz/PeerTube/issues/1741#issuecomment-514079151

darnuria:
Hi, @Chocobozzz thanks for your advise ! I reported it from #1741 here:

> @darnuria yes, but you'll also have to update the backend (add videoQuotaUsed to the array)

With @rigelk I tried this option alas videoQuotaUsed,  is not a SQL column it's a SubQuery defined here: 

https://github.com/Chocobozzz/PeerTube//blob/50b4dcce56275e57fc15ade7529f5c501cad150d/server/models/account/user.ts#L619

Chocobozzz:
Oh right. I'll update the server myself, because it's a little more complex than I thought

Chocobozzz:
Actually, it was not that hard :) https://github.com/Chocobozzz/PeerTube/commit/aa3796bda5a6776e0d21048239f25d385a8e7545 

Thanks @darnuria

darnuria:
I was busy this week, no problem for completing my work. :) :+1: 

I would have had the time this weekend to work on it!

PR #1987, opened 2019-07-25
Add default caching values

BO41:
Not caching static files hurts the load [performance](https://github.com/Chocobozzz/PeerTube/issues/1277) and also really the [seo score](https://github.com/Chocobozzz/PeerTube/issues/1080).
Especially the 2h which are configured at the moment.

This should add some [better values](https://www.digitalocean.com/community/tutorials/how-to-implement-browser-caching-with-nginx-s-header-module-on-centos-7#step-3-%E2%80%94-configuring-cache-control-and-expires-headers) for caching.

Please test it if you are running an instance (as this **isn't** tested yet).
Suggestions (for better values) are welcome :smiley: 

To see the google seo score improve, test with Lighthouse or [insight](https://developers.google.com/speed/pagespeed/insights/).
(Disclaimer: I never cached Angular and I never cached with nginx)

### DIFF
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -13,10 +13,22 @@ server {
   location / { return 301 https://$host$request_uri; }
 }
 
+# Expires map
+map $sent_http_content_type $expires {
+    default                    30d;
+    text/html                  epoch;
+    text/css                   max;
+    application/javascript     2h;
+    ~image/                    max;
+    application/font-woff2     max;
+}
+
 server {
   listen 443 ssl http2;
   listen [::]:443 ssl http2;
   server_name peertube.example.com;
+
+  expires $expires;
 
   # For example with certbot (you need a certificate to run https)
   ssl_certificate      /etc/letsencrypt/live/peertube.example.com/fullchain.pem;


BO41:
If merged, the release notes should feature this change, so that running instances can update their nginx config.

>To enable the new configuration, restart Nginx.
 $   `sudo systemctl restart nginx`

Nutomic:
I like the idea, but I think it would be much cleaner to handle cache headers inside Peertube. You should also disable caching for video files, see my PR #1940 about that.

Chocobozzz:
I'm not sure to understand.

Client static files are already cached, and immutable: https://github.com/Chocobozzz/PeerTube/blob/develop/support/nginx/peertube#L63

Thumbnails and avatars are cached for 2 hours on purpose because they can change: https://github.com/Chocobozzz/PeerTube/blob/develop/support/nginx/peertube#L85

BO41:
ohh, I am sorry.
I was checking https://peertube.social and they don't have any good caching enabled.
And with the checkbox in #1277 not checked, I thought this was still an issue

Nvm, the current solution is completely fine and better than my suggested :wink:

PR #1991, opened 2019-07-28
Add since parameter to peertube-import-videos

fflorent:
# Use case

As a PeerTube administrator, I would like to be able to synchronize a YT channel with a PT instance with the agreement of the videos author using a crontab.

However, I cannot rely on the name of the videos on YT. So I would like the job to only fetch the videos published since the last time I ran the peertube-import-videos job.

# Usage

> Usage: import-videos [options]                        
> 
> Options:
  -n, --video-name <name>                Video name
  -c, --category <category_number>       Category number
  -l, --licence <licence_number>         Licence number
  -L, --language <language_code>         Language ISO 639 code (fr or en...)
  -t, --tags <tags>                      Video tags
  -N, --nsfw                             Video is Not Safe For Work
  -d, --video-description <description>  Video description
  -P, --privacy <privacy_number>         Privacy
  -C, --channel-name <channel_name>      Channel name
  -m, --comments-enabled                 Enable comments
  -s, --support <support>                Video support text
  -w, --wait-transcoding                 Wait transcoding before publishing the video
  -u, --url <url>                        Server url
  -U, --username <username>              Username
  -p, --password <token>                 Password
  --target-url <targetUrl>               Video target URL
  **--since <since>                        Publication date (inclusive) since which the videos can be imported (YYYY-MM-DD)**
  -v, --verbose                          Verbose mode
  -h, --help                             output usage information

# Example

Example of a cron job (run every sunday at 20:00 fetching the videos from the last 6 days):

```shell
0 20 * * 0 /usr/bin/node /__PATH_TO_PEERTUBE__/dist/server/tools/peertube-import-videos.js \
   -u 'https://skeptikon.fr'  -U '__USER__'  --password '__PASSWORD__'  \
  --target-url 'https://www.youtube.com/channel/___CHANNEL__'  \
  --since $(date --date="-7 days" +%Y-%m-%d)
```

# TODO list

- [x] Implement the `--since` argument
- [x] Fix the `-t` argument (which now holds for `--tags`)
- [x] Also add the `--until` argument
- [ ] Mention it in some documentation (FAQ?)

### DIFF
--- a/server/tools/package.json
+++ b/server/tools/package.json
@@ -6,7 +6,6 @@
     "application-config": "^1.0.1",
     "cli-table": "^0.3.1",
     "netrc-parser": "^3.1.6",
-    "moment": "^2.24.0",
     "webtorrent-hybrid": "^3.0.1"
   },
   "summon": {
--- a/server/tools/package.json
+++ b/server/tools/package.json
@@ -6,6 +6,7 @@
     "application-config": "^1.0.1",
     "cli-table": "^0.3.1",
     "netrc-parser": "^3.1.6",
+    "moment": "^2.24.0",
     "webtorrent-hybrid": "^3.0.1"
   },
   "summon": {
--- a/server/tools/peertube-import-videos.ts
+++ b/server/tools/peertube-import-videos.ts
@@ -2,7 +2,6 @@
 require('tls').DEFAULT_ECDH_CURVE = 'auto'
 
 import * as program from 'commander'
-import * as moment from 'moment'
 import { join } from 'path'
 import { doRequestAndSaveToFile } from '../helpers/requests'
 import { CONSTRAINTS_FIELDS } from '../initializers/constants'
@@ -27,8 +26,6 @@ const processOptions = {
 let command = program
   .name('import-videos')
 
-const DATE_FORMAT = 'YYYY-MM-DD';
-
 command = buildCommonVideoOptions(command)
 
 command
@@ -37,6 +34,7 @@ command
   .option('-p, --password <token>', 'Password')
   .option('--target-url <targetUrl>', 'Video target URL')
   .option('--since <since>', 'Publication date (inclusive) since which the videos can be imported (YYYY-MM-DD)', parseDate)
+  .option('--until <until>', 'Publication date (inclusive) until which the videos can be imported (YYYY-MM-DD)', parseDate)
   .option('-v, --verbose', 'Verbose mode')
   .parse(process.argv)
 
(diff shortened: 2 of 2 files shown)


fflorent:
Also poke @floSoX @tsaquet ;)

server/tools/package.json
```diff
@@ -6,6 +6,7 @@
     "application-config": "^1.0.1",
     "cli-table": "^0.3.1",
     "netrc-parser": "^3.1.6",
+    "moment": "^2.24.0",
```
Chocobozzz: Please don't use moment. Check the date is valid manually, and create the a `Date` object:

```
const d = new Date('2019-07-29')

// 2019-07-29T00:00:00.000Z
```

gnouts:
Just a suggestion... Wouldn't it be better to keep track of the YT video ID (the one you see in a YT link) instead of a date ? Something like "this YT id matches this PT id, so don't upload again".

Nutomic:
@gnouts Agreed, that seems like a cleaner and more general solution. Something like `original-upload-url`. That could also be used by [Peertubify](https://gitlab.com/Ealhad/peertubeify) and similar projects.

fflorent:
@gnouts @Nutomic That's funny, because a friend suggested the same thing yesterday. The proposed patch is rather simple, but storing a YT video ID among the metadata may be more elegant and more features may leverage from this implementation.

Though I guess that would be more effort. @Chocobozzz what do you think?

Chocobozzz:
You can if you have time. Adding a another video field requires federation stuffs & write tests.

fflorent:
Alright, I guess I'd rather finish this enhancement and start this later as a future improvement.

server/tools/package.json
fflorent: Good idea. Done!

fflorent:
I don't know why, I get in the output:

> Video "_" has been published after "2019-01-01", don't upload it.

Is this a change from Youtube? An issue from youtube-dl?

gnouts:
@fflorent I can confirm this bug. I had the same on my upload script. Doing a pip upgrade solved the issue. Like you suggest I think it's related to youtube-dl.

fflorent:
I see. We just have to run `npm rebuild --prefix ~/projects/PeerTube youtube-dl`.

Thanks @gnouts

Chocobozzz:
LGTM. Can we merge?

fflorent:
@Chocobozzz Sure. And what about the documentation? I would like to document how to properly synchronize videos from Youtube to Peertube (I can do that in a separate PR).

Chocobozzz:
You can do it in [this document](https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/tools.md#peertube-import-videosjs)

PR #2004, opened 2019-08-04
Update CentOS 7 dependencies install

Wirebrass:
Add initialization for postgresql database (else there is an error when starting postresql service).
Enable Redis and PostreSQL services.
Replace "service" by "systemctl" (systemd).

Tested in local on a fresh CentOS 7 install.


Chocobozzz:
Thanks @Wirebrass

PR #2005, opened 2019-08-05
Add originallyPublishedAt in /videos/upload to the OpenAPI document

frankstrater:
When using the API I noticed originallyPublishedAt was missing from the API documentation, but the key/value pair is being accepted in /videos/upload as long as the string has a date-time format.


Chocobozzz:
Thanks

PR #2012, opened 2019-08-09
Add/update OpenRC service script

Wirebrass:
- Update Production.md to add use of /etc/init.d/peertube service script.
- Update default used PATH in /etc/init.d/peertube service script template.

Tested on Gentoo Linux.


Chocobozzz:
Thanks @Wirebrass

PR #2016, opened 2019-08-12
Specify the restrictions on tags array and it's string items

frankstrater:
The OpenAPI specification allows for specific array and string restrictions. See: https://swagger.io/docs/specification/data-models/data-types/


Chocobozzz:
Thanks @frankstrater

PR #2027, opened 2019-08-14
Add originallyPublishedAt to PUT videos + change HTTP response to 204

frankstrater:
Tested the API call with cURL in verbose mode.


Chocobozzz:
thanks @frankstrater

PR #2045, opened 2019-08-17
peertube-import-videos.ts: add --tmpdir, --first, --last and --verbose [level] parameters

Yetangitu:
The `--tmpdir <tmpdir>` parameter can be used to designate a working directory for downloading and converting imported videos.

Without this parameter PT will unceremoniously (attempt to) dump videos into its own _tools_ directory - where the user might not even have write access.

I refrained from creating a randomly-named tmpdir inside the designated directory as the file names used by pt-i-v are already randomised and as such the chance of a file name collision is negligible.

The program will error out with `--tmpdir directory_name_here: directory does not exist or is not accessible` if the designated directory does not exist.

The `--first n` and `--last n` parameters can be used to limit processing to the first or last _n_ elements of the playlist returned by `youtube-dl`. These parameters can be used (optionally in combination with `--since YYYY-MM-DD` and/or `--until YYYY-MM-DD`) to limit the number of videos which PT needs to process to mirror a channel. This is a somewhat hackish workaround for the fact that `youtube-dl` can only return an entire channel playlist (which can contains thousands of videos) which `peertube-import-videos` then needs to crawl through to import the last few items. An example says more than many words:

Given a mirrored Youtube channel which starts with _500_ videos and daily adds a single video:
- `peertube-import-videos` would need to trudge through _500_ elements on the second day to import element _501_. The next day it would need to walk over _501_ elements to import numner _502_, etc.
- using `--since <date-last-sync> --last 3` this would be limited to a maximum of _3_ elements per day where the automatic duplicate detection would take care of filtering out those videos which have already been imported.

With `--verbose [verbosity]` the chattiness of `peertube-import-videos` can be set from `0` (_only report on errors_) to `4` (_dump all available data_). Use `--verbose 1` to only be notified of successfully imported videos, `--verbose 0` only reports on errors. These levels are useful when `peertube-import-videos` is run in a cron job to mirror a channel as it (at level `1`) only sends mail when new videos are imported or (at level `0`) when errors occur. The default level is `2` (info), when `--verbose` is used without the optional parameter the logging level is set to `3` (debug) to emulate the way this switch worked before this change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,6 @@
     "jsonld": "~1.1.0",
     "jsonld-signatures": "https://github.com/Chocobozzz/jsonld-signatures#rsa2017",
     "lodash": "^4.17.10",
-    "loglevel": "^1.6.3",
     "lru-cache": "^5.1.1",
     "magnet-uri": "^5.1.4",
     "memoizee": "^0.4.14",
--- a/package.json
+++ b/package.json
@@ -130,6 +130,7 @@
     "jsonld": "~1.1.0",
     "jsonld-signatures": "https://github.com/Chocobozzz/jsonld-signatures#rsa2017",
     "lodash": "^4.17.10",
+    "loglevel": "^1.6.3",
     "lru-cache": "^5.1.1",
     "magnet-uri": "^5.1.4",
     "memoizee": "^0.4.14",
--- a/server/tools/peertube-import-videos.ts
+++ b/server/tools/peertube-import-videos.ts
@@ -36,6 +36,8 @@ command
   .option('--tmpdir <tmpdir>', 'Working directory')
   .option('--since <since>', 'Publication date (inclusive) since which the videos can be imported (YYYY-MM-DD)', parseDate)
   .option('--until <until>', 'Publication date (inclusive) until which the videos can be imported (YYYY-MM-DD)', parseDate)
+  .option('--first <first>', 'Process first n elements of returned playlist')
+  .option('--last <last>', 'Process last n elements of returned playlist')
   .option('-v, --verbose', 'Verbose mode')
   .parse(process.argv)
 
@@ -88,10 +90,17 @@ async function run (url: string, user: UserInfo) {
 
     // Normalize utf8 fields
     if (Array.isArray(info) === true) {
-      infoArray = info.map(i => normalizeObject(i))
+      if (program[ 'first' ]) {
+        infoArray = info.slice(0, program[ 'first' ]).map(i => normalizeObject(i))
+      } else if (program[ 'last' ]) {
+        infoArray = info.slice(0 - program[ 'last' ]).map(i => normalizeObject(i))
+      } else {
+        infoArray = info.map(i => normalizeObject(i))
+      }
     } else {
       infoArray = [ normalizeObject(info) ]
     }
+
     console.log('Will download and upload %d videos.\n', infoArray.length)
 
     for (const info of infoArray) {
--- a/server/tools/peertube-import-videos.ts
+++ b/server/tools/peertube-import-videos.ts
@@ -2,6 +2,7 @@
 require('tls').DEFAULT_ECDH_CURVE = 'auto'
 
 import * as program from 'commander'
+import * as log from 'loglevel'
 import { join } from 'path'
 import { doRequestAndSaveToFile } from '../helpers/requests'
 import { CONSTRAINTS_FIELDS } from '../initializers/constants'
@@ -38,13 +39,13 @@ command
   .option('--until <until>', 'Publication date (inclusive) until which the videos can be imported (YYYY-MM-DD)', parseDate)
   .option('--first <first>', 'Process first n elements of returned playlist')
   .option('--last <last>', 'Process last n elements of returned playlist')
-  .option('-v, --verbose', 'Verbose mode')
+  .option('-v, --verbose [verbosity]', 'Verbosity, from 0 (only report errors) to 4 (for debugging), default is 2 (info)')
   .parse(process.argv)
 
 getServerCredentials(command)
   .then(({ url, username, password }) => {
     if (!program[ 'targetUrl' ]) {
-      console.error('--targetUrl field is required.')
+      log.error('--targetUrl field is required.')
 
       process.exit(-1)
     }
@@ -53,12 +54,35 @@ getServerCredentials(command)
       program[ 'tmpdir' ] = __dirname;
     } else {
       if (!existsSync(program[ 'tmpdir' ])) {
-        console.error('--tmpdir %s: directory does not exist or is not accessible', program[ 'tmpdir' ])
+        log.error('--tmpdir %s: directory does not exist or is not accessible', program[ 'tmpdir' ])
 
         process.exit(-1)
       }
     }
 
+    log.setLevel('info')
+    if (program[ 'verbose' ]) {
+      switch (program[ 'verbose' ]) {
+        case '0':
+          log.setLevel('error')
+          break
+        case '1':
+          log.setLevel('warn')
+          break
+        case '2':
+          log.setLevel('info')
+          break
+        case '3':
+          log.setLevel('debug')
+          break
+        case '4':
+          log.setLevel('trace')
+          break
+        default:
+          log.setLevel('debug')
+          break
+      }
+    }
 
     removeEndSlashes(url)
     removeEndSlashes(program[ 'targetUrl' ])
@@ -67,7 +91,7 @@ getServerCredentials(command)
 
     run(url, user)
       .catch(err => {
-        console.error(err)
+        log.error(err)
         process.exit(-1)
       })
   })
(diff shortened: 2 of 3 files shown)


server/tools/peertube-import-videos.ts
```diff
@@ -68,30 +106,37 @@ async function run (url: string, user: UserInfo) {
...
 
     let infoArray: any[]
 
     // Normalize utf8 fields
     if (Array.isArray(info) === true) {
-      infoArray = info.map(i => normalizeObject(i))
+      if (program[ 'first' ]) {
+        infoArray = info.slice(0, program[ 'first' ]).map(i => normalizeObject(i))
+      } else if (program[ 'last' ]) {
+        infoArray = info.slice(0 - program[ 'last' ]).map(i => normalizeObject(i))
```
fflorent: I guess you can omit the `0` here :)

server/tools/peertube-import-videos.ts
```diff
@@ -33,27 +34,64 @@ command
...
   .then(({ url, username, password }) => {
     if (!program[ 'targetUrl' ]) {
-      console.error('--targetUrl field is required.')
+      log.error('--targetUrl field is required.')
 
       process.exit(-1)
     }
 
+    if (!program['tmpdir']) {
+      program[ 'tmpdir' ] = __dirname;
```
fflorent: Commander allows specifying the default value. Rather adding this kind of conditions, we'd probably rather do:
```js
command
  // …
  .option('--tmpdir <tmpdir>', 'Working directory', __dirname)
  // …
```

server/tools/peertube-import-videos.ts
```diff
@@ -33,27 +34,64 @@ command
...
+    } else {
+      if (!existsSync(program[ 'tmpdir' ])) {
```
fflorent: If we use `existsSync`, we may not notice when the `tmpdir` have no write access.

What about using [`accessSync`](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_accesssync_path_mode)?

server/tools/peertube-import-videos.ts
```diff
@@ -33,27 +34,64 @@ command
...
+          log.setLevel('info')
+          break
+        case '3':
+          log.setLevel('debug')
+          break
+        case '4':
+          log.setLevel('trace')
+          break
+        default:
+          log.setLevel('debug')
```
fflorent: Shouldn't it be `info`? (the default value as reported in the commander help)

server/tools/peertube-import-videos.ts
```diff
@@ -33,27 +34,64 @@ command
...
+          break
+      }
+    }
```
fflorent: Wouldn't be possible to use a map object? Or even an array? Like:
```js
const logLevels = ['error', 'warn', 'info', 'debug', 'trace']
const level = parseInt(program[ 'verbose' ]) > 4 ? 2 : program[ 'verbose' ]
log.setLevel(level)
```

But I wonder if we shouldn't let the user specify directly the verbosity (`error`, `info`, …) which are self-explanatory values (and raise an error if the passed value is not among the ones expected).

server/tools/peertube-import-videos.ts
```diff
@@ -68,30 +106,37 @@ async function run (url: string, user: UserInfo) {
...
       process.exit(1)
     }
 
     let infoArray: any[]
 
     // Normalize utf8 fields
     if (Array.isArray(info) === true) {
-      infoArray = info.map(i => normalizeObject(i))
+      if (program[ 'first' ]) {
+        infoArray = info.slice(0, program[ 'first' ]).map(i => normalizeObject(i))
```
fflorent: Shouldn't it be `info.slice(program[ 'first' ])`?

server/tools/peertube-import-videos.ts
```diff
@@ -68,30 +106,37 @@ async function run (url: string, user: UserInfo) {
...
+      } else if (program[ 'last' ]) {
+        infoArray = info.slice(0 - program[ 'last' ]).map(i => normalizeObject(i))
+      } else {
+        infoArray = info.map(i => normalizeObject(i))
+      }
```
fflorent: I may nitpick here: I guess we shouldn't repeat the `.map(i => normalizeObject(i))`.

```js
if (Array.isArray(info) === true) {
  if (program[ 'first' ]) {
    infoArray = info.slice(0, program[ 'first' ])
  } else if (…) {
    infoArray = info.slice(- program[ 'last' ])
  } else {
    infoArray = info
  }
  infoArray = info.map(i => normalizeObject(i))
}
```

Also we can use this trick to avoid the `if (Array.isArray(info) === true) {` condition (`concat` can be used to force a value to be an array):
```js
infoArray = [].concat(info);
if (program[ 'first' ]) {
  infoArray = infoArray.slice(0, program[ 'first' ])
} else if (program[ 'last' ]) {
  infoArray = infoArray.slice(- program[ 'last' ])
}
infoArray = infoArray.map(i => normalizeObject(i))
```

package.json
```diff
@@ -130,6 +130,7 @@
     "jsonld": "~1.1.0",
     "jsonld-signatures": "https://github.com/Chocobozzz/jsonld-signatures#rsa2017",
     "lodash": "^4.17.10",
+    "loglevel": "^1.6.3",
```
Chocobozzz: Please don't add another dependency. If you want to manage log levels, create your own helper (or use winston)

server/tools/peertube-import-videos.ts
Yetangitu: True, I more or less hacked this change into the program to be able to mirror a channel which has been blocked for no good reason, then unblocked again. This is clear from many of the additions, including...

server/tools/peertube-import-videos.ts
Yetangitu: ...this... I'll change it.

server/tools/peertube-import-videos.ts
Yetangitu: The default log level is set to `info` just above the `switch`, this is used when there is no `--verbose` option on the command line. When the program is called with `--verbose` (without a following parameter) the log level is set to `debug` to emulate the way it worked before this change. Using `--verbose` with the optional parameter makes it possible to select a level of verbosity.

server/tools/peertube-import-videos.ts
Yetangitu: The former, sure, that is cleaner. The latter, well, it works but is isn't cleaner or clearer than the original. I don't really care though, as long as it works and looks somewhat maintainable.

package.json
Yetangitu: OK, I'll look for another option. I do believe this functionality - i.e. user-selectable verbosity - should be included in all the tools as it is makes the CLI-tools far more usable in cron jobs etc. Would you accept a patch to the other tools which includes this? I made this for my own use but I'm sure there are others who want to control their PT-instance(s) using cron jobs and other automated tools which generally prefer less chatty/more concise output from client programs.

server/tools/peertube-import-videos.ts
Yetangitu: The starting zero is there for clarity, it is optional according to the specs.

server/tools/peertube-import-videos.ts
Yetangitu: This cat can be skinned in many ways, sure.

server/tools/peertube-import-videos.ts
Yetangitu: Certainly, I put it there for clarity - dangling minuses can cause confusion.

package.json
Chocobozzz: > Would you accept a patch to the other tools which includes this

Sure, but I think you should just use winston with a Console transporter

Yetangitu:
The `--verbose <level>` parameter has moved to `cli.ts` where it is used in combination with `getLogger(loglevel)` to provide a winston-based logger to other cli tools. The `<level>` parameter is no longer optional, use `--verbose 4` or the equivalent `--verbose debug` to get _verbose_ output.

- At level `0`(or `error`) it only shows errors
- At level `1` (or `warn`) it shows successfully downloaded videos (and errors, of course). Use this in _cron_ jobs etc.
- At level `2` (or `info`) it shows an overview of available videos, whether they will be downloaded, the download process and the subsequent upload to PT. This is the default level.
- At level `4` (or `debug`) if shows raw JSON dumps etc.

Yetangitu:
The last commit mostly serves to kick travis into action to see whether it works this time - there seem be be some problems here lately?

server/tools/peertube-import-videos.ts
```diff
@@ -355,24 +363,27 @@ async function getAccessTokenOrDie (url: string, user: UserInfo) {
...
+    exit_error(`Invalid date passed: ${dateAsStr}. See help for usage.`);
   }
   return date;
 }
 
 function formatDate (date: Date): string {
   return date.toISOString().split('T')[0];
 }
+
+function exit_error (message:string, ...meta: any[]) {
```
Chocobozzz: `exitError` and then it's ready to merge :)

Yetangitu:
Have at it

Chocobozzz:
Thanks @Yetangitu

PR #2071, opened 2019-08-22
Feature/strong model types

Chocobozzz:
Strong typing of database models. It's not very beautiful, but provides better safety and less headache when we manipulate these objects.


Chocobozzz:
Merged manually in https://github.com/Chocobozzz/PeerTube/commit/5c5e587307a27e173333789b5b5167d35f468b01

PR #2073, opened 2019-08-22
fixes 2065: optimisation

JohnXLivingston:
don't write json multiple times.


JohnXLivingston:
> maybe name it `notificationMessage` or something like this wink

Done.

PR #2088, opened 2019-08-28
Bug fix: logical and/or boolean selector

frankstrater:
Not sure how to describe it, but just think what happens if you post 'false' and what result it gets when evaluating `false || true` (the API user wants it to be false...)


Chocobozzz:
Oop's, thanks @frankstrater!

PR #2091, opened 2019-08-28
Feature/admin responsability

Chocobozzz:
screen_2019-09-03-11:25:10
screen_2019-09-03-11:25:03
screen_2019-09-03-11:24:34
screen_2019-09-03-11:24:23

### DIFF
--- a/client/src/app/+about/about-instance/about-instance.component.html
+++ b/client/src/app/+about/about-instance/about-instance.component.html
@@ -8,9 +8,25 @@
     </div>
 
     <div class="short-description">
-      <div>{{ shortDescription }}</div>
+      <div class="block short-description">{{ shortDescription }}</div>
 
-      <div *ngIf="isNSFW" class="dedicated-to-nsfw">This instance is dedicated to sensitive/NSFW content.</div>
+      <div *ngIf="isNSFW" class="block dedicated-to-nsfw">This instance is dedicated to sensitive/NSFW content.</div>
+
+      <div class="block">
+        <div class="section-title" *ngIf="languages.length !== 0">Main instance languages</div>
+
+        <ul>
+          <li *ngFor="let language of languages">{{ language }}</li>
+        </ul>
+      </div>
+
+      <div class="block">
+        <div class="section-title" *ngIf="categories.length !== 0">Main instance categories</div>
+
+        <ul>
+          <li *ngFor="let category of categories">{{ category }}</li>
+        </ul>
+      </div>
     </div>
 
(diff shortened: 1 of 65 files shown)


Chocobozzz:
Added some screens

Nutomic:
Looks good! I also added some info about technical details (server provider etc) [on my instance](https://peertube.social/about/instance), maybe we could add a field for that too? Or an option to add extra fields?

Can this info be edited by mods, or only by admins? And could you add a screenshot how users see this info?

Chocobozzz:
> Looks good! I also added some info about technical details (server provider etc) on my instance, maybe we could add a field for that too?

It's a good idea. I'll add it. 

> Can this info be edited by mods, or only by admins?

Only by admins

> And could you add a screenshot how users see this info?

screen_2019-09-04-16:35:01
screen_2019-09-04-16:34:27

mayel:
Will this information be made available via an API, such as nodeinfo or in some ActivityPub endpoint? It would be great because we plan on implementing something similar for [MoodleNet](https://new.moodle.net/) and would be nice to make it interoperable. It would also allow for fediverse directory sites to index and display this information.

Chocobozzz:
I don't think there is a standardized place where we could put these information. For now it will be in our REST API: `/api/v1/about`. But if we find a place to put them, we'll do it :+1:

mayel:
Great :) nodeinfo2 might be an option? I've opened an issue there about adding fields: https://git.feneas.org/jaywink/nodeinfo2/issues/2

jaywink:
Hey, just to sync my reply here as well:

---

> Hi @mayel ! I think it's a good idea to provide a possibility for admins to provide extra information about their instance. This is after all one of the reasons for NodeInfo type documents being created.
> 
> NodeInfo2 was a kind of temporary step towards something that would be better thought out and truly cross-protocol compatible. I've been drafting a thing called ServerInfo (but will be renamed to ServiceInfo in the next draft). I think it makes sense to push that forward instead of iterating on NodeInfo2. I believe we can get a longer lasting result. I also intend to submit the next draft for comments to the SocialCG group.
> 
> You can see the discussion and link to the draft in our Forum: https://talk.feneas.org/t/serverinfo-specification-for-server-metadata/99
> 
> Opinions would be very much welcome! I'll anyway have a peek into the linked PR on PeerTube and try incorporate those ideas into the next draft, based on this feedback.
> 
> I hope to build a working parser and serializer for the next draft soon as well to get some live working versions out there.
> 
> Does this sound good to you?

PR #2096, opened 2019-08-29
Add missing downloadEnabled option

frankstrater:
Added the downloadEnabled option to the OpenAPI doc, but did not define a default value (see below), because posted value can be any 'truthy' or 'falsey' string, as far as I know.

```yaml
downloadEnabled:
  description: Enable or disable download for this video
  type: string
  default: 'true'
```

### DIFF
--- a/support/doc/api/openapi.yaml
+++ b/support/doc/api/openapi.yaml
@@ -878,6 +878,9 @@ paths:
                 commentsEnabled:
                   description: Enable or disable comments for this video
                   type: string
+                downloadEnabled:
+                  description: Enable or disable download for this video
+                  type: string
                 originallyPublishedAt:
                   description: Date when the content was originally published
                   type: string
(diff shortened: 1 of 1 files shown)


Chocobozzz:
I'm not sure to understand why it is not a `boolean` type?

frankstrater:
Because it's a `multipart/form-data` request, so all non-files are strings?

frankstrater:
and downloadEnabled has the same type as commentsEnabled... (consistency)

Chocobozzz:
But the update endpoint is also compatible with a classic `json` body. Could we use the `oneOf` key? https://swagger.io/docs/specification/data-models/data-types/

frankstrater:
Ok, agreed, there's still more improvements to be done. I noticed that the API is rather flexible/forgiving in terms of what it accepts, which kind of complicates writing the spec in OpenAPI and takes time testing it. 

I'll close this pull request and will work on a more complete improvement next monday.

Chocobozzz:
Thanks!

PR #2131, opened 2019-09-17
CLI tools documentation: Change cli parameter from -t to --target-url

pjoubert-:
This just corrects the documentation for the video importer: the actual parameter has become --target-url and not anymore -t


Chocobozzz:
thanks

Issue #1979, opened 2019-07-18
About the unsustainable use of online video


sbernard31:
Regarding last report of [shift project](https://theshiftproject.org) about _**“Climate crisis: the unsustainable use of online video”**_, streaming is a real problem ! :sob: 

Here the report :page_facing_up: :
:gb: : https://theshiftproject.org/en/article/unsustainable-use-online-video/
:fr: : https://theshiftproject.org/article/climat-insoutenable-usage-video/

**Peertube** could follow some simple rules to try to limit its impacts :earth_africa: :
**a)** do not download video if user don't really want to watch it.
**b)** encourage usage of best encoding format like H.264.
**c)** encourage user to use low resolution.

This rules could be declined in features like :
1. for a) no autostart !
2. for b) forbid worst encoding format.
3. for b) add some advice or notation system about encoding format at upload time to nudge user to the right format and resolution. (CO²/bandwish/storage saved if it used a better encoding format)
4. for c) let instance owner define the max resolution.
5. for c) encourage instance owner to use a max resolution as lower as possible.
6. for c) about resolution choice,  auto mode could prefer the lower one.
7. for c) if it exist a resolution lower than the one chosen, a message could be displayed with the CO² saved if the lower resolution was chosen.
8. for c) add an audio only mode.

This features list is just some ideas to illustrate my point, this is not a real TODO list. 
Up to you(we) to find the better feature which will help to follow the above rules.

(**If you vote down, please try to explain why** in a constructive way. Maybe your point of view makes sense and it could be valuable to share it)

lapineige:
Another idea for a) : don't load the full video when you connection is fast enough to load the missing part when watching it.
If you open a video, start to watch it, but pause it and come back later, maybe you will stop and close the tab for some reason. But the player would load the full video anyway.
Instead, the player could load the next (for instance) 10%, or 2min, and then stop until the user resume the playback (or come back to the tab ?)

(this is especially relevant for long videos, talks, movies…)

Chocobozzz:
@lapineige This will be fixed with the HLS player

lapineige:
What does it changes ?

Chocobozzz:
It has a max buffer size, so it won't load the entire video

njean42:
+1 on no autostart (possibly configurable by user?)
+1 on audio-only play mode

sbernard31:
[...]

PR #2137, opened 2019-09-18
Autoplay next recommended video

LoveIsGrief:
This is a continuation of #1370 that was abandoned.

I just made it work again and addressed some comments on the PR.

The one thing I didn't do (maybe quite major), was mess with the recommendation algorithm. I feel that it's out of the scope of this PR and should be done in another one to separate concerns. Changing the recommendation algorithm would entail a bigger discussion imo, since there are many way of doing it.
I hope you agree

So, this PR simply adds a checkbox for the user to autoplay recommended videos once the current video has stopped playing.

A final TODO before it's ready:

 - [x] test is if playlists will still work properly with this

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,8 @@
 /scripts/i18n/generate-iso639-target.ts
 
 # Other
+/dump.rdb
+/.theia/
 /profiling/
 /*.zip
 /*.tar.xz
--- a/client/src/app/videos/+video-watch/video-watch.component.ts
+++ b/client/src/app/videos/+video-watch/video-watch.component.ts
@@ -35,7 +35,7 @@ import { getStoredTheater } from '../../../assets/player/peertube-player-local-s
 import { PluginService } from '@app/core/plugins/plugin.service'
 import { HooksService } from '@app/core/plugins/hooks.service'
 import { PlatformLocation } from '@angular/common'
-import { randomInt } from '@shared/core-utils/miscs/miscs';
+import { randomInt } from '@shared/core-utils/miscs/miscs'
 
 @Component({
   selector: 'my-video-watch',
@@ -222,7 +222,7 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
   onRecommendations (videos: Video[]) {
     if (videos.length > 0) {
       // Pick a random video until the recommendations are improved
-      this.nextVideoUuid = videos[randomInt(0,videos.length-1)].uuid
+      this.nextVideoUuid = videos[randomInt(0,videos.length - 1)].uuid
     }
   }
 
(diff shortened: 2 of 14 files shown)


Nutomic:
Awesome! So the checkbox is on the video watch page? Maybe you could post a screenshot.

Nutomic:
Also, is the setting stored permanently? In light of #1979, I think it would be better if the setting only applies to the current session (and is disabled by default).

LoveIsGrief:
@Nutomic The original PR introduced it as a user setting (which I'm quite content with) and it's false by default.
afbeelding

.gitignore
```diff
@@ -37,6 +37,8 @@
 /scripts/i18n/generate-iso639-target.ts
 
 # Other
+/dump.rdb
+/.theia/
```
Chocobozzz: What is it?

client/src/app/videos/+video-watch/video-watch.component.ts
```diff
@@ -217,6 +218,12 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
     return this.video.tags
   }
 
+  onRecomendations(videos: Video[]){
```
Chocobozzz: run `npm run lint` so you can fix code format

Chocobozzz: And it should be `onRecommendations`

client/src/app/videos/+video-watch/video-watch.component.ts
```diff
@@ -470,6 +477,10 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
 
       this.player.on('customError', ({ err }: { err: any }) => this.handleError(err))
 
+      if(this.user && this.user.autoPlayNextVideo){
```
Chocobozzz: put this in a `else if` after https://github.com/Chocobozzz/PeerTube/pull/2137/files#diff-b7342a12f863e2ea4fe04b8f4f704fbfR489

.gitignore
LoveIsGrief: Gitpod created those

server/helpers/custom-validators/users.ts
```diff
@@ -65,6 +65,10 @@ function isUserBlockedValid (value: any) {
   return isBooleanValid(value)
 }
 
+function isUserAutoPlayNextVideoValid (value: any) {
+  return isBooleanValid(value)
```
BO41: shouldn't there be a single function with `return isBooleanValid(value)` instead of those redundant functions? @Chocobozzz

Chocobozzz: No I prefer to wrap them in these functions, because if you need this validator in different places you just use the wrapper instead of thinking "wait so what is the type of this field?". Moreover it's easier to add other validators of this field in the future (for example if it must always be true etc)

LoveIsGrief:
No other changes have been requested, so I removed the "WIP" prefix. I'm open to other suggestions...

Chocobozzz:
Thanks @LoveIsGrief

LoveIsGrief:
Thanks for integrating @Chocobozzz and thanks @BO41 for the initial PR!

PR #2141, opened 2019-09-20
fix a few typos

Findus23:
Hi,

I noticed a few spelling errors in the `angular_en_US.xml`

### DIFF
--- a/server/tests/api/check-params/users.ts
+++ b/server/tests/api/check-params/users.ts
@@ -901,7 +901,7 @@ describe('Test users API validators', function () {
       await makePostBodyRequest({ url: server.url, path: registrationPath, token: server.accessToken, fields })
     })
 
-    it('Should fail with a channel name that is the same than user username', async function () {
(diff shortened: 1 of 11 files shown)


Chocobozzz:
Hi,

Thanks but it's a generated file. You have to update the HTML instead :)

Findus23:
Ah, that makes a lot of sense.
I'll update it.

server/tests/api/check-params/users.ts
```diff
@@ -901,7 +901,7 @@ describe('Test users API validators', function () {
       await makePostBodyRequest({ url: server.url, path: registrationPath, token: server.accessToken, fields })
     })
 
-    it('Should fail with a channel name that is the same than user username', async function () {
+    it('Should fail with a channel name that is the same as user username', async function () {
```
BO41: is "user username" correct? sounds a bit strange

Findus23: You are totally right, I missed that one.

Chocobozzz:
Thanks!

Issue #1648, opened 2019-02-19
Make RSS Feeds links <a href.. links


mattcoxonline:
**What happened?**
Every single instance of the RSS feed links I encounter seem to be constructed in some weird way, which means I click it (in Firefox) and it opens a dialog asking me to save the file.

I right-click it and you don't get the standard context menu for a link - with options such as "Copy link to clipboard" - because the way it's constructed is not a link.

This means I am currently unable to subscribe to any RSS feeds.

**Additional information**
* PeerTube version or URL: https://peertube.fr/video-channels/chrisweredigital@share.tube/videos
* Browser name/version: Firefox, latest stable, Mac OS

rigelk:
@mattcoxonline what version of Firefox do you have?

mattcoxonline:
Ah, now I am on a Windows PC - using Firefox - and I see how it's meant to behave:

<img src="https://user-images.githubusercontent.com/673542/53042296-3e7aad80-3486-11e9-99f7-c80050736070.png" alt="A screenshot of the RSS dropdown on PeerTube" width="250" />

On Firefox Mac, this didn't happen. On the mac, I didn't see the dropdown. It just brought up a dialog box like this:

<img src="https://user-images.githubusercontent.com/673542/53042609-f14b0b80-3486-11e9-8d12-b09a9984c2e6.png" alt="A screenshot of the RSS dropdown on PeerTube" width="250" />

As you can see, I can replicate this on Firefox for Windows - but this time via this menu:

<img src="https://user-images.githubusercontent.com/673542/53042502-b47f1480-3486-11e9-9ba3-b105e9cf2141.png" alt="A screenshot of the RSS dropdown on PeerTube" width="250" />

I guess both the RSS icon and the option in the Subscribe dropdown should react in the same way / give the same options?

buoyantair:
@mattcoxonline  The "Subscribe via RSS" clearly states to subscribe _via_ RSS. Instead I think it would be better to have a subscribe button which pulls up a dropdown list.

rigelk:
@buoyantair that's already a dropdown in a dropdown. UI-wise, it's not great.

buoyantair:
@rigelk Right... How about just a popup that allows you to select the subscription medium? We can use the popup to replace the other dropdown too.

frankstrater:
The problem here (and also on other web applications) is the routing with using file extensions for format selection. See examples below:
```
[...]

PR #2145, opened 2019-09-24
Fix subscribe via RSS button

frankstrater:
Currently the button to subscribe via RSS to a channel is not working when using the Firefox browser. This is due to the automatic content-disposition Firefox assigns on links ending with file extensions (.xml and .atom), resulting in a download modal instead of showing it inline in a browser window or tab.

A screenshot of the RSS dropdown on PeerTube

This pull request does not fix the automatic content-disposition, but makes the button into a link, so that Firefox users can at least use the right-click context menu and copy the link location.

This problem was already mentioned in issue #1648 , but not fixed and further discussion let it to be labeled as a future enhancement. I suggest to implement this fix now and leave the future enhancement (adding Subscribe via Atom and Subscribe via JSON) for later.

### DIFF
--- a/client/src/app/shared/user-subscription/subscribe-button.component.html
+++ b/client/src/app/shared/user-subscription/subscribe-button.component.html
@@ -46,7 +46,7 @@
       <div class="dropdown-divider"></div>
 
       <h6 class="dropdown-header" i18n>Using a syndication feed</h6>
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Thanks @frankstrater!

Issue #2007, opened 2019-08-07
Ability to link to a certain time in video description


cdauth:
I hope this hasn't been requested before, at least I couldn’t find anything related.

In YouTube, when you type in a timestamp (for example `5:12`) into the video description or a comment, it is transformed into a link. When clicking the link, the player jumps to that time stamp and starts playing from there. This is commonly used to create a table of contents in the video description, or to comment about a certain moment in the video.

It would be nice if PeerTube would support this as well. Currently, the only way that I could find is to manually create a link to the video at the given start time (for example `[0:20](/videos/watch/abcdef?start=20)`). This is less than ideal, because:
* The video ID has to be specified (simply using `?start=20` as the URL does not work, the base URL seems to be that of the instance, not of the video)
* The link will open in a new tab, causing the video to play in two different tabs at the same time

Chocobozzz:
Thanks for reporting this issue, but this is a duplicate of https://github.com/Chocobozzz/PeerTube/issues/1312  :)

Issue #1728, opened 2019-03-22
Convert video time to links in comments


roipoussiere:
It could be interesting to allow users to post a comment with a link to a specific time on the current video.

Something as simple as:

> [0:42]() nice

From a quick look, since PeerTube already allows to put a time on the url with `?t=<time_offset>`, it's mainly a regex on the comment system. Maybe something like `(\d?\d[:'])?\d?\d[:']\d\d"?` that parses 
`0:42`, `0'42"`, `00:42`, `00'42"`, `13:37`, `1:13:37` and `01:13:37`.

Chocobozzz:
Thanks for reporting this issue, but this is a duplicate of https://github.com/Chocobozzz/PeerTube/issues/1312

PR #2157, opened 2019-10-02
Add hyperlink video timestamps in description

### LINKED ISSUE #1312: Automatically hyperlink video timecodes in description and comment text
This is a request for a new feature, where common video timecode formats (like 3:96, 1:23:40) will be identified automatically and hyperlinked to a timecode within the video on the current page.

This shorthand may come naturally to people, and other platforms support this sort of hyperlinking. It's especially useful for commenting on a specific moment in a video (lacking a specific way to do this?) or for writing descriptions that include a written table-of-contents.

For example, this is a long (12 hour!) video that includes a description I originally wrote for YouTube. On that platform, clicking the timecode will seek the player to that location. Not perfect, but it's nice to have a freeform solution for use cases like this that the software doesn't directly support yet.

https://diode.zone/videos/watch/2f1544f6-3a3b-4cd0-bb72-c57d2e2d1e9b

Lesterpig:
Fix #1312 (duplicates: #1728 and #2007)
The modification is also applied to comments and video editing.

### DIFF
--- a/client/src/app/shared/angular/timestamp-route-transformer.directive.ts
+++ b/client/src/app/shared/angular/timestamp-route-transformer.directive.ts
@@ -1,0 +1,45 @@
+import { Directive, ElementRef, HostListener, Output, EventEmitter } from '@angular/core'
+import { Router } from '@angular/router'
+
+type ElementEvent = Omit<Event, 'target'> & {
+  target: HTMLInputElement
+}
+
+@Directive({
+  selector: '[timestampRouteTransformer]'
+})
+export class TimestampRouteTransformerDirective {
+  @Output() timestampClicked = new EventEmitter<number>()
+
+  constructor (private el: ElementRef, private router: Router) { }
+
+  @HostListener('click', ['$event'])
+  public onClick ($event: ElementEvent) {
+    if ($event.target.hasAttribute('href')) {
+      const ngxLink = document.createElement('a')
+      ngxLink.href = $event.target.getAttribute('href')
+
+      // we only care about reflective links
+      if (ngxLink.host !== window.location.host) return
+
+      const ngxLinkParams = new URLSearchParams(ngxLink.search)
+      if (ngxLinkParams.has('start')) {
+        const separators = ['h', 'm', 's']
+        const start = ngxLinkParams
+          .get('start')
+          .match(new RegExp('(\\d{1,9}[' + separators.join('') + '])','g')) // match digits before any given separator
+          .map(t => {
+            if (t.includes('h')) return parseInt(t, 10) * 3600
+            if (t.includes('m')) return parseInt(t, 10) * 60
+            return parseInt(t, 10)
+          })
+          .reduce((acc, t) => acc + t)
+        this.timestampClicked.emit(start)
+      }
+
+      $event.preventDefault()
+    }
+
+    return
+  }
+}
--- a/client/src/app/shared/renderer/markdown.service.ts
+++ b/client/src/app/shared/renderer/markdown.service.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@angular/core'
 import { MarkdownIt } from 'markdown-it'
+import { buildVideoLink } from '../../../assets/player/utils'
 import { HtmlRendererService } from '@app/shared/renderer/html-renderer.service'
 
 type MarkdownParsers = {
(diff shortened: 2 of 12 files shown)


alcalyn:
Works well ! Although for now, it is reloading the full page when clicking on a timecode

Capture d’écran - 2019-10-07 à 18 34 41

Lesterpig:
Yep, if some has any experience with the player to avoid that refresh it would awesome :)

alcalyn:
I can make it works by adding in `client/src/app/videos/+video-watch/video-watch.component.ts`:

``` typescript
  bindTimecodeLinks () {
    const timecodeLinks = this.elementRef.nativeElement.querySelectorAll('.timecode')

    timecodeLinks.forEach((timecodeLink: HTMLElement) => {
      this.bindTimecodeLink(timecodeLink)
    })
  }

  bindTimecodeLink (timecodeLink: HTMLElement) {
    timecodeLink.addEventListener('click', (event: MouseEvent) => {
      event.preventDefault()

      const timecode: number = this.getTimecodeFromText((event.target as HTMLInputElement).innerText)

      this.player.currentTime(timecode)
      this.player.play()
    })
  }

  getTimecodeFromText (timecode: string): number {
    const tokens: number[] = timecode.split(':').map(n => parseInt(n, 10))

    const s: number = tokens.pop()
    const m: number = tokens.pop()
    const h: number = tokens.pop()

    return s + (m ? m * 60 : 0) + (h ? h * 3600 : 0)
  }
```

and a class to your timecode links:

``` diff
-      return `<a href="${url}">${str}</a>`
+      return `<a class="timecode" href="${url}">${str}</a>`
```

But I don't know when to proper bind button, the only solution I found for now is:

``` typescript
  ngAfterViewInit () {
    setTimeout(() => {
      this.bindTimecodeLinks()
    }, 1000)
  }
```

Also, I think we could create a markdownIt plugin to move the methods inside.

Lesterpig:
Thanks for taking the time to look at this :+1: 
I'll try to apply your method to my commits during the week-end

Chocobozzz:
Instead of using query selectors, you could also use a directive. See for example: https://stackoverflow.com/a/49006667

kontrollanten:
Great work! I'd propose that we accept the full page reload for now, and then creates an issue for making it smoother in the future.

client/src/app/shared/renderer/markdown.service.ts
```diff
@@ -130,7 +139,7 @@ export class MarkdownService {
   private avoidTruncatedTags (html: string) {
     return html.replace(/\*\*?([^*]+)$/, '$1')
       .replace(/<a[^>]+>([^<]+)<\/a>\s*...((<\/p>)|(<\/li>)|(<\/strong>))?$/mi, '$1...')
-      .replace(/\[[^\]]+\]\(([^\)]+)$/m, '$1')
+      .replace(/\[[^\]]+\]?\(?([^\)]+)$/, '$1')
```
Chocobozzz: ?

client/src/app/videos/+video-watch/comment/video-comment.component.ts
```diff
@@ -78,16 +79,16 @@ export class VideoCommentComponent implements OnInit, OnChanges {
   }
 
   isRemovableByUser () {
-    return this.comment.account && this.isUserLoggedIn() &&
+    return this.isUserLoggedIn() &&
```
Chocobozzz: why did you remove `this.comment.account`?

client/src/app/videos/+video-watch/comment/video-comment.component.ts
```diff
@@ -78,16 +79,16 @@ export class VideoCommentComponent implements OnInit, OnChanges {
...
       (
         this.user.account.id === this.comment.account.id ||
         this.user.hasRight(UserRight.REMOVE_ANY_VIDEO_COMMENT)
       )
   }
 
   private async init () {
-    this.sanitizedCommentHTML = await this.markdownService.textMarkdownToHTML(this.comment.text, true)
-
+    const safeHTML = await this.htmlRenderer.toSafeHtml(this.comment.text)
```
Chocobozzz: I think there is a rebase issue here no?

client/src/app/shared/angular/timestamp-route-transformer.directive.ts
```diff
@@ -0,0 +1,34 @@
...
+      const ngxLink = document.createElement('a')
+      ngxLink.href = $event.target.getAttribute('href')
+
+      // we only care about reflective links
+      if (ngxLink.host !== window.location.host) return
+
+      const ngxLinkParams = new URLSearchParams(ngxLink.search)
+      if (ngxLinkParams.has('start')) this.timestampClicked.emit(parseInt(ngxLinkParams.get('start'), 10))
+
+      this.router.navigateByUrl(ngxLink.pathname + ngxLink.search + ngxLink.hash)
```
Chocobozzz: I'm not sure we should update the link (youtube does not seem to update it), because you could have some undesired effects: you handle URL param change + handle the link click in the watch page

client/src/app/videos/+video-watch/comment/video-comment.component.ts
rigelk: Ah, that's an artifact from the messy conflict resolution I went through. Sorry about that.

client/src/app/shared/angular/timestamp-route-transformer.directive.ts
rigelk: I can remove the link update altogether.

Chocobozzz: I think keeping the link is fine (in case where your open the link in another tab)

Chocobozzz:
Thanks @rigelk & @Lesterpig

Issue #507, opened 2018-04-14
Mark deleted comment with replies rather than removing completely


ChameleonScales:
When you delete a comment that has replies, it also deletes the replies, which may not be ideal.
A suggestion would be to mark the comment with a "Removed by user" in order to keep its replies and not break the tree of comments.

BO41:
I would like to try to solve this issue. But I don't find where the actual comment gets deleted.
somewhere at `server/controllers/api/videos/comment.ts`, am I right?

Instead of removing the comment and its thread, I should be able to change the owner or the content to "removed by user" and leave the thread as it is.

Chocobozzz:
It's removed here: https://github.com/Chocobozzz/PeerTube/blob/develop/server/controllers/api/videos/comment.ts#L146

You could add a boolean column to the video comment model (and create a new migration script) `deleted` for example.

Chocobozzz:
Implemented by @alcalyn in https://github.com/Chocobozzz/PeerTube/commit/b5206dfc455c119b0dcb897bd7144be6ea4d999d & https://github.com/Chocobozzz/PeerTube/commit/69222afac8f8c41d90295b33f0695bbff352851e

PR #2165, opened 2019-10-05
Soft delete comments so that entire thread is not deleted

alcalyn:
Related to #507 

Currently, when deleting a comment that have replies, the comment and all the replies were removed.

This PR only soft deleted comments so that replies stay visible.

Screenshot_2019-11-25 date-toggle-demo

- [x] Back: Update tests to assert comment is still there, but marked as deleted
- [x] Back: Update server to soft delete comment instead of destroy
- [x] Back: Update API to no longer returns comment text and account if soft deleted
- [x] Back: Migration to make `comment.accountId` optional and add the `deletedAt` column
- [x] Federation: Replace deleted comments with a `Tombstone` object
- [x] Federation: Handle comment deletion federation requests by marking comments as deleted
- [x] Federation: Handle deleted comments when following a new instance and fetching all comments for the first time
- [x] Front: Display deleted comment as deleted on page load
- [x] Front: Display deleted comment as deleted dynamically when clicking on Delete

### DIFF
--- a/client/src/app/videos/+video-watch/comment/video-comment.model.ts
+++ b/client/src/app/videos/+video-watch/comment/video-comment.model.ts
@@ -12,6 +12,8 @@ export class VideoComment implements VideoCommentServerModel {
   videoId: number
   createdAt: Date | string
   updatedAt: Date | string
+  deletedAt: Date | string
+  isDeleted: boolean
   account: AccountInterface
   totalReplies: number
   by: string
@@ -28,14 +30,18 @@ export class VideoComment implements VideoCommentServerModel {
     this.videoId = hash.videoId
     this.createdAt = new Date(hash.createdAt.toString())
     this.updatedAt = new Date(hash.updatedAt.toString())
+    this.deletedAt = hash.deletedAt ? new Date(hash.deletedAt.toString()) : null
+    this.isDeleted = hash.isDeleted
     this.account = hash.account
     this.totalReplies = hash.totalReplies
 
-    this.by = Actor.CREATE_BY_STRING(this.account.name, this.account.host)
-    this.accountAvatarUrl = Actor.GET_ACTOR_AVATAR_URL(this.account)
+    if (this.account) {
+      this.by = Actor.CREATE_BY_STRING(this.account.name, this.account.host)
+      this.accountAvatarUrl = Actor.GET_ACTOR_AVATAR_URL(this.account)
 
-    const absoluteAPIUrl = getAbsoluteAPIUrl()
-    const thisHost = new URL(absoluteAPIUrl).host
-    this.isLocal = this.account.host.trim() === thisHost
+      const absoluteAPIUrl = getAbsoluteAPIUrl()
+      const thisHost = new URL(absoluteAPIUrl).host
+      this.isLocal = this.account.host.trim() === thisHost
+    }
   }
 }
--- a/server/models/video/video-comment.ts
+++ b/server/models/video/video-comment.ts
@@ -507,17 +507,6 @@ export class VideoCommentModel extends Model<VideoCommentModel> {
   }
 
   toActivityPubObject (this: MCommentAP, threadParentComments: MCommentOwner[]): VideoCommentObject | ActivityTombstoneObject {
-    if (this.isDeleted()) {
-      return {
-        id: this.url,
-        type: 'Tombstone',
-        formerType: 'Note',
-        published: this.createdAt.toISOString(),
-        updated: this.updatedAt.toISOString(),
-        deleted: this.deletedAt.toISOString()
-      }
-    }
-
     let inReplyTo: string
     // New thread, so in AS we reply to the video
     if (this.inReplyToCommentId === null) {
@@ -526,8 +515,22 @@ export class VideoCommentModel extends Model<VideoCommentModel> {
       inReplyTo = this.InReplyToVideoComment.url
     }
 
+    if (this.isDeleted()) {
+      return {
+        id: this.url,
+        type: 'Tombstone',
+        formerType: 'Note',
+        inReplyTo,
+        published: this.createdAt.toISOString(),
+        updated: this.updatedAt.toISOString(),
+        deleted: this.deletedAt.toISOString()
+      }
+    }
+
     const tag: ActivityTagObject[] = []
     for (const parentComment of threadParentComments) {
+      if (!parentComment.Account) continue
+
       const actor = parentComment.Account.Actor
 
       tag.push({
--- a/server/models/video/video-comment.ts
+++ b/server/models/video/video-comment.ts
@@ -1,5 +1,5 @@
 import { AllowNull, BelongsTo, Column, CreatedAt, DataType, ForeignKey, Is, Model, Scopes, Table, UpdatedAt } from 'sequelize-typescript'
-import { ActivityTagObject } from '../../../shared/models/activitypub/objects/common-objects'
+import { ActivityTagObject, ActivityTombstoneObject } from '../../../shared/models/activitypub/objects/common-objects'
 import { VideoCommentObject } from '../../../shared/models/activitypub/objects/video-comment-object'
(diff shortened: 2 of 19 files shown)


alcalyn:
I'm having trouble with:

- `comment.totalReplies` which is still either `0` or `undefined`
- I don't know how to run the migration I created (tried to restart server, to `npm run build:server`...)

rigelk:
> I don't know how to run the migration I created (tried to restart server, to `npm run build:server`...)

Have you incremented LAST_MIGRATION_VERSION in `server/initializers/constants.ts`?

alcalyn:
Ha, didn't aware of this constant. Just incremented it, run `npm run build:server` and it worked! Thank you

server/models/video/video-comment.ts
```diff
@@ -1,4 +1,4 @@
-import { AllowNull, BelongsTo, Column, CreatedAt, DataType, ForeignKey, Is, Model, Scopes, Table, UpdatedAt } from 'sequelize-typescript'
+import { AllowNull, BelongsTo, Column, CreatedAt, DataType, DeletedAt, ForeignKey, Is, Model, Scopes, Table, UpdatedAt } from 'sequelize-typescript'
```
LeoMouyna: @alcalyn This line seems too long according to travis :upside_down_face:

client/src/app/videos/+video-watch/comment/video-comment.model.ts
```diff
@@ -28,6 +30,8 @@ export class VideoComment implements VideoCommentServerModel {
     this.videoId = hash.videoId
     this.createdAt = new Date(hash.createdAt.toString())
     this.updatedAt = new Date(hash.updatedAt.toString())
+    this.deletedAt = null === hash.deletedAt ? null : new Date(hash.deletedAt.toString())
```
LeoMouyna: nitpick: I'm not sure this equal checker is need.
'has.deletedAt ? new Date(hash.deletedAt.toString()) : null' should do the same.

alcalyn: thanks for the review :)

Chocobozzz:
@alcalyn Did you manage to fix your `totalReplies` issue?

I think it would be simpler to not use the paranoid mode and instead manually update the `deletedAt` field, because we also need to remove the `text` field (set to null or empty string).

We also have to deal with federation:
 * Add a `deleted` field and use a `Tombstone` object: https://www.w3.org/TR/activitypub/#delete-activity-outbox
 * Correctly process this `deleted` field

alcalyn:
Hello, yes I could add the scope `ScopeNames.ATTRIBUTES_FOR_API` but it works only for top-level comments. I couldn't have replies count for replies.

I was not sure if we should also erase comment from db. I'll try to do it without paranoid.

Same for account that posted the comment: if we also delete it, we have to make it optionnal, and make many changes

Chocobozzz:
> Same for account that posted the comment: if we also delete it, we have to make it optionnal, and make many changes

I don't think we have the choice. Users expects that we don't keep this information if they delete their comment. If you did not plan to make these changes, we can merge your work in a WIP branch and I'll continue it when I'll have time :)

alcalyn:
I may have time to finish it, I'll redo it without `paranoid`

alcalyn:
**I get issues** when running them, it looks like the migration I created is not run on testing database.

I'm running tests with:

``` bash
$ npm run build -- --light
$ npm run clean:server:test

$ npm run mocha -- --exit --require ts-node/register/type-check --bail server/tests/api/videos/video-comments.ts 

> peertube@2.0.0 mocha /var/www/PeerTube
> mocha "--exit" "--require" "ts-node/register/type-check" "--bail" "server/tests/api/videos/video-comments.ts"

  Test video comments
    ✓ Should not have threads on this video (42ms)
    ✓ Should create a thread in this video (141ms)
    ✓ Should list threads of this video (84ms)
    ✓ Should get all the thread created (38ms)
    ✓ Should create multiple replies in this thread (467ms)
    ✓ Should get correctly the replies (52ms)
    ✓ Should create other threads (193ms)
    ✓ Should list the threads (44ms)
    1) Should delete a reply

  8 passing (19s)
  1 failing

  1) Test video comments
       Should delete a reply:
     Error: expected 204 "No Content", got 500 "Internal Server Error"
```

The error is `null value in column "accountId" violates not-null constraint`:

```
$ tail -n 2 test1/logs/peertube.log 
{"err":"SequelizeDatabaseError: null value in column \"accountId\" violates not-null constraint [...] }
{"message":"127.0.0.1 - - [23/Nov/2019:13:26:37 +0000] \"DELETE /api/v1/videos/1/comments/4 HTTP/1.1\" 500 - \"-\" \"node-superagent/3.8.3\"\n","level":"info","label":"localhost:9001","timestamp":"2019-11-23T13:26:37.059Z"}
```

I think I have to run migrations on test database but I don't know how to.

### About Tombstone object

The api now returns a Tombstone for deleted comments:

**before deletion**:
```
$ curl -X GET http://0.0.0.0:9000/videos/watch/e7c647ea-a9bc-4eee-99b2-82109ec70a5c/comments/17 -H "accept: application/activity+json" | jq

{
  "type": "Note",
  "id": "http://localhost:9000/videos/watch/e7c647ea-a9bc-4eee-99b2-82109ec70a5c/comments/17",
  "content": "Hello",
  "inReplyTo": "http://localhost:9000/videos/watch/e7c647ea-a9bc-4eee-99b2-82109ec70a5c",
  "updated": "2019-11-23T13:14:39.228Z",
  "published": "2019-11-23T13:14:39.199Z",
  "url": "http://localhost:9000/videos/watch/e7c647ea-a9bc-4eee-99b2-82109ec70a5c/comments/17",
  "attributedTo": "http://localhost:9000/accounts/root",
  "tag": [],
  "to": [
    "https://www.w3.org/ns/activitystreams#Public"
  ],
  "cc": [
    "http://localhost:9000/accounts/root/followers"
  ],
  "@context": ["..."]
}
```

**after deletion**:
```
$ curl -X GET http://0.0.0.0:9000/videos/watch/e7c647ea-a9bc-4eee-99b2-82109ec70a5c/comments/17 -H "accept: application/activity+json" | jq

{
  "@context": ["..."],
  "id": "http://localhost:9000/videos/watch/e7c647ea-a9bc-4eee-99b2-82109ec70a5c/comments/17",
  "type": "Tombstone",
  "formerType": "Note",
  "published": "2019-11-23T13:14:39.199Z",
  "updated": "2019-11-23T13:14:55.596Z",
  "deleted": "2019-11-23T13:14:55.592Z"
}
```

Chocobozzz:
You need to allow `null`  [here](https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/video/video-comment.ts#L180) and [here](https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/account/account.ts#L204)

alcalyn:
Thanks it worked, migrations was actually run. I'm working on others tests about federation.

alcalyn:
Should be ready to review

Chocobozzz:
Great job @alcalyn! Just one thing: you'll have to handle Tombstone objects in [this file](https://github.com/Chocobozzz/PeerTube/blob/develop/server/lib/activitypub/video-comments.ts), so that when you follow a new server and crawl its outbox, you create the appropriate tree. Without that, you'll miss some comment trees because the remote body won't be considered [as valid](https://github.com/Chocobozzz/PeerTube/blob/develop/server/lib/activitypub/video-comments.ts#L129)

alcalyn:
Ok I didn't tried this use case yet. I'm adding tests and fix it

alcalyn:
I fixed crawl of deleted comments when subscribing a new instance in this commit: https://github.com/Chocobozzz/PeerTube/pull/2165/commits/b1f7c4f9d3222a116b5d8305131653651e23c364

Chocobozzz:
Thanks @alcalyn :+1:

PR #2180, opened 2019-10-17
Update user-edit.component.html

### LINKED ISSUE #1943: Missing autocomplete attribute on user's form fields
**What happened?**

As an administrator, when I edit a user to change his quota, my browser keep the e-mail in memory, and propose it to me on other websites login forms.

You should add an autocomplete="off" on fields. I'm not quite sure it is enought (see https://developer.mozilla.org/fr/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#The_autocomplete_attribute_and_login_fields)

**Additional information**
* PeerTube version or URL: 1.3.1
* Browser name/version: Firefox

Knackie:
fixes #1943


Chocobozzz:
Thanks

Issue #2014, opened 2019-08-11
Allow user to disable video in the player


ZettaScript:
**Enhancement**
Add an option in the player to allow user to disable video (play and download only soundtrack).

**Why?**
When playing music videos, I don't care about the image. Disable video would reduce used bandwidth and CPU.

**Next step?**
Allow to upload sound files? Allow to disable video by default when editing video settings.

Chocobozzz:
Hi,

I guess that if you're on another tab, your web browser optimize video playing so the CPU overhead is insignificant. Regarding bandwidth, h264 is very good to optimizing static images so again, the overhead is insignificant. We don't want to add this complexity in PeerTube for so little gain.

> Allow to upload sound files

People can already upload audio files if their admin allows it.

Silmathoron:
@Chocobozzz the CPU overhead might be insignificant but what about data transfer? Is h264 that efficient?

Nutomic:
Also, music videos might not necessarily have a static image. There could be animation, or an actual music video playing, which would take a lot of bandwidth and cpu for transmission and decoding.

Yetangitu:
I see a few possibilities here which could help to improve the utility of PT in these situations:

 - have the transcoder optionally create an audio-only version of imported material using either a new parameter (--extract-audio or something along those lines) which is added to the resolution chooser in the player. This audio-only version can be generated using the video thumbnail for a static image.
 - The same as above without the static image, i.e. an audio-only mp4 container. I have not tried yet whether the web player supports these but I know that regular players (vlc, mplayer, mpv et al) have not problems with such files or streams.

While h264 is an effective codec it still adds overhead for any static image added to the stream. Testing with Firefox (nightly, 72.0a1) on my lowly Thinkpad T42p makes the thing jump to 90% CPU when playing audio tracks with static imagery. Playing a true audio-only stream takes no more than 5% so there is a marked difference.

I'll do some experiments around this theme as such a facility would be a marked improvement for one of my use cases: video lectures. Listening to such while on the road gobbles up ~400MB/hour for a low-res video version while audio-only versions don't take much more than ⅒ of that.

Yetangitu:
I made a proof of concept of how this could work by adding a `0p` transcoding step to the list of available resolutions which runs a new `presetAudioSplit()` preset in _ffmpeg-utils_ which runs _ffmpeg_ with `-vn -acodec copy`, producing `mp4` files without video streams.

[...]

Issue #1437, opened 2018-11-29
Create a sound file when transcoding


roipoussiere:
Sometime I want to download only the sound of a video, typically to get a podcast from a conference in order to listen it later.

It could be nice to have a music format (such as mp3 and/or ogg) available in the formats list when we download a video.

Chocobozzz:
Thanks for the issue but it is a duplicate of https://github.com/Chocobozzz/PeerTube/issues/939

zicmama:
[REC] Machine from https://www.copylaradio.com is showing how to make a DIY webradio player / recorder that could do the job (through Rompr and REC function. It is still a simple draft...
Hope soon it uses IPFS as global backup & collective Jukebox App.

Yetangitu:
I implemented a proof of concept for this feature, see https://github.com/Chocobozzz/PeerTube/issues/2014#issuecomment-547968994 and https://github.com/Chocobozzz/PeerTube/issues/2014#issuecomment-548726861. The implementation can be found in the `audio-only` branch of my _PeerTube_ fork:

https://github.com/Yetangitu/PeerTube/tree/audio-only

PR #2213, opened 2019-11-01
Add audio-only option to transcoders and player

Yetangitu:
This patch adds an audio-only option to PeerTube by means of a new transcoding configuration which creates mp4 files which only contain an audio stream. This new transcoder has a resolution of '0' and is presented in the preferences and in the player resolution menu as 'Audio-only' (localised). When playing such streams the player shows the file thumbnail as background and disables controls autohide.

Audio-only files can be shared and streamed just like any other file. They can be downloaded as well, the resulting file will be an mp4 container with a single audio stream.

Here's how audio-only files are presented in the player:

image image

...and in the transcoding preferences:

image

This patch is a proof of concept to show the feasbility of 'true' audio-only support. There are better ways of doing this which also enable multiple audio streams for a given video stream (e.g. DASH) but as this would entail a fundamental change in the way PeerTube works it is a bridge too far for a simple proof of concept.

Earlier discussion on this patch can be found in #2014.


Yetangitu:
Travis failures are caused by the thing running out of space:

`Error: createdb: database creation failed: ERROR:  could not write to file "base/113554/2657": No space left on device `

Yetangitu:
Kicked Travis but it still fails with timeouts (probably related to), running out of disk space:

`UnhandledPromiseRejectionWarning: Error: createdb: database creation failed: ERROR:  could not write to file "base/112149/2609_fsm": No space left on device`

Chocobozzz:
Thnaks @Yetangitu, it's interesting. 

But is this change worth it? I mean, I understand that with low computers playing a video consumes more CPU than an audio file. But when you play an audio video on PeerTube, I think you don't stay in the tab. Does your web browser still consumes a lot of CPU if you are on another tab, or reduce your web browser?

Regarding bandwidth, do you have some hints about what is the overhead compared to a static image merged with an audio file?

I think we could merge this PR depending on the answers of the questions above, but I'm afraid it would be a duplicate with https://github.com/Chocobozzz/PeerTube/issues/1437 (people don't expect to have a mp4 file when they want to download the audio stream of a video). Moreover, I think that with our new HLS player (that uses [hls.js](https://github.com/video-dev/hls.js/)), we can split the audio and the video streams, and maybe only play the audio stream (but I did not test it).

Silmathoron:
@Chocobozzz there is a noticeable difference (see [comment in previous discussion](https://github.com/Chocobozzz/PeerTube/issues/2014#issuecomment-546741491)) and apparently the [audio download can be m4a](https://github.com/Chocobozzz/PeerTube/issues/2014#issuecomment-548736720)

Nutomic:
The HLS player still seems to be far from production ready as far as I can tell. So in that case, it would make sense to use this implementation as a temporary workaround for audio only playback.

Yetangitu:
If the HLS player can be brought up relatively soon - as in months, not years - I'd say wait for that. Should this not be feasible this patch can be used as a stop-gap measure but in that case I'll go over it with a fine-toothed comb to make it 'production-ready'. Once the HLS player is ready the audio-only files created by this patch can be replaced with whatever format that player requires.

Yetangitu:
> Thanks @Yetangitu, it's interesting.
> 
> But is this change worth it? I mean, I understand that with low computers playing a video consumes more CPU than an audio file. But when you play an audio video on PeerTube, I think you don't stay in the tab. Does your web browser still consumes a lot of CPU if you are on another tab, or reduce your web browser?

Video increases CPU/GPU load when in a foreground tab or on a mobile device. If this load can be avoided... and more importantly, if the bandwidth requirements can be lowered...

> Regarding bandwidth, do you have some hints about what is the overhead compared to a static image merged with an audio file?

For a 5-minute piece of music the storage/bandwidth required for the video stream (at 720p) was slightly higher than that required by the audio stream:

```
audio-only:              4876271 bytes
audio+360p static image: 7526283 bytes
audio+720p static image: 10540647 bytes
```

Part of the appeal of an audio-only 'resolution' is that it is indelibly linked to the video source and its metadata and (by virtue of being generated automatically) it is always available, this in contrast to 'manually uploaded' audio+static_image versions which are separate videos as far as the system is concerned and as such get/need their own metadata.

> I think we could merge this PR depending on the answers of the questions above, but I'm afraid it would be a duplicate with #1437 (people don't expect to have a mp4 file when they want to download the audio stream of a video). Moreover, I think that with our new HLS player (that uses [hls.js](https://github.com/video-dev/hls.js/)), we can split the audio and the video streams, and maybe only play the audio stream (but I did not test it).

How are those audio and video streams served? For 'true' audio-only they should be served as two separate streams which are multiplexed in the client, otherwise the 'audio-only' version will still consume the same bandwidth as the AV-version.

Chocobozzz:
> The HLS player still seems to be far from production ready as far as I can tell.

@Nutomic Why?

Nutomic:
@Chocobozzz Because it is still considered experimental. But I havent tried it, so maybe I'm wrong.

Yetangitu:
> Moreover, I think that with our new HLS player (that uses [hls.js](https://github.com/video-dev/hls.js/)), we can split the audio and the video streams, and maybe only play the audio stream (but I did not test it).

I enabled HLS and ran `npm run optimize-old-videos`, it optimised the one video on the test instance. Restarted and ran it through some tests -> no difference that I could notice. What incantations should be done to test this _new HLS player_?

Nutomic:
@Yetangitu That script doesnt do anything about HLS, it optimizes videos which have a too high bitrate. But it seems like there is no way to convert old videos to HLS, so that would be worth an issue.

https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/tools.md#optimize-old-videosjs

Yetangitu:
After re-enabling HLS (which got disabled in the config file by... something...) and re-uploading a video it now generates HLS playlists. The good news is that a true audio-only file (as in 'feed PeerTube an audio-only file') plays in the HLS player, so far so good. It does not yet work with a static background (which should be an easy fix) and - here's the real problem - it doesn't separate audio and video at the server. In other words, the HLS player as currently implemented does not fit the purpose of serving true audio-only versions. Maybe this can be changed in some way, I'll have a look at this but I don't know if it is an easy fix.

Chocobozzz:
> - it doesn't separate audio and video at the server. In other words, the HLS player as currently implemented does not fit the purpose of serving true audio-only versions

Yes, for now we only generate 1 mp4 file that includes audio & video, so we can serve it (statically of so people download the file). In the future we could imagine giving the choice to admins.

Yetangitu:
I assume the final implementation will do away with the double set of transcoded files? The `-fragmented` versions work just fine for the normal player so there is no need to generate 'non-fragmented' versions for non-HLS purposes as is done now.

Yetangitu:
Here's the resulting files for a single video (4:41 long, audio with a still image):

```
frank@sloth:/srv/peertube/storage$ ls -l videos/1a0fb9d8-6501-4b50-bab4-80a5513449d9-*
-rw-r--r-- 1 frank src  4439139 Nov  8 13:24 videos/1a0fb9d8-6501-4b50-bab4-80a5513449d9-0.mp4
-rw-r--r-- 1 frank src 12527544 Nov  8 13:25 videos/1a0fb9d8-6501-4b50-bab4-80a5513449d9-360.mp4
-rw-r--r-- 1 frank src 17618147 Nov  8 13:24 videos/1a0fb9d8-6501-4b50-bab4-80a5513449d9-480.mp4
frank@sloth:/srv/peertube/storage$ ls -l streaming-playlists/hls/1a0fb9d8-6501-4b50-bab4-80a5513449d9/
total 33860
-rw-r--r-- 1 frank src     7491 Nov  8 13:25 0.m3u8
-rw-r--r-- 1 frank src  4450369 Nov  8 13:25 1a0fb9d8-6501-4b50-bab4-80a5513449d9-0-fragmented.mp4
-rw-r--r-- 1 frank src 12511450 Nov  8 13:25 1a0fb9d8-6501-4b50-bab4-80a5513449d9-360-fragmented.mp4
-rw-r--r-- 1 frank src 17602021 Nov  8 13:25 1a0fb9d8-6501-4b50-bab4-80a5513449d9-480-fragmented.mp4
-rw-r--r-- 1 frank src     7737 Nov  8 13:25 360.m3u8
-rw-r--r-- 1 frank src     7754 Nov  8 13:25 480.m3u8
-rw-r--r-- 1 frank src      236 Nov  8 13:25 master.m3u8
-rw-r--r-- 1 frank src    18319 Nov  8 13:25 segments-sha256.json
```

Chocobozzz:
> The -fragmented versions work just fine for the normal player so there is no need to generate 'non-fragmented' versions for non-HLS purposes as is done now.

videostream, the library we use to stream videos from webtorrent does not support fragmented mp4

Chocobozzz:
I'll merge manually this PR.

Yetangitu:
OK

Chocobozzz:
Merged. But we have an issue with the HLS player, hls.js does not seem to be able to switch between a container with audio + video to a container without audio.

Yetangitu:
> Merged. But we have an issue with the HLS player, hls.js does not seem to be able to switch between a container with audio + video to a container without audio.

I'll look into this one of these days, a bit busy now.

PR #2215, opened 2019-11-02
Remove unused function searchVideoWithPagination

fflorent:
It seems that this function is used nowhere. Do you think it's fine to remove it?

Florent


Chocobozzz:
Thanks

PR #2219, opened 2019-11-04
Fix hours in peertube-import-videos

fflorent:
Youtube doesn't store the hour of the publish date so
set it to zero in order to compare it.

### DIFF
--- a/server/tools/peertube-import-videos.ts
+++ b/server/tools/peertube-import-videos.ts
@@ -375,6 +375,7 @@ function parseDate (dateAsStr: string): Date {
     exitError(`Invalid date passed: ${dateAsStr}. Expected format: YYYY-MM-DD. See help for usage.`);
   }
   const date = new Date(dateAsStr);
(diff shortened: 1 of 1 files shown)


Chocobozzz:
I'm not sure to understand. Youtube uses a format like `YYYY-MM-DD` without hours/minutes/seconds?

fflorent:
> Youtube uses a format like YYYY-MM-DD without hours/minutes/seconds?

Yes… Actually `YYYYMMDD` but youtube-dl formats it correctly to a timestamp.

Florent

Chocobozzz:
Okay but hours, minutes and seconds should be automatically set to 0 when you don't provide them:

```
> new Date('2019-11-04')
2019-11-04T00:00:00.000Z
```

fflorent:
Oh, true, but it depends on the locale timezone… That's why I propose this patch (I should have mentioned it, sorry).

PR #2231, opened 2019-11-12
Fix link for account creation on another instance

Yorwba:
The link currently points to https://joinpeertube.org/en/#instances-list , but the instance list has been moved to https://joinpeertube.org/instances#instances-list


Chocobozzz:
Thanks @Yorwba

PR #2247, opened 2019-11-14
Sticking to one env-var management system

Leopere:
Starting with the Postgres container so that you only have to set overlapping environment variables in one place.


support/docker/production/docker-compose.yml
```diff
@@ -47,15 +47,16 @@ services:
...
+    env_file:
+      - .env
     volumes:
       - ./docker-volume/db:/var/lib/postgresql/data
     restart: "always"
     labels:
       traefik.enable: "false"
+    networks:
+      default:
+        ipv4_address: 172.20.0.3
```
Chocobozzz: Why is it necessary?

support/docker/production/.env
```diff
@@ -14,6 +18,7 @@ PEERTUBE_SMTP_FROM=noreply@domain.tld
 PEERTUBE_SMTP_TLS=false
 PEERTUBE_SMTP_DISABLE_STARTTLS=false
 PEERTUBE_ADMIN_EMAIL=admin@domain.tld
+POSTFIX_myhostname=${PEERTUBE_WEBSERVER_HOSTNAME}
```
Chocobozzz: Where does this variable is used?

support/docker/production/docker-compose.yml
Leopere: My instance was struggling to get the internal docker DNS resolution to work. I figured since you'd taken the time to define the network's internal IP when we could do something like my definition for my Traefik proxy network.
```YAML
networks:
  default:
    ipam:
      driver: default
      config:
      - subnet:  172.20.0.0/16
  traefik:
    external: true
```
I figured why not I could remove that or you could before this merges, but the rest should probably merge so that we can have one unified location for configuration if not in the `docker-compose.yml` itself the `.env` can be used.

Leopere: Removed other change so the envvar change can be merged.

Chocobozzz:
Thanks

PR #2258, opened 2019-11-16
Clarify and expand on admin credentials detail

Leopere:
It felt unclear so I clarified and suggested a method for obtaining the credentials easier.


Chocobozzz:
Thanks

PR #2283, opened 2019-11-24
Add missing button roles for the language chooser and keyboard shortcut menu items

### LINKED ISSUE #2282: Accessibility: Change Language and Show Keyboard Commands buttons are missing the "button" role
**What happened?**

My screen reader only announced "Change language - clickable" and "Show Keyboard Shortcuts - clickable".

My screen reader should have announced the above mentioned items as "buttons", as it does with others like the "Share" or "Subscribe" buttons.

**Steps to reproduce:**

1. Visit any PeerTube instance with your browser, like Safari on Mac or Firefox on Windows. Can also be Chrome if need be.

2. Open the Accessibility Inspector in Firefox and turn on the Accessibility Engine. Then, right-click the Show Keyboard Shortcuts item and select "Inspect Accessibility".

3. The role of that item should be "Button", not "Text frame".

**Additional information**
* PeerTube version or URL: v2.0.0.
* Browser name/version: Firefox 70
* NodeJS version: N/A

* Link to browser console log if useful: N/A
* Link to server log if useful (journalctl or /var/www/peertube/storage/logs/): N/A

MarcoZehe:
Fixes #2282 .

### DIFF
--- a/client/src/app/menu/menu.component.html
+++ b/client/src/app/menu/menu.component.html
@@ -104,11 +104,11 @@
 
     <div class="footer d-flex justify-content-between">
       <span class="language">
-        <span tabindex="0" (keyup.enter)="openLanguageChooser()" (click)="openLanguageChooser()" i18n-title title="Change the language" class="icon icon-language"></span>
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks @MarcoZehe

PR #2288, opened 2019-11-26
Update production.md

DontUseGithub:
small inconsistency fix

### DIFF
--- a/support/doc/production.md
+++ b/support/doc/production.md
@@ -127,7 +127,9 @@ Now you have the certificates you can reload nginx:
 $ sudo systemctl reload nginx
 ```
 
-**FreeBSD**
+**On FreeBSD**
+
+```
 On FreeBSD you can use [Dehydrated](https://dehydrated.io/) `security/dehydrated` for [Let's Encrypt](https://letsencrypt.org/)
 
 ```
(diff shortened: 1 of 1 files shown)


Chocobozzz:
I don't understand the purpose of this PR

DontUseGithub:
The patter on that page is like this - bold & big text and then from a new line is the regular text, example:

### **Database**

Create the production database and a peertube user inside PostgreSQL:

____

But those two situations from the PR are different, they're in line:

**FreeBSD** On FreeBSD you can use Dehydrated security/dehydrated for Let's Encrypt

____

But should be like this:

### **FreeBSD** 

On FreeBSD you can use Dehydrated security/dehydrated for Let's Encrypt

DontUseGithub:
The pattern on that page is like this - bold & big text and then from a new line is the regular text, example:

### **Database**

Create the production database and a peertube user inside PostgreSQL:

____

But what we currently have is this (they're in line):

**FreeBSD** On FreeBSD you can use Dehydrated security/dehydrated for Let's Encrypt

____

But should be like this:

### **FreeBSD** 

On FreeBSD you can use Dehydrated security/dehydrated for Let's Encrypt

____

BTW, why have you deleted that META issue? That is important. I remeber in Chocobozzz/PeerTube
there was a closed issue about SEO. Now if you search for it, you [won't find it](https://github.com/Chocobozzz/PeerTube/search?q=seo&type=Issues). I bet that person has deleted his account. Who knows how many more issues like these will just get lost.

Chocobozzz:
Please check your changes then, it breaks production.md file.

PR #2302, opened 2019-12-03
improve video-watch title display on small screens

rigelk:
This PR is only modifying css/html for the video-watch page, and only regarding the title block and the recommendations block. Its aim is to improve:

- long title display on small screens
- space taken when no recommendation is available

Below, on a 1100px-wide viewport before the PR, and after the PR:

Screenshot_2019-12-03 Soutenons notre Internet(2)

Screenshot_2019-12-03 Soutenons notre Internet(3)

### DIFF
--- a/client/src/app/videos/+video-watch/video-watch.component.html
+++ b/client/src/app/videos/+video-watch/video-watch.component.html
@@ -39,70 +39,82 @@
     <div class="video-info">
       <div class="video-info-first-row">
         <div>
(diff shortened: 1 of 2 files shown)


Chocobozzz:
This breaks the layout 
screen_2019-12-04-08:39:29

rigelk:
> This breaks the layout
> screen_2019-12-04-08:39:29

that should be fixed now that I removed the part about the recommendations.

Issue #1928, opened 2019-06-20
Password Reset Link expires without warning


realityfabric:
**What happened?**
I used the "Forgot My Password" feature. I received an email with the link, but I did not open the link until 20+ minutes after I had received it. (The email had been sent to spam)

When attempting to reset my password, I was able to fill in the password and confirm password fields, but received an error message when I clicked to submit.

I tested multiple lengths of password, just in case I was using a password that was too long.

I used the feature again and used the new link almost immediately (less than 3 minutes) and was able to change my password and login without any errors or issues.

**What do you expect to happen instead?**
I expected to be able to change my password. I think a timeout is a good idea, but there should be a warning *before* filling out the password reset form. Either in the email, or on the reset form itself (preferably not even showing the form if the link is expired), and ideally both places. At the very least the error message should be more clear about why the submission isn't working.

**Steps to reproduce:**
1. Be logged out of any PeerTube accounts.
2. Go to the login page.
3. Click "I forgot my password"
4. Enter your account email address
5. Confirm that you received an email, and wait 20+ minutes (possibly less)
6. Follow the link in the email. You will be taken to a password reset form.
7. Fill out the password reset form (with a new password and the password confirmation)
8. Click submit. You will receive an error message.
9. Try various passwords if you want, they should all receive the same error message.

**Additional information**
* PeerTube version or URL: https://peertube.social
* Browser name/version: Firefox 60.7.1esr

NassimBounouas:
Hi @realityfabric,

I'm working on your issue. I found that there is a life time of 5 minutes set [there](https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/constants.ts#L474).

What I'm suggesting as a first step is to add this life time on the reset form as you requested and add it to the email sent to the user.

Then it could be interesting to call a validator inspired by [this one](https://github.com/Chocobozzz/PeerTube/blob/develop/server/middlewares/validators/users.ts#L318) when the user is using the reset link in the email. By doing this, we can display him/her an error message instead of a reset form.

Question to @Chocobozzz : Currently the `USER_PASSWORD_RESET_LIFETIME` is hard coded to 5 minutes. It could be interesting to retrieve this value from the configuration to let the instance owner to increase or decrease it. What do you think about this idea ?

Nutomic:
Anyway 5 minutes seems way too short for this. I think it would be fine to set the timeout to one hour or more.

NassimBounouas:
[...]

PR #2305, opened 2019-12-04
Feature/password reset link expiration

NassimBounouas:
#1928


NassimBounouas:
@Chocobozzz : The token is, now, stored into redis with a TTL of 3600 seconds. The frontend indicated the hard coded 1 hour limitation (notification & mail). 
Since the commits are really messy, the PR requires a squash. 

As you suggested, we can plan to use another key in the future.

Chocobozzz:
Thanks @NassimBounouas

Issue #1654, opened 2019-02-21
Suggestions for search filtering improvements


AutomCoding:
I have played around a bit with the search filters today and have some suggestions for improvements. They are small and somewhat related, so I collected them all here in a discussion type issue.

1. After searching with the filters **Published date** or **Duration**, it is not possible to reset the filter to display all without reloading the page (and lose all other parameters). I have considered the following options, of which the first is my favorite since it is probably the most intuitive and I like the second least because it is easier to see all options (when they are so few).
    - Add **Show all** radio buttons
    - Replace radio buttons with dropdown menus
    - Add **Reset** buttons (and make it visually obvious it doesn't apply to the entire form)
2. When filtering on language, videos with subtitles in that language are still excluded from the result. Translating videos take a lot of time and effort and at the moment, there is no way to discover them. Of my two proposals I like the second one the most, as the user probably would enter the same language in both dropdowns. The only problem is if the user *only wants to see videos with subtitles* in the selected language.
    - Add another dropdown menu for subtitle language
    - Add a **Show videos with subtitles (in this language)** checkbutton
    - Include subtitled videos in the result and leave the form as it is
3. The number-of-applied-filters counter goes up when you select something from a dropdown menu, but doesn't go down when selecting the blank, upper-most option. If I do that on the category dropdown, does that mean the result will be videos from all categories, or only uncategorized videos? This is also true for the other filters, but they do not have a selectable default value.
    - Make the counter tick down and replace the empty label with something more descriptive: **All categories**, **All licences**, et cetera
    - Make the counter tick down
    - Make the search engine function as the UI currently implies
4. It is not obvious how to separate the tags under **All of these tags** and **One of these tags**. I think the second option is better, but it will require much more time to make.
    - Write a help text, possibly hidden with one of those question mark symbols triggering a popup
    - Create an interactive thing [like this page selector](https://tools.wmflabs.org/pageviews/) which creates a deletable box for each tag (page in the linked example)

It is possible and probable there are better solutions than my proposals to these problems. So, what do you think?

Chocobozzz:
Thanks for this interesting issue. My opinion:

 1. Add **Show all** radio buttons
 2. Add a **Show videos with subtitles (in this language)** checkbox
 3. Make the counter tick down **(including the "Show all" radio buttons)** and replace the empty label with something more descriptive: **All categories**, **All licences**, et cetera
[...]

PR #2306, opened 2019-12-04
search filtering improvements per #1654

rigelk:
Only thing I left for later: the "Show videos with subtitles (in this language)" checkbox.


rigelk:
Note that the global reset is not perfect: it doesn't reset "Published date" or "Duration" yet.

PR #2316, opened 2019-12-08
lighter box-shadow for .dropdown-menu

rigelk:
Before/After

image(2)


Chocobozzz:
Thanks, I prefer this new version!

PR #2324, opened 2019-12-10
Add support for HLS in the Docker container.

BenLubar:


support/docker/production/Dockerfile.stretch
```diff
@@ -1,4 +1,5 @@
-FROM node:10-stretch-slim
+# use buster for new enough ffmpeg
```
rigelk: Please remove the comment, and instead specifiy that kind of extraneous information in the body of your PR.

Chocobozzz:
Thanks @BenLubar!

support/docker/production/Dockerfile.stretch
BenLubar: I added it as a comment because the file is named stretch

rigelk:
@BenLubar could you rename the file to end in "buster"? That would make more sense.

Chocobozzz:
I'll do it

PR #2325, opened 2019-12-10
"autoplay next video" switch for both user and visitors in video watch view

### LINKED ISSUE #1891: Autoplay switch on videos pages
An enable/disable autoplay button on each video page can be usefull.

rigelk:
edit: the autoplay is temporary, unless logged-in.

Screenshot_2019-12-10

resolves #1891

---

_also included in commit_: change to the placeholder for the search bar.


rigelk:
@Chocobozzz should I switch from localStorage to sessionStorage, so that the parameter is automatically forgotten and resets to falsy once the tab is closed? Or do you want persistence accross tabs (in which case I have to implement an expiration to the localStorage wrapper we use)?

Chocobozzz:
I don't understand why you used the local storage and the user settings. IMO you should just use the user settings. But I think we should have two different parameters: one to autoplay the next recommended video, and another one to autoplay playlists. Because as a peertube user, I don't want to autoplay the next recommended video but I want to autoplay playlists

rigelk:
@Chocobozzz I used the user settings, and added the localStorage to provide the functionality for visitors.

Any idea (UX-wise) how to differentiate the parameter for playlists and the one for regular videos?

Chocobozzz:
> @Chocobozzz I used the user settings, and added the localStorage to provide the functionality for visitors.

Oh right, then I think we can use the local storage :+1: I don't think we care about the local storage expiration.

> Any idea (UX-wise) how to differentiate the parameter for playlists and the one for regular videos?

Maybe **Autoplay next video** and **Autoplay playlist**?

rigelk:
I added the requested switch, albeit under a different form:

Screenshot from 2019-12-11 22-16-34

As you can see I also added a loop playlist setting.

I also finished to use sessionStorage for the loop playlist and autoplay video setting.

client/src/app/videos/+video-watch/video-watch-playlist.component.html
```diff
@@ -14,6 +14,26 @@
         <span>{{ currentPlaylistPosition }}</span><span>{{ playlistPagination.totalItems }}</span>
       </div>
     </div>
+
+    <div class="playlist-controls">
+      <my-global-icon
```
Chocobozzz: IMO the switch with a "auto play playlist" would be better

client/src/app/videos/+video-watch/video-watch-playlist.component.html
```diff
@@ -14,6 +14,26 @@
...
+        iconName="videos"
+        [class.active]="autoPlayNextVideoPlaylist"
+        (click)="switchAutoPlayNextVideoPlaylist()"
+        [ngbTooltip]="autoPlayNextVideoPlaylistSwitchText"
+        placement="bottom auto"
+        container="body"
+      ></my-global-icon>
+
+      <my-global-icon
```
Chocobozzz: :+1:

client/src/app/videos/+video-watch/video-watch-playlist.component.ts
```diff
@@ -23,14 +29,29 @@ export class VideoWatchPlaylistComponent {
...
     private videoPlaylist: VideoPlaylistService,
     private router: Router
-  ) {}
+  ) {
+    this.autoPlayNextVideoPlaylist = this.auth.isLoggedIn()
+      ? this.auth.getUser().autoPlayNextVideoPlaylist
+      : peertubeLocalStorage.getItem(VideoWatchPlaylistComponent.LOCAL_STORAGE_AUTO_PLAY_NEXT_VIDEO_PLAYLIST) !== 'false'
+    this.setAutoPlayNextVideoPlaylistSwitchText()
+
+    this.loopPlaylist = peertubeSessionStorage.getItem(VideoWatchPlaylistComponent.SESSION_STORAGE_AUTO_PLAY_NEXT_VIDEO_PLAYLIST) === 'true'
```
Chocobozzz: Why is it `!== 'false'` above and now `=== 'true'`?

client/src/app/videos/+video-watch/video-watch-playlist.component.ts
```diff
@@ -119,6 +140,54 @@ export class VideoWatchPlaylistComponent {
       const start = next.startTimestamp
       const stop = next.stopTimestamp
       this.router.navigate([],{ queryParams: { videoId: next.video.uuid, start, stop } })
+    } else if (this.loopPlaylist) {
+      this.currentPlaylistPosition = 0
+      this.navigateToNextPlaylistVideo(this.playlistElements.find(e => e.position === this.currentPlaylistPosition))
```
Chocobozzz: Please create a dedicated method to find the next element of a playlist depending on the position

client/src/app/videos/+video-watch/video-watch.component.ts
```diff
@@ -435,15 +436,24 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
 
       this.player.one('ended', () => {
         if (this.playlist) {
-          this.zone.run(() => this.videoWatchPlaylist.navigateToNextPlaylistVideo())
-        } else if (this.user && this.user.autoPlayNextVideo) {
+          if (
+            this.user && this.user.autoPlayNextVideoPlaylist ||
```
Chocobozzz: put this condition in a dedicated method `isPlaylistAutoPlayEnabled()` for example

client/src/app/videos/+video-watch/video-watch.component.ts
```diff
@@ -559,7 +569,7 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
   }) {
     const { video, videoCaptions, urlOptions, user } = params
 
-    let startTime = timeToInt(urlOptions.startTime) || (video.userHistory ? video.userHistory.currentTime : 0)
+    let startTime = timeToInt(urlOptions.startTime) || (video.userHistory && !this.playlist ? video.userHistory.currentTime : 0)
```
Chocobozzz: Same here, it's now too complex for a single line: please create a method

client/src/app/videos/+video-watch/video-watch.module.ts
```diff
@@ -12,14 +12,16 @@ import { NgbTooltipModule } from '@ng-bootstrap/ng-bootstrap'
...
 
 @NgModule({
   imports: [
     VideoWatchRoutingModule,
     SharedModule,
     NgbTooltipModule,
     QRCodeModule,
-    RecommendationsModule
+    RecommendationsModule,
+    InputSwitchModule
```
Chocobozzz: I think you'll have issues when you'll load your account page (after a watch page load), since we already import this module in the my account module.

client/src/app/videos/recommendations/recommended-videos.component.html
```diff
@@ -1,12 +1,20 @@
...
+        <p-inputSwitch [(ngModel)]="autoPlayNextVideo" (ngModelChange)="switchAutoPlayNextVideo()"></p-inputSwitch>
+      </div>
     </div>
 
-    <div *ngFor="let video of (videos$ | async)">
+    <div *ngFor="let video of (videos$ | async); let i = index; let length = count">
       <my-video-miniature [video]="video" [user]="user" (videoBlacklisted)="onVideoRemoved()" (videoRemoved)="onVideoRemoved()">
       </my-video-miniature>
+      
+      <hr *ngIf="!playlist && i == 0 && length > 1" />
```
Chocobozzz: If length === 0 this block should not be displayed (because there are no iteration in the ngfor)

client/src/assets/images/global/repeat.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="feather feather-repeat"><polyline points="17 1 21 5 17 9"></polyline><path d="M3 11V9a4 4 0 0 1 4-4h14"></path><polyline points="7 23 3 19 7 15"></polyline><path d="M21 13v2a4 4 0 0 1-4 4H3"></path></svg>
```
Chocobozzz: Where does this icon come from?

client/src/app/videos/+video-watch/video-watch.module.ts
rigelk: I didn't notice errors there.

client/src/app/videos/+video-watch/video-watch-playlist.component.html
rigelk: I strongly disagree. The "playlist-controls" area is bound to have multiple buttons like this one: loop, autoplay, random, etc. Having 3 switches will clutter the space, whereas with icons like here, the area stays lean.

client/src/app/videos/+video-watch/video-watch-playlist.component.ts
rigelk: The difference lies with how we interpret non-existent values (default values).

client/src/app/videos/+video-watch/video-watch-playlist.component.html
Chocobozzz: I would just keep the switch for the auto play, and the icons for the other features. But as you want, I don't have a strong opinion here

client/src/app/videos/+video-watch/video-watch-playlist.component.ts
Chocobozzz: Then please put a comment (`default is "false"` for example) or abstract this in a separate method/class

client/src/assets/images/global/repeat.svg
rigelk: https://www.iconfinder.com/icons/5402418/arrow_cycle_loop_repeat_replay_icon

client/src/app/videos/recommendations/recommended-videos.component.html
rigelk: I'm only interested in the case where this is the first element of a list that contains more than one video recommended, and the current video played is not part of a playlist.

Chocobozzz: Yep sorry :+1:

rigelk:
@Chocobozzz note I also changed the search bar border to give it a more modern look.

PR #2330, opened 2019-12-11
Slightly relax Cipher Suite hardening in nginx example config

mqus:
This enables legacy software like apps on android 4.4.2 and matches the traefik configuration, where the specific cipher suite is already allowed.

This problem mostly stemmed from trying to get the Newpipe Peertube support working on older android devices (https://github.com/TeamNewPipe/NewPipe/pull/2792). E.g. Framatube.org seems to use the example config and won't work with android apps on android 4.4.2 which don't implement the crypto on their own but use the system libraries. 

The solution for this would be to add a single cipher suite which does work with android kitkat (`TLS_RSA_WITH_AES_256_CBC_SHA`) and is already enabled in the [traefik config](https://github.com/Chocobozzz/PeerTube/blob/develop/support/docker/production/config/traefik.toml#L27) for the docker file. SSLLabs calls this Ciphersuite "weak" but not "insecure". Other possibilities for a reasonably secure Ciphersuite which are supported by Android <=4.4.2 are:
```
TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA
TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA
TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA
TLS_RSA_WITH_AES_128_CBC_SHA
TLS_RSA_WITH_AES_256_CBC_SHA <-this is the one currently in question
```

To be clear, in the future this change should probably be reversed again, but for now there are still many devices running 4.4 and 4.2 out there. Also: this only matters for client apps like NewPipe (and probably Thorium, if it would support Android <5), not Web Browsers, as they come with their own, far more current crypto implementations.

Please let me know what you think and if I should take another cipher, I'm no crypto expert (nor a peertube expert) by any means. I just have a smartphone with Android 4.4.2 and want things to work 😉 

If this is merged, It would be great to let instance owners know and adapt their configurations accordingly.


Chocobozzz:
Thanks @mqus. You might want to update our [nginx template too](https://github.com/Chocobozzz/PeerTube/blob/develop/support/nginx/peertube#L28) too I guess

mqus:
Thanks!

PR #2349, opened 2019-12-21
Design suggestion for the action buttons row in video watch page

rigelk:
I've been wondering about the look and feel of buttons in the watch page. To me, the row of buttons looks bulky in a part of the UI where visual space is scarce. The colors (especially with the support button adding yet another color) don't mix well with the gray of the buttons' background:
Screenshot_2019-12-21 UWwVx62

Option 1: transparent background and inverted support button
Screenshot_2019-12-20 UWwVx62(2)

Option 2: transparent background and normal support button that reverts back to the current design on hover
Screenshot_2019-12-20 UWwVx62(3)

Option 3: (my preference and the current content of the PR) option 2 + neutralize the green/red of like/dislike buttons, and (resp.) use the main color and a darker shade of the foreground color.
Screenshot_2019-12-20 UWwVx62(4)

Note I switched text to uppercase since it balances the lack of background in terms of noticeability of the buttons.

### DIFF
--- a/client/src/app/videos/+video-watch/video-watch.component.scss
+++ b/client/src/app/videos/+video-watch/video-watch.component.scss
@@ -220,66 +220,73 @@ $video-info-margin-left: 44px;
           .action-button:not(:first-child),
           .action-dropdown,
           my-video-actions-dropdown {
-            margin-left: 10px;
+            margin-left: 5px;
           }
 
-          .action-button {
+          ::ng-deep.action-button {
             @include peertube-button;
-            @include grey-button;
             @include button-with-icon(21px, 0, -1px);
-            @include apply-svg-color($grey-foreground-color);
+            @include apply-svg-color(var(--actionButtonColor));
 
-            font-size: 15px;
+            font-size: 100%;
             font-weight: $font-semibold;
             display: inline-block;
             padding: 0 10px 0 10px;
             white-space: nowrap;
+            background-color: transparent !important;
+            color: var(--actionButtonColor);
+            text-transform: uppercase;
 
             &::after {
               display: none;
             }
 
-            .action-button-like,
-            .action-button-dislike {
+            &:hover {
+              opacity: 0.9;
+            }
+
+            &.action-button-like,
+            &.action-button-dislike {
+              filter: brightness(120%);
+
               .count {
                 margin-right: 5px;
               }
             }
 
             &.action-button-like.activated {
-              background-color: $green;
-
               .count {
-                color: #fff;
+                color: $activated-action-button-color;
               }
 
               my-global-icon {
-                @include apply-svg-color(#fff);
+                @include apply-svg-color($activated-action-button-color);
               }
             }
 
             &.action-button-dislike.activated {
-              background-color: $red;
-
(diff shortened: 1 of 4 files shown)


Chocobozzz:
Option 1: :+1: 
Option 2: :+1: but we need to find a way to emphasis the support button
Option 3: :-1: because I don't think it's a good idea to use the main instance color (that could be green or red or dark etc).

rigelk:
I'll implement option 1 in develop since it is a first step for all other options anyway.

DontUseGithub:
> I'll implement option 1 in develop since it is a first step for all other options anyway.

I'd suggest to not make the " support button" overlap with the upvote/downvote line. IMO it should be next to it, not on top of it.

rigelk:
@Chocobozzz WDYT about that support button color and color zone?

Screenshot_2019-12-24 UWwVx62(1)

The above also tries a smaller likes/dislikes line, @DontUseGithub.

DontUseGithub:
@rigelk If I may, I'd like to give my opinions as well. I think having three different colors so close to each other is too much. Maybe make the pink heart either green or red.

rigelk:
> I think having three different colors so close to each other is too much. Maybe make the pink heart either green or red.

I'd take it the other way around. Make the like/dislike buttons darkgrey/grey, so that the support button stands out (?)

DontUseGithub:
> > I think having three different colors so close to each other is too much. Maybe make the pink heart either green or red.
> 
> I'd take it the other way around. Make the like/dislike buttons darkgrey/grey, so that the support button stands out (?)

Sounds great to me!

alcalyn:
I like the pink heart in the last screenshot, even if there is 3 differents colors close to each others.

Or at most the heart could be green to reflect with the +1 button

rigelk:
Alright. Since I kind of agree using the main color of an instance can lead to various colours being used and blur the meaning of the buttons, I removed the colors to provide relative emphasis to the support button:

Screenshot_2019-12-26 UWwVx62(1)
Screenshot_2019-12-26 UWwVx62(2)

I'll default with the green support button.

client/src/app/videos/+video-watch/video-watch.component.scss
```diff
@@ -212,6 +212,8 @@ $video-info-margin-left: 44px;
         width: max-content;
         margin-left: auto;
 
+        $activated-action-button-color: black;
```
Chocobozzz: Please put this in the variables files

Chocobozzz:
Thanks all for this PR. @rigelk Merge this when you want

DontUseGithub:
@rigelk I think the second screenshot looks better.
Also, what if we hide (move) the SHARE and SAVE buttons under the 'three dots' button to that the UI look less bulky?

rigelk:
> what if we hide (move) the SHARE and SAVE buttons under the 'three dots' button to make the UI look less crowded/bulky?

The support button takes precedence, and I would argue that the share button is more important than the like/dislike buttons in PeerTube.

DontUseGithub:
Makes sense, what about the SAVE button?

Issue #1250, opened 2018-10-12
Jobs list too large for screen


rhaamo:
**What happened?**

Jobs list is too large and the job queue selector is hidden on the right

**What do you expect to happen instead?**

Jobs list fit on screen or move the dropdown selector in a always-visible part

**Steps to reproduce:**

1. Go to jobs list

**Additional information**
* PeerTube version or URL: PeerTube v1.0.0...d9d1989b (current develop)
* Browser name/version: Firefox
* Screen: 12.5" 1366x768 (minus gnome dock on left)

![image](https://user-images.githubusercontent.com/8625/46853494-8c0f4b80-cdfe-11e8-8d48-43f1c0dce3ad.png)

Chocobozzz:
You can remove the menu, maybe you'll have more space to display the table

I don't know how to handle the admin tables on small screen :thinking:

rigelk:
@Chocobozzz maybe we can cheat a little and use primeng's selectable columns (a select a the top of the table that lets the user the possibility to toggle the display of some columns). This would allow us to hide a column or two on small displays, but still allow the sure to see them provided they toggle them.

scanlime:
Re how any of this should look on a smaller screen:

1. A "mobile-first" design might choose to, instead of making this a wide tabular format, to have line-items for each item that can be expanded to reveal more details

2. A responsive design might have two modes, a tabular mode and a mode like in (1)

3. The simplest usable solution would be to set minimum width for the table to something that leaves enough room to interact, and accept that the table will need to scroll horizontally on mobile.

alcalyn:
Hi, I made a PR about that (see link just above), you can view screenshots and give feedback

PR #2362, opened 2019-12-27
Enhance jobs list displaying on smaller screens

alcalyn:
Related issue: https://github.com/Chocobozzz/PeerTube/issues/1250

Jobs list, now:
jobs

Now, in dark theme:
jobs-dark

This PR makes jobs list table less large so that it can fit in small screens.

- It removes columns `processedOn` and `finishedOn` because not always used, and put them in details instead (on toggle).
- It removes the arrow toggle and make the entire row clickable to open details.
- It moves job current state to the end and display with an icon and colored.
- It fixes columns width and responsiveness


rigelk:
I merged manually in 427121217 and eb1c46e9a. Thanks for the UI rework @alcalyn, it is way clearer and more compact now!

Issue #520, opened 2018-04-21
Private instance (family, restricted community...)


AntoineMazuyer:
Hello,

My question can be weird, but it is possible to made video accessible only for logged user on my instance ? I would like to make a platform to share videos ONLY with my family and I don't know if peertube could be suitable for this use.

Thanks in advance !

mokas01:
Would be awesome, and complementary to https://github.com/Chocobozzz/PeerTube/issues/371

Chocobozzz:
For now you have 2 solutions:
 * Create 1 account that you'll share with your family and upload the videos in `private` privacy
 * Upload your videos in `unlisted` (they will be federated but people will need to find the correct URL to watch it) and share the URL with your family

To have a real restricted instance, we would need:
 * An admin option to forbid followers
 * An admin option to allow only logged users to list/watch videos

benrob0329:
Wouldn't this be equivalent to Mastodon's followers only privacy setting? With the account set up for non-automatic follow approval.

anmol26s:
On YunoHost you can keep the web interface behind the SSO. So only the users who are logged in on YunoHost will be able to access the web interface. Further only group of users on YunoHost can be given access to the web interface,all other users will be treated same as pubic users and will not have access to the web interface of PeerTube.

vincib:
I tried to use HTTP AUTH in the Nginx in front of the peertube, but any call by the js client to /api/ would not use the HTTP auth :/ 
no luck then. If anyone has a better idea :) 
(or I will end up with a port knocker on location / through a crude php script...)

hauspie:
> For now you have 2 solutions:
> 
>     * Create 1 account that you'll share with your family and upload the videos in `private` privacy
> 
>     * Upload your videos in `unlisted` (they will be federated but people will need to find the correct URL to watch it) and share the URL with your family
> 
> 
> To have a real restricted instance, we would need:
> 
>     * An admin option to forbid followers
> 
>     * An admin option to allow only logged users to list/watch videos

I do not think that a full private instance was what the OP has in mind. 

Actually what I would really like in an instance is the possibility to set a video as only viewable by the users of my instance, on a per video basis. This would allow a semi-public instance that can be federated with others but with some videos (or channels maybe) that are only listed if you are connected on the local instance (and of course not federated).

[...]

PR #2383, opened 2020-01-04
Private instance mode

Bricklou:
This PR is a solution for #520 which instance owner to make their instance private. But I think it is missing some static file and route blocking (for example videos or preview blocking when user is not logged in)

If private mode is enabled, my changes;
- Force the user to be logged in to access to videos
- Disable follow system
- Allow admin to update the mode from the configuration page


Chocobozzz:
What's the advantage of merging this PR compared to the new internal privacy setting?

Bricklou:
This merging lock all access by default, so users can put their videos in public and only logged user can access to them. If user is not logged in, all routes redirect to the login page. It is equivalent to the new system but it is for all the instance by default and can't be bypassed by users.

Chocobozzz:
> so users can put their videos in public and only logged user can access to them.

It's an inconsistent behaviour then. 

I don't think this is the right way to do it. It's a little bit complex and error prone: we need to add middlewares for many routes (in the client **and** in the server). We'll possibly forgot one resulting in a security issue. IMHO locking the privacy modes for your current instance is a cleaner and safer way to achieve the same goal.

Bricklou:
It's just a try, if it is necessary to rework all the system to avoid security problems, I can. But globally, the idea is there: ensure that access is restricted to guest persons on an Peertube instances.

And when I say that user can publish public video, I meant available for every logged in peoples.

rigelk:
@Bricklou as mentioned before, the current implementation changes confidentiality definitions, is hard to maintain in the long term and error-prone doing so. Since we integrated a new confidentiality mode for videos restricting access to logged-in users, the issue is less pressing.

Please open a new issue to properly discuss your envisioned functionality.

derMart:
@rigelk This confidentiality mode is really nice news. Having trouble to find this in the documentation, could you link me there?

ManuelBarrioLinares:
So... What is the best current way to achieve this level of privacy for an organization?

PR #2393, opened 2020-01-08
Add outbox page size parameter

### LINKED ISSUE #2230: Allow GET `size` parameter for OrderedCollectionPage
Going through the outbox of a PeerTube instance's outbox (or any route paginating a Collection, really), as pages are limited to 10 activities at per page, can take some time. I suggest adding a GET parameter to specify the number of activities contained per page (within lower/upper boundaries).

rigelk:
resolves #2230

### DIFF
--- a/server/middlewares/validators/activitypub/pagination.ts
+++ b/server/middlewares/validators/activitypub/pagination.ts
@@ -1,0 +1,23 @@
+import * as express from 'express'
+import { query } from 'express-validator'
+import { logger } from '../../../helpers/logger'
+import { areValidationErrors } from '../utils'
+
+const apPaginationValidator = [
+  query('page').optional().isInt({ min: 1 }).withMessage('Should have a valid page number'),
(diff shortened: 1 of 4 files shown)


Chocobozzz:
Limit the max pagination to 50 entries please, it's an heavy endpoint

server/middlewares/validators/activitypub/pagination.ts
```diff
@@ -0,0 +1,23 @@
+import * as express from 'express'
+import { query } from 'express-validator'
+import { logger } from '../../../helpers/logger'
+import { areValidationErrors } from '../utils'
+
+const apPaginationValidator = [
+  query('page').optional().isInt({ min: 1 }).withMessage('Should have a valid page number'),
+  query('size').optional().isInt({ min: 10, max: 50 }).withMessage('Should have a valid page size (min: 10, max: 50)'),
```
Chocobozzz: put them in https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/constants.ts#L182

also i don't think a minimum is really useful

Issue #2149, opened 2019-09-27
Code review (accessibility)


BenLubar:
(bugs marked with 🐛)

https://github.com/Chocobozzz/PeerTube/blob/9a18a6252071cf21b18f82a24bb63078abb75bc1/client/src/app/shared/video/video-thumbnail.component.html#L5

🐛 This should be using `aria-label` as the video title is not the ID of an element. Alternatively, since another nearby UI element already has the same action and label, just set it to `aria-hidden="true"` so screen readers don't say the name twice.

https://github.com/Chocobozzz/PeerTube/blob/1198edf4bb06ce5f1668b97cf9ca8fb483fe3f41/client/src/index.html#L1

This should have a `lang` attribute equal to the current language. Otherwise, screen readers will assume an arbitrary language (likely the system default) which could be wrong.

https://github.com/Chocobozzz/PeerTube/blob/457bb213b273a9b206cc5654eb085cede4e916ad/client/src/app/header/header.component.html#L2

Needs an `aria-label` attribute (something like "video search" or "search" would be fine)

[pretty much every use of `my-global-icon`]

Should have an `aria-label` attribute, and has a weird ➕-shaped hitbox in the browser if it's taller than the nearby text.

[wherever the glyphicons font-face is defined]

🐛 Should have `font-display: swap;` to avoid browsers blocking on this file.

[wherever link color is defined in css]

🐛 `#007ad9` is too dark to be easily readable in the dark theme - this should be changed to use a variable.

https://github.com/Chocobozzz/PeerTube/blob/5819e694ba238b5e4ee97347c6cf5ee7a82725b5/client/src/sass/include/_variables.scss#L9-L12

🐛 This shade of grey cannot easily be read on a dark background - this should be changed to use a variable.

https://github.com/Chocobozzz/PeerTube/blob/bfbd912886eba17b4aa9a40dcef2fddc685d85bf/client/src/app/videos/%2Bvideo-watch/video-watch-playlist.component.scss#L16

🐛 This makes playlist titles unreadable on the dark theme.

https://github.com/Chocobozzz/PeerTube/blob/0ba5f5baade94a051710d9d9d408ac8083c14ac6/client/src/app/%2Bmy-account/my-account-settings/my-account-change-password/my-account-change-password.component.html#L6-L25

[...]

PR #2400, opened 2020-01-10
Accessibility fixes for #2149

rigelk:
I left out the color changes for links, as well as anything related to contrast on the dark theme.

*note*: I took the opportunity to include plural correction for channel pages.

### DIFF
--- a/server/lib/client-html.ts
+++ b/server/lib/client-html.ts
@@ -26,7 +26,9 @@ export class ClientHtml {
   }
 
   static async getDefaultHTMLPage (req: express.Request, res: express.Response, paramLang?: string) {
-    const html = await ClientHtml.getIndexHTML(req, res, paramLang)
+    const html = paramLang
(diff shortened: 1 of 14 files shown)


server/lib/client-html.ts
```diff
@@ -129,6 +130,10 @@ export class ClientHtml {
     return join(__dirname, '../../../client/dist/' + buildFileLocale(lang) + '/index.html')
   }
 
+  private static addHtmlLang (htmlStringPage: string, paramLang?: string) {
```
Chocobozzz: `paramLang` should not be optional

server/lib/client-html.ts
```diff
@@ -98,6 +98,7 @@ export class ClientHtml {
 
     let html = buffer.toString()
 
+    if (paramLang) html = ClientHtml.addHtmlLang(html)
```
Chocobozzz: you forgot to put `paramLang` as param

PR #2411, opened 2020-01-14
Add video file metadata to download modal, via ffprobe

### LINKED ISSUE #479: Add video format/codec informations
Hello devs.

I've thought about a feature, which can be really useful, for when someone wanted to download a video: giving to viewer the ability to know technical information about video.
I give you an example:
> Video informations:
1920x1080px 1fps yuv420p vp9
opus 96kbits/s stereo
WebM - 4.4Mo

I've done that, manually, for [this video](https://peertube.xyz/videos/watch/b80c711a-c455-4535-b1f0-2d0cf1290e62).

I find that really interesting for downloaders, knowing size, codecs and technical stuff.

PS: maybe it's not a good idea, because download are into mp4 format (transcoding is done on server somewhere). Let me know about my idea :) (I think it's still a good idea :))

Thanks for all the Peertube work :+1:

rigelk:
resolves #479

Screenshot_2020-01-14 Nettoyer les têtes de son magnétoscope VHS


rigelk:
@Chocobozzz concerning the AP representation of a video file, it is usually as a `Link` entity, but [Link](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-link) that could contain metadata or link to its representation.

server/initializers/migrations/0475-video-file-metadata.ts
```diff
@@ -0,0 +1,24 @@
+import * as Sequelize from 'sequelize'
+
+async function up (utils: {
+  transaction: Sequelize.Transaction,
+  queryInterface: Sequelize.QueryInterface,
+  sequelize: Sequelize.Sequelize
+}): Promise<void> {
+
+  const data = {
+    type: Sequelize.JSON,
```
Chocobozzz: JSONB

server/lib/video-transcoding.ts
```diff
@@ -230,11 +231,13 @@ export {
 async function onVideoFileTranscoding (video: MVideoWithFile, videoFile: MVideoFile, transcodingPath: string, outputPath: string) {
   const stats = await stat(transcodingPath)
   const fps = await getVideoFileFPS(transcodingPath)
+  const metadata = await getMetadataFromFile(transcodingPath)
 
   await move(transcodingPath, outputPath)
 
   videoFile.size = stats.size
   videoFile.fps = fps
+  videoFile.metadata = metadata
```
Chocobozzz: please select the field you want to include in the json, it's better for security and safety

Chocobozzz:
> @Chocobozzz concerning the AP representation of a video file, it is usually as a Link entity, but Link cannot contain metadata or link to its representation. Should I create a new Link type to point to the metadata?

I think you could use `tag`, but I'm afraid the JSON is already big enough. Ideally, these metadata should not be stored by remote servers, but instead lazy loaded. It's the same for the web browser clients: do we really need these information when we fetch a video? IMO it would be better to fetch the metadata on demand

server/initializers/migrations/0475-video-file-metadata.ts
rigelk: type is `JSON` in model, should I change all to JSON? It is not like we are going to build queries on it.

drzraf:
Many API work using `response fields filtering` for that purpose.
For example: [Youtube parts](https://developers.google.com/youtube/v3/docs/videos/list#part):
Some of theses subfields object and corresponding quota (= retrieval cost) for requesting that data:
-     id: 0
-     fileDetails: 1
-     processingDetails: 1
-     contentDetails: 2
-     snippet: 2
-     statistics: 2
-     status: 2
There is also a [field](https://developers.google.com/youtube/v3/getting-started#fields) that defines how to further filter subfields of the response (to lower the `JSON payload size` cost).

That would be a nice and generic way to limit standard payload size increase without limiting (meta)data extension.

rigelk:
> Many API work using `response fields filtering` for that purpose.

The problem it is solving is arguably different. AP requests for the video object want the whole object representation as they don't know what part of it will be needed by their users.

server/models/video/video-format-utils.ts
```diff
@@ -193,7 +194,9 @@ function videoFilesModelToFormattedJSON (
         torrentUrl: model.getTorrentUrl(videoFile, baseUrlHttp),
         torrentDownloadUrl: model.getTorrentDownloadUrl(videoFile, baseUrlHttp),
         fileUrl: model.getVideoFileUrl(videoFile, baseUrlHttp),
-        fileDownloadUrl: model.getVideoFileDownloadUrl(videoFile, baseUrlHttp)
+        fileDownloadUrl: model.getVideoFileDownloadUrl(videoFile, baseUrlHttp),
+        metadata: videoFile.metadata,
```
Chocobozzz: I don't think we should include metadata in this JSON. Clients should always fetch the `metadaUrl` to get these information.

server/models/video/video-format-utils.ts
```diff
@@ -220,6 +223,17 @@ function addVideoFilesInAPAcc (
       fps: file.fps
     })
 
+    if (file.metadata) {
```
Chocobozzz: So we don't have to fetch metadata every time we want to create an AP JSON object, I think we should always include it in the object.

server/tests/api/videos/video-transcoder.ts
```diff
@@ -458,6 +458,46 @@ describe('Test video transcoding', function () {
...
+        // expected stream properties
+        'stream[0].codec_long_name',
+        'stream[0].profile',
+        'stream[0].width',
+        'stream[0].height',
+        'stream[0].display_aspect_ratio',
+        'stream[0].avg_frame_rate',
+        'stream[0].pix_fmt'
+      ]) {
+        expect(metadata).to.have.nested.property(p)
```
Chocobozzz: You also need to get on every instance:
 * The video
 * Then find the `metadataUrl` of every file
 * Get every `metadataUrl` and check its validity (just check 1 or 2 fields for example)

This way, you'll be able to see it federation works correctly.

client/src/app/shared/video/modals/video-download.component.ts
```diff
@@ -67,10 +75,26 @@ export class VideoDownloadComponent {
   getLink () {
     return this.type === 'subtitles' && this.videoCaptions
       ? this.getSubtitlesLink()
-      : this.getVideoLink()
+      : this.getVideoFileLink()
   }
 
-  getVideoLink () {
+  async onResolutionIdChange () {
+    this.videoFile = await this.getVideoFile()
```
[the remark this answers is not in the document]
rigelk: you mean "doesn't" ?

client/src/app/shared/video/modals/video-download.component.ts
```diff
@@ -104,4 +134,64 @@ export class VideoDownloadComponent {
...
+    delete format.tags
+
+    return mapValues(
+      pick(format, Object.keys(keyToTranslateFunction)),
+      (val, key) => keyToTranslateFunction[key](val)
+    )
+  }
+
+  getMetadataStream (streams: FfprobeStream[], type: 'video' | 'audio') {
+    const stream = streams.find(s => s.codec_type === type) || undefined
```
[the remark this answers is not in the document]
rigelk: ah right. I don't know why I did this :man_shrugging:

server/lib/video-transcoding.ts
```diff
@@ -230,11 +233,13 @@ export {
 async function onVideoFileTranscoding (video: MVideoWithFile, videoFile: MVideoFile, transcodingPath: string, outputPath: string) {
   const stats = await stat(transcodingPath)
   const fps = await getVideoFileFPS(transcodingPath)
+  const metadata = await getMetadataFromFile(transcodingPath)
 
   await move(transcodingPath, outputPath)
 
   videoFile.size = stats.size
   videoFile.fps = fps
+  videoFile.metadata = pick(metadata, [ 'streams', 'format', 'chapters' ])
```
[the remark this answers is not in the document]
rigelk: I want to be strict in what I store. If ever ffprobe changes adds keys, there will not be extra bloat added.
Arguably `chapters` is not exploited yet in the interface, but this is because I lacked an example to work on.

client/src/app/shared/video/modals/video-download.component.ts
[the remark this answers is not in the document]
Chocobozzz: yes sorry

server/helpers/ffmpeg-utils.ts
```diff
@@ -169,24 +169,27 @@ async function getVideoFileFPS (path: string) {
...
 
-async function getVideoFileBitrate (path: string) {
-  return new Promise<number>((res, rej) => {
+async function getMetadataFromFile<T> (path: string, cb = metadata => metadata) {
+  return new Promise<T>((res, rej) => {
     ffmpeg.ffprobe(path, (err, metadata) => {
       if (err) return rej(err)
 
-      return res(metadata.format.bit_rate)
+      delete metadata.format.filename // we want to remove information about the filesystem
```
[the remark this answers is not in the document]
rigelk: I want to keep the ffmpeg what it is: an util. Removing the filename could be done via a dedicated object, but elsewhere, for instance in that stage: https://github.com/Chocobozzz/PeerTube/pull/2411#discussion_r384820930

server/middlewares/validators/videos/videos.ts
```diff
@@ -192,6 +192,20 @@ const videosCustomGetValidator = (fetchType: 'all' | 'only-video' | 'only-video-
...
 
+const videoFileMetadataGetValidator = getCommonVideoEditAttributes().concat([
+  param('id').custom(isIdOrUUIDValid).not().isEmpty().withMessage('Should have a valid id'),
+  param('videoFileId').custom(isIdValid).not().isEmpty().withMessage('Should have a valid videoFileId'),
+
+  async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    logger.debug('Checking videoFileMetadataGet parameters', { parameters: req.params })
+
+    if (areValidationErrors(req, res)) return
+    if (!await doesVideoExist(req.params.id, res)) return
```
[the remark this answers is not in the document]
rigelk: That was a simplification on my part to test things, I had forgotten about it. Thanks for spotting it!

server/models/video/video-format-utils.ts
rigelk: What do you mean?

server/models/video/video-format-utils.ts
rigelk: Why? Updates will refresh the metadata anyway.

Chocobozzz: Let the remote instance decide (some of them may not be interested in metadata).

server/middlewares/validators/videos/videos.ts
```diff
@@ -198,6 +204,21 @@ const videosCustomGetValidator = (
...
```
Chocobozzz: I don't think we need this.

Instead, transform `doesVideoExistForVideoFile`  to `doesVideoFileOfVideoExist` and fetch `videoFile` according to an `id` and `videoId`. Just fetch `videoFile`, as you don't need `video`.

Also, instead of using `toInt` here you can use it line 209: `isEmpty().toInt()`

server/models/video/video-file.ts
```diff
@@ -29,6 +31,33 @@ import { MVideoFile, MVideoFileStreamingPlaylistVideo, MVideoFileVideo } from '.
...
+export enum ScopeNames {
+  WITH_VIDEO = 'WITH_VIDEO',
+  WITH_METADATA = 'WITH_METADATA'
+}
+
+const METADATA_FIELDS = [ 'metadata', 'metadataUrl' ]
+
+@DefaultScope(() => ({
+  attributes: {
+    exclude: METADATA_FIELDS
```
Chocobozzz: I think we need `metadataUrl`, or we won't be able to fetch metadata no?

I think your tests work because most of the time we use `VideoFileModel.unscoped()` in joins. I think you can remove the `unscoped()` call

server/tests/api/videos/video-transcoder.ts
```diff
@@ -458,6 +467,70 @@ describe('Test video transcoding', function () {
...
+      }
+      expect(metadata).to.not.have.nested.property('format.filename')
+    }
+
+    const otherServers = servers
```
Chocobozzz: Why do you exclude server 2 ? It's interesting to test it

server/models/video/video-file.ts
rigelk: Do you mean to remove the `unscoped()` call line 50?

Chocobozzz: No, I mean removing `VideoFileModel.unscoped()` in `models/video/video.ts` and `models/redundancy/video-redundancy.ts`

server/middlewares/validators/videos/videos.ts
rigelk: I still need to check for the video, since it gives me an `id` of the video whereas without it I just have its `uuid`.

rigelk: Ah, figured it out. nevermind.

server/models/video/video-file.ts
```diff
@@ -157,17 +194,20 @@ export class VideoFileModel extends Model<VideoFileModel> {
...
-    const options = {
-      include: [
-        {
-          model: VideoModel.unscoped(),
-          required: true
-        }
-      ]
-    }
+  static async doesVideoExistForVideoFile (id: number, videoIdOrUUID: number | string) {
+    const videoFile = await VideoFileModel.loadWithVideo(id)
```
Chocobozzz: `loadWithVideo` does not load the video file with the streaming playlist video. Please make a SQL request 

```
include: [
        {
          model: video model
          required: false,
          where { uuid or id: ... }
        },
        {
          model: streaming playlist model,
          required: false,
          include: [
                    { model: video model, required: true, where { uuid or id ... } }
          ]
      ],
```

PR #2412, opened 2020-01-14
Downsample to the closest divisor standard framerate

### LINKED ISSUE #2350: Convert 50fps videos to 25fps instead of 30fps
Converting 50fps videos to 30fps produces a jerky result. This could be avoided by setting the fps limit for low resolutions to 25fps is this case. What do you think ?

rigelk:
resolves #2350

The PR defines two sets of acceptable framerates: standard and hd standard framerates, which are used to check a video's conformance to player standards. If a video uses a non-standard framerate, we check if even downsampling is possible before rejecting.

### DIFF
--- a/server/helpers/ffmpeg-utils.ts
+++ b/server/helpers/ffmpeg-utils.ts
@@ -263,6 +263,10 @@ async function canDoQuickTranscode (path: string): Promise<boolean> {
   return true
 }
 
+function getClosestFramerateStandard (fps: number, hd = false): number {
+  return VIDEO_TRANSCODING_FPS[hd ? 'HD_STANDARD' : 'STANDARD'].slice(0).sort((a, b) => fps % a - fps % b)[0]
+}
+
 // ---------------------------------------------------------------------------
 
 export {
@@ -291,11 +295,11 @@ async function buildx264Command (command: ffmpeg.FfmpegCommand, options: Transco
     options.resolution !== undefined &&
     options.resolution < VIDEO_TRANSCODING_FPS.KEEP_ORIGIN_FPS_RESOLUTION_MIN &&
     fps > VIDEO_TRANSCODING_FPS.AVERAGE ||
-    // If the video is doesn't match had standard
-    !VIDEO_TRANSCODING_FPS.HD_STANDARD.map(value => fps % value).includes(0)
+    // If the video is doesn't match hd standard
+    !VIDEO_TRANSCODING_FPS.HD_STANDARD.some(value => fps % value === 0)
   ) {
     // Get closest standard framerate by modulo: downsampling has to be done to a divisor of the nominal fps value
-    fps = VIDEO_TRANSCODING_FPS.STANDARD.sort((a, b) => fps % a - fps % b)[0]
+    fps = getClosestFramerateStandard(fps)
   }
 
   command = await presetH264(command, options.inputPath, options.resolution, fps)
@@ -308,7 +312,7 @@ async function buildx264Command (command: ffmpeg.FfmpegCommand, options: Transco
 
   if (fps) {
     // Hard FPS limits
-    if (fps > VIDEO_TRANSCODING_FPS.MAX) fps = VIDEO_TRANSCODING_FPS.HD_STANDARD.sort((a, b) => fps % a - fps % b)[0]
+    if (fps > VIDEO_TRANSCODING_FPS.MAX) fps = getClosestFramerateStandard(fps, true)
     else if (fps < VIDEO_TRANSCODING_FPS.MIN) fps = VIDEO_TRANSCODING_FPS.MIN
 
     command = command.withFPS(fps)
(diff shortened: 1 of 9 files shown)


Chocobozzz:
We don't want to reject some videos with a specific framerate. Most users won't even understand the error message :)

rigelk:
@Chocobozzz That's what YT and Vimeo do, though.

server/helpers/ffmpeg-utils.ts
```diff
@@ -286,13 +286,16 @@ export {
...
     options.resolution !== undefined &&
     options.resolution < VIDEO_TRANSCODING_FPS.KEEP_ORIGIN_FPS_RESOLUTION_MIN &&
-    fps > VIDEO_TRANSCODING_FPS.AVERAGE
+    fps > VIDEO_TRANSCODING_FPS.AVERAGE ||
+    // If the video is doesn't match had standard
+    !VIDEO_TRANSCODING_FPS.HD_STANDARD.map(value => fps % value).includes(0)
   ) {
-    fps = VIDEO_TRANSCODING_FPS.AVERAGE
+    // Get closest standard framerate by modulo: downsampling has to be done to a divisor of the nominal fps value
+    fps = VIDEO_TRANSCODING_FPS.STANDARD.sort((a, b) => fps % a - fps % b)[0]
```
Chocobozzz: `sort` mutates `VIDEO_TRANSCODING_FPS.STANDARD`. To avoid side effects, please create a dedicated function that return the closest standard framerate value (using a simple loop for example)

server/helpers/ffmpeg-utils.ts
```diff
@@ -286,13 +286,16 @@ export {
...
   let fps = await getVideoFileFPS(options.inputPath)
-  // On small/medium resolutions, limit FPS
   if (
+    // On small/medium resolutions, limit FPS
     options.resolution !== undefined &&
     options.resolution < VIDEO_TRANSCODING_FPS.KEEP_ORIGIN_FPS_RESOLUTION_MIN &&
-    fps > VIDEO_TRANSCODING_FPS.AVERAGE
+    fps > VIDEO_TRANSCODING_FPS.AVERAGE ||
+    // If the video is doesn't match had standard
+    !VIDEO_TRANSCODING_FPS.HD_STANDARD.map(value => fps % value).includes(0)
```
Chocobozzz: You can use `some`

server/helpers/ffmpeg-utils.ts
```diff
@@ -286,13 +286,16 @@ export {
...
 async function buildx264Command (command: ffmpeg.FfmpegCommand, options: TranscodeOptions) {
   let fps = await getVideoFileFPS(options.inputPath)
-  // On small/medium resolutions, limit FPS
   if (
+    // On small/medium resolutions, limit FPS
     options.resolution !== undefined &&
     options.resolution < VIDEO_TRANSCODING_FPS.KEEP_ORIGIN_FPS_RESOLUTION_MIN &&
-    fps > VIDEO_TRANSCODING_FPS.AVERAGE
+    fps > VIDEO_TRANSCODING_FPS.AVERAGE ||
+    // If the video is doesn't match had standard
```
Chocobozzz: Typo

server/helpers/ffmpeg-utils.ts
```diff
@@ -305,7 +308,7 @@ async function buildx264Command (command: ffmpeg.FfmpegCommand, options: Transco
 
   if (fps) {
     // Hard FPS limits
-    if (fps > VIDEO_TRANSCODING_FPS.MAX) fps = VIDEO_TRANSCODING_FPS.MAX
+    if (fps > VIDEO_TRANSCODING_FPS.MAX) fps = VIDEO_TRANSCODING_FPS.HD_STANDARD.sort((a, b) => fps % a - fps % b)[0]
```
Chocobozzz: Same that the above comment

server/tests/api/videos/video-transcoder.ts
```diff
@@ -416,6 +417,42 @@ describe('Test video transcoding', function () {
...
+    await uploadVideo(servers[1].url, servers[1].accessToken, videoAttributes)
+
+    await waitJobs(servers)
+
+    for (const server of servers) {
+      const res = await getVideosList(server.url)
+
+      const video = res.body.data.find(v => v.name === videoAttributes.name)
+
+      for (const resolution of ['240']) {
```
Chocobozzz: what's the point of a loop here?

server/helpers/ffmpeg-utils.ts
rigelk: or I can `.slice(0)` the arry beforehand.

Chocobozzz: a loop in a function is simpler to read/understand and more efficient that copying an object to sort it and take the first element as a oneliner

PR #2415, opened 2020-01-16
Fix: unregistring a plugin did not remove properly his hooks.

JohnXLivingston:
I noticed that when developing a plugin, the hooks where called multiple times. Sometimes with old code. I found the issue here. It works with plugin under development. I did not try with plugins installed directory from npm, but I'm pretty sure it will be ok.


Chocobozzz:
Thanks @JohnXLivingston

PR #2416, opened 2020-01-16
Adding the request ip for hook filter:api.user.signup.allowed.result

JohnXLivingston:
The `filter:api.user.signup.allowed.result` hook is called in two places : on the register form submit, and in getConfig.
I propose passing the request IP in the getConfig call too.
The reason why: I'm developing a plugin that will allow/disallow the registration for some countries. So I can test before the form submit if I want. 
The plugin can still know if we are in getConfig or on the form submit, by testing `params.body`.


Chocobozzz:
Thanks @JohnXLivingston!

PR #2421, opened 2020-01-19
Add manage buttons for own channels and account, video counts

rigelk:
Own channels/account "Manage" buttons for quick access to administration:

Screenshot_2020-01-19 Video channel videos - PeerTube
Screenshot_2020-01-19 Account video channels - PeerTube

Video counts:

Screenshot_2020-01-19 Account videos - PeerTube
Screenshot_2020-01-19 Account playlists - PeerTube

### DIFF
--- a/client/src/app/+accounts/accounts.component.html
+++ b/client/src/app/+accounts/accounts.component.html
@@ -33,7 +33,7 @@
       </div>
 
       <div class="right-buttons">
-        <a *ngIf="isManageable" routerLink="/my-account" class="btn btn-outline-primary mr-2" i18n>Manage</a>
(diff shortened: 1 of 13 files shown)


Chocobozzz:
:+1: for this use case. But do we really want to add this blue button in our default theme? We don't have any blue for now

Instead of a badge, what about putting the number of elements before the title? `9 videos` or `17 playlists` for example (I don't know if it's really better, it's just to discussion)

rigelk:
@Chocobozzz I would care about the number being more discrete if the space were constrained, but that is not the case yet in these views, so a badge is fine for now. 

I guess it is time to have a new color to our theme, complementary to the main color 🤷🏽‍♂️

rigelk:
Screenshot_2020-01-20 Account video channels - PeerTube(1)

PR #2446, opened 2020-02-02
Channel identifier text description

DontUseGithub:
>  The channel name is a unique identifier of your channel on this and all the other instances.

this is true if we consider "cats@domain.com" as "a channel". If by "a channel" you only mean "cats" then the text needs to be changed to something like:

> The channel name is a unique identifier of your channel on this instance. Together with the domain name - it becomes unique across all instances.

In any case, English native speakers I'm open to suggestions on how to improve this :)

### DIFF
--- a/client/src/app/+signup/+register/register-step-channel.component.html
+++ b/client/src/app/+signup/+register/register-step-channel.component.html
@@ -40,7 +40,7 @@
     </div>
 
     <div class="name-information" i18n>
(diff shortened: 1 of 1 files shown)


Chocobozzz:
> The channel name is a unique identifier of your channel on this instance. Together with the domain name - it becomes unique across all instances.

:+1:

rigelk:
I will do the modifications after the merge. Thanks @DontUseGithub and @Aluriak!

PR #2447, opened 2020-02-02
Account identifier text description

DontUseGithub:
>  The username is a unique identifier of your account on this and all the other instances.

this is true if we consider "john@domain.com" as "an account". If by "an account" you only mean "john" then the text needs to be changed to something like:

> The username is a unique identifier of your account on this instances. Together with the domain name - it becomes unique across all instances.

In any case, English native speakers I'm open to suggestions on how to improve this :)

### DIFF
--- a/client/src/app/+signup/+register/register-step-user.component.html
+++ b/client/src/app/+signup/+register/register-step-user.component.html
@@ -29,7 +29,7 @@
     </div>
 
     <div class="name-information" i18n>
-      The username is a unique identifier of your account on this instance. It's like an address mail, so other people can find you.
+      The username is a unique identifier of your account on this and all the other instances. It's as unique as an email address, which makes it easy for other people to find it.
     </div>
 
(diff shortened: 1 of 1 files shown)


Aluriak:
Good improvement, crystal clear explanation. Why a dash instead of a comma ?

DontUseGithub:
@Aluriak Thank you. Right, the dash feels unnecessary, I'll remove it. 

Have you checked both commits, do you think they both look good?

- https://github.com/Chocobozzz/PeerTube/pull/2447/commits/bf94eb0e0527eb5467c942b75b4eab9036b206d1
- https://github.com/Chocobozzz/PeerTube/pull/2446/commits/09c982dbefac401b9c62c5229f4dc69788834e07

rigelk:
@DontUseGithub next time please do it in one PR, as they are closely related.

Aluriak:
> Have you checked both commits, do you think they both look good?

yes, i should have make the same remark there !

PR #2455, opened 2020-02-04
Restyle toast notifications to tone down colors

rigelk:
Screenshot from 2020-02-04 17-37-42


Chocobozzz:
What about notifications that display errors or warnings?

rigelk:
@Chocobozzz I added the missing `warn` style.

PR #2456, opened 2020-02-04
Move administration under user dropdown, disable hotkey cheatsheet on touchscreens

rigelk:


Chocobozzz:
Admins (including me) won't like to have to go in the dropdown to go in the administration page. It's an often used entry.

Issue #2385, opened 2020-01-06
Want to add to youtube-dl extractor to enable better 'Import with URL' results


bluecat999:
**What happened?**
I imported a video from a youtube-dl-recognized PeerTube instance (peertube.co.uk) and there were some missing or unexpected results:

- Description is truncated
- Language missing (see https://github.com/Chocobozzz/PeerTube/issues/2252)
- License missing (see https://github.com/Chocobozzz/PeerTube/issues/2252)
- Captions are missing (see https://github.com/Chocobozzz/PeerTube/issues/1016)
- Original publish date is overwritten
(Support was untested, I don't know what that is for :/ )

**What do you expect to happen instead?**
I expect these to be properly extracted by youtube-dl and imported by PeerTube.

### I intend to help fix this by improving the youtube-dl extractor.
It already extracts the license properly. I can probably add fields to extract the language and (if we determine it to be expected behaviour) the original publish date if it exists. I am unsure about the captions but I believe that is possible. I will need feedback on the support section.

After this, is should be simple enough for someone to fix the PeerTube import script.

BlueCatYellowCat:
Note: I believe this is important to fix soon due to the news of peertube.video, a large instance, closing.

Chocobozzz:
You'll need to patch youtube-dl to support these new fields

rigelk:
hi @bluecat999 - please provide the example video that demonstrates your bugs.

Description is truncated by design (https://github.com/Chocobozzz/PeerTube/blob/develop/server/helpers/youtube-dl.ts#L221-L229), but within a limit of 10k chars (!). Is the description you refer to longer?

Same goes for original publish date (see same file).

rigelk:
@Chocobozzz @bluecat999 note that these new fields are being added in https://github.com/ytdl-org/youtube-dl/pull/23657. License is already present, though, so we might want to investigate why it failed so far.

rigelk:
with https://github.com/ytdl-org/youtube-dl/pull/23657 now merged, we should expect things to improve noticeably as soon as youtube-dl releases a new version.

PR #2473, opened 2020-02-11
peertube-import-videos passes remaining options down to youtube-dl

### LINKED ISSUE #2377: peertube-import-videos pass arbitrary options to youtube-dl
**What happened?**

I'd like to import using IPv6/IPv4 exclusively

I can't

I think that providing youtube-dl full-options support would add flexibility to the import process and adapt to the many possible edge-cases.

### LINKED ISSUE #2471: Ability to select timecode of a video to import
Improvement proposal: ability to select timecode of video to import.

When you import a video from another website, Peertube import all the stream. But it is possible that only a part of this video has an interest.
For example, https://video.lqdn.fr/videos/watch/b6544b93-1ff7-4bb8-a0a5-dceab779915a comes from the french Assemblée Nationale, and start only from minute 19. I would like to be able to say at importation to skip this 19 first minutes (and potentially the last xx minutes).

drzraf:
- Fixes #2377 
- Possibly fixes #2471 via youtube-dl `exec` or `--postprocessor-args` option
- Possibly help fixing #2385

As an additional effect, this PR dumps youtube-dl **error** output rather than only the `exec()` (error) status which makes way easier to debug youtube-dl failures.

### DIFF
--- a/server/tools/peertube-import-videos.ts
+++ b/server/tools/peertube-import-videos.ts
@@ -38,6 +38,7 @@ command
   .option('--first <first>', 'Process first n elements of returned playlist')
   .option('--last <last>', 'Process last n elements of returned playlist')
   .option('-T, --tmpdir <tmpdir>', 'Working directory', __dirname)
+  .usage("[global options] [ -- youtube-dl options]")
   .parse(process.argv)
 
 const log = getLogger(program['verbose'])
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks for the PR @drzraf 

I'd rather prefix all youtube-dl options by `ydl-myoption` for better maintainability. For example, you could add a youtubedl option like `--ydl--postprocessor-args`. In the script, just check every parameter sent to commander and add options that start with `--ydl-` (and replace the key using `/^--ydl//`)

drzraf:
For `commander` to work, options must
- either be **defined** (and I initially tried such an implementation that but it's definitely not something you want : We are speaking about literally **hundreds** of yt-dl options!),
- either no validation happens (which is not desirable either) : https://github.com/tj/commander.js#action-handler-subcommands
  >  Any unknown options will be reported as an error.

If the fear is about options conflict, I don't see any risk in this implementation.
The ` -- ` syntax passing down all additional options and arguments is a multi-decades old syntax supported by all argument parsers out there : Options following ` -- ` will simply not be considered by `commander` and won't conflict in any way.

You can think of ` -- ` as a global (and convenient) `--yt-*` prefix applying to all subsequent options and arguments with less layering/magic happening what also allow to simpler yt-dl options copy/pasting.

Chocobozzz:
Okay then :+1:

Issue #2348, opened 2019-12-19
when importing, publishedAt should be set to originalPublishedAt


drzraf:
**What happened?**

All videos appear to be published `some minutes ago` on the channel page.

**What do you expect to happen instead?**

Videos get the exact same set of metadata
 
**Steps to reproduce:**

1. import a Youtube channel using `node dist/server/tools/peertube-import-videos.js `

**Additional information**
* PeerTube version or URL: 2.0.0

**Others**
Import is currently the narrowest features to real `mirroring`. And also extra-metadata can be used to keep track of original source/channel, I think that most if not all "standard" metadata should be copied over as-is. In particular the `published date`. That's what users expect.

drzraf:
/edit I saw that even the metadata API does not offer a way to change the published date. Even though I maintain the suggestion in the OP, at least a workaround could be to `display` and `sort` the channel by `originalPublishedAt`.

Chocobozzz:
The published date is automatically set by peertube, and cannot be updated by the users so we have a reliable attribute. We added `originalPublishedAt` to not lose the original publication date information. You can filter with this attribute in the search page

frankstrater:
The Peertube API offers a way to change the `originalPublishedAt` by placing a PUT on the imported video. It's a workaround and needs an extra API call, but it works. See this Python example script at:

https://github.com/beeldengeluid/peertube-docs/blob/master/api/videos-imports.py

lines 138- 153

drzraf:
But the `originalPublishedAt` which is what is actually useful when importing, is not used in the UI.
Neither on the channel/trending pages. I'd expect a channel's option to use it a the first choice (and use `publishedAt` as a default)

Chocobozzz:
I think there is a UI/UX addition that could be made so users can change the sort of some videos pages (channels videos, local videos...). Ping @rigelk

Chocobozzz:
I added ability to sort by `originallyPublishedAt` in https://github.com/Chocobozzz/PeerTube/commit/2fd59d7d89d1c389446ee67838c821e2622fc8ca

A plugin can hook the videos list sort API call to replace `publishedAt` by `originallyPublishedAt` if your want

kontrollanten:
I've created a plugin for this; [peertube-plugin-sort-originally-published-at](https://framagit.org/kontrollanten/peertube-plugin-sort-originally-published-at) PeerTube package repository should've indexed it within 24 hours.

Issue #2006, opened 2019-08-06
Allow setting a date for originallyPublishedAt when importing video


frankstrater:
**What happened?**

Importing a video through the API with [POST /videos/imports](https://docs.joinpeertube.org/api-rest-reference.html#tag/Video/paths/~1videos~1imports/post) takes the date of the video file as a value for originallyPublishedAt.

**What do you expect to happen instead?**

I would expect that you can set a value for originallyPublishedAt, just like with POST /videos/upload and only use the date of the video file if it's not posted as a value.

Chocobozzz:
Fixed in https://github.com/Chocobozzz/PeerTube/commit/8a86e5dc4f04d7fb07d012fbe2cb7ab3baadbf19

PR #2474, opened 2020-02-11
peertube-import-videos using arbitrary additional metadata

### LINKED ISSUE #2385: Want to add to youtube-dl extractor to enable better 'Import with URL' results

### LINKED ISSUE #2467: Ability to store the original video URL a meta for imports
**What happened?**

When importing a video, this line tests for title equality.
https://github.com/Chocobozzz/PeerTube/blob/a15871560f80e07386c1dabb8370cd2664ecfd1f/server/tools/peertube-import-videos.ts#L140

This is suboptimal
1. Because title can change at the source of the destination, resulting in duplicates.
2. Because Peertube imported video does not hold the original location/unique id of the source which may be useful (for establishing a mapping for example)

The importer could fill-in a meta like "Original URL" or "Original ID" containing the Youtube/Vimeo/... URL/ID.

**Additional information**
* PeerTube version or URL: 2.0.0

drzraf:
- Fixes #2467
- Probably fixes #2385
- Offers a workaround for #2348 (and #2006 ?)

Bundles #2473

### DIFF
--- a/server/tools/peertube-import-videos.ts
+++ b/server/tools/peertube-import-videos.ts
@@ -38,6 +38,7 @@ command
   .option('--first <first>', 'Process first n elements of returned playlist')
   .option('--last <last>', 'Process last n elements of returned playlist')
   .option('-T, --tmpdir <tmpdir>', 'Working directory', __dirname)
+  .usage("[global options] [ -- youtube-dl options]")
   .parse(process.argv)
 
 const log = getLogger(program['verbose'])
@@ -71,10 +72,11 @@ async function run (url: string, user: UserInfo) {
 
   const youtubeDL = await safeGetYoutubeDL()
 
-  const options = [ '-j', '--flat-playlist', '--playlist-reverse' ]
+  const options = [ '-j', '--flat-playlist', '--playlist-reverse', ...command.args ]
+
   youtubeDL.getInfo(program['targetUrl'], options, processOptions, async (err, info) => {
     if (err) {
-      exitError(err.message)
+      exitError(err.stderr + ' ' + err.message)
     }
 
     let infoArray: any[]
@@ -146,7 +148,7 @@ function processVideo (parameters: {
 
     log.info('Downloading video "%s"...', videoInfo.title)
 
-    const options = [ '-f', 'bestvideo[ext=mp4]+bestaudio[ext=m4a]/best', '-o', path ]
+    const options = [ '-f', 'bestvideo[ext=mp4]+bestaudio[ext=m4a]/best', ...command.args, '-o', path ]
     try {
       const youtubeDL = await safeGetYoutubeDL()
       youtubeDL.exec(videoInfo.url, options, processOptions, async (err, output) => {
--- a/server/tools/peertube-import-videos.ts
+++ b/server/tools/peertube-import-videos.ts
@@ -32,6 +32,7 @@ command
   .option('-u, --url <url>', 'Server url')
   .option('-U, --username <username>', 'Username')
   .option('-p, --password <token>', 'Password')
+  .option('--metadata <metadata>', 'Additional metadata to add to the videos to be uploaded')
   .option('--target-url <targetUrl>', 'Video target URL')
   .option('--since <since>', 'Publication date (inclusive) since which the videos can be imported (YYYY-MM-DD)', parseDate)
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks for the PR but why don't you use the commander parameter instead? For example: `--video-name '@video.name`

Also, I'm not sure to understand how this fix https://github.com/Chocobozzz/PeerTube/issues/2467 since you don't add another column in the database

drzraf:
> Thanks for the PR but why don't you use the commander parameters instead? For example: `--video-name '@video.name`

Because it didn't sound flexible enough for *any* metadata. We would have to specify any possible existing or future metadata as an allowed option on the command-line syntax.

> Also, I'm not sure to understand how this fixes #2467 since you don't add another column in the database.

That's the point I misunderstood and didn't checked it.
I wrongly believed that any arbitrary metadata would be stored in a kind of "opaque" metadata fieldset.
I didn't thought each and every metadata should be a defined field by itself, allowed by the app' beforehand.

(I find that situation rather cumbersome. Isn't possible to provide an opaque `array` fieldset for arbitrary, user-provided, metadata? Could namespaced them under `@user.` if really needed ? Or maybe hijacking/extending tags for this ? :))

Chocobozzz:
> (I find that situation rather cumbersome. Isn't possible to provide an opaque array fieldset for arbitrary, user-provided, metadata? Could namespaced them under @user. if really needed ? Or maybe hijacking/extending tags for this ? :))

You could try to use a plugin to store additional video metadata I think. But we don't really want to store non-known additional metadata for now for different reasons (we use a relational DB, AP schema etc).

drzraf:
I understand your concern, but I'm pretty sure there is an agnostic (and controlled) way to provide such extensibility:

- An `extra`/`user` object which contains two keys: `public` and `private`, whose serialization would be limited, to say, 64 kB and stored in a newly created, dedicated [Sequelize.JSONB](https://sequelize.org/master/class/lib/data-types.js~JSONB.html) column.

- During storage, the `{private: {opaque object}, public: {opaque object}}` keys are stored as-is (implicitly serialized/ordered/deduplicated using JSONB)
- During `/video/` API calls, data is retrieved (implicitly `JSON.parse`'d) and provided as-is. (Using `private/public` according to API call permissions). Data would be available in the API response via something like `metadata.extra.public/private`.

Don't you think it'd would be a safe enough and non-intrusive way to provide the needed extensibility and resolve problems like "`storing extra-meta during import`". It'd would also be a way for plugins to extend the data-model (naive and limited but better than nothing).

Eg: `metadata.user.private.@myplugin.anything = whatever`
According to [Sequalize](https://github.com/sequelize/sequelize/issues/4963) [stackoverflow](https://stackoverflow.com/questions/49613110) and [blog posts](https://www.compose.com/articles/is-postgresql-your-next-json-database), PostgreSQL even provide a way to query/search/filter based on these nested JSON data attributes (!)

Wouldn't this sound better than messing-up with [Sequelize hooks](https://sequelize.org/v5/manual/hooks.html) to play with database schema (and upgrade/downgrade scripts) that would be way more complex to developers.

Chocobozzz:
We designed a plugin system where you can store data you want, and so create a hashmap with `{ [ videoId: number ]: metadata: any }`. It misses dedicated API route so the client get this data but we'll try to implement this in the future.

Or you could use https://github.com/Chocobozzz/PeerTube/pull/2411 and add you metadata in the public `metadata` field create by @rigelk.

Chocobozzz:
Closing due to inactivity. Feel free to comment!

PR #2476, opened 2020-02-11
Update default user theme to instance-default

ipbc-dev:
Hello,

This will update new users theme to be instance-default


Chocobozzz:
Thanks!

PR #2480, opened 2020-02-12
Fixing user dropdown menu position for languages with long text

ipbc-dev:
This is the user dropdown menu if you open it in spanish version. This happens because the text in spanish is longer than english.
dropdownMenuBroken

I been working on a final solution (for all languages). This is the result in spanish
dropdownMenuFixedES

And this is the result in English
dropdownMenuFixedEN

I'm using a fixed position just to be sure that this menu can overlap any other element in the DOM

### DIFF
--- a/client/src/app/menu/menu.component.scss
+++ b/client/src/app/menu/menu.component.scss
@@ -88,18 +88,25 @@ menu {
         }
       }
 
-      .dropdown-item {
-        @include dropdown-with-icon-item;
-
-        my-global-icon {
-          width: 22px;
-          height: 22px;
-
-          &[iconName="sign-out"] {
-            position: relative;
-            right: -1px;
-            height: 21px;
-            width: 21px;
+      .dropdown-menu{
+        position: fixed!important;
+        top: 110px!important;
+        left: 50px!important;
+        will-change: unset !important;
+        transform: none !important;
+        .dropdown-item {
+          @include dropdown-with-icon-item;
+  
+          my-global-icon {
+            width: 22px;
+            height: 22px;
+  
+            &[iconName="sign-out"] {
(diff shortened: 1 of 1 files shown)


rigelk:
Please avoid fixed positions, as they don't mix well with different screen sizes.

rigelk:
@ipbc-dev Thanks for bringing up the issue - I fixed the issue in eb7c7a5 with a more adaptable solution.

ipbc-dev:
Hello, thank you for taking care of this really fast! @rigelk 

This is what I see after pulling your approach and build
##### ENGLISH
dropdownMenuNewApproach
##### SPANISH
dropdownMenuNewApproach2

I can see your properties in my developer console so, your fixes are in. However, I think something is wrong.

Fixed positions mix perfect with different screen sizes if the position is really fixed... in this case, the dropdown-menu can be shown always in the same position from 0,0 (upper left corner), like the header and the menu-wrapper that are using a fixed position as well. I've tested it in several sizes and my solution works fine (in some mobile sizes, it's not positioned in the best place, but still usable). Anyway, I can tweak it and add a media query to leave mobile devices as they are now. Let me know about your preffered solution and thanks a lot!

rigelk:
@ipbc-dev thanks for pointing it out! Turns out a recent commit on develop increased the menu z-index over that of the dropdowns - which shouldn't happen. We don't have a unique place where z-indexes are managed right now, so this was bound to happen. I'll fix that soon, no need to go with fixed positions.

rigelk:
@ipbc-dev fixed it in 36f2981f7.

ipbc-dev:
Thank you a lot @rigelk Amazing approach for centralize z-indexes :)

PR #2485, opened 2020-02-13
Extend build commands

ipbc-dev:
Hello! 
I had CSS issues for a specific language (Spanish) and I needed to add a '--light-es' parameter to the client.sh in order to test it without losing 20 minutes for each compilation. It can be useful to have a --light parameter for each language so when developers face problems for a specific languages they don't need to  build the entire project.


Chocobozzz:
I hope we'll be able to extract `$localize` strings in Angular 9.1 so we can use the `--localize` option and stop loosing so much time to build all PeerTube languages. Thanks @ipbc-dev

Issue #2304, opened 2019-12-04
Docker Improvements


jamesmeneghello:
I have some time coming up over the holidays and quite a lot of experience in Docker / Kubernetes. Is there any interest in improving Docker support in the project? There are quite a few things that could be improved:

- Environment variable handling (while it's undergone some changes lately) could be streamlined a lot
- Using multi-stage builds to combine development/production containers and slim down the prod container further (incl. the changes made today)
- Improving the compose file for development (allowing development to occur directly in the containers)
- Providing k8s manifests for deployments

I'd be more than happy to take this on and maintain it into the future, but any feedback or suggestions or pain points would be great as to project direction! :)

Chocobozzz:
Hi,

I would suggest that you create your own repository that we can link in the [documentation website](https://docs.joinpeertube.org/#/install-unofficial) (for unofficial k8s support).

Regarding the dev docker file, I wanted to delete it since I don't have time to maintain the workflow. And since we now have gitpod that works, I don't think it's really interesting to keep it.

dirkkelly:
@Chocobozzz Just wanted to chime in here, the reason I've been in the docs lately is because I'm trying to get started so that I can provide development support.

I've given Gitpod a go and it was great for getting a feel for the project. However I'm really looking for a way to develop without relying on a third party private hosted service, especially as I am often working on an Australian internet connection.

Additionally I'm trying to get a manageable and scalable production instance together and would prefer to have this running on k8s rather than bespoke built and maintained servers.

Is it possible we could consider keeping docker support in the this repository if we were able to show consistent support and maintenance for it?

Nutomic:
>Regarding the dev docker file, I wanted to delete it since I don't have time to maintain the workflow. 

@Chocobozzz  I dont think there is anything wrong with having other contributors maintain the Docker configuration.

Chocobozzz:
> Is it possible we could consider keeping docker support in the this repository if we were able to show consistent support and maintenance for it?

Yep, no pb

> I dont think there is anything wrong with having other contributors maintain the Docker configuration.

Sure

kimsible:
I explore PeerTube code and I wonder why a docker build based on debian buster is not proposed for the master branch ? 

I mean, if I'm not wrong the HLS support is only avalaible on ffmpeg >= 4.1 , so on debian buster.

The master branch has a debian stretch build with ffmpeg 3.

[...]

PR #2494, opened 2020-02-14
Chore/docker-documentation-improvements

kimsible:
This an update `/support/docker/production/.env` file example since the recent changes of environment variables (https://github.com/Chocobozzz/PeerTube/pull/2247) in the release v2.1.

Also, more details and some improvements for the docker documentation `/support/doc/docker.md` to simplify the installation, some parts were unclear for non-familiar with docker-compose and traefik : https://github.com/kimsible/PeerTube/blob/chore/improve-docker-documentation/support/doc/docker.md

Plus, it moves ACME email and domain list from `/support/docker/production/config/traefik.toml` to `/support/docker/production/.env` and `/support/docker/production/docker-compose.yml` as a command of traefik service.


kimsible:
Thanks for adding the labels @rigelk. 

I also wonder if `./docker-volume/traefik/traefik.toml` could be placed in `./docker-volume/config/traefik.toml` as it is provided by the peertube docker image `chocobozzz/peertube:production-buster` and as the `[acme] email` and `[[acme.domains]]` configurations lines have been dropped and now defined as a traefik command in `docker-compose.yml`
```
services:
  reverse-proxy:
    volumes:
      [...]
      - ./docker-volume/traefik/acme.json:/etc/acme.json
      - ./docker-volume/traefik/traefik.toml:/traefik.toml
 ```
Also, for `./docker-volume/traefik/acme.json`, is it really usefull to link as a volume it in `docker-compose.yml` ? I mean, docker will not erase until the traefik container is recreated.

If I don't miss anything it could simplify more the configuration.

Chocobozzz:
Thanks for the PR. Could you fix the conflict please?

kimsible:
Ok thanks @Chocobozzz for replying to my first comment in the issue https://github.com/Chocobozzz/PeerTube/issues/2304.

Here `.env` has minor changes to make documentation clearer, as it is the main file to configure.
I see it as an example file.

For the extrating of `[acme] email` and `[[acme.domains]]` of  `./docker-volume/traefik/traefik.toml` to `.env` / `docker-compose.yml`, it is more to follow the recent changes of the v2.1, to unify main env var in the same file `.env`.

In my opinion, some environement variables should have been left in docker-compose.yml (POSTGRES and POSTFIX) but I suppose they've been extracted to make docker-compose.yml simpler to maintain.

LecygneNoir:
Hello,

I have just noticed (already present in the old doc it seems) that the files used for "preview" are different from the files downloaded.

Indeed, to download with curl, you use /master/, but for the link to the previews, you use /develop/  

I guess we should use /master/ for all?

Could you fix it in the PR before merging? (Else I may create a new PR once this one is merged ^^)

Anyway, thanks for the improvements! 👍

rigelk:
> Indeed, to download with curl, you use /master/, but for the link to the previews, you use /develop/ […] I guess we should use /master/ for all?

I personally don't understand why you don't use relative references, since you are already in the peertube directory. You don't need to curl these files, you can just edit them locally.

LecygneNoir:
> 
> I personally don't understand why you don't use relative references, since you are already in the peertube directory. You don't need to curl these files, you can just edit them locally. The same goes for links: you replace local references with references to the github.

Nope, do not forgot it's a docker usage, so in this documentation you vurl the file. You cannot assume the entire git repo has been cloned, as ideally, you just need the docker-compose and some config files and the docker image do all the "app" job for you.

The main advantage  of using docker is to avoid cloning and building the app manually.

So, you need to curl the files as you do not have them locally if you have not git cloned the repo :)  
Or I am totally misunderstanding your answer sorry :sweat_smile: 

My point was, for the `curl` part and the `View the source of the file you're about to download:` we should use same URL, as long as we continue to provide these 2 lines in doc.

Valid only if we keep both lines, which I think we should (one should always look at what he download)

rigelk:
ah, I see. Thanks for putting that into context @LecygneNoir.

kimsible:
> I have just noticed (already present in the old doc it seems) that the files used for "preview" are different from the files downloaded.
> Indeed, to download with curl, you use /master/, but for the link to the previews, you use /develop/

Thanks for the review @LecygneNoir and @rigelk, it's fixed.

PR #2495, opened 2020-02-14
Fix missing python dependency in Dockerfile.buster

kimsible:
**Issue** : https://github.com/Chocobozzz/PeerTube/issues/2491

PR #2497, opened 2020-02-15
Update postgres/redis in docker-compose 

shleeable:
Redis 5 is backwards compatible with 4.
Postgres 12 needs additional testing, but I've installed without issue on my test.

### DIFF
--- a/support/docker/production/docker-compose.yml
+++ b/support/docker/production/docker-compose.yml
@@ -54,6 +54,13 @@ services:
     restart: "always"
     labels:
       traefik.enable: "false"
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "postgres"]
+    environment:
+      - POSTGRES_DB=postgres
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+    shm_size: 64M
 
   redis:
     image: redis:5-alpine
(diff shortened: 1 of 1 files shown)


shleeable:
Should be able to use the ${PEERTUBE_DB_USERNAME} ${POSTGRES_PASSWORD} to unhardcode those username/password values.

kimsible:
> Should be able to use the ${PEERTUBE_DB_USERNAME} ${PEERTUBE_DB_PASSWORD} to unhardcode those username/password values.

There is another PR about the env vars : https://github.com/Chocobozzz/PeerTube/pull/2495

support/docker/production/docker-compose.yml
```diff
@@ -46,22 +46,32 @@ services:
     restart: "always"
 
   postgres:
-    image: postgres:10-alpine
+    image: postgres:12
```
Nutomic: Why not `12-alpine`?

support/docker/production/docker-compose.yml
```diff
@@ -46,22 +46,32 @@ services:
...
     env_file:
       - .env
     volumes:
       - ./docker-volume/db:/var/lib/postgresql/data
     restart: "always"
     labels:
       traefik.enable: "false"
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "postgres"]
+    environment:
```
Chocobozzz: Why do you need these env variables?

Chocobozzz:
Closing due to inactivity. Feel free to comment!

PR #2499, opened 2020-02-15
Add python-dev and ca-certificates to docker build

### LINKED ISSUE #2491: Python missing in production-buster docker image
**What happened?**
Since I upgraded my instance running in docker to 2.1.0 with the `production-buster` image, I can't use imports anymore. I'm using the API but it's the same from the browser.  
And opening the logs, I see the error that it can't find python.

Peertube importing my video. 😛 

**Steps to reproduce:**

1. Set `chocobozzz/peertube:production-buster` for the service `peertube` in docker-compose

2. Pull image and restart stack

3. Try to import a video (in this case from Twitch)

**Additional information**
* PeerTube version or URL: v2.1.0
* Browser name/version: REST API, Chrome 79
* NodeJS version: v10.19.0

* Link to server log if useful (journalctl or /var/www/peertube/storage/logs/): https://gist.github.com/mkody/79eff40c23561fed53c99566eae3092d

mkody:
Tested and running on my instance.  
Fixes #2491

After I added `python-dev` (like it is found in [the dependencies page](https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/dependencies.md#debian--ubuntu-and-derivatives)) the missing python in env was fixed.

Also had to install ca-certificates or I'd get certificate errors.

### DIFF
--- a/support/docker/production/Dockerfile.buster
+++ b/support/docker/production/Dockerfile.buster
@@ -7,7 +7,7 @@ ARG NPM_RUN_BUILD_OPTS
 
 # Install dependencies
 RUN apt update \
- && apt install -y --no-install-recommends openssl ffmpeg gnupg gosu \
+ && apt install -y --no-install-recommends openssl python-dev ca-certificates ffmpeg gnupg gosu \
  && gosu nobody true \
  && rm /var/lib/apt/lists/* -fR
 


Chocobozzz:
Thanks for your PR, but it should already be fixed by https://github.com/Chocobozzz/PeerTube/pull/2495

mkody:
@Chocobozzz Oh right, but ca-certificate is missing tho

Chocobozzz:
Thanks, added in https://github.com/Chocobozzz/PeerTube/commit/8d3fe90c3f3bfd43366ea41a2519431b9001e744

PR #2503, opened 2020-02-16
Explain view calculation on short videos, rewording

3risian:
FAQ previously did not explain how views are calculated for videos under 30 seconds. Minor rewording fixes based on feedback from non-technical person.


Chocobozzz:
Thanks

Issue #2506, opened 2020-02-18
Translation of title tooltips


AutomCoding:
Is it possible to make a tags' title attributes translatable? Like [this one](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/login/login.component.html#L50), for example.

rigelk:
@AutomCoding yes, using `i18n-title` tag.

AutomCoding:
@rigelk I'm searching through the code for more instances of this, will make a PR soon.

PR #2507, opened 2020-02-18
Mark additional strings as translatable

AutomCoding:
I have searched through all HTML files in /client/scr/app for elements with a title attribute, but not i18n-title (#2506). Doing that, I also found some more strings that weren't tagged as translatable.

### DIFF
--- a/client/src/app/+my-account/my-account-settings/my-account-video-settings/my-account-video-settings.component.html
+++ b/client/src/app/+my-account/my-account-settings/my-account-video-settings/my-account-video-settings.component.html
@@ -11,9 +11,9 @@
 
     <div class="peertube-select-container">
       <select id="nsfwPolicy" formControlName="nsfwPolicy">
-        <option i18n-value value="do_not_list">Do not list</option>
-        <option i18n-value value="blur">Blur thumbnails</option>
-        <option i18n-value value="display">Display</option>
+        <option i18n value="do_not_list">Do not list</option>
+        <option i18n value="blur">Blur thumbnails</option>
+        <option i18n value="display">Display</option>
       </select>
     </div>
   </div>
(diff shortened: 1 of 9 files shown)


client/src/app/+my-account/my-account-settings/my-account-video-settings/my-account-video-settings.component.html
```diff
@@ -11,9 +11,9 @@
 
     <div class="peertube-select-container">
       <select id="nsfwPolicy" formControlName="nsfwPolicy">
-        <option i18n value="do_not_list">Do not list</option>
-        <option i18n value="blur">Blur thumbnails</option>
-        <option i18n value="display">Display</option>
+        <option i18n-value value="do_not_list">Do not list</option>
+        <option i18n-value value="blur">Blur thumbnails</option>
+        <option i18n-value value="display">Display</option>
```
rigelk: The value is not the target of translation here. Please revert this.

client/src/app/shared/video-playlist/video-add-to-playlist.component.html
```diff
@@ -59,7 +59,7 @@
     </div>
   </div>
 
-  <div class="new-playlist-button dropdown-item" (click)="openCreateBlock($event)" [hidden]="isNewPlaylistBlockOpened">
+  <div class="new-playlist-button dropdown-item" (click)="openCreateBlock($event)" [hidden]="isNewPlaylistBlockOpened" i18n>
```
rigelk: `i18n` basically adds the content of the element to the translation. To avoid confusing our translators as to what needs to be effectively translated, we try to avoid having code in our translatable input. Please wrap the text to translate with a `<span i18n> … </span>` instead.

rigelk:
THanks @AutomCoding!

PR #2531, opened 2020-03-03
Add nginx in docker-compose

rigelk:
Currently, traefik is the default reverse-proxy (!) in our docker-compose. At first it was an easy choice as a drop-in cert generator tailored for docker. However, it doesn't benefit from optimisations and features held in the nginx configuration (mostly url rewrites that allow nginx to directly handle serving static files, which greatly reduces load on the peertube runtime), which leads me to <strike>replace traefik by nginx</strike> add nginx as a webserver behind traefik in this PR. I made sure we keep roughly the same feature set: email, domain can both be set in the `.env` without needing to touch the nginx conf.

The PR poses a few questions: how is the docker-compose commonly used? Do you rely heavily on traefik? Do you use it without modifications? or do you disable it to setup your own reverse-proxy (nginx?)? Hopefully this PR helps setups with little to no modification<strike>, at the cost of - potentially - hurting those that relied on traefik</strike>. What is your situation?

TODO:
- [x] test postgres migration 10 -> ~~12~~ 13
- [x] mount storage and assets in reverse-proxy
- [x] test live deployment
- [x] update documentation

SIDE EFFECTS:
- nginx config now has placeholders that `envsubst` can target.
- redis and postgres images are updated - postgres needs manual data backup/restore. 

BREAKING:
- if you rely on peertube's nginx template, beware: I changed `peertube.example.com` to `${WEBSERVER_HOST}` and `127.0.0.1:9000` to `${PEERTUBE_HOST}`.


rigelk:
@LecygneNoir @Nutomic @kimsible I might need your review on this one, as the PR changes the `docker-compose.yml` rather heavily.

support/docker/production/docker-compose.yml
```diff
@@ -49,29 +39,23 @@ services:
     restart: "always"
 
   postgres:
-    image: postgres:10-alpine
+    image: postgres:12-alpine
```
LecygneNoir: Hello,  
Be careful, updating postgres could mean manual manipulation, have you tested the upgrade?

support/docker/production/docker-compose.yml
```diff
@@ -3,24 +3,20 @@ version: "3.3"
...
+    env_file:
+      - .env
     ports:
       - "80:80"     # The HTTP port
       - "443:443"   # The HTTPS port
     volumes:
-      - /var/run/docker.sock:/var/run/docker.sock # So that Traefik can listen to the Docker events
-      - ./docker-volume/traefik/acme.json:/etc/acme.json
-      - ./docker-volume/traefik/traefik.toml:/traefik.toml
+      - type: bind
```
LecygneNoir: Hello,
This means we need to get the nginx configuration directly in the peertube repo, where before we only get the docker-compose part (with the traefik conf inside). Perhaps you should mention it in the documentation?

rigelk: Yes, that's why it is in TODO.

LecygneNoir: Okay great, I was unsure if it covers new deployment or update existing :+1:

support/docker/production/docker-compose.yml
rigelk: I am testing it at the moment.

rigelk: Migrating beetween major postgres version requires some manual dump/restore, but otherwise works:

```
docker-compose -f support/docker/production/docker-compose.yml exec -T postgres pg_dumpall -U postgres > pgdump
```

And after updating:

```
docker-compose -f support/docker/production/docker-compose.yml exec -i postgres psql -U postgres < pgdump
```

Restart the postgres service and delete `pgdump`.

support/docker/production/docker-compose.yml
```diff
@@ -30,16 +30,11 @@ services:
...
-    # remove them if you are using another reverse proxy.
-    labels:
-      traefik.enable: "true"
-      traefik.frontend.rule: "Host:${PEERTUBE_WEBSERVER_HOSTNAME}"
-      traefik.port: "9000"
     # If you don't want to use a reverse proxy (not suitable for production!)
     # ports:
       # - "80:9000"
     volumes:
+      - ./docker-volume/assets:/client/dist
```
Nutomic: Any reason why this new volume is needed?

support/docker/production/.env
```diff
@@ -20,10 +20,11 @@ PEERTUBE_SMTP_TLS=false
...
 PEERTUBE_ADMIN_EMAIL=<MY EMAIL ADDRESS>
 POSTFIX_myhostname=<MY DOMAIN>
-TRAEFIK_ACME_EMAIL=<MY EMAIL ADDRESS>
-# If you need to obtain ACME certificates for more than one DOMAIN
-# pass them as a comma separated string
-TRAEFIK_ACME_DOMAINS=<MY DOMAIN>
+CERTBOT_EMAIL=<MY EMAIL ADDRESS>
+# Variables to substitute in the nginx template, space-separated
+ENVSUBST_VARS=FQDN PEERTUBE_HOST
+FQDN=<MY DOMAIN>
```
Nutomic: What does this even stand for? Why not just `DOMAIN`?

rigelk: That's a fully qualified domain name, but you're right, it should just be `DOMAIN` now.

support/docker/production/docker-compose.yml
rigelk: Yes, to allow this block to be functional: https://github.com/Chocobozzz/PeerTube/blob/develop/support/nginx/peertube#L62-L67

rigelk: But I made a mistake: it should be bound to `/app/client/dist`.

Nutomic: Ah my nginx config looks pretty differently, I just cache all the frontend files.

https://yerbamate.dev/nutomic/peertube-compose/src/branch/master/templates/nginx.conf

rigelk: I didn't think about that. I'll try this.

kimsible:
@rigelk as you say, traefik is useful especially for a drop-in cert generator.
I undersdand the point about optimization and statics but how we will generate cert now with docker ?

Spass1966:
Not to be too caustic but this is a terrible idea. Traefik is by far the superior reverse proxy. It's easy, no fuss, set it and forget it.

Tackle the larger issues with the docker first, like passed PUID and PGID values not being respected. That's a huge issue for server admins like me, since PeerTube currently expects everything to be owned by itself as root.

I'd look at turning the container into something that follows the LinuxServer container model way before swapping out Traefik. I use Traefik for all 36 and counting containers of mine. It's virtually drop-in. Nginx is not.

rigelk:
> Not to be too caustic but this is a terrible idea. Traefik is by far the superior reverse proxy. It's easy, no fuss, set it and forget it.

I would argue the point of this issue is to have not just a reverse-proxy but a webserver, for its ability to serve static files. c.f. the PR description, c.f. our nginx configuration. Performance improvements and features that you miss with having a direct Traefik <-> PeerTube setup.

> Tackle the larger issues with the docker first, like passed PUID and PGID values not being respected. That's a huge issue for server admins like me, since PeerTube currently expects everything to be owned by itself as root.

That is outside the scope of the issue, and I would rather discuss it in a proper issue. But maybe I missed the relevance.

> I'd look at turning the container into something that follows the LinuxServer container model way before swapping out Traefik. I use Traefik for all 36 and counting containers of mine. It's virtually drop-in. Nginx is not.

I don't ask you to remove Traefik from your infrastructure though: I'm merely changing the *default* option. And I'm open to keeping Traefik for SSL termination, with Nginx as webserver doing the heavy lifting behind it.

kimsible:
I don't know what to say about replacing traefik with nginx+cerbot that would help you except it would be nice to have multiple exemples of docker-compose setup.

Maybe the best choice would be having a support only for the basic production configuration of peertube server and the main Docker image and create another repository for docker-compose.

Or, another solution we can support full docker as Mastodon does, but it should be the main tool for development and productions.

rigelk:
@kimsible thanks a lot for your answers to the initial questions.

> it would be nice to have multiple exemples of docker-compose setup.

I don't think this is the direction we want to take, as that would increase our maintenance load.

We have two deployment (not counting the docker image) targets that we need to support: "baremetal", and docker. In each, we need to provide settings that are: secure (hence the reverse-proxy defaults and security headers), and performant (hence the current PR).

---

Aside from that, I refactored the PR: it doesn't replace traefik anymore, but instead just adds nginx in the mix. Call it separation of concerns if you will; the advantage compared to the first version of my PR is that swapping the RP is easier. Nginx here just acts as a webserver, and leaves the TLS termination/cert gen to traefik.

LecygneNoir:
> Aside from that, I refactored the PR: it doesn't replace traefik anymore, but instead just adds nginx in the mix. Call it separation of concerns if you will; the advantage compared to the first version of my PR is that swapping the RP is easier. Nginx here just acts as a webserver, and leaves the TLS termination/cert gen to traefik.

Good idea indeed! Like that, someone will directly profit from the performance work done on Nginx and it does not break work already done with traefik for people already using it :+1: 

Though I'll need to add this nginx possibility in the Kubernetes helm chart now :joy: 

As we use an official nginx image, we still need to download the Peertube nginx configuration in addition to the traefik one, but I think it's acceptable.

kimsible:
> I don't think this is the direction we want to take, as that would increase our maintenance load.
>
> We have two deployment (not counting the docker image) targets that we need to support [...]

Ok @rigelk, thanks for clarifying that point, I understand, the "baremetal" is indeed as important as docker deployment, and one official small and secure docker setup is really enough.

Another point linked to the deployment, I'm working on a deployment script based on this docker setup : https://github.com/kimsible/install-peertube/

I'm also thinking to add a "server to server migration" feature to that script.

I will for sure follow this nginx integration to keep it up to date.

Nutomic:
@kimsible why don't you use ansible instead of the manual scripts? Here is my repo for reference, literally takes one command to install.  

https://yerbamate.dev/nutomic/peertube.social

kimsible:
> @kimsible why don't you use ansible instead of the manual scripts? Here is my repo for reference, literally takes one command to install.
> 
> https://yerbamate.dev/nutomic/peertube.social

@Nutomic thanks for the link, it's a really interesting tool. But shortly their goals are a little bit different, I'm trying to develop a simple tool without any dependency (except docker and docker-compose).

fflorent:
**TL;DR**: Maybe that's irrelevant here, but I talk about my experience of using docker-compose and especially Traefik for PeerTube. 

My testimony for using docker-compose.yml (for the skeptikon.fr website):
 - I personally like the ease for updates, for administrating env variables and for the transparent letsencrypt negotiation ;
 - We **really** suffer from the `network_mode: host`, which prevents use from installing other services like nextcloud on the same server because of port conflicts on the host due to this mode (I have tried to hack a bit to workaround this issue with no luck) ;
 - I have been told that nginx has nice cache and optimization features that traefik doesn't (this PR seems to confirm that), and I wonder if that's not the cause of slowness when requesting webpages sometimes ;
 - it seems that we are stuck with the version 1.7.X version of traefik (at least for a moment), because the latest major version introduces breaking changes and we can't yet afford to adapt our configuration for them ;

I wanted to study and try to remove Docker, mostly because of the 2nd point above, but I lack of time for that.

If we can get rid of Traefik and use Nginx instead (with the certbot service), it would probably spare me lots of time. Keeping using Traefik in front of Nginx seems to me overkill, but I am not an expert (maybe you have good arguments for this choice).

----

> how is the docker-compose commonly used? 

I must have answered the question above.

> Do you rely heavily on traefik? 

Only for PeerTube. I cannot install much services because of that (due to the network_mode as explained above).

> Do you use it without modifications?

With the least modifications, thus we can follow the updates without pain.

> or do you disable it to setup your own reverse-proxy (nginx?)? 

Nope :).

Thanks for your amazing work! :pray:

support/docker/production/docker-compose.yml
kimsible: It may seems obvious but I'd add that before running the postgres service again and execute the restore command, we need to delete all db files in the docker-volume to re-init the db after restarting the service : 
```bash
rm -rf ./docker-volume/db
```
Otherwise we got this error even after the restore command : 
```
postgres_1       | 
postgres_1       | PostgreSQL Database directory appears to contain a database; Skipping initialization
postgres_1       | 
postgres_1       | 2020-03-26 19:21:19.547 UTC [1] FATAL:  database files are incompatible with server
postgres_1       | 2020-03-26 19:21:19.547 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 10, which is not compatible with this version 12.2.
```

support/docker/production/docker-compose.yml
```diff
@@ -2,44 +2,65 @@ version: "3.3"
...
-    # Traefik labels are suggested as an example for people using Traefik,
-    # remove them if you are using another reverse proxy.
+    # If you provide your own reverse-proxy, otherwise not suitable for production:
+    #ports:
+    #  - "80:80"
+    volumes:
+      - type: bind
+        # Switch sources if you downloaded the nginx configuration without the whole repository
+        #source: ./peertube
+        source: ../../nginx/peertube
```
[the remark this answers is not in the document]
kimsible: What I actually mean is that `source: ../../nginx/peertube` should be commented to follow the rest of the Compose file and the comment should be as bellow : 
```yaml
      - type: bind
        # Switch sources if you downloaded the nginx configuration with the whole repository
        #source: ../../nginx/peertube
        source: ./docker-volume/nginx/peertube
```

support/docker/production/docker-compose.yml
kimsible: > Migrating beetween major postgres version requires some manual dump/restore, but otherwise works. Execute on the running peertube service before updating the image:
> 
> ```
> docker-compose -f support/docker/production/docker-compose.yml exec -T postgres pg_dumpall -U postgres > pgdump
> ```
> 
> And after updating the image, run the postgres service and execute the restore command:
> 
> ```
> docker-compose -f support/docker/production/docker-compose.yml exec -i postgres psql -U postgres < pgdump
> ```
> 
> Restart the postgres service and delete `pgdump`.

You actually don't need a `pg_dumpall` because restarting the postgres service will recreate roles and peertube database template according to the docker-compose setup.

Dump command : 
> docker-compose exec -T postgres_service pg_dump -U postgres_user postgres_db > pgdump

Restore command (after deleting `./docker-volume/db` mounted database and `compose up postgres_service`):
> docker-compose exec -T postgres_service pg_restore -U postgres_user -d postgres_db < pgdump

support/docker/production/docker-compose.yml
```diff
@@ -2,44 +2,65 @@ version: "3.3"
...
+    #  context: .
+    #  dockerfile: ./support/docker/production/Dockerfile.buster
+    image: chocobozzz/peertube:production-buster
+    env_file:
+      - .env
+    # If you provide your own webserver and reverse-proxy, otherwise not suitable for production:
+    #ports:
+    #  - "80:9000"
     volumes:
+      - assets:/app/client/dist
```
LecygneNoir, line 63: Where does the assets directory comes?

It should be already available in the docker image (as the client is builded with the server in docker), but I am unsure where you get it here.

is this a way to "share" this volume with the nginx image, and populate it when the peertube docker starts?

rigelk:
Yes.


The directory comes from the PeerTube docker image. If you know of another way to double bind a directory from a docker container, I'm all ears.

LecygneNoir:
> Yes.
> 
> The directory comes from the PeerTube docker image. If you know of another way to double bind a directory from a docker container, I'm all ears.

Ok thanks, I was unsure :sweat_smile: 

Honestly I do not know about a better way to do that. I have the same problem in kubernetes (accessing the assets from nginx) and as volumes work slightly differently, I finally write an init script that cp the assets to a network shared volume...

Your way with docker-compose sharing volume on same host is better ^^

support/docker/production/docker-compose.yml
rigelk: @LecygneNoir it allows to share the directory beetween the PeerTube container and the webserver (in read-only for the latter), allowing file-serving optimizations within the Nginx config to really shine.

> is this a way to "share" this volume with the nginx image, and populate it when the peertube docker starts?

That's exactly what I did ;)

Issue #2532, opened 2020-03-03
Comments not being shown


cabaleiro8:
**What happened?**
Videos with several comments does not show all of them

**What do you expect to happen instead?**
Have a load more button, load more comments on scroll or some other approach to be able to see ALL comments

**Steps to reproduce:**
1. https://bittube.video/videos/watch/a29e7fdb-2fa5-4973-92bc-dfa5dd13e4c7
In this URL you can notice that there are 16 comments but only 10 are shown

2. https://peervideo.club/videos/watch/f3cb1945-a8f7-481f-a465-946c6f884e50
In this URL there are 30 comments, but only 20 are shown (if you include replies as comments, there are 28 comments)

**Additional information**
* PeerTube version or URL: 2.1.1
* Browser name/version: Chrome (did not test others)

DontUseGithub:
https://github.com/Chocobozzz/PeerTube/issues/1887

Chocobozzz:
Fixed by @ipbc-dev in https://github.com/Chocobozzz/PeerTube/pull/2535

PR #2535, opened 2020-03-04
Fixing load more comments on infinite scroll

ipbc-dev:
Fix for issue #2532 (https://github.com/Chocobozzz/PeerTube/issues/2532)

I saw that 'onNearOfBottom' function is incrementing the current comments page before check if it 'hasMoreItems'. This is getting an uncomplete comments set because it skip always last comments page. Another approach would be set the default value of 'this.componentPagination.currentPage' to 0 instead of 1. But, in my opinion, this is more reliable.

### DIFF
--- a/client/src/app/videos/+video-watch/comment/video-comments.component.ts
+++ b/client/src/app/videos/+video-watch/comment/video-comments.component.ts
@@ -218,7 +218,6 @@ export class VideoCommentsComponent implements OnInit, OnChanges, OnDestroy {
       this.componentPagination.totalItems = null
 
       this.syndicationItems = this.videoCommentService.getVideoCommentsFeeds(this.video.uuid)
-      
       this.loadMoreThreads()
     }
   }
--- a/client/src/app/videos/+video-watch/comment/video-comments.component.ts
+++ b/client/src/app/videos/+video-watch/comment/video-comments.component.ts
@@ -192,11 +192,10 @@ export class VideoCommentsComponent implements OnInit, OnChanges, OnDestroy {
     return this.authService.isLoggedIn()
   }
 
-  onNearOfBottom () {
-    this.componentPagination.currentPage++
-
+  onNearOfBottom () {    
     if (hasMoreItems(this.componentPagination)) {
       this.loadMoreThreads()
+      this.componentPagination.currentPage++
     }
   }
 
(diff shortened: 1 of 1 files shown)


client/src/app/videos/+video-watch/comment/video-comments.component.ts
```diff
@@ -192,11 +192,10 @@ export class VideoCommentsComponent implements OnInit, OnChanges, OnDestroy {
     return this.authService.isLoggedIn()
   }
 
-  onNearOfBottom () {
-    this.componentPagination.currentPage++
-
+  onNearOfBottom () {    
     if (hasMoreItems(this.componentPagination)) {
       this.loadMoreThreads()
+      this.componentPagination.currentPage++
```
Chocobozzz: Should be above `this.loadMoreThreads()` no?

ipbc-dev: I've done a pair of tests and for me makes no difference, as the 'loadMoreThreads' function does not use 'currentPage'. The important fact is to be below 'hasMoreItems'. No problem if you want to change it @Chocobozzz ;) Thanks a lot for a quick response (as always)

Chocobozzz: `loadMoreThreads` uses `this.componentPagination.currentPage` in https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/videos/%2Bvideo-watch/comment/video-comments.component.ts#L115 and https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/videos/%2Bvideo-watch/comment/video-comment.service.ts#L59 so yes please do the modification. And thanks to you for the bug fix :)

client/src/app/videos/+video-watch/comment/video-comments.component.ts
```diff
@@ -219,7 +218,7 @@ export class VideoCommentsComponent implements OnInit, OnChanges, OnDestroy {
       this.componentPagination.totalItems = null
 
       this.syndicationItems = this.videoCommentService.getVideoCommentsFeeds(this.video.uuid)
-
+
```
Chocobozzz: Please remove this extra space

Chocobozzz:
Thanks @ipbc-dev

PR #2538, opened 2020-03-05
Fixes docker-compose .env

kimsible:
**Small fixes in docker environment variables :** 
- Remove `<MY POSTGRES DB>` and set default `POSTGRES_DB` to `peertube` as it is non-editable
- Add a comment about database suffix (`PEERTUBE_DB_SUFFIX`)
- Update some comments
- Comment `PEERTUBE_WEBSERVER_HTTPS` and `PEERTUBE_WEBSERVER_PORT` as the same values are defined in `production.yml`


kimsible:
@rigelk I didn't see these micro mistakes in the last PR https://github.com/Chocobozzz/PeerTube/pull/2494

rigelk:
@LecygneNoir aren't `PEERTUBE_WEBSERVER_HTTPS` and `PEERTUBE_WEBSERVER_PORT` extraneous now?

kimsible:
Also why not remove `PEERTUBE_SMTP_USERNAME`, `PEERTUBE_SMTP_PASSWORD`, `PEERTUBE_SMTP_PORT`, `PEERTUBE_SMTP_TLS`,  `PEERTUBE_SMTP_DISABLE_STARTTLS` since postfix is the default  smtp relay for docker ?

It could be in `docker-volume/config/production.yml` :

```
smtp:
  hostname: null
  port: 25
  username: null
  password: null
  tls: false
  disable_starttls: false
  ca_file: null 
```
At last, `docker-volume/config/production.yml` may be intended to disappear for docker setup and define every environment variables in `.env` ?

Same for `PEERTUBE_TRUST_PROXY` might it be in `docker-volume/config/production.yml` ? @LecygneNoir  @rigelk ?

kimsible:
Finally, I created an issue for Postfix / OpenDKIM signature, it seems to be more of lots work than expected to implement simply : https://github.com/Chocobozzz/PeerTube/issues/2545

LecygneNoir:
Hello,

Be careful, env variables are not mandatory, but if present they overwrite the production.yaml

In some context (Docker swarm or Kubernetes), it's easier to set env variables than to edit `production.yml`, that's why variables such as `PEERTUBE_TRUST_PROXY`, `PEERTUBE_WEBSERVER_HTTPS/PORT` or SMTP are added.

We may delete them in the default configuration, but in this case modifying the entrypoint to read them (if they are here) and populating the production.yml may be necessary to avoid problem during upgrade.

Or add warning in the release for admin to modifying their production.yml manually when upgrading (like it was done for manual postgres request during the v2.1.0)

As an example, on kubernetes, my production.yml is near empty, and I used env variable in the deployment directly: https://git.lecygnenoir.info/LecygneNoir/peertube-helm/src/branch/master/values.yaml

I may understand the desire to reduce env variable number, but it's a docker practice to use them massively instead of files (as files needs persistent volumes).  
I will not vote on the fact that it's a good practice or not thought :joy: 

Just be careful if we remove env variable, we need to be precise which one, how to add them in production.yml instead, and in which fields precisely (as docker admin may have never opened production.yml yet)

LecygneNoir:
> Also why not remove `PEERTUBE_SMTP_USERNAME`, `PEERTUBE_SMTP_PASSWORD`, `PEERTUBE_SMTP_PORT`, `PEERTUBE_SMTP_TLS`, `PEERTUBE_SMTP_DISABLE_STARTTLS` since postfix is the default smtp relay for docker ?

I think not on this one, as these parameters allow to use an external email server and not the local relay, directly by using the smtp library in Peertube.

For example you may connect to your email account and use it to send your Peertube email, that's the point of this configuration both in the production.yaml and with the env variable.

Removing this and force using the local docker postfix relay may lead to classify lots of email as spam, as email would be send directly with the public IP the docker use (default gateway?).

Or perhaps I am mistaking your point @kimsible ? :pray:

kimsible:
Thanks a lot for the details @LecygneNoir , I understand now, I've missed the point about Docker swarm Kubernetes.
For `PEERTUBE_WEBSERVER_HTTPS` and `PEERTUBE_WEBSERVER_PORT` do you confirm it's a good practice to let them commented instead of override them with the same value than `production.yml` as it was before ?

LecygneNoir:
I have made some tests, and it seems that env variable just overwrite everything else configured, either in production.yml or in the backend interface Oo

For example it's possible to set `PEERTUBE_ADMIN_EMAIL` as env var, if you do that, you cannot change the admin email in the web backend, once you reload the page it's resetted to the value in the env variable :scream: 

Same for `PEERTUBE_SIGNUP_ENABLED` or `PEERTUBE_TRANSCODING_ENABLED`, while these env vars are absents from the `.env` file, I do not know from where I got them :thinking: 

So I guess the env vars overwrite the production.yml in any case.

With this at hands, you may let them commented, if the admin configured its production.yml the server will work, and if he already used them, then env vars will overwrite any default configuration in the production.yml, so the server will work too.

I am not very familiar with the usage of docker-compose (sorry) but I guess if you customize your `.env`, you'll not let it being replaced by a new one during an upgrade :-)

Hope it helps! 
(not sure to be clear, please do not hesitate if I something needs to be more explicated)

kimsible:
>For example it's possible to set PEERTUBE_ADMIN_EMAIL as env var, if you do that, you cannot 
>
> change the admin email in the web backend, once you reload the page it's resetted to the value in the env variable scream

@LecygneNoir, yes for the admin user email but for the contact form the mail is indeed changed.
 
> Same for PEERTUBE_SIGNUP_ENABLED or PEERTUBE_TRANSCODING_ENABLED, while these env vars are absents from the .env file, I do not know from where I got them thinking

They may have been inserted to the database after the first start, we can add a comment to inform it's not editable after the first start ?

For `PEERTUBE_WEBSERVER_HTTPS` and `PEERTUBE_WEBSERVER_PORT`, I think these are extraneous, because in production the best practice is to use https except if you don't use the default reverse proxy and use the one of your cloud / hosting, but who uses PeerTube on a cloud ? It's a little bit nonsense according to the basic needs of the app ?

kimsible:
@rigelk @LecygneNoir would you see anything to add or edit on this PR ?

LecygneNoir:
Seems ok for me, thanks! :+1:

kimsible:
@Chocobozzz this PR wouldn't be included in the next 2.2 release ? maybe you wait a stable 2.2 to merge it as a patch ?

rigelk:
@kimsible it won't land in 2.2. Patches are for hotfixes.

kimsible:
> @kimsible it won't land in 2.2. Patches are for hotfixes.

ok

Chocobozzz:
This PR is not related to peertube releases (not bundled inside the archive but downloaded manually by administrators).

kimsible:
> This PR is not related to peertube releases (the files changed are not bundled inside the archive but downloaded manually by administrators).

Thanks for clarifying =).

PR #2539, opened 2020-03-05
Add action hook, comments loaded

ipbc-dev:
I'm adding donate buttons to the comments via a plugin (peertube-plugin.bittube-airtime-module) 
DonateButtonDark

I needed to add a hook to know when the comments are loaded and when the load more comments on scroll is being triggered. Would be nice to add it to the Federation in order to keep my plugin compatible with it. I also may be usefull for somebody else. Feel free to make suggestions and thanks a lot!

### DIFF
--- a/client/src/app/videos/+video-watch/comment/video-comments.component.ts
+++ b/client/src/app/videos/+video-watch/comment/video-comments.component.ts
@@ -130,6 +130,7 @@ export class VideoCommentsComponent implements OnInit, OnChanges, OnDestroy {
         this.componentPagination.totalItems = res.total
 
         this.onDataSubject.next(res.data)
+        this.hooks.runAction('action:video-watch.comments.loaded', 'video-watch', { data: this.componentPagination })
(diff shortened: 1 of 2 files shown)


client/src/app/videos/+video-watch/comment/video-comments.component.ts
```diff
@@ -130,6 +130,7 @@ export class VideoCommentsComponent implements OnInit, OnChanges, OnDestroy {
         this.componentPagination.totalItems = res.total
 
         this.onDataSubject.next(res.data)
+        this.hooks.runAction('action:video-watch.comments.loaded', 'video-watch', { data: this.componentPagination })
```
Chocobozzz: `action:video-watch.video-threads.loaded`

ipbc-dev: Done! Thanks a lot @Chocobozzz  !!!

PR #2542, opened 2020-03-07
Servicify menu, close menu on admin for medium screens

rigelk:


client/src/app/core/routing/menu-guard.service.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  static close () {
+    MenuGuards.guards.push(CloseMenuGuard)
+    return CloseMenuGuard
+  }
+}
+
+abstract class MenuGuard implements CanActivate, CanDeactivate<any> {
+  public display = true
```
Chocobozzz: Remove `public`

client/src/app/core/routing/menu-guard.service.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+  canActivate (): boolean {
+    // small screens already have the site-wide onResize from screenService
+    // > medium screens have enough space to fit the administrative menus
+    if (!this.screen.isInMobileView() && this.screen.isInMediumView()) {
+      this.menu.isMenuDisplayed = this.display
+    }
+    return true
+  }
+  public canDeactivate = this.canActivate
```
Chocobozzz: Remove `public` and put attributes below `display = true`

client/src/app/core/routing/menu-guard.service.ts
```diff
@@ -0,0 +1,45 @@
...
+    MenuGuards.guards.push(CloseMenuGuard)
+    return CloseMenuGuard
+  }
+}
+
+abstract class MenuGuard implements CanActivate, CanDeactivate<any> {
+  public display = true
+  constructor (protected menu: MenuService, protected screen: ScreenService, display: boolean) {
+    this.display = display
+  }
```
Chocobozzz: Add ` \n`

client/src/app/core/routing/menu-guard.service.ts
```diff
@@ -0,0 +1,45 @@
...
+  canActivate (): boolean {
+    // small screens already have the site-wide onResize from screenService
+    // > medium screens have enough space to fit the administrative menus
+    if (!this.screen.isInMobileView() && this.screen.isInMediumView()) {
+      this.menu.isMenuDisplayed = this.display
+    }
+    return true
+  }
```
Chocobozzz: Add ` \n`

client/src/app/core/routing/menu-guard.service.ts
```diff
@@ -0,0 +1,45 @@
...
+import { CanActivate, CanDeactivate } from '@angular/router'
+import { MenuService } from '@app/core/menu'
+import { ScreenService } from '@app/shared/misc/screen.service'
+
+@Injectable()
+export class MenuGuards {
+  public static guards: any = []
+
+  static open () {
+    MenuGuards.guards.push(OpenMenuGuard)
```
Chocobozzz: Why do you need to push `OpenMenuGuard`?

rigelk: To make the providers in the module definition properly consider them.

client/src/app/core/routing/menu-guard.service.ts
rigelk: This is a function, so it should be placed below the function it copies.

Chocobozzz: No: https://travis-ci.org/Chocobozzz/PeerTube/jobs/659521986

rigelk: That's news to me !

client/src/app/core/routing/menu-guard.service.ts
Chocobozzz: Okay then please use a `Set` or check duplicates so you don't add multiple times the same guard

client/src/app/core/routing/menu-guard.service.ts
```diff
@@ -0,0 +1,47 @@
+import { Injectable } from '@angular/core'
+import { CanActivate, CanDeactivate } from '@angular/router'
+import { MenuService } from '@app/core/menu'
+import { ScreenService } from '@app/shared/misc/screen.service'
+
+@Injectable()
+export class MenuGuards {
+  public static guards: any = []
```
Chocobozzz: I think you can type this to `MenuGuard[]`

rigelk:
@Chocobozzz I figured the current use is simple enough to just list both cases explicitly for now.

PR #2546, opened 2020-03-08
Settings and administration panels design rebalancing: grid, colors, font-{size, weight}

rigelk:
The objective is to better use the page width, and more properly signal different sections. The whole is tailored to be responsive. The left column in particular leaves place for `form-group` descriptions instead of `help` buttons which clutter the current design and fail at providing a direct insight.

Colors and `font-size`s are rebalanced in the navs and tabs as well as labels' `font-weight`, to better reflect an element's hierarchy in the page instead of basic emphasis relative to its child/parent element.

Screenshot_2020-03-09 Edit custom configuration - PeerTube(1)
Screenshot_2020-03-09 Edit custom configuration - PeerTube

### DIFF
--- a/client/src/app/+about/about.component.html
+++ b/client/src/app/+about/about.component.html
@@ -2,11 +2,11 @@
   <div class="sub-menu">
 
     <div class="links">
-      <a i18n routerLink="instance" routerLinkActive="active" class="title-page">Instance</a>
+      <a i18n routerLink="instance" routerLinkActive="active" class="title-page title-page-about">Instance</a>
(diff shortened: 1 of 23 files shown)


Chocobozzz:
Please paste some screens of the new interface

Chocobozzz:
And tests fails: https://travis-ci.org/Chocobozzz/PeerTube/jobs/659885313

rigelk:
@Chocobozzz I fixed tests and added screenshots.

PR #2550, opened 2020-03-09
Action hook: thread replies loaded

ipbc-dev:
Hi!
As I did last week for comments, I added a donation button to comment replies (only if you have peertube-plugin-bittube-airtime-module)
image
In order to do this from my plugin, I needed to add this hook for comment replies loaded, returning the array of replies. I would be nice to keep my plugin compatible with the federation by adding this PR. It also may be useful for somebody else. Thanks in advance!

### DIFF
--- a/shared/models/plugins/client-hook.model.ts
+++ b/shared/models/plugins/client-hook.model.ts
@@ -67,6 +67,8 @@ export const clientActionHookObject = {
   'action:video-watch.player.loaded': true,
   // Fired when the video watch page comments(threads) are loaded and load more comments on scroll
   'action:video-watch.video-threads.loaded': true,
+  // Fired when a user click on 'View x replies' and they're loaded
+  'action:video-watch.video-threads.reply-loaded': true,
 
   // Fired when the search page is being initialized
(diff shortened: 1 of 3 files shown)


shared/models/plugins/client-hook.model.ts
```diff
@@ -67,6 +67,8 @@ export const clientActionHookObject = {
   'action:video-watch.player.loaded': true,
   // Fired when the video watch page comments(threads) are loaded and load more comments on scroll
   'action:video-watch.video-threads.loaded': true,
+  // Fired when a user click on 'View x replies' and they're loaded
+  'action:video-watch.video-threads.reply-loaded': true,
```
Chocobozzz: `action:video-watch.video-thread-replies.loaded` for consistency with `filter:api.video-watch.video-thread-replies.list.params` filter

ipbc-dev: Done! Thanks @Chocobozzz !!

Chocobozzz:
Thanks @ipbc-dev

Issue #2545, opened 2020-03-08
Docker postfix - wrong OpenDKIM rights on keys folder - mails fail


kimsible:
**What happened?**

With the actual postfix image `wader/postfix-relay` in docker-compose, we can configure OpenDKIM keys generation and make them persistent by adding `OPENDKIM_DOMAINS` and a mounted volume on `/etc/opendkim/keys`.

Unfortunately, when mounting the volume the image does not handle strict permissions to `opendkim` user on the keys folder (https://github.com/wader/postfix-relay/pull/10).  So postfix can't get the private key and **email sending fails**.

**Steps to reproduce:**

1. Add `OPENDKIM_DOMAINS=<MY DOMAIN>=<selector>` to `.env` 

2. Add `- ./docker-volume/opendkim/keys:/etc/opendkim/keys` volume to the postfix service in `docker-compose.yml`

3. Start your containers `docker-compose up -d`

4. Add the DNS record in you domain found here `cat ./docker-volume/opendkim/keys/*/*.txt`

5. Make sure your registrar updated the DNS records `dig TXT <selector>._domainkey.<MY DOMAIN>` (without any subdomain)

6. Send an email with PeerTube (by lost password link for example)

7. Log PeerTube and Postfix containers  `docker-compose logs peertube` and `docker-compose logs postfix`, If you don't see any logs for postfix you need to connect inside and log `/var/log...` 

**Solutions**
First of all, the image `wader/postfix-relay` seems  not very maintained and is based on a debian image. I found another postfix image based on alpine linux https://github.com/bokysan/docker-postfix. Seems more complete on some features like using the port `587` and lighter than the previous one except there is also an issue about OpenDKIM rights on the keys folder : https://github.com/bokysan/docker-postfix/issues/11 and no generation keys implemented yet.

The solutions I propose are : 
- Handle DKIM signature in PeerTube with emailer / nodemailer and manual generation keys : https://github.com/Chocobozzz/PeerTube/compare/develop...kimsible:feat/dkim-emailer?expand=1
- Handle DKIM signature with the image `boky/postfix` and auto-generation keys / rights fixed in a `Dockerfile.postfix` https://github.com/Chocobozzz/PeerTube/compare/develop...kimsible:feat/dkim-custom-postfix-alpine-docker-image?expand=1

**Questions**
- Maybe, DKIM is not intended to be supported by PeerTube itself or official deployment setups ? **but this is another issue that not receiving email because of the domain verification**.
[...]

PR #2551, opened 2020-03-10
Add DKIM to docker-compose

kimsible:
Issue: https://github.com/Chocobozzz/PeerTube/issues/2545

Add DKIM generation keys and DKIM signature sending with the current image `mwader/postfix-relay`.

### DIFF
--- a/support/docker/production/.env
+++ b/support/docker/production/.env
@@ -20,6 +20,9 @@ PEERTUBE_SMTP_TLS=false
 PEERTUBE_SMTP_DISABLE_STARTTLS=false
 PEERTUBE_ADMIN_EMAIL=<MY EMAIL ADDRESS>
 POSTFIX_myhostname=<MY DOMAIN>
+# If you need to generate a list of sub/DOMAIN keys
+# pass them as a whitespace separated string <DOMAIN>=<selector>
+OPENDKIM_DOMAINS=<MY DOMAIN>=peertube
 TRAEFIK_ACME_EMAIL=<MY EMAIL ADDRESS>
 # If you need to obtain ACME certificates for more than one DOMAIN
 # pass them as a comma separated string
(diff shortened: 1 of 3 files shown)


support/docker/production/.env
```diff
@@ -20,6 +20,9 @@ PEERTUBE_SMTP_TLS=false
 PEERTUBE_SMTP_DISABLE_STARTTLS=false
 PEERTUBE_ADMIN_EMAIL=<MY EMAIL ADDRESS>
 POSTFIX_myhostname=<MY DOMAIN>
+# If you need to generate a list of sub/DOMAIN keys
+# pass them as a whitespace separated string <DOMAIN>=<selector>
+OPENDKIM_DOMAINS=<MY DOMAIN>=peertube
```
Chocobozzz: I think we should disable DKIM by default

kimsible: Without DKIM signature, as Postfix is the default smtp relay, DMARK verification fails so every email go into the spams, that's my initial though about enable DKIM. There has been 2 fixes on the image `mwader/postfix-relay`.

Chocobozzz: I'm not sure to understand: most people won't have a DMARC record. While I agree this is better to have SPF/DKIM and DMARC, we cannot force it by default.

kimsible: Enabling DKIM with that image simply means that DKIM signature will be sent according to the RSA key automatically generated. 

- With DKIM enabled and no DMARC record, RSA keys will be generated, mails sent but no guarantee to not be received as a spam.
- Without DKIM enabled and no DMARC record, no RSA keys, mails are sent and the same thing happens : spam.

Enable it won't change the final behaviour, sending mail even if there is no DMARC record.

I think enabling that feature is not forcing to set a DMARC record but showing a way for a good practice with Postfix as a smtp.

I also think that if Postfix is just partially setup (as a local smtp), we also should disable it by default and let in any case the user configure all the mail part alone.

Nutomic:
FYI since I set the correct reverse dns for my instance, I didnt have any complaints about receiving mail from peertube.social. So DKIM doesnt seem to be required from my perspective.

PR #2558, opened 2020-03-15
Add overview of a user's actions in user-edit

rigelk:
The PR adds common stats to have an easier overview of a user's actions. It also uses the new layout for settings, and displays a user's avatar and display name. I use breadcrumbs to distinguish edit/create user views, and help for navigation within future listings related to a user.

Screenshot_2020-03-15 Update a user - PeerTube

### DIFF
--- a/client/src/app/shared/users/user.model.ts
+++ b/client/src/app/shared/users/user.model.ts
@@ -51,11 +51,11 @@ export class User implements UserServerModel {
   videoQuotaDaily: number
   videoQuotaUsed?: number
   videoQuotaUsedDaily?: number
-  videoCount?: number
-  videoAbuseCount?: number
-  videoAbuseAcceptedCount?: number
-  videoAbuseCreatedCount?: number
-  videoCommentCount?: number
+  videosCount?: number
+  videoAbusesCount?: number
+  videoAbusesAcceptedCount?: number
+  videoAbusesCreatedCount?: number
+  videoCommentsCount?: number
 
   theme: string
 
@@ -84,11 +84,11 @@ export class User implements UserServerModel {
     this.videoQuotaDaily = hash.videoQuotaDaily
     this.videoQuotaUsed = hash.videoQuotaUsed
     this.videoQuotaUsedDaily = hash.videoQuotaUsedDaily
-    this.videoCount = hash.videoCount
-    this.videoAbuseCount = hash.videoAbuseCount
-    this.videoAbuseAcceptedCount = hash.videoAbuseAcceptedCount
-    this.videoAbuseCreatedCount = hash.videoAbuseCreatedCount
-    this.videoCommentCount = hash.videoCommentCount
+    this.videosCount = hash.videosCount
+    this.videoAbusesCount = hash.videoAbusesCount
+    this.videoAbusesAcceptedCount = hash.videoAbusesAcceptedCount
+    this.videoAbusesCreatedCount = hash.videoAbusesCreatedCount
+    this.videoCommentsCount = hash.videoCommentsCount
 
     this.nsfwPolicy = hash.nsfwPolicy
     this.webTorrentEnabled = hash.webTorrentEnabled
--- a/client/src/app/shared/users/user.model.ts
+++ b/client/src/app/shared/users/user.model.ts
@@ -51,6 +51,11 @@ export class User implements UserServerModel {
   videoQuotaDaily: number
   videoQuotaUsed?: number
   videoQuotaUsedDaily?: number
+  videoCount?: number
+  videoAbuseCount?: number
+  videoAbuseAcceptedCount?: number
+  videoAbuseCreatedCount?: number
+  videoCommentCount?: number
 
   theme: string
 
@@ -79,6 +84,11 @@ export class User implements UserServerModel {
     this.videoQuotaDaily = hash.videoQuotaDaily
     this.videoQuotaUsed = hash.videoQuotaUsed
     this.videoQuotaUsedDaily = hash.videoQuotaUsedDaily
+    this.videoCount = hash.videoCount
+    this.videoAbuseCount = hash.videoAbuseCount
+    this.videoAbuseAcceptedCount = hash.videoAbuseAcceptedCount
+    this.videoAbuseCreatedCount = hash.videoAbuseCreatedCount
+    this.videoCommentCount = hash.videoCommentCount
 
     this.nsfwPolicy = hash.nsfwPolicy
     this.webTorrentEnabled = hash.webTorrentEnabled
--- a/server/models/account/user.ts
+++ b/server/models/account/user.ts
@@ -87,6 +87,7 @@ const literalVideoQuotaUsed: any = [
 ]
 
 enum ScopeNames {
+  FOR_ME_API = 'FOR_ME_API',
   WITH_VIDEOCHANNELS = 'WITH_VIDEOCHANNELS',
   WITH_STATS = 'WITH_STATS'
 }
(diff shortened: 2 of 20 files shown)


rigelk:
@Chocobozzz is your time to make a review is limited, I think the critical part to review is the addition of scopes/literal to the UserModel, affecting `listUsers` and `getUser`.

client/src/app/shared/users/user.model.ts
```diff
@@ -51,6 +51,11 @@ export class User implements UserServerModel {
   videoQuotaDaily: number
   videoQuotaUsed?: number
   videoQuotaUsedDaily?: number
+  videoCount?: number
```
Chocobozzz: Please use plural `videosCount`, `videoAbusesCount` etc

server/models/account/user.ts
```diff
@@ -106,12 +123,66 @@ enum ScopeNames {
...
+      include: [
+        literalVideoQuotaUsed,
+        [
+          literal(
+            '(' +
+              'SELECT COUNT("video"."id") ' +
+              'FROM "video" ' +
+              'INNER JOIN "videoChannel" ON "videoChannel"."id" = "video"."channelId" ' +
+              'INNER JOIN "account" ON "account"."id" = "videoChannel"."accountId" ' +
+              'WHERE "account"."actorId" = "UserModel"."id"' +
```
Chocobozzz: `"account"."userId"`

server/models/account/user.ts
```diff
@@ -106,12 +123,66 @@ enum ScopeNames {
...
+            '(' +
+              'SELECT ("abuses", "acceptedAbuses")::text ' +
+              'FROM (' +
+                'SELECT COUNT("videoAbuse"."id") AS "abuses", ' +
+                       `COUNT(CASE WHEN "videoAbuse"."state" = ${VideoAbuseState.ACCEPTED} THEN 1 END) AS "acceptedAbuses" ` +
+                'FROM "videoAbuse" ' +
+                'INNER JOIN "video" ON "videoAbuse"."videoId" = "video"."id" ' +
+                'INNER JOIN "videoChannel" ON "videoChannel"."id" = "video"."channelId" ' +
+                'INNER JOIN "account" ON "account"."id" = "videoChannel"."accountId" ' +
+                'WHERE "account"."actorId" = "UserModel"."id"' +
```
Chocobozzz: `"account"."userId"`

server/models/account/user.ts
```diff
@@ -106,12 +123,66 @@ enum ScopeNames {
...
+          ),
+          'videoCount'
+        ],
+        [
+          literal(
+            '(' +
+              'SELECT ("abuses", "acceptedAbuses")::text ' +
+              'FROM (' +
+                'SELECT COUNT("videoAbuse"."id") AS "abuses", ' +
+                       `COUNT(CASE WHEN "videoAbuse"."state" = ${VideoAbuseState.ACCEPTED} THEN 1 END) AS "acceptedAbuses" ` +
```
Chocobozzz: Try to use `FILTER`: https://medium.com/little-programming-joys/the-filter-clause-in-postgres-9-4-3dd327d3c852

server/models/account/user.ts
```diff
@@ -106,12 +123,66 @@ enum ScopeNames {
...
+          ),
+          'videoAbuseCount'
+        ],
+        [
+          literal(
+            '(' +
+              'SELECT COUNT("videoAbuse"."id") ' +
+              'FROM "videoAbuse" ' +
+              'INNER JOIN "account" ON "account"."id" = "videoAbuse"."reporterAccountId" ' +
+              'WHERE "account"."actorId" = "UserModel"."id"' +
```
Chocobozzz: `"account"."userId"`

server/models/account/user.ts
```diff
@@ -106,12 +123,66 @@ enum ScopeNames {
...
+          ),
+          'videoAbuseCreatedCount'
+        ],
+        [
+          literal(
+            '(' +
+              'SELECT COUNT("videoComment"."id") ' +
+              'FROM "videoComment" ' +
+              'INNER JOIN "account" ON "account"."id" = "videoComment"."accountId" ' +
+              'WHERE "account"."actorId" = "UserModel"."id"' +
```
Chocobozzz: `"account"."userId"`

server/models/account/user.ts
```diff
@@ -87,7 +104,7 @@ enum ScopeNames {
   ]
 }))
 @Scopes(() => ({
-  [ScopeNames.FOR_ME_API]: {
+  [ScopeNames.WITH_VIDEOCHANNELS]: {
```
Chocobozzz: You still need a `ScopeNames.FOR_ME_API` scope that includes `UserNotificationSettingModel`

server/models/account/user.ts
```diff
@@ -637,6 +695,14 @@ export class UserModel extends Model<UserModel> {
   toFormattedJSON (this: MUserFormattable, parameters: { withAdminFlags?: boolean } = {}): User {
     const videoQuotaUsed = this.get('videoQuotaUsed')
     const videoQuotaUsedDaily = this.get('videoQuotaUsedDaily')
+    const videoCount = this.get('videoCount')
+    const videoAbuseCount = this.get('videoAbuseCount') as string
+    const videoAbuseCreatedCount = this.get('videoAbuseCreatedCount')
+    const videoCommentCount = this.get('videoCommentCount')
+
+    function parseAbuseCount (videoAbuseCount: string, position: number) {
```
Chocobozzz: I'm not sure to understand what is `videoAbuseCount` or `position`. Try to build your data directly using postgresql instead

server/models/account/user.ts
```diff
@@ -431,7 +486,10 @@ export class UserModel extends Model<UserModel> {
   }
 
   static loadById (id: number): Bluebird<MUserDefault> {
-    return UserModel.findByPk(id)
+    return UserModel.scope([
+      ScopeNames.WITH_VIDEOCHANNELS,
+      ScopeNames.WITH_STATS
```
Chocobozzz: I think it would be better to include stats on demand (for example with a query parameter `?includeStats=true/false`

server/models/account/user.ts
rigelk: oh, that's why some stats were bogus :sweat:

server/models/account/user.ts
rigelk: I'm just reusing the `INNER JOIN` and making `COUNT` with different conditions, so the literal output is in the form `(x,y)` (a string) and this function extracts the stat at position z. for instance, for `z = 0`, we get `x`. For `z = 1`, we get `y`. But that is only worth the trouble if making another join from the ground up is costly. It is, right? Right? :sweat_smile: I have no idea how/if sequelize optimises literals.

server/models/account/user.ts
rigelk: `withStats`? That way, it matches the scope name.

Chocobozzz: :+1:

server/models/account/user.ts
Chocobozzz: I think you can just do multiple `COUNT` with their own `FILTER` that are mapped to a specific name. I don't think it will change a lot the query performance

Chocobozzz:
Please add unit tests

server/models/account/user.ts
```diff
@@ -637,6 +725,14 @@ export class UserModel extends Model<UserModel> {
   toFormattedJSON (this: MUserFormattable, parameters: { withAdminFlags?: boolean } = {}): User {
     const videoQuotaUsed = this.get('videoQuotaUsed')
     const videoQuotaUsedDaily = this.get('videoQuotaUsedDaily')
+    const videosCount = this.get('videosCount')
+    const videoAbusesCount = this.get('videoAbusesCount') as string
+    const videoAbusesCreatedCount = this.get('videoAbusesCreatedCount')
+    const videoCommentsCount = this.get('videoCommentsCount')
+
+    function parseAbuseCount (videoAbusesCount: string, position: number) {
```
Chocobozzz: const [ videoAbusesCount, videoAbusesAcceptedCount ] = videoAbusesCount.split(':')

server/tests/api/users/users.ts
```diff
@@ -273,6 +273,12 @@ describe('Test users', function () {
 
       expect(userMe.specialPlaylists).to.have.lengthOf(1)
       expect(userMe.specialPlaylists[0].type).to.equal(VideoPlaylistType.WATCH_LATER)
+
+      // Check stats are included with withStats
+      expect(userGet.videosCount).to.be.a('number')
```
Chocobozzz: Checking field types is not enough: upload videos, add comments, abuses, update abuse states and check stats are correct

Chocobozzz:
Tests fail

rigelk:
@Chocobozzz failed tests seem to be unreleated to the PR. relaunching.

rigelk:
Failed tests seem to be caused by the machine running out of disk space. Should I split the tests some more @Chocobozzz?

PR #2594, opened 2020-04-02
[Docker] Move streaming-playlists to /data

xcffl:
Everything under `/app/storage` should be moved to `/data`. However, `streaming-playlists` is not listed yet. So there will still be `/app/storage/streaming-playlists`. This PR fixes the problem.

### DIFF
--- a/support/docker/production/config/production.yaml
+++ b/support/docker/production/config/production.yaml
@@ -53,6 +53,7 @@ storage:
   captions: '../data/captions/'
   cache: '../data/cache/'
   plugins: '../data/plugins/'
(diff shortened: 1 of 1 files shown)


xcffl:
By the way, `/app` is not persistant, which has caused so many hls-related problems...

Chocobozzz:
Oup's. Sorry and thank you!

Chocobozzz:
Fixed a typo in https://github.com/Chocobozzz/PeerTube/commit/4fd98756e246a8fd5818580cf09f0ba5f78a48fa

PR #2615, opened 2020-04-09
Add sendmail

immae:
This PR adds support for sendmail along smtp for e-mail configuration.

### DIFF
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -53,6 +53,10 @@ redis:
 
 # SMTP server to send emails
 smtp:
+  # smtp or sendmail
+  transport: smtp
+  # Path to sendmail command. Required if you use sendmail transport
+  sendmail: null
   hostname: null
   port: 465 # If you use StartTLS: 587
   username: null
--- a/server/lib/emailer.ts
+++ b/server/lib/emailer.ts
@@ -41,33 +41,43 @@ class Emailer {
     this.initialized = true
 
     if (isEmailEnabled()) {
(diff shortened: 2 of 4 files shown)


config/production.yaml.example
```diff
@@ -53,6 +53,8 @@ redis:
 
 # SMTP server to send emails
 smtp:
+  transport: smtp
```
Chocobozzz, line 56: Please add these keys in https://github.com/Chocobozzz/PeerTube/blob/develop/config/default.yaml too

server/lib/emailer.ts
```diff
@@ -41,33 +41,41 @@ class Emailer {
...
+          port: CONFIG.SMTP.PORT,
+          secure: CONFIG.SMTP.TLS,
+          debug: CONFIG.LOG.LEVEL === 'debug',
+          logger: bunyanLogger as any,
+          ignoreTLS: CONFIG.SMTP.DISABLE_STARTTLS,
+          tls,
+          auth
+        })
+      } else { // sendmail
+        this.transporter = createTransport({
```
Chocobozzz, line 73: Please add log "Using sendmail to send emails"

config/production.yaml.example
Chocobozzz, line 56: Add a comment `smtp or sendmail`

config/production.yaml.example
```diff
@@ -53,6 +53,8 @@ redis:
...
+  sendmail: null
```
Chocobozzz, line 57: Add a comment `Path to sendmail command. Required if you use sendmail transport`

immae:
Fixed @Chocobozzz thanks for the review!

Chocobozzz:
Thanks @immae

PR #2621, opened 2020-04-10
Add server plugin filter hooks for import with torrent and url

rigelk:


server/lib/job-queue/handlers/video-import.ts
```diff
@@ -144,6 +151,29 @@ async function processFile (downloader: () => Promise<string>, videoImport: MVid
     }
     videoFile = new VideoFileModel(videoFileData)
 
+    // Check we accept this video
+    const acceptParameters = {
+      videoFilePath: tempVideoPath,
+      videoFile,
+      user: videoImport.User
+    }
+    const acceptedResult = options.type === 'youtube-dl'
```
Chocobozzz, line 160: Instead of using a ternary on the `wrapFun` call:

`let hookName = ... ? 'filter:api.video.post-import-url.accept.result': '...'`

And then just call the function `wrapFun(isPostImportVideoAccepted, acceptParameters, hookName)`

shared/models/plugins/server-hook.model.ts
```diff
@@ -9,9 +9,13 @@ export const serverFilterHookObject = {
   // Used to get detailed video information (video watch page for example)
   'filter:api.video.get.result': true,
 
-  // Filter the result of the accept upload function
+  // Filter the result of the accept upload, import via torrent or url functions
   // If this function returns false then the upload is aborted with an error
   'filter:api.video.upload.accept.result': true,
+  'filter:api.video.pre-import-url.accept.result': true,
```
Chocobozzz, line 15: Please specify the difference between a `pre-import` and a `post-import`

rigelk: :+1:

shared/models/plugins/server-hook.model.ts
```diff
@@ -9,9 +9,13 @@ export const serverFilterHookObject = {
   // Used to get detailed video information (video watch page for example)
   'filter:api.video.get.result': true,
 
-  // Filter the result of the accept upload function
+  // Filter the result of the accept upload, import via torrent or url functions
```
kimsible, line 12: It means `filter:api.video.upload.accept.result` will also hook an import after downloading the video inside a job, how to inform the client ?

rigelk: No, it will only work for the direct upload.

rigelk: You cannot inform the client. I should check if a notification is sent.

kimsible: Maybe, we should inform with persistent message until the user delete the inserted import infos.

Via a simple alert in the video-watch view with a message like "Video import has been rejected."
[](https://postimg.cc/ctJhyrNQ)

And a status message in the video-list view: 

[](https://postimg.cc/CBG9QdVh)

if it's too much work, I can submit another PR ?

rigelk: WDYT @Chocobozzz ?

Chocobozzz: Not a bad idea: I think we can add another entry in the `VideoState` enum

shared/models/plugins/server-hook.model.ts
```diff
@@ -9,9 +9,13 @@ export const serverFilterHookObject = {
...
   // If this function returns false then the upload is aborted with an error
   'filter:api.video.upload.accept.result': true,
+  'filter:api.video.pre-import-url.accept.result': true,
+  'filter:api.video.pre-import-torrent.accept.result': true,
+  'filter:api.video.post-import-url.accept.result': true,
+  'filter:api.video.post-import-torrent.accept.result': true,
```
kimsible, line 18: Are these name filters definitive ?

rigelk: Not sure - let me check that on my local branch

rigelk: Yup, they *should* be.

kimsible: thanks !

shared/models/plugins/server-hook.model.ts
kimsible: @Chocobozzz @rigelk sorry to come back on this PR but the client message should be implemented in another PR ?

Chocobozzz: The `errorMessage` of your hook is displayed in the video import table

kimsible: Thanks @Chocobozzz, it sounds logic.

Issue #2602, opened 2020-04-04
Max resolution and max length / size for video and audio uploading


kimsible:
It's really a big issue for moderation. 

Actually we only can fix a daily and max stockage quota per user.

I wonder how can we implement easily : 

- A max resolution and max length or size for each video uploading (maybe per extension)
- A max length or size for each audio uploading (maybe per extension)

Would you have an idea to create this limitation without implementing a full-feature, maybe with ffmpeg ?

rigelk:
Could you please detail the moderation use case and how urgent the need is?

More precisely:
- how is unbridled video resolution currently problematic for you?
- wouldn't bitrate be a more fitting metric to consider (instead of resolution)? would it also apply to audio in your use case?
- how is unbridled video/audio duration currently problematic for you?
- what is the use case for different limits per extension?
- what is the use case for different limits between audio/video?

kimsible:
I simply need to not block all the threads if lots of videos / audio are being sent / transcoded at the same time according to the limitations of a server CPU / RAM.

> how is unbridled video resolution currently problematic for you?

> how is unbridled video/audio duration currently problematic for you?

I thought about **2K and 4K resolutions** and **duration lower than 1 hour for videos and 4~5 hours for audio** for a Youtube Standard HD format /  FLAC audio as metrics understandable for most of video makers but maybe it's not the right approach.

> what is the use case for different limits per extension?

> what is the use case for different limits between audio/video?

Actually only for audio, we sugguest audio live replay instead of video live replay; we want to encourage audio compression for audio with long duration, especially for lowless compression WAV or FLAC, but can be applied for all audio, again it may not be the right approach.

> wouldn't bitrate be a more fitting metric to consider (instead of resolution)? would it also apply to audio in your use case?

Yes, bitrate could be the solution for both audio and video. Do you think about setting the maximum bitrate in ffmpeg ?

rigelk:
> I simply need to not block all the threads if lots of videos / audio are being sent / transcoded at the same time according to the limitations of a server CPU / RAM.

I see. This is moderation in its broadest sense then: making use of resources in moderation - I'm re-labeling the issue accordingly.

> we sugguest _audio live replay_ instead of _video live replay_

What are they? If they correspond to a standard, I'm unfamiliar with both terms.

> we want to encourage audio compression for audio with long duration

I see, but complexity-wise, even long uncompressed audio streams are significantly lower on CPU than short video transcodes. It might not be worth the trouble.

[...]

PR #2627, opened 2020-04-10
Add notifier to plugin helpers

kimsible:
This is two helpers needed for the plugin `upload-limits`.

- ~getBaseClientScriptsRoute~
- notifier (ToastModule); asked in this issue : https://github.com/Chocobozzz/PeerTube/issues/2602

### DIFF
--- a/client/src/app/core/plugins/plugin.service.ts
+++ b/client/src/app/core/plugins/plugin.service.ts
@@ -258,11 +258,6 @@ export class PluginService implements ClientHook {
         return environment.apiUrl + `${pathPrefix}/${plugin.name}/${plugin.version}/static`
       },
 
-      getBaseClientScriptsRoute: () => {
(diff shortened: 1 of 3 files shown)


client/src/app/core/plugins/plugin.service.ts
```diff
@@ -257,6 +258,11 @@ export class PluginService implements ClientHook {
         return environment.apiUrl + `${pathPrefix}/${plugin.name}/${plugin.version}/static`
       },
 
+      getBaseClientScriptsRoute: () => {
```
Chocobozzz, line 261: Why do you need the client scripts path since peertube auto loads your scripts?

kimsible: Well, it was for a web-assembly file, but it can be put in the static assets and lazy-load path scripts are defined in Webpack config via the `publicPath`.

I delete it so.

Chocobozzz:
Thanks!

Issue #1016, opened 2018-09-04
import youtube subtitles 


PhieF:
Hi !
it seems possible to import youtube subs
https://github.com/rg3/youtube-dl/#subtitle-options

but I guess peertube doesn't do it, do you think it could be added ?

Thanks !

Booteille:
Is integrating this feature planned anytime soon?
I often have to import subtitles manually actually, that's a bit annoying.

Chocobozzz:
@Booteille No sorry

Booteille:
Okay! Thanks for the answer!

benpro:
For reference, this is how you do it with youtube-dl.

```
youtube-dl "https://www.youtube.com/watch?v=XXX" --all-subs --skip-download
```

I must say, auto-downloading and importing YT subtitles would be nice. Some language detection should be done too.

drzraf:
May #2473 + #2474 be sufficient to resolve this one their own?

Chocobozzz:
Implemented by @kimsible in https://github.com/Chocobozzz/PeerTube/commit/50ad0a1c1699fb1799c9ba2a99bf888894f88df4

PR #2629, opened 2020-04-11
Add import youtube subtitles

kimsible:
**Issue**: https://github.com/Chocobozzz/PeerTube/issues/1016


server/helpers/youtube-dl.ts
```diff
@@ -45,6 +51,35 @@ function getYoutubeDLInfo (url: string, opts?: string[]): Promise<YoutubeDLInfo>
...
+  return new Promise<YoutubeDLSubs>((res, rej) => {
+    const cwd = CONFIG.STORAGE.TMP_DIR
+    const options = opts || { all: true, format: 'vtt', cwd }
+
+    safeGetYoutubeDL()
+      .then(youtubeDL => {
+        youtubeDL.getSubs(url, options, (err, files) => {
+          if (err) return rej(err)
+
+          const subtitles = files.reduce((acc, filename) => {
```
Chocobozzz, line 64: Maybe a `.filter` based on the regex and a `.map` would be simpler?

Chocobozzz:
Thanks for the PR. Please could you add unit tests?

In https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/videos/video-imports.ts

Run you test using `npm run mocha -- --exit -r ts-node/register -r tsconfig-paths/register --bail server/tests/your-test.ts`

I added subtitles to the youtube video we use to test youtube imports: https://www.youtube.com/watch?v=msX3jv1XdvM

server/helpers/youtube-dl.ts
kimsible: A `.reduce` seems more appropriate to match the language code ? 
With `.filter` and `.map` we'll need two `.filter` to validate extension / language code and to match the language code.

Chocobozzz: oup's, you're right :+1:

kimsible:
> Thanks for the PR. Please could you add unit tests?
> 
> In https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/videos/video-imports.ts
> 
> Run you test using `npm run mocha -- --exit -r ts-node/register -r tsconfig-paths/register --bail server/tests/your-test.ts`
> 
> I added subtitles to the youtube video we use to test youtube imports: https://www.youtube.com/watch?v=msX3jv1X

> npm run mocha -- --exit -r ts-node/register -r tsconfig-paths/register --bail server/tests/your-test.ts

Thanks for adding subtitles to the youtube video :)

I added some unit tests, but can't confirm it's working, mocha does not pass the first timeout and output : 

```bash
  1) Test video imports
       "before all" hook for "Should import videos on server 1":
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/kimmy/peertube-git/server/tests/api/videos/video-imports.ts)
```

These tests are passed through the CI ?

Chocobozzz:
I'll fix the tests. But it's weird you have have this timeout issue :thinking:

rigelk:
Thanks a bunch @kimsible - this feature will definitely prove useful to many!

kimsible:
> I'll fix the tests. But it's weird you have have this timeout issue thinking

Thanks, I'll explore my install of PeerTube, maybe it's caused by a lack of memory.

PR #2639, opened 2020-04-14
Refactor follow/mute as modals in admin, add actions in abuse list

rigelk:
This PR refactors the follow page to a *reusable* modal, and uses it to allow preemptive domain mutes.
Screenshot_2020-04-14 Following list - PeerTube
Screenshot_2020-04-14 Muted instances - PeerTube(1)
Screenshot_2020-04-14 Muted instances - PeerTube

This PR also *begins* adding moderation actions from the actions button for video abuses.

Screenshot_2020-04-14 Video abuses list - PeerTube


client/src/app/+admin/follows/follows.routes.ts
```diff
@@ -40,15 +39,6 @@ export const FollowsRoutes: Routes = [
           }
         }
       },
-      {
-        path: 'following-add',
```
Chocobozzz, line 44: Maybe we should put a redirection here instead of deleting the route

PR #2654, opened 2020-04-17
Add markdown support to plugins

kimsible:
This PR adds to plugins :

- `markdown-text` and `markdown-enhanced` setting type to the server helpers
- `markdownRenderer` to the client helpers

As mentioned in this issue https://github.com/Chocobozzz/PeerTube/issues/2602, markdown is rendered with another library (marked) lazy-loaded over the internal markdown-it of PeerTube.

```js
registerSetting({
  name: 'my-markdown-area',
  label: 'Markdown text',
  type: 'markdown-text',
  default: false
})
```

```js
const { markdownRenderer } = peertubeHelpers

await markdownRenderer.textMarkdownToHTML('**My Bold Text**')
// return <strong>My Bold Text</strong>

await markdownRenderer.enhancedMarkdownToHTML('![alt-img](http://.../my-image.jpg)')
// return <img alt=alt-img src=http://.../my-image.jpg />
```

### DIFF
--- a/client/src/app/+admin/plugins/plugin-show-installed/plugin-show-installed.component.html
+++ b/client/src/app/+admin/plugins/plugin-show-installed/plugin-show-installed.component.html
@@ -10,6 +10,18 @@
       <label *ngIf="setting.type !== 'input-checkbox'" [attr.for]="setting.name" [innerHTML]="setting.label"></label>
       <input *ngIf="setting.type === 'input'" type="text" [id]="setting.name" [formControlName]="setting.name" />
       <textarea *ngIf="setting.type === 'input-textarea'" type="text" [id]="setting.name" [formControlName]="setting.name"></textarea>
+      <my-help *ngIf="setting.type === 'markdown-text'" helpType="markdownText"></my-help>
+      <my-help *ngIf="setting.type === 'markdown-enhanced'" helpType="markdownEnhanced"></my-help>
+      <my-markdown-textarea
+        *ngIf="setting.type === 'markdown-text'"
+        markdownType="text" [id]="setting.name" [formControlName]="setting.name" textareaWidth="500px" [previewColumn]="false"
+        [classes]="{ 'input-error': formErrors['settings.name'] }"
+      ></my-markdown-textarea>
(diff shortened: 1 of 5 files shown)


client/src/types/register-client-option.model.ts
```diff
@@ -27,5 +27,10 @@ export type RegisterClientHelpers = {
     confirm?: { value: string, action?: () => void }
   }) => void
 
+  markdownRenderer: {
+    textMarkdownToHTML: (textMarkdown: string) => void
```
Chocobozzz, line 31: `    textMarkdownToHTML: (textMarkdown: string) => Promise<string>`

client/src/types/register-client-option.model.ts
```diff
@@ -27,5 +27,10 @@ export type RegisterClientHelpers = {
...
+    enhancedMarkdownToHTML: (enhancedMarkdown: string) => void
```
Chocobozzz, line 32: `Promise<string>`

client/src/app/+admin/plugins/plugin-show-installed/plugin-show-installed.component.html
```diff
@@ -10,6 +10,18 @@ <h2>
       <label *ngIf="setting.type !== 'input-checkbox'" [attr.for]="setting.name" [innerHTML]="setting.label"></label>
       <input *ngIf="setting.type === 'input'" type="text" [id]="setting.name" [formControlName]="setting.name" />
       <textarea *ngIf="setting.type === 'input-textarea'" type="text" [id]="setting.name" [formControlName]="setting.name"></textarea>
+      <my-help *ngIf="setting.type === 'markdown-text'" helpType="markdownText"></my-help>
```
Chocobozzz, line 13: Please add `\n` to separate blocks

client/src/types/register-client-option.model.ts
kimsible: Thanks I also updated the documentation according to the returned type.

Chocobozzz:
Thanks

Issue #2673, opened 2020-04-24
Hidden dropdown sub-menu on touch devices


kimsible:
**What happened?**

The sub-menu in `/my-account/*` paths is displayed with an `overflow-x` to make it scroll on touch devices (small and mobile screens) . When we click on a multiple menu entry, the dropdown appears hidden and the sub-menu must scroll also vertically.

 
![my-account_nav_mobile](https://user-images.githubusercontent.com/1877318/80215223-62aebd00-863c-11ea-82c7-0996948a23a7.png)

I suggest we use a modal for displays that not fully fit the sub-menu horizontally, small and mobile views (screen width lower than 800px).

PR #2674, opened 2020-04-24
Use modal instead of dropdown menu in small/mobile views

kimsible:
Here is the solution suggested for this issue :  https://github.com/Chocobozzz/PeerTube/issues/2673

Before:
<div>

</div>

Now:
<div>

</div>

Edit : it may be an issue of NgbDropdown, on small screens sometimes the dropdown is inside sometimes outside but in any case dropdown on a horizontal overflow menu is such a pain on touch screens, and not really mobile-friendly.

### DIFF
--- a/client/src/app/shared/menu/top-menu-dropdown.component.ts
+++ b/client/src/app/shared/menu/top-menu-dropdown.component.ts
@@ -1,7 +1,5 @@
 import {
-  AfterViewInit,
   Component,
-  HostListener,
   Input,
   OnDestroy,
   OnInit,
@@ -31,7 +29,7 @@ export type TopMenuDropdownParam = {
   templateUrl: './top-menu-dropdown.component.html',
   styleUrls: [ './top-menu-dropdown.component.scss' ]
 })
-export class TopMenuDropdownComponent implements OnInit, OnDestroy, AfterViewInit {
+export class TopMenuDropdownComponent implements OnInit, OnDestroy {
   @Input() menuEntries: TopMenuDropdownParam[] = []
 
   @ViewChild('modal', { static: true }) modal: NgbModal
@@ -39,7 +37,6 @@ export class TopMenuDropdownComponent implements OnInit, OnDestroy, AfterViewIni
   suffixLabels: { [ parentLabel: string ]: string }
   hasIcons = false
   container: undefined | 'body' = undefined
-  isInSmallView = false
   isModalOpened = false
   currentMenuEntryIndex: number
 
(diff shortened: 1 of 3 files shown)


client/src/app/shared/menu/top-menu-dropdown.component.ts
```diff
@@ -85,6 +105,31 @@ export class TopMenuDropdownComponent implements OnInit, OnDestroy {
     this.openedOnHover = false
   }
 
+  @HostListener('window:resize')
+  onWindowResize () {
+    this.isInSmallView = !!this.screen.isInSmallView()
+  }
```
rigelk, line 111: Why not use a getter? This would prevent having to setup another listener, which is why we use a global service like ScreenService…

```
get isInSmallView () {
  return this.screen.isInSmallView()
}
```

client/src/app/shared/menu/top-menu-dropdown.component.ts
```diff
@@ -85,6 +105,31 @@ export class TopMenuDropdownComponent implements OnInit, OnDestroy {
...
+
+  openModal (index: number) {
+    this.currentMenuEntryIndex = index
+    this.isModalOpened = true
+
+    this.modalService.open(this.modal, {
+      centered: true, beforeDismiss: async () => {
```
rigelk, line 118:
```suggestion
      centered: true,
      beforeDismiss: async () => {
```

client/src/app/shared/menu/top-menu-dropdown.component.ts
```diff
@@ -23,38 +32,49 @@ export type TopMenuDropdownParam = {
   templateUrl: './top-menu-dropdown.component.html',
   styleUrls: [ './top-menu-dropdown.component.scss' ]
 })
-export class TopMenuDropdownComponent implements OnInit, OnDestroy {
+export class TopMenuDropdownComponent implements OnInit, OnDestroy, AfterViewInit {
   @Input() menuEntries: TopMenuDropdownParam[] = []
 
+  @ViewChild('modal', { static: true }) modal: ElementRef
```
rigelk, line 38:
```suggestion
  @ViewChild('modal', { static: true }) modal: NgbModal
```

rigelk: I know we are not very consistent with this throughout the application, but that gives a bit more info type-wise.

client/src/app/shared/menu/top-menu-dropdown.component.ts
kimsible: oh ! thanks I never thought about getter ! it's magic :)

rigelk:
Thanks a bunch @kimsible!

Issue #2676, opened 2020-04-24
Broken tabs nav of videos add form on small devices


kimsible:
**What happened?**
------
When at least 2 tabs of videos add methods (Upload, URL & Torrent Import) are enabled (especially in French language), the tabs appear broken on small devices (width lower than 800px).

<div>
<img width=49% alt=videos-upload-small src=https://user-images.githubusercontent.com/1877318/80231974-7108d300-8654-11ea-8f39-5334dc8e5541.png />
<img width=49% alt=videos-upload-mobile src=https://user-images.githubusercontent.com/1877318/80231976-71a16980-8654-11ea-83b6-cb74c9da7265.png />
</div>

**Solution**
------
The simplest way to handle this is to make the `.video-add-nav` tabs container horizontal scrollable and hide background and border of the active tab only on small views : 

<div>
<img width=49% alt=videos-upload-mobile--fr-fix src=https://user-images.githubusercontent.com/1877318/80232352-0e640700-8655-11ea-9fcb-0378875fa168.png />
<img width=49% alt=videos-upload-small-fr-fix src=https://user-images.githubusercontent.com/1877318/80232356-0e640700-8655-11ea-870c-9dda3abc4f8f.png />
<img width=49% alt=videos-upload-small-fix src=https://user-images.githubusercontent.com/1877318/80232357-0efc9d80-8655-11ea-9ba0-8633c123639d.png />
<img width=49% alt=videos-upload-mobile-fix src=https://user-images.githubusercontent.com/1877318/80232358-0efc9d80-8655-11ea-8d16-06c17761329f.png />

</div>

Chocobozzz:
Fixed by @kimsible

PR #2677, opened 2020-04-24
Make video-add-nav tabs scrollable on small devices

kimsible:
**Issue** : https://github.com/Chocobozzz/PeerTube/issues/2676

Before
------

<div>

</div>

After
------

<div>

</div>
<div>

</div>


Chocobozzz:
Thanks @kimsible

Issue #2675, opened 2020-04-24
Hidden dropdown on playlist video (watch view) caused by scroll


kimsible:
**What happened?**

When we watch our own playlist from `/videos/watch/playlist/:playlist-uuid`, we can access to a dropdown controls (Edit start/stop at and Delete from playlist) by clicking at right. 

Unfortunately, the fact that the playlist container is scrolling with a defined `max-height `under the current video playing, the dropdown of the two last videos may appear hidden and cause somtimes a weird effect with the scrollbar.

<img style="float:right" width=52% alt=videos-watch-playlist src=https://user-images.githubusercontent.com/1877318/80218182-229e0900-8641-11ea-9c58-9f8d9006a3c0.png />

**Solution**

Though these controls are already available in `/my-account/video-playlists/:playlist-uuid`, we may simply hide those ones on the watch view ?

<img style="float:right" width=52% alt=videos-watch-playlist-from-my-account src=https://user-images.githubusercontent.com/1877318/80218824-f636bc80-8641-11ea-8453-4742a66fb444.png />

Note than on YouTube, only the "delete trash icon" is visible on hover.

PR #2678, opened 2020-04-24
Avoid dropdown in scrolling own playlist watch (touchscreens)

kimsible:
This is a proposition for the issue : https://github.com/Chocobozzz/PeerTube/issues/2675

Make only the "delete option" available as YouTube does on watching playlist view to avoid dropdown menu in a scrolling content.

<div>

</div>

### DIFF
--- a/client/src/app/+my-account/my-account-video-playlists/my-account-video-playlist-elements.component.scss
+++ b/client/src/app/+my-account/my-account-video-playlists/my-account-video-playlist-elements.component.scss
@@ -48,3 +48,11 @@
     margin-top: -$sub-menu-margin-bottom-small-view;
   }
 }
+
+@media not all and (hover: hover) and (pointer: fine) {
+  .video {
+    .more {
+      opacity: 1;
+    }
+  }
+}
--- a/client/src/app/shared/video-playlist/video-playlist-element-miniature.component.html
+++ b/client/src/app/shared/video-playlist/video-playlist-element-miniature.component.html
@@ -36,7 +36,9 @@
     </div>
   </a>
 
-  <div *ngIf="owned" class="more" ngbDropdown #moreDropdown="ngbDropdown" placement="bottom-right"
+  <my-edit-button *ngIf="owned && touchScreenEditButton" [routerLink]="[ '/my-account', 'video-playlists', playlist.uuid ]"></my-edit-button>
+
+  <div *ngIf="owned" class="more" ngbDropdown #moreDropdown="ngbDropdown" placement="bottom auto"
        (openChange)="onDropdownOpenChange()" autoClose="outside"
   >
     <my-global-icon iconName="more-vertical" ngbDropdownToggle role="button" class="icon-more" (click)="$event.preventDefault()"></my-global-icon>
@@ -82,8 +84,9 @@
       </ng-container>
 
       <span class="dropdown-item" (click)="removeFromPlaylist(playlistElement)">
-            <my-global-icon iconName="delete"></my-global-icon> <ng-container i18n>Delete from {{ playlist?.displayName }}</ng-container>
-          </span>
+        <my-global-icon iconName="delete"></my-global-icon>
+        <ng-container i18n>Delete from {{ playlist?.displayName }}</ng-container>
+      </span>
     </div>
   </div>
 </div>
--- a/client/src/app/shared/video-playlist/video-playlist-element-miniature.component.scss
+++ b/client/src/app/shared/video-playlist/video-playlist-element-miniature.component.scss
@@ -36,6 +36,12 @@ my-video-thumbnail,
     }
   }
 
+  @media not all and (hover: hover) and (pointer: fine) {
+    .more {
+      opacity: 1 !important;
+    }
+  }
+
   &.playing {
     background-color: rgba(0, 0, 0, 0.02);
   }
--- a/client/src/app/shared/video-playlist/video-playlist-element-miniature.component.scss
+++ b/client/src/app/shared/video-playlist/video-playlist-element-miniature.component.scss
@@ -88,12 +88,15 @@ my-video-thumbnail,
     @include ellipsis;
   }
 
-  .more {
+  .more, my-edit-button {
     justify-self: flex-end;
     margin-left: auto;
     cursor: pointer;
-    opacity: 0;
     min-width: 24px;
+  }
+
+  .more {
+    opacity: 0;
 
     &.show {
       opacity: 1;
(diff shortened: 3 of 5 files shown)


Chocobozzz:
I think the dropdown is useful to quickly update video timestamps. @rigelk do you know how we could fix this issue?

kimsible:
> I think the dropdown is useful to quickly update video timestamps. @rigelk do you know how we could fix this issue?

@Chocobozzz I think the only way to fix that issue would be to detach the dropdown menu from the component, Souncloud uses that way in the sound playlist. But when you scroll after clicking on the more button the dropdown will be fixed, it's not very adapted to touch devices.

rigelk:
@kimsible this issue is commonly fixed by making the component bind to `body` instead of using its parent. See https://ng-bootstrap.github.io/#/components/dropdown/api and more specifically `container`. Putting `container="body"` on the dropdown element should do the trick.

kimsible:
> @kimsible this issue is commonly fixed by making the component bind to `body` instead of using its parent. See https://ng-bootstrap.github.io/#/components/dropdown/api and more specifically `container`. Putting `container="body"` on the dropdown element should do the trick.

@rigelk thanks for the trick ! unfortunately it does not work or didn't get how to do.. 
I put `container="body"` on the dropdown element (one with `ngbDropdown` and the other `ngbDropdownMenu`) and I can't reproduce the dropdown outsite the container scrolled.

By the way, this trick will fix the dropdown to the body ? so if you scroll at the same time on touch devices you need to listen an event to make the dropdown disappear ?

rigelk:
The library does this for us :)You can take the example of how we do it on the menu component with the user dropdown.

kimsible:
> The library does this for us :)You can take the example of how we do it on the menu component with the user dropdown.

Thanks @rigelk the trick on the user dropdown of the menu component works, the dropdown is not inside anymore, but it is worst depending on the playlist size the dropdown may flee go to the bottom or the top of the page.

I understand the useful side as @Chocobozzz  described but this useful dropdown only works on a small playlist and on a very large screen (when the playlist stick to the right of the page).

From my point of view, dropdown should never be used on moving/overflow content and touch screens.

Sure, the proposition remove the ability to quickly update video timestamps but it still available on another page.

I thought about the modal but if a video is playing we should stop it while displaying the modal.

I really don't see any quick other solutions to fix it but redesign this view.

video-watch-playlist-dropdown outside - 3
video-watch-playlist-dropdown outside - 2
video-watch-playlist-dropdown outside

Chocobozzz:
> Sure, the proposition remove the ability to quickly update video timestamps but it still available on another page.

I understand your issue, and I agree we should find another way to design this component. But I'm still convinced we need to quickly set the timestamp here, because playlists creator will check their timestamp in this view (using the player), and probably adapt them (increase one or two seconds etc).

kimsible:
> I understand your issue, and I agree we should find another way to design this component. But I'm still convinced we need to quickly set the timestamp here, because playlists creator will check their timestamp in this view (using the player), and probably adapt them (increase one or two seconds etc).

@Chocobozzz yes, the better way should be to re-design this view. But my proposition is a good compromise no ? In our instance it's a big issue especially because we have playlists creator that just can't access easily to that dropdown menu (especially the delete option) but it's not really editable from a CSS hack.

I think a user will for sure flee in front something buggy instead of a clean simple thing.

It's not like the feature has disappeared, it is just disabled temporally waiting when someone here would have time to re-design it.

To be frank, it would be a pleasure for me to do it as I proposed lots of things for PeerTube in a small time and several time far more for the community than the requests of our instance, I'm unpaid and volunteer here. It's too much energy for me to debate several days on one this simple thing, open source projects must be careful and not hurry to implement features, but it does not mean to be centralized decision making and obtuse. Questioning openly our own code and past decisions should be part of a project developed with a community.

rigelk:
@kimsible voluntary/unpaid time and energy is scarce, and I am in the same boat so be sure I understand the impatience when PRs take more time than expected to merge over seemingly trivial things.

Your code and PRs shows how much care went into them, and this is precious to us and the PeerTube community as a whole. It raises the bar for us, and this is why we want the best solution for this issue - and as a community, we have to discuss issues, sometimes agreeing, sometimes not. I know it is frustrating: we have to spend energy to convince others of the need for an issue that is painfully obvious to us, and then convince others of *how* it should be adressed.

You might be right on the design decision in this PR (more on that below). Regardless, you opened a PR without prior discussion in #2675 (chocobozzz had just time to label it as a bug), so we can discuss the right solution only *after* you spent energy coding yours. Here Chocobozzz even took some of his free time to answer this PR last saturday.

Issues/forum/IRC/Matrix/Mastodon might seem overkill, but they are there to test the waters, and avoid most cases of misunderstanding and wasted effort. They are also your best option at making the decision a community-driven process. This is even more true for UX problems, which can lead to a plethora of design changes and often need to be backed by user stories.

Most issues are not immediately tackled after they are open, so usually the discussion has time to take place before the implementation. Claiming an issue can be a good way to press the discussion to happen, too.

We cannot always answer immediately, and we might need to be called back to some issues, voluntary time or not. We might even prioritize other issues in the meantime. We are humans after all, so please bear with us :slightly_smiling_face: 

---

> I think a user will for sure flee in front something buggy instead of a clean simple thing.

Yes, definitely! In defense of chocobozzz, I think we have a distorted perception of what is buggy as maintainers, as we tend to see so many big and small that one like that went unnoticed. Thanks for spotting it!

I understand you take issue of the dropdown making the video list in the playlist mode on small screens expand and thus appear buggy to the user. I see a few options:

1. have a buttons row insead of the dropdown (though in that case we would still have the timestamps edition dropdown, which leads to 4.)
2. have a single button instead of the dropdown, but provide (a) a quick access to the playlist edition view above (a link to `/my-account/video-playlists/:uuid`)
3. <strike>use `container="body"`</strike> (wasn't good enough as you demonstrated)
4. use `placement="bottom auto"`

I would have tried 4. first, but 2. is good too so long as it features (a).

> From my point of view, dropdown should never be used on moving/overflow content and touch screens.

This is a good point. I would say "on moving content _with_ touchscreens", don't you think? That leads me to a potential alternative: have 2. (a) on touchscreens, and 4. on regular screens. What do you think?

kimsible:
@rigelk Thanks, I won't come back on that point that you've summarized well. My intention was not to hurry anyone, but for sure I will take time to open more issues in the future.
__________________________________________

I agree with everything, I tested your proposition which is a good compromise : 

- On regular screens the option 4. (`placement="bottom auto"`)  it seems ok 
videos-watch-playlist-edit-button-large-screen-fix

- On touchscreens the option  2. (a) (a quick access to the playlist edition view above, a link to `/my-account/video-playlists/:uuid`)
I propose an edit button, with label on tablet (small views) and hidden label on mobile, visible on video `.playing` . What do you think about that ?

videos-watch-playlist-edit-button-tablet-fix
videos-watch-playlist-edit-button-mobile-fix

Issue #2397, opened 2020-01-10
Add Followers and Followings count (UI) in the About page and a button to reveal the full lists


ghost:
Add Followers and Following count (UI) in the About page:
![adwq](https://user-images.githubusercontent.com/57924087/72138053-9ec35800-3394-11ea-8067-de2bbdd49e19.png)

Add a button to reveal the full lists:
I'd suggest showing only the first 20, 25 urls and then where the list ends, add a button that says - "show full list"
![full-list](https://user-images.githubusercontent.com/57924087/72138920-7a687b00-3396-11ea-9363-73563d5dc270.png)

Chocobozzz:
Why?

DontUseGithub:
> Add Followers and Following count (UI) in the About page

Makes the page more insightful. Later maybe we could also add a way to sort the lists by different factors such as age, nr. of local videos each instance has, etc.


> Add a button to reveal the full lists

The long lists don't look very appealing UI/UX-wise

MahdiTurki:
Hi, is this still not done? If not, may I work on it?

Chocobozzz:
@MahdiTurki yep you can work on it :+1:

MahdiTurki:
This seems more complicated than I thought. 
I am sorry, I have tried but I can't do it.

MahdiTurki:
I am having difficulties, I can't figure out how to create instances so that I can test what I have done. Since there are no actual instances in my database.

Chocobozzz:
Use https://peertube2.cpy.re in https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/environments/environment.hmr.ts#L4

MahdiTurki:
I managed to make the 'show full list' works so that when the users clicks on it it displays all the followers or followings. However I did not find a way to display the number of followers and followings.
I thought about importing all the followers and followings on init but that may take too long to load and it does not seem convenient.

PR #2688, opened 2020-04-27
Add button to display full list in About page

MahdiTurki:
#2397 

Instead of scrolling down to get more followers/followings, the user can press show full list that will display the full list. The button will disappear when pressed and the full list is displayed.
I haven't managed to display the number of followers/followings.

### DIFF
--- a/client/src/app/+about/about-follows/about-follows.component.html
+++ b/client/src/app/+about/about-follows/about-follows.component.html
@@ -1,7 +1,7 @@
 <div class="row">
   <h1 class="sr-only" i18n>Follows</h1>
   <div class="col-xl-6 col-md-12">
-    <h2 i18n class="subtitle">Followers instances</h2>
+    <h2 i18n class="subtitle">Followers instances ({{ followersPagination.totalItems }})</h2>
 
     <div i18n class="no-results" *ngIf="followersPagination.totalItems === 0">This instance does not have instances followers.</div>
 
@@ -13,13 +13,13 @@
         {{ follower }}
       </a>
     </ng-container>
-    
-    <a i18n class="showMore" *ngIf="!showMoreFollowers" (click)="loadAllFollowers()" (click)="showMoreFollowers=true">Show full list</a>
-    <a i18n class="showMore" *ngIf="showMoreFollowers" (click)= "showMoreFollowers=false">Show less</a>
+
+      <button i18n class="showMore" *ngIf="!showMoreFollowers && followersPagination.totalItems > 0" (click)="loadAllFollowers()" (click)="showMoreFollowers=true">Show full list</button> 
+      <button i18n class="showMore" *ngIf="showMoreFollowers" (click)= "showMoreFollowers=false">Show less</button> 
   </div>
 
   <div class="col-xl-6 col-md-12">
-    <h2 i18n class="subtitle">Followings instances</h2>
+    <h2 i18n class="subtitle">Followings instances ({{ followingsPagination.totalItems }})</h2>
 
     <div i18n class="no-results" *ngIf="followingsPagination.totalItems === 0">This instance is not following any other.</div>
 
@@ -32,8 +32,8 @@
       </a>
     </ng-container>
 
-    <a i18n class="showMore" *ngIf="!showMoreFollowings" (click)="loadAllFollowings()" (click)="showMoreFollowings=true">Show full list</a>
-    <a i18n class="showMore" *ngIf="showMoreFollowings" (click)="showMoreFollowings=false">Show less</a>
+    <button i18n class="showMore" *ngIf="!showMoreFollowings && followingsPagination.totalItems > 0" (click)="loadAllFollowings()" (click)="showMoreFollowings=true">Show full list</button>
+    <button i18n class="showMore" *ngIf="showMoreFollowings" (click)="showMoreFollowings=false">Show less</button>
   </div>
 
 </div>
--- a/client/src/app/+about/about-follows/about-follows.component.html
+++ b/client/src/app/+about/about-follows/about-follows.component.html
@@ -6,10 +6,16 @@
     <div i18n class="no-results" *ngIf="followersPagination.totalItems === 0">This instance does not have instances followers.</div>
 
     <a *ngFor="let follower of followers" [href]="buildLink(follower)" target="_blank" rel="noopener noreferrer">
-      {{ follower }}
+      {{follower}}
     </a>
-
-    <a i18n class="showMore" *ngIf="!showMoreFollowers" (click)="loadAllFollowers()" (click)= "showMoreFollowers=true">Show full list</a>
+    <ng-container *ngIf="showMoreFollowers">
+      <a *ngFor="let follower of moreFollowers" [href]="buildLink(follower)" target="_blank" rel="noopener noreferrer">
+        {{ follower }}
+      </a>
+    </ng-container>
+    
+    <a i18n class="showMore" *ngIf="!showMoreFollowers" (click)="loadAllFollowers()" (click)="showMoreFollowers=true">Show full list</a>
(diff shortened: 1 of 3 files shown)


DontUseGithub:
> The button will disappear when pressed and the full list is displayed.

I think the user would expect that button to go to the bottom of that list and have a reversed action and name ("hide full list").

client/src/app/+about/about-follows/about-follows.component.html
```diff
@@ -1,22 +1,26 @@
...
     <div i18n class="subtitle">Followings instances</div>
 
     <div i18n class="no-results" *ngIf="followingsPagination.totalItems === 0">This instance does not have instances followings.</div>
-
+    
     <a *ngFor="let following of followings" [href]="buildLink(following)" target="_blank" rel="noopener noreferrer">
       {{ following }}
     </a>
+    
+    <a id="showMore" href="javascript:;" *ngIf="!showMoreFollowings" (click)="loadAllFollowings()" (click)= "showMoreFollowings=true">Show full list</a>
```
rigelk, line 23: why do you use `href="javascript:;"`?

rigelk: please use a class instead of an id ; they benefit from the component encapsulation so ids are not as useful.

client/src/app/+about/about-follows/about-follows.component.scss
```diff
@@ -16,3 +16,7 @@ a {
 .no-results {
   justify-content: flex-start;
 }
+
+#showMore {
+  color: black;
```
rigelk, line 21: please use something more generic, like the `peertube-button-link` mixin.

client/src/app/+about/about-follows/about-follows.component.ts
```diff
@@ -47,24 +50,20 @@ export class AboutFollowsComponent implements OnInit {
     this.loadMoreFollowings()
   }
 
-  onNearOfBottom () {
-    this.onNearOfFollowersBottom()
+  loadAllFollowings () {
+    while(hasMoreItems(this.followingsPagination)) {
```
rigelk, line 54:
```suggestion
    while (hasMoreItems(this.followingsPagination)) {
```

client/src/app/+about/about-follows/about-follows.component.ts
```diff
@@ -47,24 +50,20 @@ export class AboutFollowsComponent implements OnInit {
...
-    this.loadMoreFollowers()
-  }
-
-  onNearOfFollowingsBottom () {
-    if (!hasMoreItems(this.followingsPagination)) return
-
-    this.followingsPagination.currentPage += 1
-    this.loadMoreFollowings()
+  loadAllFollowers () {
+    while(hasMoreItems(this.followersPagination)) {
```
rigelk, line 62:
```suggestion
    while (hasMoreItems(this.followersPagination)) {
```

client/src/app/+about/about-follows/about-follows.component.ts
```diff
@@ -47,24 +50,20 @@ export class AboutFollowsComponent implements OnInit {
...
+      this.followersPagination.currentPage += 1
+      
+      this.loadMoreFollowers();
```
rigelk, line 65:
```suggestion
      this.loadMoreFollowers()
```

client/src/app/+about/about-follows/about-follows.component.ts
```diff
@@ -16,6 +16,9 @@ export class AboutFollowsComponent implements OnInit {
   followers: string[] = []
   followings: string[] = []
 
+  showMoreFollowers = false;
+  showMoreFollowings = false;
```
rigelk, line 20:
```suggestion
  showMoreFollowers = false
  showMoreFollowings = false
```

client/src/app/+about/about-follows/about-follows.component.html
rigelk: please use the `i18n` attribute.

rigelk: please remove trailing spaces.

client/src/app/+about/about-follows/about-follows.component.html
```diff
@@ -1,22 +1,26 @@
-<div class="row" myInfiniteScroller [autoInit]="true" (nearOfBottom)="onNearOfBottom()" [dataObservable]="onDataSubject.asObservable()">
+<div class="row">
   <div class="col-xl-6 col-md-12">
     <div i18n class="subtitle">Followers instances</div>
 
     <div i18n class="no-results" *ngIf="followersPagination.totalItems === 0">This instance does not have instances followers.</div>
-
+
```
rigelk, line 6: please remove trailing spaces.

MahdiTurki:
> > The button will disappear when pressed and the full list is displayed.
> 
> I think the user would expect that button to go to the bottom of that list and have a reversed action and name ("hide full list").

I think it won't matter that much. Why would he want to hide the full list if he chose to display it?

DontUseGithub:
> 
> I think it won't matter that much. Why would he want to hide the full list if he chose to display it?

Various reasons, maybe he has clicked it by mistake or perhaps he wanted to search for a specific name (via ctrl+F) and then hide one list and open the second one. 

But it just feels right to have a way of reverting back whatever you've done, I even remember reading about this somewhere, it's how good designs are made.

> Make actions reversible – be forgiving
> 
> This rule means that the user should always be able to quickly backtrack whatever they are doing. This allows users to explore the product without the constant fear of failure — when a user knows that errors can be easily undone, this encourages exploration of unfamiliar options. On the contrary, if a user has to be extremely careful with every action they take, it leads to a slower exploration and nerve-racking experience that no one wants.

https://xd.adobe.com/ideas/process/ui-design/4-golden-rules-ui-design/

MahdiTurki:
You are totally right. I will keep that in mind in the future.
I will think of a way on how to hide the list and will try to implement it.

DontUseGithub:
>  I will think of a way on how to hide the list and will try to implement it.

I hope you meant not to hide it totally, just revert it back to how it was before the user has clicked that button :)

MahdiTurki:
> I hope you meant not to hide the entire list, just revert it back to how it was before the user has clicked that button :)

My bad! Yes that's what I meant.

MahdiTurki:
Now the user has the option to show the full list, and when it's pressed, the user can scroll down, hide the full list and will only see the first 20 followers/followings.

test2a:
a quick question here. how about we display server stats along with the video stored size on the server? i mean there are servers on the stats page with thousands of videos. If we had a way of knowing the cpu size or the ram of that server, one could understand the transcoding their server has to do, the people it has to serve and other stuff

kontrollanten:
@rigelk Is there anything more that needs to be done on this (except solving the file conflict)?

rigelk (changes requested):
@kontrollanten yes, showing the count of instances. The UI for empty lists has not been worked on either (buttons should not appear), and actions should match the color scheme for links.

kontrollanten:
@MahdiTurki Thanks for your contribution! Do you think that you can do the last fixes requested by @rigelk ? It would be nice to be able to merge this :)

MahdiTurki:
Hi, sure thing. I will start working on them soon.
> and actions should match the color scheme for links. 


@rigelk, I don't get what you mean exactly, could you please elaborate more?

rigelk:
@MahdiTurki just have links use the main color :wink:

kontrollanten:
> @MahdiTurki just have links use the main color

Is it the Show more/less links that you're thinking of? I guess we should add the `button-grey` mixin and remove the `padding: 0`;

image

rigelk:
Either is fine

MahdiTurki:
I managed to display the count of instances. I tested with 2 different instances and it worked fine.
I also managed to hide the show/hide button when there are no followers or followings.

Chocobozzz:
Merged manually. Thanks @MahdiTurki

PR #2691, opened 2020-04-28
Remove border-bottom on header to prevent header redimensioning

kimsible:
Small fix preventing header size change when clicking on the left menu; the header elements (Logo, Title, Search Bar and Upload Button) have a weird effect of being resized / re-align vertically.

If that `border-bottom` is really useful on the left and right header elements, we can put it on the parent, so the `.header` element as proposed.

On the contrary, if that `border-bottom` on expanded view is useless (it is just a grey line inside a grey gradient...) we can drop it ? @rigelk  ?

### DIFF
--- a/client/src/app/app.component.html
+++ b/client/src/app/app.component.html
@@ -2,7 +2,7 @@
 
 <my-hotkeys-cheatsheet></my-hotkeys-cheatsheet>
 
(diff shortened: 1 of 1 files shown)


rigelk:
Thanks for spotting the issue @kimsible - it is indeed extraneous.

PR #2692, opened 2020-04-28
Fix broken buttons on mobile history view

kimsible:
Small to fix broken buttons on mobile history view, for example in French language :

<div>

</div>


rigelk:
Nice catch! Thanks @kimsible

Issue #2694, opened 2020-04-28
Unreadable tables of user-account on small devices


kimsible:
All tables in user-account are unreadable on small devices (small and especially mobile views).

The quickest way to make it readable and accessible may be to fix a `min-with` to the table then `overflow-x: auto` and `width: 100vw` on it's wrapper.

![my-account-tables-1](https://user-images.githubusercontent.com/1877318/80494659-34481f00-8967-11ea-87b6-d42f17044981.png)

PR #2695, opened 2020-04-28
Fix tables width, paginators, button on mobile / touchscreens

kimsible:
Proposition for the issue : https://github.com/Chocobozzz/PeerTube/issues/2694

It also fixes the display of actions buttons in an `overflow` table view.

**Before**
my-account-tables-1

**After**
my-account-tables-1-fix

### DIFF
--- a/client/src/app/+my-account/my-account.component.scss
+++ b/client/src/app/+my-account/my-account.component.scss
@@ -1,5 +1,3 @@
-
-
 .row {
   flex-direction: column;
   width: 100%;
--- a/client/src/app/shared/buttons/action-dropdown.component.html
+++ b/client/src/app/shared/buttons/action-dropdown.component.html
@@ -1,4 +1,4 @@
-<div class="dropdown-root" ngbDropdown [placement]="placement" container="body" autoClose="outside" *ngIf="areActionsDisplayed(actions, entry)">
+<div class="dropdown-root" ngbDropdown [placement]="placement" container="body" *ngIf="areActionsDisplayed(actions, entry)">
   <div
     class="action-button" [ngClass]="{ small: buttonSize === 'small', grey: theme === 'grey', orange: theme === 'orange', 'button-styled': buttonStyled }"
     ngbDropdownToggle role="button"
--- a/client/src/app/shared/buttons/action-dropdown.component.html
+++ b/client/src/app/shared/buttons/action-dropdown.component.html
@@ -1,4 +1,4 @@
-<div class="dropdown-root" ngbDropdown [placement]="placement" *ngIf="areActionsDisplayed(actions, entry)">
+<div class="dropdown-root" ngbDropdown [placement]="placement" container="body" autoClose="outside" *ngIf="areActionsDisplayed(actions, entry)">
   <div
     class="action-button" [ngClass]="{ small: buttonSize === 'small', grey: theme === 'grey', orange: theme === 'orange', 'button-styled': buttonStyled }"
     ngbDropdownToggle role="button"
--- a/client/src/sass/application.scss
+++ b/client/src/sass/application.scss
@@ -340,6 +340,16 @@ table {
         width: 100% !important;
       }
     }
+
+    &.expanded {
+      @include overflow-datatable($small-view, $expanded-horizontal-margins/3);
+    }
+  }
+}
+
+@media screen and (max-width: #{$small-view + $menu-width}) {
+  .main-col:not(.expanded) {
+    @include overflow-datatable($small-view, $not-expanded-horizontal-margins);
   }
 }
 
--- a/client/src/sass/application.scss
+++ b/client/src/sass/application.scss
@@ -267,7 +267,7 @@ table {
   top: #{-($header-height + 20px)};
 }
 
-@media screen and (max-width: #{map-get($grid-breakpoints, xxl)}) {
+@media screen and (max-width: #{breakpoint(xxl)}) {
   .main-col {
     &.expanded {
       .margin-content {
@@ -278,7 +278,7 @@ table {
   }
 }
 
-@media screen and (max-width: #{map-get($grid-breakpoints, lg)}) {
+@media screen and (max-width: #{breakpoint(lg)}) {
   /* the following applies from 500px to 900px and is partially overriden from 500px to 800px by changes below to $small-view */
   .main-col {
     &, &.expanded {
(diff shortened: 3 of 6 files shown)


client/src/app/+my-account/my-account.component.scss
```diff
@@ -8,3 +8,63 @@
...
+              margin-left: 5px;
+            }
+
+            :first-child {
+              margin-left: 0px;
+            }
+          }
+        }
+      }
+    }
```
rigelk, line 38: This block might be factorized in a mixin.

rigelk:
Having a table that displays properly on mobile is something that we might want to apply not just to that view, but also to all instances of `p-table` used throughout PeerTube. Your fix is generic enough to improve that situtation on other tables by placing it in `application.scss`.

It is a ton of work to check all other tables for regressions, but the potential value is huge for mobile users. Potentially in another PR or by someone else, of course :slightly_smiling_face: What do you think?

kimsible:
> Having a table that displays properly on mobile is something that we might want to apply not just to that view, but also to all instances of `p-table` used throughout PeerTube. Your fix is generic enough to improve that situtation on other tables by placing it in `application.scss`.


Thanks for the review, I first thought about all tables, but admins have `dropdown` menus (so it will rase the same issue than touchscreens on the playlist :/). 

> It is a ton of work to check all other tables for regressions, but the potential value is huge for mobile users. Potentially in another PR or by someone else, of course slightly_smiling_face What do you think?

Yes ! I agree, plus it will be very useful for moderators to have an access on touchscreens.

I can place that code in  `application.scss` and exclude admin tables ?

rigelk:
> Thanks for the review, I first thought about all tables, but admins have `dropdown` menus (so it will raise the same issue than touchscreens on the playlist :/).

You are only affecting the container's `overflow-x` this time, so that problem will not repeat itself there. Probably. :sweat_smile: 

> I can place that code in `application.scss` and exclude admin tables ?

Maybe more `primeng-custom.scss` since the change mainly affects `p-table`. You can do a first version excluding the admin tables if you fear regressions there, but you should test visually any other affected table - because that's what I will have to do anyway :wink: 

> I forgot to change the `:hover` event to a continuously display for touchscreens

What do you mean?

kimsible:
> > Thanks for the review, I first thought about all tables, but admins have `dropdown` menus (so it will raise the same issue than touchscreens on the playlist :/).
> 
> You are only affecting the container's `overflow-x` this time, so that problem will not repeat itself there. Probably. sweat_smile
> 
With the same trick of the user dropdown on the left menu I suppose ?

> > I forgot to change the `:hover` event to a continuously display for touchscreens
> 
> What do you mean?

Action-buttons appear on `td:hover` this event does not exist on touchscreen, no ?

rigelk:
> > You are only affecting the container's `overflow-x` this time, so that problem will not repeat itself there. Probably. sweat_smile
> 
> With the same trick of the user dropdown on the left menu I suppose ?

No, I mean that the dropdown doesn't have the issue currently (no overflow rule present) and won't have it with your additions since you only affect the horizontal overflow, and the dropdown… well… drops vertically. Am I overlooking something?

> Action-buttons appear on the line hover event `tr:hover` this event does not exist on touchscreen, no ?

Ah, for action buttons. You're right.

kimsible:
These last commits apply this on every tables (including admin), also :

- Keep displaying and fix action-cell elements on touchscreens (not :hover)

my-account-tables-ownership-mobile


- Fix paginator on mobile views

admin-tables-showing
admin-tables-showing-fix-2



- Apply dropdown trick the `dropdown-root` on action-cell 

admin-tables-dropdown
admin-tables-dropdown-fix

kimsible:
@rigelk I fixed the `min-width` of the table to `800px / $small-view`, do you think it is enough for admins tables ?

rigelk:
> @rigelk I fixed the `min-width` of the table to `800px / $small-view`, do you think it is enough for admins tables ?

It might be a bit short. 900px/lg breakpoint is better.

kimsible:
@rigelk thanks, just to inform that this PR may depend on that one about contents responsive styles (expanded/not-expanded) : https://github.com/Chocobozzz/PeerTube/pull/2698

client/src/sass/application.scss
```diff
@@ -331,3 +331,38 @@ table {
...
+@media screen and (max-width: #{map-get($grid-breakpoints, lg)}) {
+  .main-col {
+    &.expanded {
+      @include overflow-datatable(map-get($grid-breakpoints, lg), $expanded-horizontal-margins/3);
+    }
+
+    &:not(.expanded) {
+      // TODO: $horizontal-margins might change to $expanded-horizontal-margins/3 + $menu-width/2
+      @include overflow-datatable(map-get($grid-breakpoints, lg), $not-expanded-horizontal-margins + $menu-width/2);
+    }
```
kimsible, line 358: Required the PR https://github.com/Chocobozzz/PeerTube/pull/2698 to change the `$horizontal-margins` and compact this block :

rigelk: I merged the aforementioned PR.

client/src/app/shared/buttons/action-dropdown.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="dropdown-root" ngbDropdown [placement]="placement" *ngIf="areActionsDisplayed(actions, entry)">
+<div class="dropdown-root" ngbDropdown [placement]="placement" container="body" autoClose="outside" *ngIf="areActionsDisplayed(actions, entry)">
```
rigelk, line 1:
```suggestion
<div class="dropdown-root" ngbDropdown [placement]="placement" container="body" *ngIf="areActionsDisplayed(actions, entry)">
```

You cannot use `autoClose="outside"`: the action-dropdown then doesn't close on menu clicks, which is required in some scenarios (coming to mind, the addition of a video in a playlist, where the previous dropdown is kept open as the new one opens…).

client/src/sass/application.scss
rigelk: It got merged :slightly_smiling_face:

kimsible:
Thanks @rigelk, I think about the paginator on mobile, the items are well limited to 5 ?

kimsible:
> @kimsible what de you mean?

More we have items more we have pages in the paginator (seems to be limited to 5) but there is a solution using the "Programmatic Control" https://www.primefaces.org/primeng/showcase/#/table/page

rigelk:
> More we have items more we have pages in the paginator (seems to be limited to 5) but there is a solution using the "Programmatic Control" [primefaces.org/primeng/showcase/#/table/page](https://www.primefaces.org/primeng/showcase/#/table/page)

IMHO simpler: use `pageLinkSize="n"` on the `p-paginator` component, where `n` is the number of pages we wish to display.

kimsible:
> > More we have items more we have pages in the paginator (seems to be limited to 5) but there is a solution using the "Programmatic Control" [primefaces.org/primeng/showcase/#/table/page](https://www.primefaces.org/primeng/showcase/#/table/page)
> 
> IMHO simpler: use `pageLinkSize="n"` on the `p-paginator` component, where `n` is the number of pages we wish to display.
> 
> EDIT: c.f. https://www.primefaces.org/primeng/showcase/#/paginator

Thanks, maybe the Programmatic Control in another PR later ?

rigelk:
> Thanks, maybe the Programmatic Control in another PR later ?

Isn't limiting the paginator via `pageLinkSize` enough? At size 1 it gives the same behaviour and is one component less to maintain :slightly_smiling_face:

kimsible:
> Isn't limiting the paginator via `pageLinkSize` enough? At size 1 it gives the same behaviour and is one component less to maintain slightly_smiling_face

size 1 is enough, we need the screenService on each table to change dynamically the pageLinkSize

rigelk:
> size 1 is enough, we need the screenService on each table to change dynamically the pageLinkSize

It would be more manageable in another PR.

Could you resolve conflicts and rebase on develop first?

kimsible:
> > size 1 is enough, we need the screenService on each table to change dynamically the pageLinkSize
> 
> It would be more manageable in another PR.

Yes it seems not so easy to implement..., btw, 5 pages-links appears on a screen width ~400px.

> Could you resolve conflicts and rebase on develop first?

I did it on the last commit

kimsible:
> LGTM! Thanks so much for taking the time to go the extra mile to make the table improvement generic smiley

Yeah ! next time the paginator, you're welcome, using PeerTube on tablets and big smartphones is very asked :)

client/src/sass/application.scss
```diff
@@ -364,51 +363,16 @@ table {
     }
 
     &:not(.expanded) {
-      // TODO: $horizontal-margins might change to $expanded-horizontal-margins/3 + $menu-width/2
       @include overflow-datatable(breakpoint(lg), $not-expanded-horizontal-margins + $menu-width/2);
```
kimsible, line 366: @rigelk when you rebased you forgot to change the `$horizontal-margins` to `$expanded-horizontal-margins/3` here, no ?

rigelk: Ah, right. I assumed you had forgotten to remove the TODO since your PR wasn't a draft anymore.

rigelk: I'll change that in the next commit I do.

kimsible: Thanks !

rigelk: @kimsible I integrated the change in c2a89b70c

PR #2671, opened 2020-04-22
Fix broken views on mobile and small screens

kimsible:
## textareas

Fix `width: 100%` to textareas in small and mobile screens as it is applied for inputs  

## accounts/user channels
Better display of channels headers of users.
@rigelk is this respond to this issue https://github.com/Chocobozzz/PeerTube/issues/2440 ?
<div>

</div>

## my-account/videos-channels
`padding-bottom` on channel item + center every infos
<div>

</div>

## my-account/subscriptions
center every infos + `flex-direction: column` to mobile and small screens
<div>

</div>

## my-account/video-channels/create
input name width 100% instead of auto
<div>

</div>

## my-account/settings 
quota progress higher width than screen width instead of 100%
<div>
 
 
</div>

## my-account/playlists
`flex-direction` header `column`
<div>
 

</div>

## my-account/videos
better display of control buttons + `flex-direction` header `column`
also add the $minimum-view (320px) in Sass variables
<div>
 
 
</div>

## my-account/notifications
 `flex-direction` header `column` + margin-bottom on first button
<div>

</div>

## search channels
reproduce same css as mobile screen for small screen
<div>

</div>

## my-account/video-channels
margin-top of `.playlist-info` same as `.sub-menu` margin-bottom in small-screens
<div>

</div>

PR #2701, opened 2020-04-29
Fix select width on view < 800px (mobile)

kimsible:
Very small fix linked to this merged PR : https://github.com/Chocobozzz/PeerTube/pull/2671

The peertube-select where not included in the CSS rules for screen width under 800px, it caused an `overflow-x` on mobile view (between ~ `340px`  and  ~`400px` screen width)

### DIFF
--- a/client/src/sass/application.scss
+++ b/client/src/sass/application.scss
@@ -325,7 +325,8 @@ table {
       input[type=text],
       input[type=password],
       input[type=email],
-      textarea {
+      textarea,
+      .peertube-select-container {
         width: 100% !important;
       }
     }


test2a:
https://ibb.co/d4cLZkj
hey. just found this visual bug on peertub2.cpy.re on my mobile phone just minutes ago.
observations:
1. the stats page does not expand to 100%,  as you mentioned. 
2. the header has what appears to be fixed background and that hides the text and other elements as it sticks above all. 
3. if i resize the browser window, the sidebar does NOT appear in the sidebar on a phone but i think it assumes i have a different resolution and shows up as a smaller window which is not intuitive

rigelk:
@test2a please port your comment over to https://github.com/Chocobozzz/PeerTube/pull/3004, which works on a fix for this.

PR #2702, opened 2020-04-29
Make admin sub-navs, selects, buttons responsive

kimsible:
Small PR that makes the rest of the admin responsive (sub-navs, selects, buttons), what do you think ? 

<div>

</div>

_____

<div>

</div>

_____

<div>

</div>

____

<div>

</div>

### DIFF
--- a/client/src/sass/application.scss
+++ b/client/src/sass/application.scss
@@ -322,19 +322,16 @@ table {
         margin-bottom: $sub-menu-margin-bottom-small-view;
       }
 
-      .admin-sub-header:first {
+      .admin-sub-header {
         flex-direction: column;
 
         .admin-sub-nav {
           width: 100%;
         }
 
-        > :first-child {
+        .form-sub-title {
           margin-right: 0px !important;
           margin-bottom: 10px;
-        }
-
-        .form-sub-title {
           text-align: center;
         }
 
--- a/client/src/sass/application.scss
+++ b/client/src/sass/application.scss
@@ -322,6 +322,36 @@ table {
         margin-bottom: $sub-menu-margin-bottom-small-view;
       }
 
+      .admin-sub-header:first {
+        flex-direction: column;
+
+        .admin-sub-nav {
+          width: 100%;
+        }
+
+        > :first-child {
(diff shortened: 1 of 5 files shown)


rigelk:
I can't seem to obtain the same result of moderation/follows `admin-sub-header`.

kimsible:
> @kimsible I can't seem to obtain the same result of moderation/follows `admin-sub-header`.

@rigelk Sorry, I commit a wrong version you can now test with the last commit.

Also, delete the "p-table" on the console, these are not fully responsive, the other PR is not merged yet : https://github.com/Chocobozzz/PeerTube/pull/2695

kimsible:
By the way, there still another thing to fix on the jobs view

client/src/sass/application.scss
```diff
@@ -338,3 +361,32 @@ table {
...
+          text-align: center;
+        }
+
+        .admin-sub-nav {
+          display: block;
+          overflow-x: auto;
+          white-space: nowrap;
+          height: 50px;
+          padding: 10px 0;
+          width: calc(100vw - #{$menu-width + ($expanded-horizontal-margins/3) *2});
```
rigelk, line 383: since this is the only varying part in the whole block and that you repeat twice, generally make a mixin to factorize it. Mixin are inexpensive but greatly improve maintainability (since then you spot immediately what the pivoting part between two usages is.

kimsible: I've placed the mixin into `sass/include/_mixins.scss`, right place for that kind of mixin ?

rigelk: Yes ; if a mixin is local to a component style file, we place it there, otherwise it goes in mixins. There are some exceptions for the miniature, but otherwise that's pretty much our mixin library. It only adds to a SCSS file weight if you specifically `@include` it, so you can put as many as you want there.

kimsible: thanks for the explanations and all the time you give to review all these PR :)

rigelk: You're welcome :slightly_smiling_face: Keep up the good work!

PR #2707, opened 2020-05-01
Fix regression my-account menu overflow-x on screen width < 800px

kimsible:
This PR fix a regression on the menu of `/my-account/*` which was not horizontally scrollable anymore on screen width < 800px.

account-menu-no-scroll

### DIFF
--- a/client/src/app/core/menu/menu.service.ts
+++ b/client/src/app/core/menu/menu.service.ts
@@ -7,7 +7,7 @@ import { debounceTime } from 'rxjs/operators'
 export class MenuService {
   isMenuDisplayed = true
   isMenuChangedByUser = false
-  menuWidth = 240
+  menuWidth = 240  // should be kept equal to $menu-width
 
   constructor (
     private screenService: ScreenService
--- a/client/src/app/core/menu/menu.service.ts
+++ b/client/src/app/core/menu/menu.service.ts
@@ -7,6 +7,7 @@ import { debounceTime } from 'rxjs/operators'
 export class MenuService {
   isMenuDisplayed = true
   isMenuChangedByUser = false
+  menuWidth = 240
 
   constructor (
     private screenService: ScreenService
--- a/client/src/app/shared/menu/top-menu-dropdown.component.ts
+++ b/client/src/app/shared/menu/top-menu-dropdown.component.ts
@@ -11,6 +11,7 @@ import { Subscription } from 'rxjs'
 import { NgbDropdown, NgbModal } from '@ng-bootstrap/ng-bootstrap'
 import { GlobalIconName } from '@app/shared/images/global-icon.component'
 import { ScreenService } from '@app/shared/misc/screen.service'
+import { MenuService } from '@app/core/menu'
 
 export type TopMenuDropdownParam = {
   label: string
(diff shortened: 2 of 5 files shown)


rigelk:
This prevents dropdowns in the sub-menu to work.

kimsible:
> From memory, this prevents dropdowns in the sub-menu to work.

Yes, but there is no dropdowns anymore in the sub-menu under 800px since we add a modal menu ?

client/src/sass/application.scss
```diff
@@ -383,6 +384,10 @@ table {
       .admin-sub-header {
         @include admin-sub-header-responsive($menu-width/2 + $expanded-horizontal-margins/3);
       }
+
+      .sub-menu {
+        overflow-x: auto;
+      }
```
kimsible, line 390: I think we can delete this one, when the left menu is opened, the dropdowns of the the user-menu will not be replaced by a modal with a screen width higher than 800px (cause handling this with the screenService was a bit tricky compared to css).

In any case, for that kind of small device we can lock the scroll to show dropdown ? the user will hide the left menu to see the rest of the menu ?

kimsible:
It's now fixed, it handles the left menu displayed when the screen width is higher than 800px by defining the `menuWidth` (240px) in the menu service.

client/src/app/shared/menu/top-menu-dropdown.component.ts
```diff
@@ -46,11 +46,17 @@ export class TopMenuDropdownComponent implements OnInit, OnDestroy {
...
-    private screen: ScreenService
+    private screen: ScreenService,
+    private menuService: MenuService
   ) { }
 
   get isInSmallView () {
-    return this.screen.isInSmallView()
+    let marginLeft = 0
+    if (this.menuService.isMenuDisplayed) {
+      marginLeft = this.menuService.menuWidth
```
rigelk, line 56: I know this is practical, but getters shouldn't have side-effects.

client/src/app/shared/misc/screen.service.ts
```diff
@@ -10,7 +10,12 @@ export class ScreenService {
...
 
-  isInSmallView () {
+  isInSmallView (marginLeft = 0) {
+    if (marginLeft > 0) {
+      const contentWidth = this.getWindowInnerWidth() - marginLeft
+      return contentWidth < 800
+    }
+
     return this.getWindowInnerWidth() < 800
   }
```
rigelk, line 20: Since this modifies the logic of a service function, have you checked instances of components using it for regressions?

config/test.yaml
```diff
@@ -87,8 +87,8 @@ import:
     http:
       enabled: true
       proxy:
-        enabled: false
-        url: ""
+        enabled: true
+        url: "http://188.165.225.149:7899"
```
rigelk, line 91: Please revert.

client/src/app/core/menu/menu.service.ts
```diff
@@ -7,6 +7,7 @@ import { debounceTime } from 'rxjs/operators'
 export class MenuService {
   isMenuDisplayed = true
   isMenuChangedByUser = false
+  menuWidth = 240
```
rigelk, line 10:
```suggestion
  menuWidth = 240  // should be kept equal to $menu-width
```

rigelk: This allows backtracking to the source value when looking for its name. If you say there is no clean way to do it in css, then we can only do that much :/

kimsible: Thanks !

client/src/app/shared/misc/screen.service.ts
kimsible: Yes,  every call is without any parameter, if `marginLeft = 0`  it will be the same behaviour than before because the default value of `marginLeft` is `0`.

Issue #2710, opened 2020-05-01
Menu/header not fixed in my-account/videos&playlists on touchscreen from 800px


kimsible:
These two components `my-account/videos` and `my-account/playlists` are not very responsive from a 800px screen width, that cause sometimes an horizontal scroll of the body (from 800px to 900px / 1100px).

The problem is when the left-menu is displayed, on touchscreens the menu and the header are not fixed anymore, that will break the entire page when scrolling down : 

![my-account-menu-left-between-from-800px](https://user-images.githubusercontent.com/1877318/80826159-0a4e6100-8be2-11ea-8301-5fcc611aab7a.png)

PR #2711, opened 2020-05-01
Fix non-fixed header and menu when overflow-x body on touchscreens

kimsible:
This can respond to : https://github.com/Chocobozzz/PeerTube/issues/2710

The previous solution was not good enough, it broke all the `ngbTooltips` and asked some tricks for `action-dropdown` and `help-component`.

This new solution adapts  `my-account/videos` and `my-account/channels` on screen width from 800px to large screens with menu-left displayed.

This PR depends on the resolution of https://github.com/Chocobozzz/PeerTube/pull/2707

- [x] Responsive view `my-account/videos` from 800px with menu
- [x] Responsive view  `my-account/channels` from 800px with menu
- [x] Responsive view  `search` from 800px with menu

my-account-videos-large-screens-with-menu-fix
my-account-videos-from-900px-with-menu-fix
my-account-videos-800px-900px-with-menu-fix

my-account-videos-channels-800-to-900px-with-menu-fix
my-account-videos-channels-large-screens-with-menu-fix

______

Previous solution : 

>I'm not sure of the proposition because it needs more tests with the dropdowns attached to the body container.
>
>This fix proposes to set `fixed` position to `.sub-header-container` and use `.main-col` container as the scrollable container, this way all fixed elements (`.header`, `.menu`, `.sub-header-container`) are independent each others.
>
>**EDIT**: Some mainstream GAFAM apps use this way for their layout (Twitch, Google trad...) but I wonder if it could block the SEO, @rigelk have you any ideas ?
>
>vertical-scroll-main-col-fix-header
>vertical-scroll-main-col-fix
>
>For reminder : 
>
>vertical-scroll-main-col

### DIFF
--- a/client/src/app/+my-account/my-account-video-channels/my-account-video-channels.component.html
+++ b/client/src/app/+my-account/my-account-video-channels/my-account-video-channels.component.html
@@ -19,14 +19,14 @@
 
       <div i18n class="video-channel-followers">{videoChannel.followersCount, plural, =1 {1 subscriber} other {{{ videoChannel.followersCount }} subscribers}}</div>
 
+      <div class="video-channel-buttons">
+        <my-edit-button [routerLink]="[ 'update', videoChannel.nameWithHost ]"></my-edit-button>
+        <my-delete-button (click)="deleteVideoChannel(videoChannel)"></my-delete-button>
+      </div>
+
       <div *ngIf="!isInSmallView" class="w-100 d-flex justify-content-end">
         <p-chart *ngIf="videoChannelsChartData && videoChannelsChartData[i]" type="line" [data]="videoChannelsChartData[i]" [options]="chartOptions" width="40vw" height="100px"></p-chart>
       </div>
-    </div>
-
-    <div class="video-channel-buttons">
-      <my-edit-button [routerLink]="[ 'update', videoChannel.nameWithHost ]"></my-edit-button>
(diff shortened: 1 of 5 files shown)


kimsible:
So, I repassed on dropdowns and tooltips, all appended to the body are now fixed.
Except the user-dropdown on the left menu, all the others one could be auto-closed when scroll ?

kimsible:
Thanks @rigelk for fixing regression on dropdowns.

The last commits add : 

- Append any dropdown inside a table to the body (`container"body"`) 
- Add a listener on `#content.main-col` scroll to remove any dropdown append to body

It fixes the floating dropdown problem, might be useful for https://github.com/Chocobozzz/PeerTube/pull/2678.

client/src/app/app.component.ts
```diff
@@ -93,6 +93,15 @@ export class AppComponent implements OnInit, AfterViewInit {
...
   }
 
+  @HostListener('scroll')
+  onContentScroll() {
+    const dropdown = document.querySelector('body > .dropdown')
+
+    if (dropdown) {
+      document.body.removeChild(dropdown)
+    }
+  }
```
rigelk, line 103: WDYT @Chocobozzz?

kimsible: Maybe it would be better to listen scroll event only when an `action-dropdown` is opened and destroy the event after removing the append dropdown to the body ?

PR #2720, opened 2020-05-03
Add in-text filters to video-abuse-list, fix tables pagination update

rigelk:
Inspired by GitHub/Gitlab's issue filters. It has the advantage of not overloading the interface and being very easy to extend. The parser is put in the model utils so as to be reusable for other models.

Screenshot_2020-05-03 Video abuses list - PeerTube

Also contained:
- practical "clear filter" button on table filters
- `video-abuse-list` createdAt format switched to 'short' - it retains the date precision which allowing a 50px gain in column width
- `videoAbuse` updatedAt date now only shown if different from createdAt


server/models/utils.ts
```diff
@@ -219,6 +219,59 @@ function searchAttribute (sourceField, targetField) {
...
+      } else {
+        return e.startsWith(p.prefix)
+      }
+    })).join(' '),
+    // filters defined in prefixes are added under their own name
+    ...objectMap(prefixes, p => {
+      if (typeof p === "string") {
+        return tokens.filter(e => e.startsWith(p)).map(e => e.slice(p.length)) // we keep the matched item, and remove its prefix
+      } else {
+        const _tokens = tokens.filter(e => e.startsWith(p.prefix)).map(e => e.slice(p.prefix.length)).map(p.handler)
```
Chocobozzz, line 262: :face_with_head_bandage:

PR #2758, opened 2020-05-16
Add kubernetes helm chart and README

pulpo:
I created a helm chart that allows to deploy peertube into a kubernetes cluster, into what I think it's a very simple solution. 

A README.md file was created with all the steps needed to achieve this, and how to configure (without changing the configuration, it will deploy on a domain that resolves to a 127.0.0.1 IP, with default values).

Please review it and let me know if I need to change something.


rigelk:
Hi pulpo, I'm afraid there is no way for us to maintain the chart on the long run. Rather than merging it in the main repository, I suggest you add yours to the list of applications on the documentation.

rigelk:
Also, it might be of interest to @lecygnenoir

pulpo:
Got it @rigelk, I will keep this helm chart into a repository of an opensource organization I belong to, we will maintain the helm chart on our repo, make an automated helm package and make a PR on documentation when ready.

rigelk:
@pulpo thanks for your understanding, and keep up the good work!

LecygneNoir:
Hello @pulpo !

I may be interested indeed!

I have written a chart too (https://git.lecygnenoir.info/LecygneNoir/peertube-helm), a little simpler (no integrated postgres but with nginx frontend for cache) and I would be interested to see yours (as I am basically a helm newbie :-D)

Does your repo already available, or do you want to wait for your chart to be published in helm repo?

Thanks :-)

pulpo:
Hi @LecygneNoir , nice to meet you! 
The code of the repo (and a pipeline that package the helms) is here https://gitlab.com/jsl3/helmrepo, if you see the readme you will find that you only need to run 2 commands, one to apply the repo and another to install (in case you don't use some gitops approach at your k8s).
Also, for PeerTube you can see https://gitlab.com/jsl3/helmcharts/-/tree/master/peertube details on how to customize it, you can download the values.yaml, customize it and pass like a parameter for helm with -f flag. I moved all the configuration to the values.yaml, and I customize peertube directly from there, also I've removed the nginx-proxy, I left the ingress of kubernetes to manage it for me.

Issue #2777, opened 2020-05-22
Database migration from PeerTube v2.1.1 to v.2.2.0-rc1 / docker image


kimsible:
Hi, 

Is there any tool to migrate database breaking changes ?

I tried to migrate a test instance from PeerTube v2.1.1 to v2.2.0-rc1 and I got some `SequelizeDatabaseError`.

On the watch / edit / delete video view : 
```
"err": "SequelizeDatabaseError: column VideoFileModel.metadataUrl does not exist\n    at Query.formatError (/app/node_modules/sequelize/lib/dialects/postgres/query.js:366:16)\n    at query.catch.err (/app/node_modules/sequelize/lib/dialects/postgres/query.js:72:18)\n    at tryCatcher (/app/node_modules/bluebird/js/release/util.js:16:23)\n    at Promise._settlePromiseFromHandler (/app/node_modules/bluebird/js/release/promise.js:547:31)\n    at Promise._settlePromise (/app/node_modules/bluebird/js/release/promise.js:604:18)\n    at Promise._settlePromise0 (/app/node_modules/bluebird/js/release/promise.js:649:10)\n    at Promise._settlePromises (/app/node_modules/bluebird/js/release/promise.js:725:18)\n    at _drainQueueStep (/app/node_modules/bluebird/js/release/async.js:93:12)\n    at _drainQueue (/app/node_modules/bluebird/js/release/async.js:86:9)\n    at Async._drainQueues (/app/node_modules/bluebird/js/release/async.js:102:5)\n    at Immediate.Async.drainQueues [as _onImmediate] (/app/node_modules/bluebird/js/release/async.js:15:14)\n    at runCallback (timers.js:705:18)\n    at tryOnImmediate (timers.js:676:5)\n    at processImmediate (timers.js:658:5)"
```

On video upload : 
```
[...]

PR #2780, opened 2020-05-23
Fix migration number

kimsible:
**Issue**: https://github.com/Chocobozzz/PeerTube/issues/2777


Chocobozzz:
Thanks, but I'll update this manually because I need to change the script. The release tarball contains the migration script 485

PR #2798, opened 2020-05-27
Update nginx conf to include varnish configuration

McFlat:
Add varnish configuration to nginx peertube so to improve website response times and performance overall, please see the dependencies.md doc for changes that are required for this change https://github.com/Chocobozzz/PeerTube/pull/2799

PR #2799, opened 2020-05-28
Added varnish as part of install steps

McFlat:
For performance reasons we should add varnish as part of default website install steps, instead of serving peertube through nginx on 443 we should add varnish as a middle layer to serve the website out of memory. One more commit to follow that modifies the nginx config file. https://github.com/Chocobozzz/PeerTube/pull/2798

### DIFF
--- a/support/doc/dependencies.md
+++ b/support/doc/dependencies.md
@@ -36,9 +36,36 @@
 
 ```
 sudo apt update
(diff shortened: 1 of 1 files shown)


Chocobozzz:
See https://github.com/Chocobozzz/PeerTube/pull/2798#issuecomment-635145009

We don't want to add this dependency in a classic peertube installation.

PR #2801, opened 2020-05-28
Support broadcast messages

Chocobozzz:
screen_2020-05-28-11:16:15

 * Add ability to enabled/disable message
 * Supports markdown
 * Optional dismissable message
 * 3 message levels: `info`, `warning`, `error` (using bootstrap theme)


rigelk:
I would rename it into "banner" text, since broadcast implies that you can send actual messages, that would i.e. stack in a "announcement" section or widget.

Chocobozzz:
Yes you're right, but it's just I want to be able to change how we display the message in the future, like gitlab does (using a notification etc).

Issue #944, opened 2018-08-22
Do not federate unlisted videos


supermamie:
**PeerTube version or URL:**

PeerTube v1.0.0-beta.10 

**Introduction:**

When a video is uploaded, it is always with federation in mind.
It is either private (only the uploader can see it), or federated (public or unlisted).

This is totally intended as PeerTube is a federated video tool.

**What is the behaviour I would like to have?**

But in some case, I want to upload a video to stay on my instance and not be federated.

(in my specific case, it is to upload family videos that should remain private and I am not 100% sure I can trust each and every one instance that is following mine)

Ideally, more granular rights could be added, but unlisted and not federated could be nice.

1. People on my instance with the link (unlisted, not federated)
2. People on my instance (public, not federated)
3. People with an account on my instance (I do not think it is possible now, see #520 )

* **What do you see instead?**

The normal behaviour of a federated tool :+1:

Chocobozzz:
I don't think we want to add the "federation" notion in the video form. It's too complicated, and most users won't understand it. Let's try to keep things simple: maybe a private video with the ability to set other local account that can see your video would fit your needs. Or maybe https://github.com/Chocobozzz/PeerTube/issues/520

supermamie:
#520 can do, I'll just have to create a "guest" account with no upload capacity and share it for the people I want to share the video to.
The problem is that I will have to create a new account each time I want to upload a new video with a limited audience (my family, at work, some people I talk to over the Internet, etc...)

I know it should be possible to "trust" other servers, but I already have seen some problems like this (between Mastodon and Pleroma I think) where unlisted/private toots became public.

I also know that the fediverse is not made to host private data and I should maybe install PeerTube and another video tool which is not federated instead of wanting to unfederate PeerTube.

I still have a question : why are unlisted videos sent to other servers ?
(I think the answer is "so they will become available to followers", but I follow my account via Mastodon and do not see unlisted video, but I see the number of post which is the exact total of video (public + unlisted))

Chocobozzz:
> I know it should be possible to "trust" other servers

No you're right, you should not trust the federation.

> why are unlisted videos sent to other servers ?

Well :thinking: You make a good point. The only benefit is to share an unlisted video from different instance (`https://peertube.example1.com/videos/uuid` or `https://peertube.example2.com/videos/uuid`) is it really useful?

rigelk:
Doesn't seem useful to me. Users will share the direct URL in all likelihood.

[...]

PR #2802, opened 2020-05-28
Make federation of unlisted videos an instance-level server preference

Tak:
As discussed on #944

The previous behavior is the default for the preference.
I've added tests for:
- the preference itself
- actual video federation based on the preference
- accessing unlisted videos on the origin server

### DIFF
--- a/client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
+++ b/client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
@@ -115,7 +115,6 @@ export class EditCustomConfigComponent extends FormReactive implements OnInit, A
 
         isNSFW: false,
         defaultNSFWPolicy: null,
-        federateUnlistedVideos: true,
 
         terms: null,
         codeOfConduct: null,
@@ -136,6 +135,11 @@ export class EditCustomConfigComponent extends FormReactive implements OnInit, A
         customizations: {
           javascript: null,
           css: null
+        }
+      },
+      federation: {
+        videos: {
+          federateUnlisted: true
         }
       },
       theme: {
--- a/client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
+++ b/client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
@@ -115,6 +115,7 @@ export class EditCustomConfigComponent extends FormReactive implements OnInit, A
 
         isNSFW: false,
         defaultNSFWPolicy: null,
+        federateUnlistedVideos: true,
 
         terms: null,
         codeOfConduct: null,
--- a/client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
+++ b/client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
@@ -137,11 +137,6 @@ export class EditCustomConfigComponent extends FormReactive implements OnInit, A
           css: null
         }
       },
-      federation: {
-        videos: {
-          federateUnlisted: true
-        }
-      },
       theme: {
         default: null
       },
--- a/client/src/app/core/server/server.service.ts
+++ b/client/src/app/core/server/server.service.ts
@@ -42,7 +42,6 @@ export class ServerService {
       defaultClientRoute: '',
       isNSFW: false,
       defaultNSFWPolicy: 'do_not_list' as 'do_not_list',
-      federateUnlistedVideos: true,
       customizations: {
         javascript: '',
         css: ''
@@ -58,6 +57,11 @@ export class ServerService {
       registered: [],
       registeredExternalAuths: [],
       registeredIdAndPassAuths: []
+    },
+    federation: {
+      videos: {
+        federateUnlisted: true
+      }
     },
     theme: {
       registered: [],
--- a/client/src/app/core/server/server.service.ts
+++ b/client/src/app/core/server/server.service.ts
@@ -58,11 +58,6 @@ export class ServerService {
       registeredExternalAuths: [],
       registeredIdAndPassAuths: []
     },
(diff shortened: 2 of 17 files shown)


config/default.yaml
```diff
@@ -325,6 +325,7 @@ instance:
   # By default, "do_not_list" or "blur" or "display" NSFW videos
   # Could be overridden per user with a setting
   default_nsfw_policy: 'do_not_list'
+  federate_unlisted_videos: true
```
Chocobozzz, line 328: Please put this in a dedicated key, above https://github.com/Chocobozzz/PeerTube/blob/develop/config/production.yaml.example#L183:

```
federation:
  videos:
    federated_unlisted: true
```

server/helpers/express-utils.ts
```diff
@@ -28,6 +29,19 @@ function buildNSFWFilter (res?: express.Response, paramNSFW?: string) {
   return null
 }
 
+function isPrivacyForFederation(privacy: VideoPrivacy) {
```
Chocobozzz, line 32: Why did you move these functions in this file? `VideoModel` or `helpers/videos.ts` seems more appropriate

config/default.yaml
Chocobozzz: If you want admins to be able to update this setting using the web interface, then put this at the end of the produdction.yaml file. Don't forget to also update the default.yaml file

client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
```diff
@@ -137,6 +137,11 @@ export class EditCustomConfigComponent extends FormReactive implements OnInit, A
           css: null
         }
       },
+      federation: {
+        videos: {
+          federateUnlisted: true
```
Chocobozzz, line 142: Do you want admins to update this setting on the web interface? If no, you can remove this

client/src/app/core/server/server.service.ts
```diff
@@ -58,6 +58,11 @@ export class ServerService {
       registeredExternalAuths: [],
       registeredIdAndPassAuths: []
     },
+    federation: {
+      videos: {
+        federateUnlisted: true
```
Chocobozzz, line 63: Do you want admins to update this setting on the web interface? If no, you can remove this

server/controllers/api/config.ts
```diff
@@ -348,6 +353,11 @@ function customConfig (): CustomConfig {
     theme: {
       default: CONFIG.THEME.DEFAULT
     },
+    federation: {
+      videos: {
+        federateUnlisted: CONFIG.FEDERATION.VIDEOS.FEDERATE_UNLISTED
```
Chocobozzz, line 358: Do you want admins to update this setting on the web interface? If no, you can remove this

server/initializers/config.ts
```diff
@@ -154,6 +154,11 @@ const CONFIG = {
       URL: config.get<string>('plugins.index.url')
     }
   },
+  FEDERATION: {
+    VIDEOS: {
+      get FEDERATE_UNLISTED () { return config.get<boolean>('federation.videos.federate_unlisted') }
```
Chocobozzz, line 159: Do you want admins to update this setting on the web interface? If no, don't use a getter

server/tests/api/check-params/config.ts
```diff
@@ -47,6 +47,11 @@ describe('Test config API validators', function () {
         css: 'body { background-color: red; }'
       }
     },
+    federation: {
+      videos: {
+        federateUnlisted: false
```
Chocobozzz, line 52: Do you want admins to update this setting on the web interface? If no, you can remove this

server/tests/api/videos/video-privacy.ts
```diff
@@ -164,6 +175,45 @@ describe('Test video privacy', function () {
     }
   })
 
+  it('Should upload a non-federating unlisted video to server 2', async function () {
+    this.timeout(30000)
+
+    customConfig.federation.videos.federateUnlisted = false
+    await updateCustomConfig(servers[1].url, servers[1].accessToken, customConfig)
```
Chocobozzz, line 182: If the setting cannot be updated using the web interface, take example on https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/redundancy/redundancy-constraints.ts#L124 to update the server config

client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.ts
Tak: Is there any reason not to allow this? I added it to `production.yaml.example` as well, based on your earlier comment.

Chocobozzz: > Is there any reason not to allow this?

It subjective, but it's not a setting admins will update frequently. So I don't think it's very useful to put this in the admin interface. But you're the one who does, so you're the one who decides :)

Tak: Hm, I guess the main benefit would be discoverability.
I'm open to whatever you think - if you feel like it doesn't belong in the web interface, I'm willing to do the rework.

Chocobozzz: Then let's move this out of the web interface. Admins should read the config' file when they install peertube, and we'll mention this in the changelog. You could also mention this setting in https://docs.joinpeertube.org/#/admin-following-instances?id=being-followed-by-an-instance

Tak: Ok, done

Chocobozzz:
Thanks @Tak!

Issue #2807, opened 2020-05-29
Reduce size of PeerTube deployment


chris-morgan:
The package one builds is currently around 600MB. (That’s with `yarn install --production`; if you get the dev dependencies as well it’s almost 2GB.)

I don’t like this; it’s much larger than I feel it needs to be, or should be. It’s including quite a bit of development cruft that isn’t suitable for production environments.

I write this as someone that’s experimenting with putting PeerTube on a small server for use by me only, with intent to store actual videos (which are the *really* big thing!) remotely. I have not yet finished installing PeerTube because I got distracted with this. (This was brought on in no small part by the fact that the AUR package lacks `--production` and made me run out of disk space first time through, but I would have baulked at installing a 600MB package anyway. For that matter, I *did* baulk at a similar thing, and found and installed headless versions of ffmpeg and libav so that it wouldn’t install wayland, X and a few hundred megabytes of other useless software on my server.)

Some of the immediate low-hanging fruit I see based on the 2.2.0-rc.1 package:

- The node_modules directories: they’re development tooling used to make the dist directories where the real magic happens. It does look like a few scripts may use paths in node_modules, however, so some harm *might* occur from simply deleting them outright. I don’t know—as I say, I haven’t finished installing it yet. This would save 200MB.

- client/dist/en-US/stats-es2015.json is a 47MB file used exclusively for developers of the client (scripts/client-report.sh). Normal deployments don’t need this.

Most of the remaining weight is from languages, because the client JS is built in slight variations for each language. (There’s also other resource duplication between languages which is not trivial—something around 50MB all up, at a quick guess because calculating it truly was going to take me more than fifteen seconds.)

But I don’t believe that most deployments need or even *want* those languages: if content on my site is exclusively English, I actively *don’t want* the interface to be localised for my users. So it’d be nice to have the *ability* to either restrict the available languages, or limit it to just one language.

A crude demonstration of the effects of removing languages can be seen by deleting `client/dist/??{,-*}/` and `client/dist/locale/*.??{,-*}.*` except for en-US. (This would naturally break things if people try to use a different language.)

----

**Conclusion:** it looks to me like PeerTube should be able to be *easily* reduced from 600MB to around 200MB, and that it should hopefully not be *too* difficult to allow a single-language deployment which would be under 30MB.

I may be interested in contributing to this project. It’ll depend on how my experimental deployment goes (i.e. whether I’m actually using the software).

Related: #419 concerning Docker image size a couple of years ago.

[...]

PR #2815, opened 2020-05-31
space optimisations for packages

rigelk:
As mentioned in #2807, packages have to bundle `node_modules`, and suffer from its size which is independent from the release size.

This PR replaces sharp with jimp due to sharp being over 1/3 of the server's `node_modules` size. Jimp is a pure JS alternative, and will be a bit slower. This shouldn't be a problem considering the tasks are always offset in background tasks.

Along with this change, I make the `build` script remove the generated stats file in `client` unless a flag is passed, removing it by default.

I also added `.yarnclean`, which saves another ~20MB.


server/helpers/image-utils.ts
```diff
@@ -1,7 +1,7 @@
 import 'multer'
-import * as sharp from 'sharp'
 import { readFile, remove } from 'fs-extra'
 import { logger } from './logger'
+const Jimp = require('jimp')
```
Chocobozzz, line 4: Try `import jimp from 'jimp'` instead

.yarnclean
```diff
@@ -0,0 +1,57 @@
+# test directories
```
Chocobozzz, line 1: I don't think it's really safe to remove these files... It could lead to side effects and bugs

Chocobozzz:
It seems there is an issue with the tests

server/helpers/image-utils.ts
rigelk: It requires `esModuleInterop`, which we don't seem to support(?).

rigelk:
@Chocobozzz I removed the `.yarnclean` ; I'll try it in another PR later.

As for `jimp`, I realize it could help with some installation setups as well: https://stoneartprod.xyz/@gegeweb/104308844873645664 | I tested the image generation and it generates thumbnails just fine on my end.

Chocobozzz:
> As for jimp, I realize it could help with some installation setups as well: https://stoneartprod.xyz/@gegeweb/104308844873645664

Yes, it's a good move. But tests need to be adapted. But what about image quality/size compared to sharp?

rigelk:
@Chocobozzz the relevant tests are in the `testImage` function, right? I manually checked, and indeed the size of generated thumbnails is very different as it stands now: 21kB instead of 4.7kB with sharp. I'll need to look for some more parameters.

Chocobozzz:
> the relevant tests are in the testImage function, right?

Yes, but if jimp generates images with the same size than sharp, I think you should not have to update many of them

rigelk:
@Chocobozzz I think I pinned down the reason behind the difference in size: sharp defaults to a quality of 80%. jimp needs an extra parameter for quality, which I am testing ATM.

rigelk:
Tests validate this behaviour. With the `.quality(80)` option, `jimp` produces a thumbnail of 5.4kB.

rigelk:
@Chocobozzz:
- sharp (4.3kB):
sharp
- jimp (5.4kB):
jimp

rigelk:
@Chocobozzz I tried the single tests failing on CI (`server/tests/api/check-params/video-imports.ts` and `server/tests/api/check-params/redundancy.ts`) locally and they are working just fine.

PR #2816, opened 2020-05-31
Fix regression scrollbar bgcolor mdtextarea maximized-mode

kimsible:
This PR fixes a regression on the scrollbar background-color of the markdown textarea in maximized-mode.

It should be grey instead of white.

regression-scrollbar

### DIFF
--- a/client/src/app/shared/forms/markdown-textarea.component.scss
+++ b/client/src/app/shared/forms/markdown-textarea.component.scss
@@ -214,7 +214,7 @@ $input-border-radius: 3px;
       padding: $base-padding;
       border-right: 1px dashed $input-border-color !important;
       resize: none;
-      scrollbar-color: var(--actionButtonColor) var(--textareaBackgroundColor);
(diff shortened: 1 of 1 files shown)


kimsible:
@Chocobozzz  I think posting a PR even for small fixes and features could bring the best for PeerTube, even if we think it might extend the wait and process time. It may encourage the community to review and improve code quality. In this PR case, don't hesitate to tag me if your have any doubt on code I may have touched, or if you need a review.

Chocobozzz:
> don't hesitate to tag me if your have any doubt on code I may have touched, or if you need a review.

Sorry, I did not think this PR would have change your code, but I made a mistake. I will!

I committed myself in https://github.com/Chocobozzz/PeerTube/commit/fa11e79d4ad9211848938af490493078437bf500 because I need to put this commit in the 2.2 release (and I prefer to not cherry pick, to keep the release branch just develop instead of on another path)

Issue #2018, opened 2019-08-13
Moderation table fields are truncated in view


realityfabric:
**What happened?**

When viewing the data tables in the admin/mod panels the fields become truncated if the data doesn't fit into the field. This can be reduced or eliminated by zooming out, but this makes the text difficult to read and isn't possible on smartphones or other small-screened devices.

Long email addresses and long video titles are especially problematic in this regard.

**What do you expect to happen instead?**

I expect to be able to read the data in each field without zooming out. Shortening the fields would be fine if expanding them is possible.

**Steps to reproduce:**

1. Log into an account with admin/moderator privileges.

2. View the Administrator panel.

3. If the fields are not truncated, zoom in. They will become truncated, with no option to expand.

**Additional information**
* PeerTube version or URL: v1.3.1
* Browser name/version: Google Chrome v76.0 64bit and Firefox for Android

rigelk:
Following up on #2695, @kimsible can you reproduce this issue?

kimsible:
With the new responsive scrollable tables, there is still this effect if the texts are very long, especially on video `title` and `username`, but very less reproduced than the previous view.

It also looks like the column are not resizeable, but it cannot be a solution for touch devices.

Maybe, with a css rules `flex-wrap: wrap` we can fix it, and in any case, it needs more tests.

kimsible:
It also seems ok for expanded texts

![divers-text-moderation](https://user-images.githubusercontent.com/1877318/83365731-b5e9fd00-a3aa-11ea-9780-d13feb765937.png)

PR #2817, opened 2020-05-31
Fix: make truncated moderation table field (video title) fully visible

kimsible:
Issue: https://github.com/Chocobozzz/PeerTube/issues/2018

Before: 
title-moderation-before

After:
title-moderation

### DIFF
--- a/client/src/app/+admin/moderation/moderation.component.scss
+++ b/client/src/app/+admin/moderation/moderation.component.scss
@@ -29,11 +29,11 @@
     vertical-align: top;
     text-align: right;
   }
-  
+
   .moderation-expanded-text {
     display: inline-flex;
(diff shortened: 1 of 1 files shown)


rigelk:
I would rather have the title attribute set if it isn't already the case, and/or make sure the thumbnail doesn't change size.

kimsible:
> I would rather have the title attribute set if it isn't already the case, and/or make sure the thumbnail doesn't change size.

Thanks @rigelk, it seems it changes the thumbnail size that way :-(.

rigelk:
I replaced the title value with the video name. Though not mobile-friendly, it still improves the situation for small screens.

kimsible:
Thanks for this @rigelk ! I totally forgot that PR btw.

PR #2822, opened 2020-06-02
missing lang tag, duplicate ids

rigelk:
fixes:
- `WCAG2AA.Principle3.Guideline3_1.3_1_1.H57.2` (missing lang tag in html)
- `WCAG2AA.Principle4.Guideline4_1.4_1_1.F77` (duplicate ids in svgs)

removes extraneous `alt` value in base email.


client/src/app/shared/images/global-icon.component.ts
```diff
@@ -86,7 +88,22 @@ export class GlobalIconComponent implements OnInit {
...
 
+  private getId (id: number | string = uniqueId()): string {
+    return lowerFirst(this.constructor.name) + `_${id}-`
+  }
+
+  private getUniquelyPrefixedSVG (svgStr: String) {
+    return SVGInlineLoader.getExtractedSVG(
+      svgStr,
+      {
+        idPrefix: this.getId(),
```
Chocobozzz, line 99: Do we really need to have an id for SVGs?

client/src/app/app.component.ts
```diff
@@ -170,6 +171,16 @@ export class AppComponent implements OnInit, AfterViewInit {
       filter(pathname => !pathname || pathname === '/' || is18nPath(pathname))
     ).subscribe(() => this.redirectService.redirectToHomepage(true))
 
+    navigationEndEvent.pipe(
```
Chocobozzz, line 174: Just inject `localeId` instead https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/videos/%2Bvideo-watch/video-watch.component.ts#L109

client/src/app/shared/images/global-icon.component.ts
rigelk: Yes, and anyway cleaning up all svg in advance is going to be a major pain.

Chocobozzz: > Yes

Why?

>  cleaning up all svg in advance would be a major PITA

doing this at runtime is not the appropriate solution

rigelk: Apparently, some SVG elements will have no effect without an id. Though I have yet to find an icon we use where that occurs, I deemed safer to avoid the situation altogether.

source: https://oreillymedia.github.io/Using_SVG/guide/markup.html

> > cleaning up all svg in advance would be a major PITA
>
> doing this at runtime is not the appropriate solution

The I guess we go manual.

client/src/app/app.component.ts
rigelk: Ah! Thanks

client/src/app/shared/images/global-icon.component.ts
Chocobozzz: > Then I guess we go manual.

We can also find/create a tool to cleanup svg and generate IDs, so we don't add dependencies at runtime and avoid to do things manually

rigelk: I found https://jakearchibald.github.io/svgomg/ (based on cli utility `svgo`). It fits the bill.

Chocobozzz:
Thanks, please fix directly in develop the mascot size issue in the config warning modal (when you enable signups)

rigelk:
@Chocobozzz done in 80c7779ef.

PR #2828, opened 2020-06-04
Support player hotkeys when it is not focused

Chocobozzz:
Testing a way to use player hotkeys when it is not focused without (too much) accessibility issues.

Currently blocked by https://github.com/ctd1500/videojs-hotkeys/pull/68


Chocobozzz:
Merged in https://github.com/Chocobozzz/PeerTube/commit/7ede74add2d41958f3aad0053afc812698f261a0

PR #2830, opened 2020-06-05
`fitWidth` for `video-miniature`, fluid grid

rigelk:
This PR adds another option to `video-miniature`, `fitWidth` (default: `false`), that when set to `true` adapts the miniature so that it can scale with the available space. This is particularly useful to allow a fluid display of videos in video listings, as implemented for the `abstract-video-list` component:

- before, on large screens (`>1800px`):

- after, same screen (though on local development, with less videos to display):

- before, medium screen size (`>1200px`):

- after, medium screen size (`>1200px`):

For a preview of all breakpoints: https://peertube2.cpy.re/videos/watch/e730cdda-d559-441d-8a62-e9d3a487077b

### DIFF
--- a/client/src/app/+accounts/account-video-channels/account-video-channels.component.html
+++ b/client/src/app/+accounts/account-video-channels/account-video-channels.component.html
@@ -19,6 +19,7 @@
         <div class="no-results my-5" i18n *ngIf="getVideosOf(videoChannel).length === 0">This channel doesn't have any videos.</div>
 
         <my-video-miniature
+          class="pr-2"
           *ngFor="let video of getVideosOf(videoChannel)"
           [video]="video" [user]="user" [displayVideoActions]="true"
         ></my-video-miniature>
--- a/client/src/app/shared/video/abstract-video-list.scss
+++ b/client/src/app/shared/video/abstract-video-list.scss
@@ -1,5 +1,44 @@
+@import '_variables';
 @import '_mixins';
 @import '_miniature';
+
+.avl-item {
+  // variables
+  $gutter: 10px;
+
+  // local mixin
+  @mixin avl-miniatures($n) {
+    --nb_miniatures: #{$n};
+
+    &:nth-of-type(#{$n}n) {
+      margin-right: unset;
+    }
+  }
+
+  // .avl-item
+  @media screen and (max-width: 1000px) {
+    @include avl-miniatures(2);
+  }
+
+  @media screen and (min-width: 1001px) and (max-width: 1200px) {
+    @include avl-miniatures(3)
+  }
+
+  @media screen and (min-width: 1201px) and (max-width: 1600px) {
+    @include avl-miniatures(4);
+  }
+
+  @media screen and (min-width: 1601px) and (max-width: 1800px) {
+    @include avl-miniatures(5);
+  }
+
+  @media screen and (min-width: 1801px) {
+    @include avl-miniatures(6);
+  }
+
+  margin-right: #{$gutter};
+  width: calc(100% / var(--nb_miniatures) - (#{$gutter} * (var(--nb_miniatures) - 1) / var(--nb_miniatures)));
+}
 
 .videos-header {
   display: flex;
@@ -50,8 +89,10 @@
   }
 }
 
-.margin-content {
-  @include adapt-margin-content-width;
+:host-context(.main-col:not(.expanded)) .margin-content {
+  // TODO: use mixin variant of adapt-margin-content-width
+  margin-left: 3vw !important;
+  margin-right: 3vw !important;
 }
 
 @media screen and (max-width: $mobile-view) {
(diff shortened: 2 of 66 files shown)


client/src/app/shared/video/abstract-video-list.scss
```diff
@@ -1,6 +1,45 @@
+@import '_variables';
 @import '_mixins';
 @import '_miniature';
 
+.avl-item {
```
Chocobozzz, line 5: `available-item`

client/src/app/shared/video/abstract-video-list.scss
```diff
@@ -1,6 +1,45 @@
...
+  // variables
+  $gutter: 10px;
+
+  // local mixin
+  @mixin avl-miniatures($n) {
```
Chocobozzz, line 10: `available-miniatures`

client/src/app/shared/video/abstract-video-list.scss
```diff
@@ -1,6 +1,45 @@
...
+    --nb_miniatures: #{$n};
```
Chocobozzz, line 11: `--miniatures-count`

client/src/app/shared/video/abstract-video-list.scss
```diff
@@ -1,6 +1,45 @@
...
+  @media screen and (min-width: 1601px) and (max-width: 1800px) {
+    @include avl-miniatures(5);
+  }
+
+  @media screen and (min-width: 1801px) {
+    @include avl-miniatures(6);
+  }
+
+  margin-right: #{$gutter};
+  width: calc(100% / var(--nb_miniatures) - (#{$gutter} * (var(--nb_miniatures) - 1) / var(--nb_miniatures)));
```
Chocobozzz, line 40: I think you should set a max width, because thumbnails are quite small

client/src/app/shared/video/video.model.ts
```diff
@@ -91,7 +91,7 @@ export class Video implements VideoServerModel {
     this.description = hash.description
 
     this.duration = hash.duration
-    this.durationLabel = durationToString(hash.duration)
+    this.durationLabel = durationToString(hash.duration).replace(/^0/, '')
```
Chocobozzz, line 94: Update directly `durationToString` function

client/src/app/shared/video/abstract-video-list.scss
rigelk: I ended up switching to a grid-based display, which is more maintainable and guarantees a max-size based on the minimum size specified.

rigelk:
@Chocobozzz since the grid display is set to `auto-fit`, it adds columns for as much as there is space for the minimum space required for another one. That also means we are independant from media queries for the most part, and can add elements left or right of the video display without affecting the video list display :slightly_smiling_face:

client/src/sass/include/_variables.scss
```diff
@@ -116,7 +116,7 @@ $variables: (
   --embedBigPlayBackgroundColor: var(--embedBigPlayBackgroundColor)
 );
 
-@function var($variable) {
+@function pvar($variable) {
```
Chocobozzz, line 119: I'm not sure to understand this change.

client/src/app/+accounts/account-video-channels/account-video-channels.component.html
```diff
@@ -19,6 +19,7 @@
         <div class="no-results my-5" i18n *ngIf="getVideosOf(videoChannel).length === 0">This channel doesn't have any videos.</div>
 
         <my-video-miniature
+          class="pr-2"
```
Chocobozzz, line 22: Do we want right padding on mobile?

client/src/app/shared/video/video-miniature.component.scss
```diff
@@ -109,6 +106,34 @@ $more-margin-right: 15px;
...
+
+    .video-bottom {
+      width: 100% !important;
+
+      .video-miniature-information {
+        width: calc(100% - 40px) !important;
+      }
+    }
+
+    my-video-thumbnail {
```
Chocobozzz, line 122: Could you explain this change? I'm not sure to understand why you need absolute position

client/src/sass/include/_variables.scss
rigelk: css variables are also called with 'var()' in pure css.

client/src/app/+accounts/account-video-channels/account-video-channels.component.html
rigelk: It doesn't have any impact on mobile from what I see.

client/src/app/shared/video/video-miniature.component.scss
rigelk: Yes, it is the same change I used on `.screenratio`, it provides 16:9 ratio without relying on width *and* height values being preset.

PR #2844, opened 2020-06-08
add aria-hidden to non-descriptive icons

Pandoraaa:
This PR will add aria-hidden attribute to all non descriptive icons in order to improve a11y (screen readers don't have to read icons). 

For example these icons in front of the text in the left panel menu : 

image


rigelk:
Thanks @Pandoraaa !

PR #2852, opened 2020-06-09
WIP: Add global search support

Chocobozzz:
Search index code: https://framagit.org/framasoft/peertube/search-index

Opt in. 

An admin can:
 * enable search index
 * specify index url
 * disable local search (to only enable search on the index)
 * set the search index as default (local search is selected by default on the web interface)

When coming from the search index, results `Video` have an `url: string`, `thumbnailUrl: string`, `previewUrl: string` and `embedUrl: string`  properties. Avatars also have an `url: string` property.

screen_2020-06-09-14:47:04
screen_2020-06-09-14:46:45
screen_2020-06-09-14:47:24

We display the results, and then there are 2 possibilities:
  * Users cannot search remote URI :arrow_right: they are redirected on the origin instance on video/channel click
  * Users can search remote URI :arrow_right: the client first fetch the remote object using a classic URI search (already available in PeerTube) and then redirect the user on the new created object (video or chanel)

Dropdown is disabled on search index results.


buoyantair:
This is just AWESOME!

PR #2866, opened 2020-06-11
add scope tags and aria-labels to instance table

Pandoraaa:
This PR will improve a11y with a table caption, table structure with scope to help screen readers and aria-label for its yes/no icons.

### DIFF
--- a/client/src/app/shared/instance/instance-features-table.component.scss
+++ b/client/src/app/shared/instance/instance-features-table.component.scss
@@ -14,6 +14,7 @@ table {
     }
 
     &.sub-label {
+      font-weight: $font-regular;
       padding-left: 30px;
     }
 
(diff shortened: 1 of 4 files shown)


client/src/app/shared/instance/instance-features-table.component.scss
```diff
@@ -27,6 +28,13 @@ table {
   td {
     vertical-align: middle;
   }
+
+  caption {
+    caption-side: top;
+    font-size: 16px;
+    font-weight: $font-semibold;
```
rigelk, line 35:
```suggestion
    font-size: 15px;
    font-weight: $font-bold;
```

rigelk: This is to provide the same style as `label` elements, but you might find it too bold - which do you prefer?

Pandoraaa: I find the $font-bold too bold 😄 
I'll change the font-size :)

rigelk:
Thanks again @Pandoraaa :slightly_smiling_face:

Issue #2826, opened 2020-06-04
Error logs «unknown infoHash»


JohnXLivingston:
**What happened?**
I just updated my instance to 2.2.0, and my logs are full of these:

`Jun 04 08:24:47 yiny peertube[25184]: [www.yiny.org:443] 2020-06-04 08:24:47.180 warn: Warning in tracker. {
Jun 04 08:24:47 yiny peertube[25184]:   "err": {
Jun 04 08:24:47 yiny peertube[25184]:     "stack": "Error: Unknown infoHash de3dfb71a2f8fcdca29b9e0c9b5e1df2fc1c7dac\n    at Server.<anonymous> (/var/www/peertube/versions/peertube-v2.2.0/dist/server/controllers/tracker.js:60:27)\n    at Generator.next (<anonymous>)\n    at fulfilled (/var/www/peertube/versions/peertube-v2.2.0/dist/server/controllers/tracker.js:5:58)",
Jun 04 08:24:47 yiny peertube[25184]:     "message": "Unknown infoHash de3dfb71a2f8fcdca29b9e0c9b5e1df2fc1c7dac"
Jun 04 08:24:47 yiny peertube[25184]:   }
Jun 04 08:24:47 yiny peertube[25184]: }
Jun 04 08:24:47 yiny peertube[25184]: [www.yiny.org:443] 2020-06-04 08:24:47.337 warn: Warning in tracker. {
Jun 04 08:24:47 yiny peertube[25184]:   "err": {
Jun 04 08:24:47 yiny peertube[25184]:     "stack": "Error: Unknown infoHash cb6343a4e7c3c14aeea8f5b9ed9f69218e53b4ff\n    at Server.<anonymous> (/var/www/peertube/versions/peertube-v2.2.0/dist/server/controllers/tracker.js:60:27)\n    at Generator.next (<anonymous>)\n    at fulfilled (/var/www/peertube/versions/peertube-v2.2.0/dist/server/controllers/tracker.js:5:58)",
Jun 04 08:24:47 yiny peertube[25184]:     "message": "Unknown infoHash cb6343a4e7c3c14aeea8f5b9ed9f69218e53b4ff"
Jun 04 08:24:47 yiny peertube[25184]:   }
Jun 04 08:24:47 yiny peertube[25184]: }
Jun 04 08:24:48 yiny peertube[25184]: [www.yiny.org:443] 2020-06-04 08:24:48.164 warn: Warning in tracker. {
[...]

PR #2870, opened 2020-06-12
Logging ip requesting unknown infoHash

JohnXLivingston:
As I mentioned here https://github.com/Chocobozzz/PeerTube/issues/2826 , my instance is facing a kind of DDOS: hundreds of IPs are asking for unkwnown infoHash. I had to increase nginx worker_connections, as there are thousands of connections that timeout only after 30 minutes.
Moreover, the download traffic generates an upload traffic 4 times higher (so it can be very easy to DDOS the bandwidth).
And last, but not least, it generates GB of log files. 

So, in this PR, I added the ip for clients asking for unknownInfoshash, so we can add a rule in fail2ban.

Note: the IP is logged without taking into account CONFIG.LOG.ANONYMIZE_IP. The IP was already logged on this line without taking into account the config:
 https://github.com/Chocobozzz/PeerTube/blob/435258ea3c17c10c7c2735075e59122956d45fca/server/controllers/tracker.ts#L44 
I think it is a bug. Would you like me to fix it?

### DIFF
--- a/server/controllers/tracker.ts
+++ b/server/controllers/tracker.ts
@@ -53,7 +53,7 @@ const trackerServer = new TrackerServer({
       const playlistExists = await VideoStreamingPlaylistModel.doesInfohashExist(infoHash)
       if (playlistExists === true) return cb()
 
-      return cb(new Error(`Unknown infoHash ${infoHash}`))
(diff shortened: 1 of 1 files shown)


Chocobozzz:
> I think it is a bug. Would you like me to fix it?

I'm not sure, or people will be able to easily bypass your fail2ban rule :thinking:

PR #2871, opened 2020-06-12
fix headings order or add missing ones

Pandoraaa:
This PR will improve a11y with adding correct or missing headings on each page.
Screen readers need valid title hierarchy so the user can jump on the correct section to read. 
I also fix some css to keep the same style but with the correct heading :)

### DIFF
--- a/client/src/app/+accounts/account-video-channels/account-video-channels.component.html
+++ b/client/src/app/+accounts/account-video-channels/account-video-channels.component.html
@@ -1,3 +1,4 @@
+<h1 class="sr-only" i18n>Video channels</h1>
 <div class="margin-content">
 
   <div class="no-results" i18n *ngIf="channelPagination.totalItems === 0">This account does not have channels.</div>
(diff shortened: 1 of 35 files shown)


client/src/app/+accounts/account-video-channels/account-video-channels.component.html
```diff
@@ -8,7 +9,7 @@
         <a [routerLink]="getVideoChannelLink(videoChannel)" i18n-title title="See this video channel">
           <img [src]="videoChannel.avatarUrl" alt="Avatar" />
 
-          <div>{{ videoChannel.displayName }}</div>
+          <h2 class="section-title">{{ videoChannel.displayName }}</h2>
```
rigelk, line 12: In that instance, `section-title` is adding a `border-top`, which should only be applied to the parent container.

rigelk: I fixed it in [https://github.com/Chocobozzz/PeerTube/pull/2871/commits/63c702b065e2b3bca981a3cc51035c285bc74e0a ](https://github.com/Chocobozzz/PeerTube/pull/2871/commits/63c702b065e2b3bca981a3cc51035c285bc74e0a).

rigelk:
Thanks a bunch @Pandoraaa!

PR #2873, opened 2020-06-14
cleanup extraneous timestamps in model definitions

rigelk:
per [the Sequelize documentation relative to timestamps](https://sequelize.org/master/manual/model-basics.html#timestamps), `createdAt` and `updatedAt` are automatically created on models unless explicitly disabled, which we already do.


Chocobozzz:
I'd rather keep them. IMHO it's better for people, so they can easily see there are `createdAt` and `updatedAt` fields in the model even if they don't know sequelize

PR #2875, opened 2020-06-14
factorize account/server blocklists for users and instance

rigelk:
Improves maintainability by making reusable components of the account/server mute views, so that we don't have different components for users and instance moderation.

Also, since the previous tables for users didn't have filtering nor free-form domain blocks, this will add features for them.


client/src/app/+my-account/my-account-blocklist/my-account-blocklist.component.ts
```diff
@@ -1,59 +1,15 @@
...
-import { SortMeta } from 'primeng/api'
-import { AccountBlock, BlocklistService } from '@app/shared/blocklist'
+import { Component } from '@angular/core'
+import { GenericAccountBlocklistComponent, BlocklistComponentType } from '@app/shared/blocklist'
 
 @Component({
   selector: 'my-account-blocklist',
-  styleUrls: [ './my-account-blocklist.component.scss' ],
-  templateUrl: './my-account-blocklist.component.html'
+  styleUrls: [ '../../+admin/moderation/moderation.component.scss', '../../shared/blocklist/account-blocklist.component.scss' ],
```
Chocobozzz, line 6: You should not import a file that is in a lazy loaded route.

client/src/app/shared/blocklist/server-blocklist.component.ts
```diff
@@ -0,0 +1,100 @@
+import { OnInit, ViewChild } from '@angular/core'
+import { Notifier } from '@app/core'
+import { I18n } from '@ngx-translate/i18n-polyfill'
+import { RestPagination, RestTable } from '@app/shared/rest'
+import { SortMeta } from 'primeng/api'
+import { BlocklistService, BlocklistComponentType } from './blocklist.service'
+import { ServerBlock } from '../../../../../shared/models/blocklist/server-block.model'
+import { BatchDomainsModalComponent } from '@app/+admin/config/shared/batch-domains-modal.component'
+
+export class GenericServerBlocklistComponent extends RestTable implements OnInit {
```
Chocobozzz, line 10: Why is it not an abstract class?

rigelk: Abstract classes mandate to implement all functions of the classes they extend, in that case `RestTable`.

PR #2879, opened 2020-06-15
Remove uneccessary details to link titles

Pandoraaa:
This PR will improve a11y with short link title when needed.
Link title should not have verb like "go to" for screen readers.

More improving should be done with all the target="_blank" attribute on another PR with either (new window) in title/text or adding a new window icon with css :)


Chocobozzz:
Thanks! Is there a reason you don't put an uppercase at the beginning of the title? (*plugin homepage (new window)* VS *Plugin homepage (new window)*)

Pandoraaa:
Hello ! No I was searching for "Go to" and delete it. I can add uppercase :)

Chocobozzz:
Thanks @Pandoraaa

Issue #2752, opened 2020-05-13
[Plugin/hook] to customize appearance (logo, favicon, PWA manifest name and icons)


kimsible:
I'm thinking about a plugin to customize appearance with the new helper plugin `getRouter`.
This plugin would add in `/admin/config/edit-custom`, _advanced config_ tab the ability to :
 
- upload a png or svg logo 
- customize the main color with a color picker

This plugin would also :
 
- process and use the favicon and the PWAs manifest icons with the uploaded logo.
- update PWAs manifest name and description according to the instance settings

For the last feature, I think a hook would be required to trigguer any change of the name and description instance.

kimsible:
This is a first step https://github.com/Chocobozzz/PeerTube/pull/2897.

According to the discussion on this PR, I propose these next steps :

2. Add to the admin panel the ability to upload PNG favicon and SVG logo  (without any treatment, only format / extension checking)

3. Add a script tool to generate PWA icons from SVG logo.

4. Add plugin hook when uploading SVG logo and PNG favicon

5. Add image edition lib to plugin helpers

6. Add a plugin hook in the dymamic manifest :  https://github.com/Chocobozzz/PeerTube/blob/176000efee51b0a8d91e8dc34e5d7b3c87a5c1df/server/controllers/client.ts#L160

7. Improve the router plugin : restrict writing files to an isolated directory

8. Create a plugin or update this one https://github.com/kimsible/peertube-plugin-webapp-manifest for the PWA manifest edition and icons generation.

@Chocobozzz @rigelk all these steps seem reachable to you ?

Some steps contain improvements for the plugin API, it's well in the roadmap ?

Chocobozzz:
>  all these steps seem reachable to you ?

Yes I think

> Some steps contain improvements for the plugin API, it's well in the roadmap ?

No, it's not planned in the roadmap but we can try to implement them if we have some time.

kontrollanten:
Great idea! But do we need a web UI to customize the manifest? In most cases it's just something that you configure once, therefore I'd suggest to handle this during the client build time. It should be easy to solve with webpack, I think [webpack-pwa-manifest-plugin](https://www.npmjs.com/package/webpack-pwa-manifest) solves our needs? That should be possible to use together with [angular-builders](https://www.npmjs.com/package/@angular-builders/custom-webpack).

rigelk:
@kontrollanten most servers are not (re)building their code.

kontrollanten:
[...]

PR #2897, opened 2020-06-24
Add ability to override client assets : logo - favicon - PWA icons - PWA manifest name and description

kimsible:
This PR allows to customize main assets like logo, favicon and PWA icons, PWA manifest name and description :

- [x] Add client-overrides directory storage to config files ; 
- [x] Serve overrides files (logo, favicon, PWA icons) from client-overrides dir with dist dir as fallback ;
- [x] Move logo `background-image` from CSS bundle to `<style>` tag header in `index.html` ;
- [x] Handle cache for all overrides files with a querystring `?[fileContentHash]`  ;
- [x] Make the PWA manifest a dynamic JSON to update `description`, `name`, `short_name` with instance config ;
~Add ability in the admin panel to upload a SVG / PNG logo and generate favicon - PWA icons.~

To test it the client must be built with `npm run clean:client` and `npm run dev:server`

### DIFF
--- a/server/controllers/client.ts
+++ b/server/controllers/client.ts
@@ -1,3 +1,4 @@
+import { constants, promises as fs } from 'fs'
 import * as express from 'express'
 import { join } from 'path'
 import { root } from '../helpers/core-utils'
@@ -39,18 +40,36 @@ clientsRouter.use(
 )
 
 // Static HTML/CSS/JS client files
-
 const staticClientFiles = [
   'manifest.webmanifest',
   'ngsw-worker.js',
   'ngsw.json'
 ]
+
 for (const staticClientFile of staticClientFiles) {
   const path = join(root(), 'client', 'dist', staticClientFile)
 
-  clientsRouter.get('/' + staticClientFile, (req: express.Request, res: express.Response) => {
+  clientsRouter.get(`/${staticClientFile}`, (req: express.Request, res: express.Response) => {
     res.sendFile(path, { maxAge: STATIC_MAX_AGE.SERVER })
   })
+}
+
+// Static client overrides
+const staticClientOverrides = [
+  'assets/images/logo.svg',
+  'assets/images/favicon.png',
+  'assets/images/icons/icon-36x36.png',
+  'assets/images/icons/icon-48x48.png',
+  'assets/images/icons/icon-72x72.png',
+  'assets/images/icons/icon-96x96.png',
+  'assets/images/icons/icon-144x144.png',
+  'assets/images/icons/icon-192x192.png',
+  'assets/images/icons/icon-512x512.png'
+]
+
+for (const staticClientOverride of staticClientOverrides) {
+  const overridePhysicalPath = join(CONFIG.STORAGE.CLIENT_OVERRIDES_DIR, staticClientOverride)
+  clientsRouter.use(`/client/${staticClientOverride}`, asyncMiddleware(serveClientOverride(overridePhysicalPath)))
 }
 
 clientsRouter.use('/client/locales/:locale/:file.json', serveServerTranslations)
@@ -129,4 +148,17 @@ function sendHTML (html: string, res: express.Response) {
   res.set('Content-Type', 'text/html; charset=UTF-8')
 
   return res.send(html)
+}
+
+function serveClientOverride (path: string) {
+  return async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    try {
+      await fs.access(path, constants.F_OK)
+      // Serve override client
+      res.sendFile(path, { maxAge: STATIC_MAX_AGE.SERVER })
+    } catch {
+      // Serve dist client
+      next()
+    }
+  }
 }
--- a/server/controllers/client.ts
+++ b/server/controllers/client.ts
@@ -41,7 +41,6 @@ clientsRouter.use(
 
 // Static HTML/CSS/JS client files
 const staticClientFiles = [
-  'manifest.webmanifest',
   'ngsw-worker.js',
   'ngsw.json'
 ]
@@ -53,6 +52,9 @@ for (const staticClientFile of staticClientFiles) {
     res.sendFile(path, { maxAge: STATIC_MAX_AGE.SERVER })
   })
 }
+
+// Dynamic PWA manifest
+clientsRouter.get('/manifest.webmanifest', asyncMiddleware(generateManifest))
 
 // Static client overrides
 const staticClientOverrides = [
@@ -148,6 +150,18 @@ function sendHTML (html: string, res: express.Response) {
   res.set('Content-Type', 'text/html; charset=UTF-8')
 
   return res.send(html)
+}
+
+async function generateManifest (req: express.Request, res: express.Response) {
+  const manifestPhysicalPath = join(root(), 'client', 'dist', 'manifest.webmanifest')
+  const manifestJson = await fs.readFile(manifestPhysicalPath, 'utf8')
+  const manifest = JSON.parse(manifestJson)
+
+  manifest.name = CONFIG.INSTANCE.NAME
+  manifest.short_name = CONFIG.INSTANCE.NAME
+  manifest.description = CONFIG.INSTANCE.SHORT_DESCRIPTION
+
+  res.json(manifest)
 }
 
 function serveClientOverride (path: string) {
--- a/client/src/app/app.component.scss
+++ b/client/src/app/app.component.scss
@@ -62,7 +62,7 @@
 
       .icon.icon-logo {
         display: inline-block;
-        background: url('../assets/images/logo.svg') no-repeat;
+        background-repeat: no-repeat;
         width: 23px;
         height: 24px;
         margin-right: .5rem;
--- a/client/src/index.html
+++ b/client/src/index.html
@@ -7,14 +7,14 @@
     <meta name="theme-color" content="#fff" />
     <meta property="og:platform" content="PeerTube" />
     <!-- Web Manifest file -->
-    <link rel="manifest" href="/manifest.webmanifest">
+    <link rel="manifest" href="/manifest.webmanifest?[manifestContentHash]">
 
-    <link rel="icon" type="image/png" href="/client/assets/images/favicon.png" />
+    <link rel="icon" type="image/png" href="/client/assets/images/favicon.png?[faviconContentHash]" />
 
     <!-- logo background-image -->
     <style type="text/css">
       .icon-logo {
-        background-image: url(/client/assets/images/logo.svg);
+        background-image: url(/client/assets/images/logo.svg?[logoContentHash]);
       }
     </style>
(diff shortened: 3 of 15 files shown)


kimsible:
@rigelk @Chocobozzz I can also split this PR into several, like 3 more, one for the manifest update, one for the cache, one for the front admin panel.

Do you have an idea where `server/controllers/client.ts` routes are override in develop mode ?

client/src/app/app.component.html
```diff
@@ -9,7 +9,9 @@
       <span class="icon icon-menu" (click)="menu.toggleMenu()"></span>
 
       <a class="peertube-title" [routerLink]="defaultRoute">
-        <span class="icon icon-logo"></span>
+        <object class="icon icon-logo" type="image/svg+xml" data="/client/assets/images/logo.svg">
```
rigelk, line 12: why the change to `object`? You don't motivate it.

server/controllers/client.ts
```diff
@@ -39,20 +40,47 @@ clientsRouter.use(
...
+  'icons/icon-72x72.png',
+  'icons/icon-96x96.png',
+  'icons/icon-144x144.png',
+  'icons/icon-192x192.png',
+  'icons/icon-512x512.png'
+]
+
+for (const staticClientImage of staticClientImages) {
+  let imagePhysicalPath = join(CONFIG.STORAGE.ASSETS_DIR, 'images', staticClientImage)
+  clientsRouter.use(`/client/assets/images/${staticClientImage}`, asyncMiddleware(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
```
rigelk, line 72: please split the use and definition of the middleware.

rigelk: why not use `express.static` as well here?

rigelk:
@kimsible unless you need tests running, please open PRs as WIP.

rigelk:
> Add ability in the admin panel to upload a SVG / PNG logo and generate favicon - PWA icons.

A lot of work for something that is trivial to do manually, especially since it is usually done one. I would much rather see a script in the tools, if any.

client/src/app/app.component.html
kimsible: From CSS background-url, Webpack add a unique hash in the file name like `logo.6Z9DD98KJjk9DJ98.svg`, so right no you need to override CSS to simply customize the logo. This way we'll be able to add a query-string to handle cache in data property of the object tag and store it in the storage directory. Having a separate storage for logo and favicon allows to avoid overwriting when updating PeerTube.

server/controllers/client.ts
kimsible: I thought it was better to only serve file we need, do you think we should be able to serve all the assets directory ?

kimsible:
> > Add ability in the admin panel to upload a SVG / PNG logo and generate favicon - PWA icons.
> 
> A lot of work for something that is trivial to do manually, especially since it is usually done once. I would much rather see a script in the tools, at least to begin with.

Yes, you're right, it's really asked we should keep it in mind for further PeerTube versions, changing the logo is a common habit when it comes to animate a community for a special event. And for PWAs icons, it's now in the PeerTube Core, generating them is very useful when all the tools are included to do it automatically but in my opinion (I may be wrong), the webmanifest and icons should have been a plugin.

The first big issue is the ability to update logo and favicon without doing tricky code with shell or a plugin for each restart or update of PeerTube. Really, there is not so much PeerTube instances with a custom logo for all these reasons, developing a theme is not reachable by anybody.

So, I delete this step =)

Do you think we have to create another PR for the manifest or use this one ?

kimsible:
> @kimsible unless you need tests running, please open WIP PRs as WIP.

Ok, you meant as a draft ?

rigelk:
Yes, as draft, sorry ^^'

kimsible:
> Yes, as draft, sorry ^^'

Sorry, some things need to be discussed.

server/controllers/client.ts
rigelk: If the assets directory only has the files we need to serve, then yes :slightly_smiling_face: 

It has the advantage to be pretty rich when it comes to built-ins: https://expressjs.com/en/4x/api.html#express.static

kimsible: > The assets directory only has the files we need to serve, so yes

There are all assets of the theme, there is more than we want to serve, no ?

Or you means for the `icons` ? I think favicon and logo should serve alone.

Chocobozzz:
Having another storage directory that implicitly override client assets is not something I think I like.

Another proposal:
 * Remove the manifest file from client and serve the JSON dynamically from the server (so you correctly use the peertube name/description etc from the config, and can add plugins hooks to change the icon paths)
 * Create a plugin that allows admins to easily upload images from the web interface. The plugin just overrides logos/favicon using hooks and/or css rules

Not sure if it's a better solution than the one implemented in this PR. I prefer the plugin solution, because we don't add implicit things (`assets` directory), do not add breaking changes (icon displayed using CSS) and use an existing system (plugin API, even if there are some hooks/settings buttons that are still missing). What do you think?

kimsible:
Hi @Chocobozzz , thanks for your participation. 

> Having another storage directory that implicitly override client assets is not something I think I like.

> Create a plugin that allows admins to easily upload images from the web interface. The plugin just overrides logos/favicon using hooks and/or css rules

I proposed to only override `logo.svg`, `favicon.png` and `PWA icons` in the storage assets directory.
It's not all the assets directory, if it was, yes, it could break the main theme and the instance could not benefit of the theme updates. I'm not against from a plugin but it would need some new important features.

> Remove the manifest file from client and serve the JSON dynamically from the server (so you correctly use the peertube name/description etc from the config, and can add plugins hooks to change the icon paths)

**Serving the JSON manifest dynamically is a very good idea and seems implementable with not so much efforts**, but I partially agree for the others assets :

The problem with the plugin is that the system does not allow to simply upload files for the moment, (I actually add some input file in the manifest plugin https://github.com/kimsible/peertube-plugin-webapp-manifest, and it's a real pain to to hook the client, plus sending files via plugins routes is not really secure, **the router plugin should be able to upload only in one directory isolated from the app and available only for admins**).

What we need new with plugins : 

- A real upload system with **an isolated directory, a cache naming, and an upload input file with the preview**
- Hooks to override the express route of these assets (logo, favicon and PWA icons) and the manifest icons fields
- Add the image tool as a helper (Jimp ?)

> do not add breaking changes (icon displayed using CSS) 

You mean the logo ?
In any case the logo should be displayed as a svg object, it will allow to handle cache naming with query string.

I see another problem with the plugin API, edit the logo and the favicon are really one of the most basic feature we could expect from a self-hosted webapp, the upload input file should be in the admin panel in the config Appearance part.

So, yes this solution is interesting but force us to implement / extend the plugin API support but put a basic feature into a plugin that should be in the PeerTube Core (I mean for the favicon and the logo, I don't think the manifest is a standard outside android ?).

I'm really not against this solution but **the plugin API should have all the features I mentioned and the plugin itself should be an official plugin**.

What do you think?

kimsible:
@Chocobozzz I have an (intermediate) solution : 

- Put `logo.svg` and `favicon.png` into the root of `storage/` and PWAs icons in `storage/pwa` without creating a constant config ;
- Keep the actual router fallback to `./client/dist/assets` and make it work for dev mode ;
- As @rigelk said, we could create a script in the tools to generate icons and favicon according to `./storage/logo.svg` and update cache namnig with querystrings ;
- Make the manifest dynamic ;
- Add some docs.

Chocobozzz:
Okay then. I prefer your first solution than the intermediate one, but with a more explicit name. Something like:

```
  cache: 'storage/cache/'
  plugins: 'storage/plugins/'
  # Could contain for example assets/images/favicon.png
  # If the file exists, peertube will serve it
  # If not, peertube will fallback to the default file
  client-overrides: 'storage/client-overrides/'
```

I think [your check](https://github.com/Chocobozzz/PeerTube/pull/2897/files#diff-962521466dde518282685b02da8fd82eR58) is still important, for security reasons.

> In any case the logo should be displayed as a svg object, it will allow to handle cache naming with query string.

How do you will handle cache naming?

> Do you have an idea where server/controllers/client.ts routes are override in develop mode ?

Try to update the proxy file: https://github.com/Chocobozzz/PeerTube/blob/develop/client/proxy.config.json

kimsible:
Thanks `client-overrides` with comments are more explicit =). 
But should we specify which ones are overridable ? (with a link to the documentation for example).

> How do you plan to handle cache naming?
A general global hash we can add as a querystring to the image url.
Maybe restarting PeerTube could update this hash ?

Thanks for the tips too about the proxy !

Chocobozzz:
> But should we specify which ones are overridable ? (with a link to the documentation for example).

I think a comment would be enough :+1: 

>A general global hash we can add as a querystring to the image url.

But how do you update the client to inject this querystring/hash to the logo URL? Currently, webpack does the job: updating the filename and the client HTML. If peertube automatically updates the filename on admin upload or peertube restart, we still miss the step where we inject this new logo name in the client no?

kimsible:
> But how do you update the client to inject this querystring/hash to the logo URL? Currently, webpack does the job: updating the filename and the client HTML. 

Some assets need to be updated during `runtime` and not `buildtime`, like our logo or favicon. 

> If peertube automatically updates the filename on admin upload or peertube restart, we still miss the step where we inject this new logo name in the client no?

I'm not sure I understand well your question, maybe, we could fix a naming convention and extension for the assets to override them? (btw overriding always follows the same naming convention ?)

kimsible:
Oups @Chocobozzz I forgot that the `index.html` containing favicon and manifest link was not a template, so we can't handle cache without a true templating with webpack for the favicon.

kimsible:
I didn't find how to cancel the route overriding of webpack dev server with the proxy file.

So the simplest way to test is with the client bundles `npm run clean:client` and `npm run dev:server`

Chocobozzz:
@rigelk did you review this PR?

rigelk:
@chocobozzz no, since I tought you didn't agree with the design.

JohnXLivingston:
Is there any documentation (other than the comment in production.yaml file) for this PR?

JohnXLivingston:
> Should logo.svg and favicon.png be in /var/www/peertube/storage/client-overrides/ or /var/www/peertube/storage/client-overrides/assests/images/ ?

I tried both, and it is not working.

kimsible:
@JohnXLivingston you need to restart PeerTube to update the client cache. This is an experimental feature that will be extend : https://github.com/Chocobozzz/PeerTube/issues/2752#issuecomment-649851934

JohnXLivingston:
Hum, I can't make it work... Don't know why.

kimsible:
@JohnXLivingston do you use nginx ? docker ?

JohnXLivingston:
> @JohnXLivingston do you use nginx ? docker ?

nginx

PR #2898, opened 2020-06-24
Add database.name config option

### LINKED ISSUE #1620: Make whole database name configurable
**What happened?**

Database names are required to start with `peertube`

The ability to freely chose the database name in full. In our case `"peertube"+database.suffix` https://github.com/Chocobozzz/PeerTube/blob/develop/config/production.yaml.example#L17 is not compatible with our usual naming scheme.

**Steps to reproduce:**

1. Follow the current installation instructions.

**Additional information**
* PeerTube version or URL: https://github.com/Chocobozzz/PeerTube/blob/develop/config/production.yaml.example#L17

Thank you!

gramakri:
If database.name is set, we use it as the complete database name.
If unset, we use the "peertube"+database.suffix as the complete database
name.

Fixes #1620

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -33,7 +33,7 @@ rates_limit:
 trust_proxy:
   - 'loopback'
 
-# Your database name will be "peertube"+database.suffix
+# Your database name will be database.name OR "peertube"+database.suffix
 database:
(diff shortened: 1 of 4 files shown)


thegcat:
Thank you!

PR #2904, opened 2020-06-26
Update CentOS insutructions in dependencies.md

cgarwood82:
Created section for CentOS 8 instructions as it deviates from CentOS 7

- There is no python binary that can be installed with a package so it must be symlinked
- Removed packages not needed from step 4
- Removed step concerning g++ as the version in CentOS 8 is usable
- Improved wording for Steps summary in both CentOS 8 and 7


Chocobozzz:
Thanks @cgarwood82

PR #2919, opened 2020-06-30
special display case for logged-out users to display download in watch page

rigelk:


Chocobozzz:
Put this logic in video watch component instead, because we only want this behaviour in this specific page

PR #2924, opened 2020-07-02
simplify navigation within most admin menus

rigelk:
More space and less headaches :)


client/src/app/shared/shared-main/shared-main.module.ts
```diff
@@ -81,10 +81,9 @@ import { AUTH_INTERCEPTOR_PROVIDER } from './auth'
 
     HelpComponent,
     ListOverflowComponent,
+    TopMenuDropdownComponent,
 
     UserNotificationsComponent,
-
-    FeedComponent
```
Chocobozzz, line 87: ?

rigelk: It is already present a few lines above.

Issue #2931, opened 2020-07-04
Create custom infobox in signup page


ghost:
Add a UI option to create custom infoboxes with custom title and description which will appear on the signup page and the About page.

This is what I call an infobox
![qwdqw](https://user-images.githubusercontent.com/65374098/86509011-69e90880-bded-11ea-89a8-b4603c4d769d.png)

Chocobozzz:
Implemented as a plugin hook in https://github.com/Chocobozzz/PeerTube/commit/071f3e519cbd3184b59ee728fe96c5c29b7792b9

PR #2933, opened 2020-07-04
remove Code of Conduct, add DMCA instead

k09i71:
Having both the CoC and the Terms seems redundant. CoCs are also often seen as political documents. So instead of just removing one [infobox](https://github.com/Chocobozzz/PeerTube/issues/2931) (CoC), I've decided to rename it to DMCA. If someone will want to have an infobox for the CoC, he could create one (once this https://github.com/Chocobozzz/PeerTube/issues/2931 issue will get solved).

### DIFF
--- a/client/src/app/+about/about-instance/about-instance.component.html
+++ b/client/src/app/+about/about-instance/about-instance.component.html
@@ -67,7 +67,7 @@
       <div [innerHTML]="html.moderationInformation"></div>
     </div>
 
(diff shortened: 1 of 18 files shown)


rigelk:
DMCAs are part of the US Copyright law, which not all instances might need to comply with. A DMCA infobox is not enough to comply with DMCA either, nor very helpful at that. #2931 will help in a more elegant way.

Issue #2908, opened 2020-06-28
Searching "In this instance" incorrectly implies a search of only local videos


12people:
**What happened?**
When typing a search term, "in this instance" is shown, implying that the results will be local to the current instance. However, the search terms are from multiple instances.

**What do you expect to happen instead?**
As results are collected from all instances, I would expect the search label to say "from all instances" rather than "in this instance".

**Steps to reproduce:**

Let's take the Peertube instance [P2Tube](https://tube.p2p.legal) (with 276 local videos) and the search term "brad sucks" (the name of a band) as an example.

1. Type "brad sucks" in the search field on P2Tube.
2. "In this instance" appears as the autosuggested option below. Confusingly, it's not clickable, but it at least implies that hitting Enter will search only in the current instance.
![image](https://user-images.githubusercontent.com/7361228/85945060-74ae2400-b93b-11ea-9e97-a85c7f3d07ac.png)
3. The results shown are from other instances:
![image](https://user-images.githubusercontent.com/7361228/85945090-afb05780-b93b-11ea-802d-7f84b18a1d62.png)

**Additional information**
* PeerTube version: 2.2.0
* PeerTube URL: https://tube.p2p.legal
* Browser name/version: Firefox 77

rigelk:
@12people in my tests, the option *is* clickable :/

As for your suggestion, it carries its own set of implications : "from all instances" could lead the user to think results are exhaustive in the vidiverse, which is not the case. Only instance followed by the current instance are indexed. I would suggest a "in this instance's network" instead, whereas the global search is "in the vidiverse" with a hint explaining the feature, as per #2852. WDYT?

12people:
Huh, that's strange — I've tried several instances and it's not clickable on any of them. I'm curious if we're talking about the same thing – here's what I'm doing, tested just now on [the TILVids instance](https://tilvids.com/): https://jmp.sh/8scBeKk

I agree with the suggestion — "in this instance's network" sounds good (and perhaps a hint text hidden under an (i) icon might be appropriate here as well, to explain what an instance's network is). :)

rigelk:
@12people the unclickability seems fixed in develop - probably with the refactoring done in beetween 2.2 and 2.3 to introduce the global search. You can try it on https://peertube2.cpy.re.

I also changed last week the wording.

k09i71:
- "in this instance's network" - is too long
- "in the vidiverse" - it is usually not a good practice to make up new words

[...]

PR #2937, opened 2020-07-06
text edit

k09i71:


rigelk:
Thanks, but the wording has already been discussed in https://github.com/Chocobozzz/PeerTube/issues/2908#issuecomment-650787232.

PR #2960, opened 2020-07-13
Display username in password-reset mail

### LINKED ISSUE #2954: Wrong username account displayed in reset-password mail
The password reset procedure send an email with the link to reset it, but in the mail text the username account is not displayed, the variable seems to be escaped :

Peertube-issue-account-reset-password

kimsible:
fixes #2954

### DIFF
--- a/server/controllers/api/users/index.ts
+++ b/server/controllers/api/users/index.ts
@@ -340,7 +340,7 @@ async function askResetUserPassword (req: express.Request, res: express.Response
 
   const verificationString = await Redis.Instance.setResetPasswordVerificationString(user.id)
   const url = WEBSERVER.URL + '/reset-password?userId=' + user.id + '&verificationString=' + verificationString
-  await Emailer.Instance.addPasswordResetEmailJob(user.email, url)
+  await Emailer.Instance.addPasswordResetEmailJob(user.username, user.email, url)
 
   return res.status(204).end()
 }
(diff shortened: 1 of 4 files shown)


kimsible:
@rigelk I think it is a fix for an important issue because when the user receives the notification email, when can it be know the notification concerns the user's account ? For us, as developer, it's pretty obvious yes but for the user it could be blocking.

rigelk:
Yes - I'm just renaming because prefixing with "fix :" is not describing the content of the PR.

kimsible:
!

> Yes - I'm just renaming because prefixing with "fix :" is not describing the content of the PR.

Ok, it has nothing to do with the PR but is there a label to tag fixes ?

rigelk:
For PRs, the determinant of its nature is the issue it solves, so using the `fixes #..` syntax in the end of the description or as a "stamp" in a commit message helps determine that.

kimsible:
> For PRs, the determinant of its nature is the issue it solves, so using the `fixes #..` syntax in the end of the description or as a "stamp" in a commit message helps determine that.
> 
> We use other labels to help us triage PRs (maintenance/priority), but not determine their fix/plain enhancement status.

Thanks !

kimsible:
> replace the bland `Hi,` with `Hi {username},`. It would solve the current problem and more to come.

`Hi {username}` is already used in `/server/lib/emails/common/greeting.pug`

https://github.com/Chocobozzz/PeerTube/blob/develop/server/lib/emails/common/greetings.pug#L5

```
  if username
    p Hi #{username},
  else
    p Hi,
```

rigelk:
Yes, I just meant to send inject the variable in other email templates

Chocobozzz:
Please use `release/2.3.0` as base branch since it's a fix, so we can fix this issue in the next release

PR #2962, opened 2020-07-13
disable vertical scroll instead of hide on desktop browsers

kimsible:
Small PR which is more a visual fix.

It improves UX/UI nav transitions on desktop browsers where the vertical scrollbar is not floating over the body like touch devices do.

Now, when switching **from a vertical scrolled content to a non-scrolled content / opened modal view**, the vertical scroll will be disabled instead of hide on desktop browsers.

**This way, header (upload button and search input) and content won't be involuntarily moved horizontally to the right**.

Capture d’écran du 2020-07-14 00-13-24


Chocobozzz:
Tested and happy with the result. Thanks @kimsible

PR #2963, opened 2020-07-13
correct maximized icon padding since new icons set implemented

kimsible:
Since https://github.com/Chocobozzz/PeerTube/pull/2915 the maximized icon padding in markdown textarea was not correct.

Before: 
maximized-icon

After:
maximized-icon-fixed


Chocobozzz:
Please use `release/2.3.0` as base branch since it's a fix

PR #2966, opened 2020-07-14
Add alert and hide upload view when no upload is possible

kimsible:
When user video quota or daily upload limit are set to `None - no upload possible`, an alert should inform that upload is disabled for the account and the form view not displayed once clicked on the `Upload button` and accessing to `/videos/upload`.

This PR proposes a way to alert the user and hide the upload form in this case : 

no-upload

### DIFF
--- a/client/src/app/+my-account/my-account.component.ts
+++ b/client/src/app/+my-account/my-account.component.ts
@@ -1,5 +1,5 @@
 import { Component, OnInit } from '@angular/core'
-import { AuthService, ServerService, User } from '@app/core'
+import { AuthService, ServerService, AuthUser } from '@app/core'
 import { I18n } from '@ngx-translate/i18n-polyfill'
 import { ServerConfig } from '@shared/models'
 import { TopMenuDropdownParam } from '../shared/shared-main/misc/top-menu-dropdown.component'
@@ -11,7 +11,7 @@ import { TopMenuDropdownParam } from '../shared/shared-main/misc/top-menu-dropdo
 })
 export class MyAccountComponent implements OnInit {
   menuEntries: TopMenuDropdownParam[] = []
-  user: User
+  user: AuthUser
 
   private serverConfig: ServerConfig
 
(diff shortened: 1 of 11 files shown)


rigelk:
Aside from a few errors that would need correction*, I'm pretty sure that message will never show, because the quota will never be reached *exactly to the bit*.

The idea is good though. Maybe a warning when below a certain percentage of the quota would be nice instead.

---

*you are using `videoQuota` and `videoQuotaDaily`, which don't reflect the *used* quota. For this you should use `videoQuotaUsed` and `videoQuotaUsedDaily`. I would also make the message clearer as to why the upload is not available.

kimsible:
> Aside from a few errors that would need correction*, I'm pretty sure that message will never show, because the quota will never be reached _exactly to the bit_.
> 
> The idea is good though! Maybe a warning when below a certain percentage of the quota would be nice instead.
> 
> *you are using `videoQuota` and `videoQuotaDaily`, which don't reflect the _used_ quota. For this you should use `videoQuotaUsed` and `videoQuotaUsedDaily`. I would also make the message clearer as to why the upload is not available.

Yes I thought about the quota, and it would be nice to display it. But this PR is not really about *used quota*, it's about the option *None - no upload possible* selected by the admin for a user as bellow : 

Screenshot_2020-07-15 Update a user - QueerMotion

It's a specific case, when the user has no quota at all for a reason like having a user only to comment and no uploads.

I suggest to make the *used quota* warning in another PR with the display quota as `/my-account/my-settings` does.

test2a:
Can you add a link to ping the instance mod in case you want more help/appeal the quota?
The screenshot above shows just a message. So presumably if I am disabled from uploading a video, I would have to go back and message admins about it.
Having a direct button to send a message would help the user imo

kimsible:
> Can you add a link to ping the instance mod in case you want more help/appeal the quota?
> The screenshot above shows just a message. So presumably if I am disabled from uploading a video, I would have to go back and message admins about it.
> Having a direct button to send a message would help the user imo

Thanks for your comment, you mean a link to the about page where contact form is when enabled ? it you the contact form is disabled what would you suggest to display ?

rigelk:
@kimsible users should be able to contact the admin, yes.

test2a:
@kimsible thanks. Yes. What I am saying is,
The screenshot says "Sorry, the upload feature is disabled for your account.".

What I am suggesting is "Sorry, the upload feature is disabled for your account. Click here to contact the admin for help."

This "here" could open the "contact administrator" form that shows up on the stat page
 
https://peertube2.cpy.re/about/instance


This would probably be a small papercut improvement but I do forsee the need for it. I could be wrong, maybe the team here can deliberate on this ?

kimsible:
> @kimsible thanks. Yes. What I am saying is,
> The screenshot says "Sorry, the upload feature is disabled for your account.".
> 
> What I am suggesting is "Sorry, the upload feature is disabled for your account. Click here to contact the admin for help."
> 
> This "here" could open the "contact administrator" form that shows up on the stat page
> 
> https://peertube2.cpy.re/about/instance
> 
> This would probably be a small papercut improvement but I do forsee the need for it. I could be wrong, maybe the team here can deliberate on this ?

I suggest we redirect the button to the about page, because we miss the case some instance do not have the admin form enabled.

image

What do you think @test2a  ?

test2a:
@kimsible much improvement. Thanks. Thats the gist of what i was thinking. Now, a user who is seeing unable to upload any videos gets sent to a page explaining rules. 
Earlier a user would be stuck, not knowing what to do.

Thanks for the efforts you guys put into the software. You are awesome. ?-)

client/src/app/+videos/+video-edit/video-add.component.ts
```diff
@@ -59,6 +59,12 @@ export class VideoAddComponent implements OnInit, CanComponentDeactivate {
     return { canDeactivate: true }
   }
 
+  canUpload () {
+    const { videoQuota, videoQuotaDaily } = this.auth.getUser()
+
+    return (videoQuota > 0 || videoQuota === -1) && (videoQuotaDaily > 0 || videoQuotaDaily === -1)
```
Chocobozzz, line 65: Please put the logic inside the `User` model (in a `canUpload` for example)

client/src/app/+my-account/my-account.component.ts
```diff
@@ -63,6 +65,21 @@ export class MyAccountComponent implements OnInit {
       })
     }
 
+    this.user = this.authService.getUser()
+
+    this.authService.userInformationLoaded.subscribe(
+      () => {
+        // Cannot see videos links, hide my videos and my imports
+        if (!this.canSeeVideosLink()) {
+          libraryEntries.children.splice(1, 1)
```
Chocobozzz, line 74: A `isDisplayed (): boolean` method (default is `true`) in the `TopMenuDropdownParam` type would be more robust

client/src/app/+my-account/my-account.component.ts
```diff
@@ -104,4 +121,17 @@ export class MyAccountComponent implements OnInit {
     return importConfig.http.enabled || importConfig.torrent.enabled
   }
 
+  canSeeVideosLink () {
```
Chocobozzz, line 124: Please put this logic in the `User` model

client/src/app/+my-account/my-account.component.ts
```diff
@@ -104,4 +121,17 @@ export class MyAccountComponent implements OnInit {
...
+    const { videoQuota, videoQuotaDaily, videosCount } = this.user
+
+    // can upload
```
Chocobozzz, line 127: Seems simpler:

```
if (videosCount === 0 && (videoQuota === 0 || videoQuotaDaily === 0)) return false

return true

```

client/src/app/+my-account/my-account.component.ts
kimsible: >  A isDisplayed (): boolean method (default is true) in the TopMenuDropdownParam type would be more robust

Good idea ! I sugguest optionnal isHidden () method, the default value will be undefined, so false in Angular template

client/src/app/core/users/user.model.ts
```diff
@@ -149,4 +149,21 @@ export class User implements UserServerModel {
...
+
+  canSeeVideosLink (): boolean {
+    if (this.canUpload()) {
+      return true
+    }
+
+    // cannot upload but has already some videos
+    if (this.videosCount > 0) {
+      return true
+    }
```
Chocobozzz, line 167: You forgot `return false` no?

client/src/app/menu/menu.component.ts
```diff
@@ -114,6 +114,20 @@ export class MenuComponent implements OnInit {
     }
   }
 
+  canSeeVideosLink () {
```
Chocobozzz, line 117: Use the `user` `canSeeVideosLink` method

client/src/app/shared/shared-main/misc/top-menu-dropdown.component.ts
```diff
@@ -9,11 +9,12 @@ import { NgbDropdown, NgbModal } from '@ng-bootstrap/ng-bootstrap'
 export type TopMenuDropdownParam = {
   label: string
   routerLink?: string
+  isHidden?: boolean
```
Chocobozzz, line 12: I think it should be a method so we can put a dynamic value here. (not needed now, but can be useful in the future)

client/src/app/core/users/user.model.ts
```diff
@@ -149,4 +149,21 @@ export class User implements UserServerModel {
   updateAccountAvatar (newAccountAvatar: Avatar) {
     this.account.updateAvatar(newAccountAvatar)
   }
+
+  canUpload (): boolean {
```
Chocobozzz, line 153: `isUploadDisabled` would be a better name, since this method does not check used quota (which may exceed the user quota)

client/src/app/core/users/user.model.ts
kimsible: Yes, sorry, undefined is really false

client/src/app/core/users/user.model.ts
```diff
@@ -149,4 +149,21 @@ export class User implements UserServerModel {
...
+    return (this.videoQuota === 0 || this.videoQuotaDaily === 0)
+  }
+
+  canSeeVideosLink (): boolean {
+    if (!this.isUploadDisabled()) {
+      return true
+    }
+
+    // cannot upload but has already some videos
+    if (this.videosCount > 0) {
```
kimsible, line 163: @Chocobozzz we have an issue here, it seems `videosCount` is always undefined in this context

client/src/app/shared/shared-main/misc/top-menu-dropdown.component.ts
kimsible: Ok I understood what you mean =)

client/src/app/core/users/user.model.ts
Chocobozzz: Because we don't `videosCount` in the `/me` endpoint. I'll update this PR to inject it

PR #2973, opened 2020-07-15
WIP Improve navigation user-dropdown, left-menu, my-account

kimsible:
Discussion : https://framacolibri.org/t/discussions-about-navigation-design/8599/2

user-dropdown
my-library

my-account-nav

### DIFF
--- a/client/src/app/+my-account/+my-account-video-channels/my-account-video-channel-edit.component.scss
+++ b/client/src/app/+my-account/+my-account-video-channels/my-account-video-channel-edit.component.scss
@@ -54,6 +54,8 @@ textarea {
 
 .breadcrumb {
   @include breadcrumb;
+
+  margin-top: 20px;
 }
 
 @media screen and (max-width: $small-view) {
--- a/client/src/app/+my-account/+my-account-video-channels/my-account-video-channels.component.html
+++ b/client/src/app/+my-account/+my-account-video-channels/my-account-video-channels.component.html
@@ -1,5 +1,6 @@
-<h1 class="sr-only" i18n>My channels</h1>
 <div class="video-channels-header">
+  <h1 i18n>My channels</h1>
+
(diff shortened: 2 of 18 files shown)


Chocobozzz:
Not reviewed the content, but from the UI perspective it's interesting. Just some comments:
 * A lot of people don't understand the channels concept, and don't really use it. So, should we put this in the menu?
 * I would rename `More settings` from the menu to just `More`, since it's not just about settings
 * I would keep the `More settings` below `Help share videos` in the account dropdown

kimsible:
> Not reviewed the content, but from the UI perspective it's interesting. Just some comments:
> 
> * A lot of people don't understand the channels concept, and don't really use it. So, should we put this in the menu?
> * I would rename `More settings` from the menu to just `More`, since it's not just about settings
> * I would keep the `More settings` below `Help share videos` in the account dropdown

Thanks ! it was a poc/test, this PR really needs rebasing... 

Yes the channels concept is not understandable for lots of people but maybe because it's not as visible as on the signup form, maybe it should be show up ? 

Ok for the names corrections.

Do you think entries already in the left menu should be kept in the sub menu ?

I think the real problem of design here is that left menu, user dropdown and my-account sub-menu have each one their own logic / naming, the logic / naming should be harmonized in everywhere.

Chocobozzz:
> Do you think entries already in the left menu should be kept in the sub menu ?

I'm not sure to understand. Are you talking about *I would keep the More settings below Help share videos in the account dropdown*?

test2a:
@kimsible on mobiles clicking on the profile sidebar on v 2.3.0 website throws the page to desktop mode, it does not accept the mobile resolution

kimsible:
@Chocobozzz I think this PR can be closed since you implemented the new user menu here https://github.com/Chocobozzz/PeerTube/pull/3296

PR #2981, opened 2020-07-19
Display user quota progress bars above upload form

kimsible:
This PR follows extra infos upload form of this https://github.com/Chocobozzz/PeerTube/pull/2966.
This one display the user quota progress bars above the upload form.
**The progress bar height has been doubled for more visibility and touch devices allowing to see the title on touch**

video-quota2
video-quota1


rigelk:
@kimsible doubling the bar height for visibility is a good idea! I'm not so sure about displaying the bars on the upload page though - at least not this way.

kimsible:
@rigelk what would you see to improve the UX ? an avatar of the user on the left ? an explicit title and help tooltips ?

For example Vimeo : 
image

Chocobozzz:
I think explicit titles would be better, because users won't really understand what are these progress bars when they come in the upload page for the first time

kimsible:
With an explicit title and a tooltip with quota values for accessibility : 

Capture d’écran du 2020-07-24 00-16-17
Screenshot_2020-07-24 Upload a video - mon instance

Chocobozzz:
Thanks, let's try this

Issue #2982, opened 2020-07-19
Action buttons display broken for multiple videos selection mode


kimsible:
In `My videos`, a multiple selection mode is possible, but cancel and delete actions are not displaying well.

![broken-multiple-delete](https://user-images.githubusercontent.com/1877318/87883219-3a99e480-ca06-11ea-8aa5-ed7a876849bb.png)

Plus on small screens, though checkbox are hidden to disable the feature, these action should also be hidden : on touchscreens, the screen rotation may have this buggy display : 
![broken-multiple-delete-mobile](https://user-images.githubusercontent.com/1877318/87883222-3f5e9880-ca06-11ea-9f3d-1ccd7036c72d.png)

PR #2983, opened 2020-07-19
Correct action button selection mode styles

kimsible:
Fixes https://github.com/Chocobozzz/PeerTube/issues/2982

image


Chocobozzz:
Please use `release/2.3.0` as base branch since it's a fix we would like in 2.3.0

Issue #2970, opened 2020-07-14
Can't create user from admin since v2.3.0-rc1


kimsible:
Since v2.3.0-rc1, we can't create an user from admin anymore, the select fields `Role`, `Video Quota`, `Daily Video Quota`, checkbox  `Blacklist video` are empty.

When click on send button, the error alert is empty and the user is not created.

![create-user](https://user-images.githubusercontent.com/1877318/87480504-bcb49280-c62d-11ea-9563-22b83eff6d30.png)

Browser console error.

```
ERROR TypeError: ctx.user is undefined
    UserCreateComponent_Template user-edit.component.html:97
    Angular 33
        executeTemplate
        refreshView
        refreshComponent
        refreshChildComponents
        refreshView
        refreshEmbeddedViews
        refreshView
        refreshComponent
        refreshChildComponents
        refreshView
        refreshEmbeddedViews
        refreshView
        refreshComponent
        refreshChildComponents
        refreshView
        refreshEmbeddedViews
        refreshView
        refreshComponent
        refreshChildComponents
        refreshView
        renderComponentOrTemplate
        tickRootContext
        detectChangesInRootView
        detectChanges
        tick
        next
        invoke
        onInvoke
        invoke
        run
        run
        next
        schedulerFn
    RxJS 5
        __tryOrUnsub
        next
        _next
        next
        next
    Angular 9
        emit
        checkStable
        onLeave
        onInvokeTask
        invokeTask
        runTask
        invokeTask
        invokeTask
        globalZoneAwareCallback
```

PR #2984, opened 2020-07-19
Add channel name to create-user admin form

kimsible:
Required fix : https://github.com/Chocobozzz/PeerTube/issues/2970

image

### DIFF
--- a/client/src/app/+admin/users/user-edit/user-create.component.ts
+++ b/client/src/app/+admin/users/user-edit/user-create.component.ts
@@ -44,6 +44,7 @@ export class UserCreateComponent extends UserEdit implements OnInit {
 
     this.buildForm({
       username: this.userValidatorsService.USER_USERNAME,
+      channelName: this.userValidatorsService.USER_USERNAME,
       email: this.userValidatorsService.USER_EMAIL,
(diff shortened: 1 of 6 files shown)


Chocobozzz:
Is it still a WIP?

kimsible:
> Is it still a WIP?

Yes, didn't have time to commit my local changes with the fix.

kimsible:
TODO: `channelName` must be added to `createUser` api doc as an optionnal parameter.

- On client side in the admin panel, `channelName` is required to avoid **Main *username* channel** `displayName` and unwanted unique channel name, this way we follow the signup form which forces the user to give a channel name ;
- **Server side, `channelName` stays optionnal to not break the api**.

kimsible:
Tests has been fixed.

Chocobozzz:
We don't want to autofill the channel name field depending on the user username like we do in the signup form?

kimsible:
> We don't want to autofill the channel name field depending on the user username like we do in the signup form?

You mean adding Display name fields ?

Otherwise, the goal is also to avoid the generic names "kimsible_channel" and "Main kimsible channel", I don't think automatic fill for we don't want anymore is the right thing to do.

In any case the signup form asks username and channel name (with display names) on separate steps.

Display names may be edited by the user after account created.

Issue #2987, opened 2020-07-20
Improvement generic channel UI


kimsible:
When an instance has closed registration, the only way to create accounts for the moment is to use the admin / moderation feature  « Admin > users > create user ».

The form allows the user to choose its account password by sending a mail. But we miss the unique channel name. So a generic channel name is created ***username*_channel**. Display channel name is also generated **Main *username* channel**.

The user is not as well as guided than the `signup` form when registration are opened on the instance. So the user misses all the required configuration of its own channel. I may be wrong but most of PeerTube instances have closed registrations for several reasons, so the signup form is not very used, and never for most instances then.

An example of how channel description looks like under the watch view of the video : 

![image](https://user-images.githubusercontent.com/1877318/87943789-d414d500-ca9e-11ea-85aa-621ad5ef2ffe.png)

I think we need 4 levels of improvements : 
- [x] **UI - High priority - Easy** Do not display the channel display name and avatar when both are not customized ;
   Allow old channels created to benefit at least an improvement; 
- [ ] **Admin- High priority - Average** Let the admin/modo enter a channel name, it will resolve one thing (only the unwanted generic unique channel name);
- [ ] **Admin - Medium priority - Average** Let the admin/modo send an invitation link in the admin creation form with the ability to pre-define role, quotas and blacklist video option;
- [ ] **UI + federation update - Low priority - hard** Ability to rename the unique channel name with federation / followers instances update.

kimsible:
> > Ability to rename the unique channel name with federation / followers instances update.
> 
> I don't think we'll implement this soon grin

Yes, that's why I put it as a low priority, but as I said here : https://framacolibri.org/t/edit-channel-and-account-url-name/8394/7

> Naming URLs can be an issue for creators who change their names (in case of civil state changes if they use their legal name, for example trans/non-binary people are deeply concerned).

We don't have to put them aside, and reproduce what Google/YouTube does, the right to change our name is also in the law in most of countries especially in France. Some people use a pseudonym, some people does use their legal names. It would be unfair to let this issue unchanged for a long time, when especially some of them want to go on PeerTube to avoid Google / YouTube Politic. 

The option to create a new channel and transfer all videos of the old channel to the new one puts the concerned user on an unfair choice : **Loose all your followers and channel views or be exposed with your dead-name and the fact you're transitioning**. This feature should be considered as an accessibility feature, changing name is a question of survival for some people and being forced to expose this condition is a serious social handicap.

[...]

PR #2988, opened 2020-07-20
Hide generic channel display name and avatar on watch view

### LINKED ISSUE #2953: Default ABC  channel / Main ABC channel
**What happened?**
Below the videos, peertube currently shows words "main" and "default" as a prefix to the channel name. 

Is this necessary? Can we use the words as a suffix instead because if I say check out ABC channel or I see a video, its normal to check the icon/name and I find the prefix off.
Maybe im being unreasonable,  question, is this because the user hasnt named their channels?

https://tilvids.com/videos/watch/3b4be045-bcd0-4968-a4e7-99fbde8bd281

This link says
"Main arcaneengine channel".
Can we rewrite this as 
"Arcaneengine - Main Channel"

If we really have to differentiate between channels not based on name but also on having prefix/suffix of the word channel  ?

https://peertube.cpy.re/videos/watch/da2b08d4-a242-4170-b32a-4ec8cbdca701

Here it says
"Default Chocobozzz channel"

kimsible:
First stab at : https://github.com/Chocobozzz/PeerTube/issues/2987
fixes #2953

- No avatar channel and generic display name (Main *username* channel) channel :
*the big avatar and account name redirect to the account page*
no-avatar-and-channel

- Generic display name (Main *username* channel) and a channel avatar is uploaded
*the small avatar is the channel avatar and redirect to the channel page*
no-channel-avatar-not-null

- Display name channel and avatar channel customized
*back to regular display*
n-avatar-and-channel

### DIFF
--- a/client/src/app/+videos/+video-watch/video-watch.component.ts
+++ b/client/src/app/+videos/+video-watch/video-watch.component.ts
@@ -308,9 +308,12 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
   }
 
   isChannelDisplayNameGeneric () {
-    const genericChannelDisplayName = `Main ${this.video.channel.ownerAccount.name} channel`
+    const genericChannelDisplayName = [
+      `Main ${this.video.channel.ownerAccount.name} channel`,
+      `Default ${this.video.channel.ownerAccount.name} channel`
+    ]
 
-    return this.video.channel.displayName === genericChannelDisplayName
+    return genericChannelDisplayName.includes(this.video.channel.displayName)
   }
 
   private loadVideo (videoId: string) {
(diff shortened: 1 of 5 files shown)


client/src/app/+videos/+video-watch/video-watch.component.ts
```diff
@@ -307,6 +307,12 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
     )
   }
 
+  isChannelDisplayNameGeneric () {
+    const genericChannelDisplayName = `Main ${this.video.channel.ownerAccount.name} channel`
```
Chocobozzz, line 311: At the beginning of peertube it was also `Default ... channel`. Please add this condition too (or use an array)

Chocobozzz:
Nice job :+1: Just one little thing and we can merge this

kimsible:
> Nice job +1 Just one little thing and we can merge this

Thanks for the review =). I forgot the search result, should we also adapt display for this view ?

image

Chocobozzz:
> I forgot the search result, should we also adapt display for this view ?

I don't think so, because we are explicitly searching for channels. To not confuse users, let's keep the current display in the search page

PR #3004, opened 2020-07-24
correct regressions on sub-menu for account and admin

kimsible:
Fixes : 

image
image

### DIFF
--- a/client/src/sass/application.scss
+++ b/client/src/sass/application.scss
@@ -136,6 +136,10 @@ label {
     flex-grow: 1;
   }
 
+  my-top-menu-dropdown + .margin-content {
+    margin-top: $sub-menu-height + $sub-menu-margin-bottom;
+  }
+
   .sub-menu {
     background-color: pvar(--submenuColor);
     width: 100%;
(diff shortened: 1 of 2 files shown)


client/src/sass/application.scss
```diff
@@ -145,7 +149,11 @@ label {
     padding-left: $not-expanded-horizontal-margins;
     padding-right: $not-expanded-horizontal-margins;
     position: fixed;
-    z-index: 1;
+    z-index: 200;
```
rigelk, line 152: please use the `z` function, to have z-indexes relative to https://github.com/Chocobozzz/PeerTube/blob/2e7f262724dd64a209e0bad5930ba29bb4f801c3/client/src/sass/include/_variables.scss#L123-L139

rigelk:
This PR breaks account settings.

Screenshot_2020-07-24 Account settings - PeerTube

test2a:
https://ibb.co/d4cLZkj
hey. just found this visual bug on peertub2.cpy.re on my mobile phone just minutes ago.
observations:

    the stats page does not expand to 100%, as you mentioned.
    the header has what appears to be fixed background and that hides the text and other elements as it sticks above all.
    if i resize the browser window, the sidebar does NOT appear in the sidebar on a phone but i think it assumes i have a different resolution and shows up as a smaller window which is not intuitive

kimsible:
> This PR breaks account and admin settings.

Thanks @rigelk =) it's fixed !

rigelk:
Thanks @kimsible :)

PR #3023, opened 2020-07-29
Revert user-menu dropdown active class on sub-menu and close outside-click

kimsible:
This PR reverts the recent proposition to close the dropdown sub-menu only on outside click and showing a background color on active links. 

Now, autoclose is back on click inside and outside and active class is removed, because this active effect class caused some blinking effect on the dropdown.

### DIFF
--- a/client/src/app/shared/shared-main/misc/top-menu-dropdown.component.html
+++ b/client/src/app/shared/shared-main/misc/top-menu-dropdown.component.html
@@ -4,12 +4,8 @@
     <a *ngIf="menuEntry.routerLink && isDisplayed(menuEntry)" [routerLink]="menuEntry.routerLink" routerLinkActive="active" class="title-page title-page-settings">{{ menuEntry.label }}</a>
 
     <div *ngIf="!menuEntry.routerLink && isDisplayed(menuEntry)" ngbDropdown class="parent-entry"
-      #dropdown="ngbDropdown" autoClose="outside">
+      #dropdown="ngbDropdown" autoClose="true">
       <span
-        *ngIf="isInSmallView"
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks

client/src/app/shared/shared-main/misc/top-menu-dropdown.component.html
```diff
@@ -4,12 +4,8 @@
     <a *ngIf="menuEntry.routerLink && isDisplayed(menuEntry)" [routerLink]="menuEntry.routerLink" routerLinkActive="active" class="title-page title-page-settings">{{ menuEntry.label }}</a>
 
     <div *ngIf="!menuEntry.routerLink && isDisplayed(menuEntry)" ngbDropdown class="parent-entry"
-      #dropdown="ngbDropdown" autoClose="outside">
+      #dropdown="ngbDropdown" autoClose="true">
       <span
-        *ngIf="isInSmallView"
```
Chocobozzz, line 9: Is this deletion a mistake?

Chocobozzz: I put it back in https://github.com/Chocobozzz/PeerTube/commit/805d6a354bd51fccd434c0c1b1df69c81bb90dd4 (we can revert if it was not a mistake) because it causes duplicate entries in the admin menu

kimsible: Yes, thanks. Sorry, it was a mistake, without it would indeed duplicate the entry with one for the modal view and another for the dropdown view.

Issue #3030, opened 2020-08-02
Regression on my-account / broken sub-item titles on small screens


kimsible:
On my-account sub-items titles we have a regression on small screens sub-item titles : 

![image](https://user-images.githubusercontent.com/1877318/89134431-88920a80-d525-11ea-9599-6bea1633757e.png)

![image](https://user-images.githubusercontent.com/1877318/89134465-b9723f80-d525-11ea-8a47-fbe86560dd57.png)

![image](https://user-images.githubusercontent.com/1877318/89134480-e6beed80-d525-11ea-990a-c49f6ce6e879.png)

It seems to be a mistake during a rebasing, the action button and search input was proposed to be under the sub-title.

Issue #3020, opened 2020-07-29
Different subscription icons in my-account since last changes


kimsible:
Since https://github.com/Chocobozzz/PeerTube/pull/2977/files some icons has been updated like the subscription icon to be common with the one in the left menu :

![image](https://user-images.githubusercontent.com/1877318/88801990-bd5f3400-d1aa-11ea-997c-ea17048caeb9.png)
![image](https://user-images.githubusercontent.com/1877318/88802046-d23bc780-d1aa-11ea-96aa-d7dcafec79dc.png)

And since the general user-menu effects on this PR : https://github.com/Chocobozzz/PeerTube/pull/2971, this icon has also been used on sub-titles.

![image](https://user-images.githubusercontent.com/1877318/88802265-1c24ad80-d1ab-11ea-85b6-2ebc96109260.png)

At last, since this last commit https://github.com/kimsible/PeerTube/commit/aa91835898a80e4ea142fc27b40327d8c4461686 on this PR https://github.com/Chocobozzz/PeerTube/pull/2966 about the limitation of upload form and user menus, a small refacto has been made on the user sub-menu and re-injected the old icon : 

![image](https://user-images.githubusercontent.com/1877318/88801684-52adf880-d1aa-11ea-9afe-924df5625230.png)

Anyway what icon is used, the same ones on sub-titles and sub-menu should be used.

rigelk:
fixed by #3031.

PR #3031, opened 2020-08-03
Fix regressions in my-account sub menu and sub titles

kimsible:
fixes: https://github.com/Chocobozzz/PeerTube/issues/3030 and https://github.com/Chocobozzz/PeerTube/issues/3020


Chocobozzz:
Thanks!

PR #3032, opened 2020-08-03
Add new anchors in my-settings and handle offset sub-menu height

kimsible:
All links in the user dropdown-menu refered to `/my-account/settings` with an anchor `#video-settings`, does not scroll exactly to the the target element, plus when we are on my-settings all these links can't work properly since they have the same anchor.

It also handle the offset with the fixed sub-menu.

image

### DIFF
--- a/client/src/app/app-routing.module.ts
+++ b/client/src/app/app-routing.module.ts
@@ -67,10 +67,9 @@ const routes: Routes = [
   imports: [
     RouterModule.forRoot(routes, {
       useHash: Boolean(history.pushState) === false,
-      scrollPositionRestoration: 'enabled',
+      scrollPositionRestoration: 'disabled',
       preloadingStrategy: PreloadSelectedModulesList,
-      anchorScrolling: 'enabled',
-      scrollOffset: [0, 81] // sub-menu-height
+      anchorScrolling: 'disabled'
     })
   ],
   providers: [
--- a/client/src/app/app-routing.module.ts
+++ b/client/src/app/app-routing.module.ts
@@ -67,9 +67,10 @@ const routes: Routes = [
   imports: [
     RouterModule.forRoot(routes, {
       useHash: Boolean(history.pushState) === false,
-      scrollPositionRestoration: 'disabled',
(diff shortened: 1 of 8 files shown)


client/src/app/app-routing.module.ts
```diff
@@ -67,9 +67,10 @@ const routes: Routes = [
   imports: [
     RouterModule.forRoot(routes, {
       useHash: Boolean(history.pushState) === false,
-      scrollPositionRestoration: 'disabled',
```
Chocobozzz, line 70: We disabled `scrollPositionRestoration` and `anchorScrolling` because we have our own scroll logic: https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/app.component.ts#L119

kimsible: Thanks, I've reported the changes here.

client/src/app/app.component.ts
```diff
@@ -114,6 +114,8 @@ export class AppComponent implements OnInit, AfterViewInit {
     let resetScroll = true
     const eventsObs = this.router.events
 
+    this.viewportScroller.setOffset([0, 81]) // sub-menu-height
```
Chocobozzz, line 117: Should only be applied if targeting `my-account` no?

kimsible: `my-account` and `about` ? Maybe we would add any anchors in `about/instance`

Chocobozzz: yes

kimsible:
One thing left, for example when we request two times on /my-account/settings#notifications and we scrolled between the two requests we would like to go back to the anchor #notifications scroll position with the scrollToAnchor. My knowledge of Angular Router is still limited, would you have an idea @rigelk @Chocobozzz how to proceed ? which routee event we should subscribe ?

Chocobozzz:
> One thing left, for example when we request two times on /my-account/settings#notifications and we scrolled between the two requests we would like to go back to the anchor #notifications scroll position with the scrollToAnchor

Sorry I don't know

PR #3034, opened 2020-08-04
add and document new additional video extensions supported at upload

rigelk:
In addition to documenting a bit existing extensions, add `.wmv`, a missing mainstream container. Other containers added are either low-hanging fruits (3GPP, FLV replacement, OGG (`.ogg` can also be a video container), Sony/Panasonic formats which are also found in BluRay, `.nut` as ffmpeg/mplayer developed it).

### DIFF
--- a/client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.html
+++ b/client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.html
@@ -731,7 +731,7 @@
                         i18n-labelText labelText="Allow additional extensions"
                       >
                         <ng-container ngProjectAs="description">
-                          <span i18n>Allows users to upload .mkv, .mov, .avi and .flv videos.</span>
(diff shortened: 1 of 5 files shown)


client/src/app/+videos/+video-edit/video-add-components/video-upload.component.ts
```diff
@@ -67,7 +67,7 @@ export class VideoUploadComponent extends VideoSend implements OnInit, OnDestroy
   }
 
   get videoExtensions () {
-    return this.serverConfig.video.file.extensions.join(', ')
+    return this.serverConfig.video.file.extensions.filter(e => e !== 'null').join(', ')
```
Chocobozzz, line 70: Should be handled directly in `updateWebserverConfig` function in `constants.ts` so we don't include the `null` value in the object keys

Issue #3026, opened 2020-08-01
Add the ability for channel owners to 'react' to comments


k09i71:
Add the ability for channel owners to 'react' to comments, similar to youtube's hearts. 

We should pick 5 - 6 emojis / icons and have them as the default reactions, with the ability for instance admins to add, remove or edit them.

A few FOSS emoji libraries I've found:

https://twemoji.twitter.com/
https://openmoji.org/library/

Niquarl:
What is the point of the video's owner reacting to a comment ? Shouldn't we rather want them to write a comment ? Or is the idea to 'pin' comments to the top ?
What are you thinking about the 5 default reactions ? Like a heart, a thumbs up/down, a thank you ? I'd like to know what is the intended point told by the 'reactions'.

Issue #2212, opened 2019-10-30
Can we get video responces?


anubis2814:
This may well be the worst place to post this, but I don't know where to direct this.

Video responses were one of the things that made youtube a community.  It was also a way for smaller youtubers to gain subscribers.  Once youtube killed it, whole communities collapsed.  I've been looking for a platform for years with a video response option, and I know lots of people from varying communities on youtube that would probably switch to peertube if there was a video response option.

Also you guys are probably already working on this but peertube needs better filtering by language for discover, trending and recently added.  Maybe a filter option for all languages or just your native language if you feel like it.  It really makes it hard to discover everything if most of it is in languages you don't know.

If you have another place I can post this on, please let me know, I don't want to clog up a site that is much more specific to actual problems.

Nutomic:
Could you explain how video responses (used to) work on Youtube?

robinkooli:
On the comment section there is a button “Submit video response”, that would open a list of videos and you select one and send it.

http://www.onespoonatatime.com/wp-content/uploads/2012/04/video-response-2.jpg

anubis2814:
I think that sums it up, let us know if you have other questions.  They killed the function because they wanted to make youtube a celebrity place as opposed to a community of small timers.  I was gaining around 1k subs per year.  Then in I think 2013 they killed it and in the past 5 years I've gained 1k while posting at the exact same rate if not more.  Most of my fellow community members experienced the same thing making them give up youtubing entirely.  Now its go big and flashy and make lots of money or go home.

rigelk:
@robinkooli @anubis2814 I have a few questions, to better understand the flow of actions related to video responses (or UX): are videos responses also possible to respond to a comment? Are they moderated differently from regular comments? Are they published along with a message/comment?

---

On another note, I found back [something that looks like the announcement](https://youtube-creators.googleblog.com/2013/08/so-long-video-responsesnext-up-better.html) made to retire the functionality:

> […] replacing little-used features with better ones. Currently video responses have a click-through rate of .0004% -- in other words, only 4 out of every 1 million users who sees a video response clicks on it. So, on September 12 we’re going to retire this little-used feature as we work to develop more effective fan engagement tools for creators. The team is focused on enabling you to share video links in comments. Doing this in comments will let creators and viewers add more context to a video, and more context should drive more engagement.

Not saying we should also care about the click rate like Google does. But that coud hint us toward making a better use of previews in comments, or showing embeds for comments that have a link to a video.

[...]

PR #3046, opened 2020-08-07
Add ability to delete and re-draft a comment with no replies + markdown infos + unicode emojis

kimsible:
- Hide mention of deleted comments without replies ;
- New delete and re-draft action on a comment **only on comments without replies** ;
- Move delete, delete and re-draft actions on options menu ;
- Add icons on delete, delete and re-draft and report actions in options menu ;
- Rename `Reply` value of add comment button to `Comment` **only for comment main thread textarea** ;
- Add Markdown help icon :
- Add uni-code emojis native display (depending on Operating System) https://apps.timwhitlock.info/emoji/tables/unicode.

Comments-markdown

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -61,6 +61,7 @@
     "@types/linkifyjs": "^2.1.2",
     "@types/lodash-es": "^4.17.0",
     "@types/markdown-it": "^10.0.1",
+    "@types/markdown-it-emoji": "^1.4.0",
     "@types/node": "^14.0.14",
     "@types/sanitize-html": "1.23.2",
     "@types/socket.io-client": "^1.4.32",
--- a/client/package.json
+++ b/client/package.json
@@ -61,7 +61,6 @@
     "@types/linkifyjs": "^2.1.2",
     "@types/lodash-es": "^4.17.0",
     "@types/markdown-it": "^10.0.1",
-    "@types/markdown-it-emoji": "^1.4.0",
     "@types/node": "^14.0.14",
     "@types/sanitize-html": "1.23.2",
     "@types/socket.io-client": "^1.4.32",
--- a/client/src/app/+videos/+video-watch/comment/video-comment.component.html
+++ b/client/src/app/+videos/+video-watch/comment/video-comment.component.html
@@ -1,4 +1,4 @@
-<div *ngIf="!comment.isDeleted || comment.isDeleted && comment.totalReplies !== 0" class="root-comment">
+<div *ngIf="isNotDeletedOrDeletedWithReplies()" class="root-comment">
   <div class="left">
     <a *ngIf="!comment.isDeleted" [href]="comment.account.url" target="_blank" rel="noopener noreferrer">
       <img
--- a/client/src/app/+videos/+video-watch/comment/video-comment.component.html
+++ b/client/src/app/+videos/+video-watch/comment/video-comment.component.html
@@ -1,4 +1,4 @@
-<div class="root-comment">
+<div *ngIf="!comment.isDeleted || comment.isDeleted && comment.totalReplies !== 0" class="root-comment">
   <div class="left">
     <a *ngIf="!comment.isDeleted" [href]="comment.account.url" target="_blank" rel="noopener noreferrer">
       <img
@@ -58,7 +58,7 @@
              class="comment-date">{{ comment.createdAt | myFromNow }}</a>
         </div>
 
-        <div *ngIf="comment.isDeleted" class="comment-html comment-html-deleted">
+        <div class="comment-html comment-html-deleted">
           <i i18n>This comment has been deleted</i>
         </div>
       </ng-container>
--- a/client/src/app/+videos/+video-watch/comment/video-comment.component.html
+++ b/client/src/app/+videos/+video-watch/comment/video-comment.component.html
@@ -41,12 +41,10 @@
         ></div>
 
         <div class="comment-actions">
-          <div *ngIf="isUserLoggedIn()" (click)="onWantToReply()" class="comment-action-reply" i18n>Reply</div>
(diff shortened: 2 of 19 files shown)


client/src/app/+videos/+video-watch/comment/video-comment.component.html
```diff
@@ -58,7 +57,7 @@
              class="comment-date">{{ comment.createdAt | myFromNow }}</a>
         </div>
 
-        <div *ngIf="comment.isDeleted" class="comment-html comment-html-deleted">
```
kimsible, line 61: Deleted because duplication of conditional with the parent container

rigelk:
Nice! I like the markdown icon but at the same time I'm not too sure about it's position. Usually this is right aligned in the input, and greyed like the input border.

kimsible:
> Nice! I like the markdown icon but at the same time I'm not too sure about it's position. Usually this is right aligned in the input, and greyed like the input border.

Yes, but we have the extend native icon of textarea on the right. 
We also can do as GiThub does, but we loose the one line textarea and the UI becomes more complex : 
image

kimsible:
@rigelk 
markdown-on-right

kimsible:
The last commit adds unicode emojis to markdown : 
emoji

rigelk:
@kimsible that is what I though, yes :)

Could you please fix https://github.com/Chocobozzz/PeerTube/pull/3046/checks?check_run_id=961331548#step:6:126?

kimsible:
> @kimsible that is what I though, yes :)
> 
> Could you please fix https://github.com/Chocobozzz/PeerTube/pull/3046/checks?check_run_id=961331548#step:6:126?

Thanks, I'm working on it.

For the emojis, I don't know if we must include an emoji collection into PeerTube, in the best way, yes, because some people do not have every one on their operating system (especially on desktop linux and windows). 

What do we do ?

rigelk:
> For the emojis, I don't know if we must include an emoji collection into PeerTube, in the best way, yes, because some people do not have every one on their operating system (especially on desktop linux and windows).
> 
> What do we do ?

I'm inclined to look at what Mastodon does, where they adopted a full emoji management interface that is highly extensible. Until then, only including a light set like `markdown-it-emoji` provides sounds like a possible option (+25kB for the full version or +7kB for the light version sounds *reasonable*).

It's not like PeerTube is only focusing on text and thus very reliant on emojis, but emojis have their place too in the comment palette (i.e. [an issue has already considered video reactions](https://github.com/Chocobozzz/PeerTube/issues/2212) - a much heavier development - which proves there is interest for richer expression forms to react to videos).

kimsible:
> only including a light set like markdown-it-emoji provides sounds like a possible option (+25kB for the full version or +7kB for the light version sounds reasonable).

Yes, the light version sounds reasonable thanks.

> It's not like PeerTube is only focusing on text and thus very reliant on emojis, but emojis have their place too in the comment palette (i.e. an issue has already considered video reactions - a much heavier development - which proves there is interest for richer expression forms to react to videos).

I think you meant this PR https://github.com/Chocobozzz/PeerTube/issues/3026 ?

rigelk:
@kimsible both of them actually, though they are just issues, not PRs.

kimsible:
> @kimsible both of them actually, though they are just issues, not PRs.

Ok @rigelk , so you mean adding a Popover with emojis ? 

Otherwise, I'm not sure (since I'm not an expert of Typescript) but light version emoji of markdown-it seems to not be implemented to `@types/markdown-it-emoji` https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/markdown-it-emoji/index.d.ts, the commonjs way is `require('markdown-it-emoji/light')`, do you know how to do with Typescript ?

rigelk:
> > @kimsible both of them actually, though they are just issues, not PRs.
> 
> Ok @rigelk , so you mean adding a Popover with emojis ?

no, I didn't mean anything like this. The issue I quoted was merely an example of how text is arguably not the only response type we should consider, as an argument in favor of including emojis (since you asked if we *should* include emojis).

Back on your current error, leave it at the typical `require` for the moment. We could always write types, but for now it's more work than it's worth.

kimsible:
The last commits uses Commonjs require to include markdown-it emoji plugin and to get light emoji list.
It also refactor the Markdown Pophover : 
image

image

client/package.json
```diff
@@ -61,6 +61,7 @@
     "@types/linkifyjs": "^2.1.2",
     "@types/lodash-es": "^4.17.0",
     "@types/markdown-it": "^10.0.1",
+    "@types/markdown-it-emoji": "^1.4.0",
```
kimsible, line 64: TODO : to remove

client/yarn.lock
```diff
@@ -1571,7 +1571,14 @@
   dependencies:
     "@types/node" "*"
 
-"@types/markdown-it@^10.0.1":
+"@types/markdown-it-emoji@^1.4.0":
```
kimsible, line 1574: TODO : to remove

kimsible:
Thanks for the review @Chocobozzz, I've fixed / refactored all your comments.

kimsible:
@Chocobozzz maybe this PR should have been merged before changing localization system, we need to re-pass on all files

rigelk:
@kimsible templates are not touched, just their associated components logic as the I18n service model is replaced.

kimsible:
> @kimsible templates are not touched, just their associated components logic as the I18n service model is replaced.
Thanks @rigelk, ok, in this PR we have some `i18n-title` and `i18n` tags, so we leave them as they are ?

rigelk:
Yes, since templates are not affected.

Chocobozzz:
I'll merge this PR myself.

kimsible:
> I'll merge and fix conflicts of this PR myself.

Thanks @Chocobozzz, I've already merged it but maybe you would like to check if I missed something.

Issue #3043, opened 2020-08-07
Lost scroll position when modal open


kimsible:
This PR https://github.com/Chocobozzz/PeerTube/pull/2962 proposed to disable the scrollbar instead of hide it but it brought one regression, when we open a modal we lose the scroll position : 

![issue-fixed-body gif-2](https://user-images.githubusercontent.com/1877318/89649112-88489500-d8c0-11ea-9961-fb22b89f12ea.gif)
![issue-fixed-body gif-1](https://user-images.githubusercontent.com/1877318/89649115-88e12b80-d8c0-11ea-9044-67e2e6edb49e.gif)

We have 2 solutions for modals that does not impact the new disabling scrollbar UI : 

1) Do not fix the body when modal opened, so the user can still scroll the page
![issue-fixed-body-fixed-proposition](https://user-images.githubusercontent.com/1877318/89649334-df4e6a00-d8c0-11ea-85d1-cf802c56f0a1.gif)

2) Back to hide scroll-bar only when modal opened, the user will see the upload button and search input move to the right
![issue-fixed-body-fixed-proposition-2](https://user-images.githubusercontent.com/1877318/89649690-74e9f980-d8c1-11ea-9660-0f3f61b3448a.gif)

@rigelk @Chocobozzz  what solution do you prefer ? do you have another solution ?
 

EDIT: I myself prefer the first solution, YouTube uses this ways to display modals (not that YouTube is a great example, but for UI/UX it's a good example)

PR #3047, opened 2020-08-07
Avoid lost scroll position when modal open

kimsible:
fixes https://github.com/Chocobozzz/PeerTube/issues/3043


Chocobozzz:
Thanks

Issue #2995, opened 2020-07-22
closing support button window does not start video


test2a:
**What happened?**
pressing support button while video is playing stops the video. closing the support window DOES NOT start the video on its own. it needs to be clicked to play
**What do you expect to happen instead?**
support modal should first pause and when closed, auto start the video 
**Steps to reproduce:**

1.open https://diode.zone/videos/watch/40efc093-4a5d-493b-ab8f-e54dac68a8dd

2.click on support button. modal opens, video pauses

3. close modal.  video does NOT start on its own

**Additional information**
* PeerTube version or URL: https://diode.zone/videos/watch/40efc093-4a5d-493b-ab8f-e54dac68a8dd
* Browser name/version: firefox 78.02
* NodeJS version:

* Link to browser console log if useful:
* Link to server log if useful (journalctl or /var/www/peertube/storage/logs/):

PR #3052, opened 2020-08-08
Resume video playback on close of support modal

amsharma44:
Fix issue [2995](https://github.com/Chocobozzz/PeerTube/issues/2995)


Chocobozzz:
Thanks @amsharma44

Issue #3061, opened 2020-08-10
Broadcast message moves fixed sub-menu


kimsible:
When a broadcast message is displayed, the views with fixed sub-menus (`my-account` and `about`) have these specific unhandled case.

![broadcoast-issue](https://user-images.githubusercontent.com/1877318/89816934-a8d65080-db47-11ea-9ffa-46abb0fe81a7.gif)

Since I've myself proposed this new - experimental feature I don't see a good solution against that but I can try to propose to :
- back to the original menu non-fixed
or
- do not display the broadcast message on these 2 views (But I don't think it is intended to be hidden even on these views) 

Would you have any better ideas ?

PR #3062, opened 2020-08-10
Disable fixed sub-menu with broadcast-message

kimsible:
fixes: https://github.com/Chocobozzz/PeerTube/issues/3061

This PR proposes to _unfixed_ the sub-menu when a broadcast-message is displayed.
It also brings a small refactor to improve the way we handle the offset of the content when the menu is fixed.

broadcoast-issue-resolved


Chocobozzz:
Thanks!

PR #3066, opened 2020-08-11
Add restore scroll position on user-dropdown anchors links and scroll to top on active sub-menu links

kimsible:
Restore anchor scroll position on user-dropdown account settings active links : 
user-dropdown restore scroll position

Scroll to top when sub-menu is fixed on active links : 
sub-menu active link scrollToTop

### DIFF
--- a/client/src/app/shared/shared-main/misc/top-menu-dropdown.component.ts
+++ b/client/src/app/shared/shared-main/misc/top-menu-dropdown.component.ts
@@ -94,6 +94,16 @@ export class TopMenuDropdownComponent implements OnInit, OnDestroy {
     this.isModalOpened = false
   }
 
+  onActiveLinkScrollToTop (link: HTMLAnchorElement) {
+    if (!this.isBroadcastMessageDisplayed && this.router.url.replace(/#.+$/, '').includes(link.getAttribute('href'))) {
+      window.scrollTo({
+        left: 0,
+        top: 0,
+        behavior: 'smooth'
+      })
+    }
+  }
+
   dismissOtherModals () {
     this.modalService.dismissAll()
   }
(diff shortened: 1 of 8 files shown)


client/src/app/shared/shared-main/misc/top-menu-dropdown.component.ts
```diff
@@ -94,6 +94,16 @@ export class TopMenuDropdownComponent implements OnInit, OnDestroy {
     this.isModalOpened = false
   }
 
+  onActiveLinkScrollToTop (link: HTMLAnchorElement) {
+    if (!this.isBroadcastMessageDisplayed && this.router.url.replace(/#.+$/, '').includes(link.getAttribute('href'))) {
+      window.scrollTo({
+        left: 0,
+        top: 0,
+        behavior: 'smooth'
```
kimsible, line 102: @Chocobozzz here the behavior smooth on scroll here seems to me justified because the goal is to back / scroll to top when clicking on the active link. So though the intended action is a scroll and not navigation I think the smooth can be used. What do you think ?

client/src/app/shared/shared-main/misc/top-menu-dropdown.component.ts
```diff
@@ -94,6 +94,16 @@ export class TopMenuDropdownComponent implements OnInit, OnDestroy {
     this.isModalOpened = false
   }
 
+  onActiveLinkScrollToTop (link: HTMLAnchorElement) {
+    if (!this.isBroadcastMessageDisplayed && this.router.url.replace(/#.+$/, '').includes(link.getAttribute('href'))) {
```
Chocobozzz, line 98: I don't think the `.replace` is useful since you use `.includes`

client/src/app/shared/shared-main/misc/top-menu-dropdown.component.ts
Chocobozzz: yes but why don't you use the same behaviour here: https://github.com/Chocobozzz/PeerTube/pull/3066/files#diff-e1e01cbde1b933afb4bd4324841ddcfcR206?

kimsible: here it's more for navigation when you click on a link in the user dropdown, but we can also use the smooth behavior here too, it was obvious to follow what the router does in that case but yes the view does not reload so we can uses smooth too.

Issue #3067, opened 2020-08-11
Muted player on play - iOS / Safari desktop


kimsible:
On Safari desktop or on any iOS browser (iPad for example), the player is set to muted = true when the video starts playing even if local-storage has a different muted value.

![muted-sound-safari-os](https://user-images.githubusercontent.com/1877318/89948331-2ff9f600-dc26-11ea-9281-697b283aed4d.gif)

Chocobozzz:
Do you still have the issue referenced in https://github.com/Chocobozzz/PeerTube/pull/3068#issuecomment-682400441?

kimsible:
@Chocobozzz thanks everything is good ! The only thing that we can see is sometimes when we touch the next button on the player, due to the proximity with the play / pause at the left, the next video may also be paused just after start. But this is another question of player UX/UI optimizations for touch devices =).

PR #3068, opened 2020-08-11
Fix muted sound when play and introduces autoplay on iOS 

kimsible:
fixes : https://github.com/Chocobozzz/PeerTube/issues/3067

### DIFF
--- a/client/src/assets/player/peertube-player-local-storage.ts
+++ b/client/src/assets/player/peertube-player-local-storage.ts
@@ -22,7 +22,8 @@ function getStoredMute () {
   const value = getLocalStorage('mute')
   if (value !== null && value !== undefined) return value === 'true'
 
-  return undefined
+  // By default sound is not muted
+  return false
 }
 
 function getStoredTheater () {
--- a/client/src/assets/player/peertube-player-manager.ts
+++ b/client/src/assets/player/peertube-player-manager.ts
@@ -35,7 +35,7 @@ import {
   VideoJSPluginOptions
 } from './peertube-videojs-typings'
 import { TranslationsManager } from './translations-manager'
-import { buildVideoOrPlaylistEmbed, buildVideoLink, copyToClipboard, getRtcConfig, isSafari } from './utils'
+import { buildVideoOrPlaylistEmbed, buildVideoLink, copyToClipboard, getRtcConfig, isSafari, isIOS } from './utils'
 
 // Change 'Playback Rate' to 'Speed' (smaller for our settings menu)
 (videojs.getComponent('PlaybackRateMenuButton') as any).prototype.controlText_ = 'Speed'
@@ -581,7 +581,7 @@ export class PeertubePlayerManager {
 
     // We have issues with autoplay and Safari.
     // any that tries to play using auto mute seems to work
-    if (isSafari()) return 'any'
+    if (isSafari() && !isIOS()) return 'any'
 
     return 'play'
   }
--- a/client/src/assets/player/peertube-player-manager.ts
+++ b/client/src/assets/player/peertube-player-manager.ts
@@ -35,7 +35,7 @@ import {
   VideoJSPluginOptions
 } from './peertube-videojs-typings'
 import { TranslationsManager } from './translations-manager'
-import { buildVideoOrPlaylistEmbed, buildVideoLink, copyToClipboard, getRtcConfig, isIOS, isSafari } from './utils'
+import { buildVideoOrPlaylistEmbed, buildVideoLink, copyToClipboard, getRtcConfig, isSafari } from './utils'
 
 // Change 'Playback Rate' to 'Speed' (smaller for our settings menu)
 (videojs.getComponent('PlaybackRateMenuButton') as any).prototype.controlText_ = 'Speed'
@@ -578,9 +578,6 @@ export class PeertubePlayerManager {
 
   private static getAutoPlayValue (autoplay: any) {
     if (autoplay !== true) return autoplay
-
-    // Giving up with iOS
-    if (isIOS()) return false
 
     // We have issues with autoplay and Safari.
     // any that tries to play using auto mute seems to work
--- a/client/src/assets/player/peertube-plugin.ts
+++ b/client/src/assets/player/peertube-plugin.ts
@@ -6,7 +6,7 @@ import {
   UserWatching,
   VideoJSCaption
 } from './peertube-videojs-typings'
-import { isMobile, timeToInt, isSafari, isIOS } from './utils'
+import { isMobile, timeToInt, isIOS, isSafari } from './utils'
 import {
   getStoredLastSubtitle,
   getStoredMute,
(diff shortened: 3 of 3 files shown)


Chocobozzz:
iPad should be fixed by https://github.com/Chocobozzz/PeerTube/commit/b6a8cfc5714c93d0ecf6154e5d3a25b33b50b8cd

It's videojs that automatically mutes the player to autoplay the video on safari: https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/assets/player/peertube-player-manager.ts#L587

kimsible:
> iPad should be fixed by [b6a8cfc](https://github.com/Chocobozzz/PeerTube/commit/b6a8cfc5714c93d0ecf6154e5d3a25b33b50b8cd)
> 
> It's videojs that automatically mutes the player to autoplay the video on safari: https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/assets/player/peertube-player-manager.ts#L587

@Chocobozzz the autoplay does not work on iOS plateform and this fix does not impact anything, we have iPad to test and iPad users.

And autoplay still work on safari desktop with this fix.

kimsible:
@Chocobozzz 

> iPad should be fixed by b6a8cfc

This commit includes iOS desktop mode in `isIOS` function, this way we can indeed match all iOS devices and modes.

The last commit re-introduces autoplay and muted value on iOS, mobile or desktop mode on all browsers (Safari, Chrome, Firefox), with the trigged event `play` and autoplay with `this.player.play()` when player is ready. All worked perfectly on my iPad **but for WebTorrent only videos, which not autoplay (properly) and this is an intended behaviour**.

 > It's videojs that automatically mutes the player to autoplay the video on safari: https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/assets/player/peertube-player-manager.ts#L587

I only have Safari on iPad and not on MacOS (since I don't have any Mac) but I have the browser Epiphany (Gnome Web) based on WebKit-GTK which seems to have the same behaviour. If someone have a MacOS with Safari, maybe you would confirm this trick for muted value on play works ? @rigelk  ?

Chocobozzz:
Some thoughts:
 * I cannot reproduce your issue on iOS: in my case videos (HLS/WebTorrent) don't have autoplay (iPhone, iPad desktop)
 * I can reproduce your issue on Safari Desktop with WebTorrent videos
 * We can't autoplay with sound webtorrent videos on these platforms
 * This is a very complex part of PeerTube because of 2 different player engines (HLS and WebTorrent), many web browsers (chrome, firefox, safari etc) and multiple platforms (iOS, Android, Desktop). This also involves E2E tests that are... very capricious
 * IMHO playing videos without sound is not something we want (bad UX) especially when sometimes the video is muted (webtorrent video) and other times not muted (HLS video)
 * With your patch I can autoplay a webtorrent video using Safari Desktop but it's an hack, and we aren't sure how long it will work. Moreover, it "breaks" the player on F5 (the video is stopped by the web browser but we don't have the big player button anymore).

My proposal is to force autoplay deactivation when we detect Safari Desktop, and only enable *real* autoplay (autoplay with sound or no autoplay at all).

kimsible:
> I cannot reproduce your issue on iOS: in my case videos (HLS/WebTorrent) don't have autoplay (iPhone, iPad desktop)

This PR introduces autoplay on iOS devices, I explained it. With the action `this.player.play()` when player is ready and only with HLS. Plus, if autoplay is disabled for iPhone or iPad, PeerTube playlists are unusable, it's really weird to have a portable device unsuable with playlists. Since we have more and more users on iPad (really, lots of creators).

> I can reproduce your issue on Safari Desktop with WebTorrent videos

You mean autoplay or muted sound ?

> With your patch I can autoplay a webtorrent video using Safari Desktop but it's an hack, and we aren't sure how long it will work. Moreover, it "breaks" the player on F5 (the video is stopped by the web browser but we don't have the big player button anymore).

Ok thanks, on iOS the big player button is still there.

> My proposal is to force autoplay deactivation when we detect Safari Desktop, and only enable real autoplay (autoplay with sound or no autoplay at all).

Ok I will exclude Safari from the conditional for Safari : https://github.com/Chocobozzz/PeerTube/blob/cba084246d84ee24a601de4f43f1e7b96e704d35/client/src/assets/player/peertube-plugin.ts#L78

kimsible:
@Chocobozzz  
> I cannot reproduce your issue on iOS: in my case videos (HLS/WebTorrent) don't have autoplay (iPhone, iPad desktop)

Also, the muted sound before this PR was on play only since we blocked autoplay on iOS, maybe the PR name should be renamed

Chocobozzz:
I reimplemented this PR in https://github.com/Chocobozzz/PeerTube/commit/9eccae74c844f09e9e3483db623fea00e0562365

I think this should fix your issues with playlist (tell me if you notice bugs, it's quite difficult to test)), and it's not too hacky

kimsible:
Thanks a lot @Chocobozzz I will test tomorrow with the nightly (is there any playlist hls and webtorrents on peertube2.cpy.re?) and an ipad or in 2 days with my own computer / iPad.

Chocobozzz:
You can try https://peertube2.cpy.re/videos/watch/playlist/73804a40-da9a-40c2-b1eb-2c6d9eec8f0a

The first video is in HLS mode and the second is in webtorrent mode

kimsible:
Thanks,

In this screen record The second video (webtorrent) autoplays when we load the playlist on the first video : 
IMB_TZza5F

In this one, the second video does not autoplay when we load the playlist from browser navigation (position on the second video ended, then prev, the first video autoplays but the second is paused without the big button)
IMB_lfXfKP

In this last one, same behavior then the second one, when we click on playlist link first (HLS) autoplays, second does not (Webtorrent) : 
IMB_R3wdOg

For each autoplay the sound works.

Chocobozzz:
What is this web browser?

kimsible:
> What is this web browser?

Safari

On Opera and Firefox the third test works (second video autoplays), the second test not, first yes.

On Chrome, playlist never next from the end of the first video.

Chocobozzz:
And their versions?

kimsible:
@Chocobozzz 

Here are my last tests.

Safari provided with iPadOS 13.6.1
Chrome 85.04183
Firefox 28.0 (looks already outdated but the last version on my ipad app store)

### E2E playlist 

Safari and Firefox : 
- The first video must be manual played with the big button to start playlist ;
- Sound enabled on play ;
- When playlist started the second video autoplay ;
- On browser navigation first video auto play, **the second one pauses without the big button**.

Chrome :
- The first video must be manual played with the big button to play ;
- Sound enabled on play ;
- When playlist started **the second video does not autoplay and the playlist stays at the end of the first video** ;
- On browser navigation first video auto play, **the second one pauses without the big button**.


### Test playlist

Safari, Firefox and Chrome :
- The first video must be manual played with the big button to start playlist ;
- Sound enabled on play ;
- When playlist started all videos autoplay ;
- On browser navigation *all videos pause without the big button**.

Chocobozzz:
> On browser navigation all videos pause without the big button.

Should be fixed by https://github.com/Chocobozzz/PeerTube/commit/1a568b6fcae4cd64458bb2d53543bffd1a3baabc

kimsible:
> > On browser navigation all videos pause without the big button.
> 
> Should be fixed by [1a568b6](https://github.com/Chocobozzz/PeerTube/commit/1a568b6fcae4cd64458bb2d53543bffd1a3baabc)

I've just tested again with the last nightly, and all videos still lause without the big button on browser navigation.

PR #3090, opened 2020-08-15
removed minimum width css to fit mobile devices with lesser width

test2a:
this width caused this page to scroll because the minimum width was set to 330px and phones have less width than that. i have checked the same on peertube2.cpy.re using firefox inspect element responsive design mode.
This is my first pull request, please be gentle ;-)

### DIFF
--- a/client/src/app/shared/shared-instance/instance-features-table.component.scss
+++ b/client/src/app/shared/shared-instance/instance-features-table.component.scss
@@ -7,9 +7,7 @@ table {
 
   .label,
   .sub-label {
-  
-
-    &.label {
+      &.label {
       font-weight: $font-semibold;
     }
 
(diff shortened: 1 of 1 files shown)


test2a:
fixes https://github.com/Chocobozzz/PeerTube/pull/3004#issuecomment-663616886

client/src/app/shared/shared-instance/instance-features-table.component.scss
```diff
@@ -7,7 +7,7 @@ table {
 
   .label,
   .sub-label {
-    min-width: 330px;
+
```
kimsible, line 10: The lines 10 and 11 should be deleted.

test2a: @kimsible hey. do i create a new pull request  and close this one? i removed the spaces in 10 and 11

test2a: @kimsible yeah. so the second change shows up here so i think i dont have to create a new pull request.

Chocobozzz:
Thanks

PR #3097, opened 2020-08-17
import media tooltip improvement

test2a:
tooltips on import pages showed "import mp4 file" when we can clearly import many more formats. This must have been put when we would have been only able to import mp4 files. Anyways, i have changed mp4 to say "media file" as we can import audios as well

### DIFF
--- a/client/src/app/+videos/+video-edit/video-add-components/video-import-torrent.component.html
+++ b/client/src/app/+videos/+video-edit/video-add-components/video-import-torrent.component.html
@@ -14,7 +14,7 @@
       <my-help>
         <ng-template ptTemplate="customHtml">
           <ng-container i18n>
-            You can import any torrent file that points to a mp4 file.
+            You can import any torrent file that points to a media file.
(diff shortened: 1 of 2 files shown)


test2a:
(.webm, .ogv, .mp4, .mov, .avi, .flv, .mkv, .m4v, null, .mp3, .ogg, .wma, .wav, .flac)
these are the formats currently shown in the uploads page tooltip. Found out i am unable to upload .aac file
https://filesamples.com/formats/aac
it just checks the format against this list i suppose and rejects the aac format i think. 
same for .vob and .mpeg
https://filesamples.com/categories/video

is this a transcoding limitation on these formats or have we restricted them for some reason?

rigelk:
@test2a this is not a transcoding limitation, we have restricted them because so far the list of extensions was not a list but a map to their respective mimetypes, which some extensions share, hence the choice of one to respect map uniqueness. I should check again, @chocobozzz planned to refactor that part.

If you know the mimtype of a .vob, please add it in the constants file.

test2a:
@rigelk  according to this https://filesamples.com/formats/vob
.vob has mimetype as video/mpeg which could solve both the formats as they share the same mimetype

Issue #3084, opened 2020-08-14
Allow multiple section of a same video in playlists


Chocobozzz:
Allow users to put multiple sections (`startAt`/`stopAt`) of a video in a playlist. For now we have a unique key based on the videoId, so we can't.

We'll also have to update the `videoId` url param in watch page to set the playlist position. We'll use `playlistPosition` instead (as in the embed).

PR #3101, opened 2020-08-18
Same video multiple times in the same playlist

Chocobozzz:
See https://github.com/Chocobozzz/PeerTube/issues/3084

Remove constraints so a playlist can have multiple instances of a video.

Updated the video add to playlist dropdown:
  * When a video is added in a playlist, we display an `+` icon
  * On click, it opens the *advanced settings* pane where the user can update the timestamps of the video
  * The user can also add more instances of the video in the playlist
  * When the user opens the dropdown, the *advanced settings* pane is automatically opened if the playlist contains several instances of the video **or** if the video was added to the playlist with custom timestamps

screen_2020-08-18-15:51:38

screen_2020-08-18-15:51:27


Chocobozzz:
Not a WIP anymore, just a small detail to add but the main work is there.

PR #3105, opened 2020-08-19
Hotfix/z index

Chocobozzz:


client/src/sass/include/_variables.scss
```diff
@@ -127,9 +127,9 @@ $zindex: (
   miniature    :    10,
   privacymsg   :    20,
   typeahead    :    30,
-  header       :  1000,
-  menu         : 11000,
   dropdown     : 12000,
+  header       : 12500,
+  menu         : 12600,
   popover      : 13000,
```
kimsible, line 133: popover, tooltip and help-popover should also be smaller than header and menu ?

kimsible: Also for z-index search instructions :
Capture d’écran du 2020-08-20 23-38-40

z-index value for menu should be smaller than header and the content overlay in https://github.com/Chocobozzz/PeerTube/blob/hotfix/z-index/client/src/sass/bootstrap.scss#L179 should have `z-index: z('menu') - 1;` (instead of `z('header') -1`)

client/src/sass/include/_variables.scss
```diff
@@ -124,19 +124,19 @@ $variables: (
...
-  modal        : 16000,
-  help-popover : 17000,
-  notification : 18000,
-  hotkeys      : 19000
+  miniature       :    10,
+  privacymsg      :    20,
+  dropdown        : 12000,
+  menu            : 12600,
+  search-typeahead: 12650,
+  header          : 12700,
```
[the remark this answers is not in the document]
Chocobozzz: The header has to have a bigger z-index so the typeahead is on top of the menu. I'll create another var for sub menus at `12500`

PR #3035, opened 2020-08-04
add `ng-select` for templatable select options

rigelk:
Having a slightly more customizable select component is required here to display both the label of an option and its description. [`ng-select`](https://github.com/ng-select/ng-select) component, like most select components around, features lazy-loading of remote data - which could prove useful for listing channels in the `video-edit` view in the future.

In this PR I focus on the upload/edit view, but other select usages could use option templates elsewhere.

Screenshot_2020-08-05 Upload a video - PeerTube
Screenshot_2020-08-05 Edit a video - PeerTube

fixes #3015

PR #3106, opened 2020-08-19
privacy label improved to make more sense

test2a:
improved the wording of "privacy" in upload as well as in playlists to make a little bit more sense. Attached is screenshots of new and old privacy labels. 
I have taken help of vimeo in how they make use of their text.
https://vimeo.zendesk.com/hc/en-us/articles/224817847-Privacy-settings-overview
This also helped because the first upload page only said "private" "unlisted" "public" "internal" and in the actual description page these words were absent. 
Screenshot_20200819_193125

Screenshot_20200819_193453

### DIFF
--- a/client/src/app/+my-account/my-account-video-playlists/my-account-video-playlist-edit.component.html
+++ b/client/src/app/+my-account/my-account-video-playlists/my-account-video-playlist-edit.component.html
@@ -54,7 +54,7 @@
 
       <div class="col-md-12 col-xl-6">
         <div class="form-group">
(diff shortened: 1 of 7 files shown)


test2a:
let me know if there is any other page i might have missed, that is in case you allow this change ;-)

rigelk:
hi @test2a, and thanks for your PR.

Unfortunately this is already solved by #3035.

test2a:
yeah saw that although i would still suggest the text improvements

PR #3128, opened 2020-08-27
removed pinch and zoom on touch devices

test2a:
added "initial-scale=1".  Currently the user is able to pinch and zoom on all the pages of the website when in mobile mode. this makes the application look out of place. preventing pinch and zoom makes it appear as if the app is native, like many webapps do

### DIFF
--- a/client/src/index.html
+++ b/client/src/index.html
@@ -2,7 +2,7 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0">
 
     <meta name="theme-color" content="#fff" />
     <meta property="og:platform" content="PeerTube" />
--- a/client/src/standalone/videos/embed.html
+++ b/client/src/standalone/videos/embed.html
@@ -4,7 +4,7 @@
     <title>PeerTube</title>
 
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0">
     <meta name="robots" content="noindex">
     <meta property="og:platform" content="PeerTube" />
 
--- a/client/src/app/+videos/+video-watch/video-watch.component.scss
+++ b/client/src/app/+videos/+video-watch/video-watch.component.scss
@@ -578,9 +578,6 @@ my-video-comments {
   .privacy-concerns {
     width: 100%;
 
-    strong {
-      display: none;
-    }
   }
 }
 


client/src/index.html
```diff
@@ -2,7 +2,7 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0">
```
kimsible, line 5: Thanks !

`user-scalable=0` does not work on iOS, only Javascript can prevent scaling :  https://stackoverflow.com/questions/37808180/disable-viewport-zooming-ios-10-safari

It's working on Android but I suggest we only remove pinch and zoom for the mobile mode and not for the desktop mode, otherwise we may break the touch experience on some devices (especially phones that needs zoom sometimes when we uses the desktop mode).

client/src/standalone/videos/embed.html
```diff
@@ -4,7 +4,7 @@
     <title>PeerTube</title>
 
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0">
```
kimsible, line 7: Same here.

Chocobozzz:
Hello,

I don't think disabling zoom on touch devices is a good idea: https://adrianroselli.com/2015/10/dont-disable-zoom.html

test2a:
@kimsible thanks. i did some more testing. 
@Chocobozzz hi. i used the following two links http://codepen.io/aardrian/full/ojBpjw and http://codepen.io/aardrian/full/dYNJOV which only have "width=device-width, initial-scale=1, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no" setting on one and not on the other. on my android phone, in mobile view the disable zoom worked but if i enabled the display desktop site, i was able to zoom easily. @kimsible i think at least on android, show desktop site disables this meta scalable tag and that should be enough for desktop users? 
i did some tests, 
https://www.amazon.in/ (content="width=device-width, maximum-scale=1, minimum-scale=1, initial-scale=1, user-scalable=no, shrink-to-fit=no")
https://m.youtube.com/ (content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover")
https://www.dailymotion.com/in (width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no)
https://m.facebook.com (content="user-scalable=no,initial-scale=1.0001,maximum-scale=1.0001,viewport-fit=cover")

All these major websites that people browse on their phones have this setting on for making sure the content does not stretch or allow zoom because the user thinks its not native. 

I do not have an apple device at hand but according to this
https://stackoverflow.com/questions/57821757/how-can-i-disable-zoom-and-zoom-out-for-safari-ios-13-in-my-webpage

"meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0"> " this code works now thank you so much – Kebir Asim Sep 6 '19 at 12:28

test2a:
ok. @kimsible did more testing. i used a tablet, mi pad 7 inch to access the links above, one with zoom disabled. i was able to NOT zoom as expected but if i switched to desktop mode, that functionality was gone and i was able to zoom correctly

Chocobozzz:
Sorry but at the moment we don't want to remove ability to disable zoom on peertube client.

PR #3138, opened 2020-09-04
Add docker-compose and Dockerfile

AlexHladin:
This PR adds the possibility to run all services inside Docker containers and combine them via docker-compose.
You can simply run
```
docker-compose up
```
And all infrastructure will run in a few moments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM node:12.18.3-buster-slim
+
+# Install dependencies
+RUN apt update \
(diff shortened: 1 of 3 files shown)


rigelk:
hi @AlexHladin and thanks for your PR. However we already have multiple Dockerfiles and a Docker Compose configuration available. You can read more about it in [the docker documentation](https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/docker.md).

AlexHladin:
@rigelk it was docker-compose configuration for development, not for production running.

PR #3172, opened 2020-09-22
Fix boolean types, add missing downloadEnabled on openapi.yaml 

Arman92:
Fixed boolean types which were wrongly set as string, added missing downloadEnabled for `/videos/upload` and `/videos/imports` on openapi.yaml


Chocobozzz:
Thanks @Arman92

PR #3179, opened 2020-09-25
Document dependencies version for Debian-based distributions

frankstrater:
I have only added this for Debian / Ubuntu and derivatives, because I have only tested this for this family. I'm not sure if the commands and package names are the same for other Linux distributions or OSes.


rigelk:
@frankstrater are you still planning to submit this PR? (it currently is classified as a draft)

Also, I fear the whole part teaching how to test if a command exists in the PATH is just overly verbose, or at least sysadmin 101…

frankstrater:
@rigelk No, not in this way, but I do think all the versions of the dependencies should be mentioned in dependencies.md like they are in https://github.com/Chocobozzz/PeerTube#dependencies

PR #3190, opened 2020-10-07
Fedora guide updgrade

frankstrater:
I've installed PeerTube on Fedora Workstation 32 and ran in some issues when following the dependencies and production guides. These changes address the missing parts.


frankstrater:
I'm reconsidering the nginx configuation, so I'll close this pull request for now.

PR #3198, opened 2020-10-16
Fix transcoding errors in readonly docker containers

gramakri:
ffmpeg seems to create some temporary files in the cwd. When PeerTube
is run in a read-only docker container, this causes all transcoding
to fail. As a workaround, we set the cwd to the configured tmp dir.

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     "@types/config": "^0.0.36",
     "@types/express": "^4.0.35",
     "@types/express-rate-limit": "^5.0.0",
-    "@types/fluent-ffmpeg": "^2.1.8",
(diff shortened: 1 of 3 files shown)


gramakri:
The exact error is in the forum post here - https://forum.cloudron.io/topic/3414/transcode-after-import-fails (EROFS)

gramakri:
I think this fails because we have to update the `@types/fluent-ffmpeg` to 2.1.14 (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48421)

Chocobozzz:
Thanks @gramakri

PR #3201, opened 2020-10-18
Add possibility to share origin URL to video if it's not local

jeena:
It's very difficult to share the original URL for a video when you watch it on your
own instance while it's not from it but from a different instance. Right now when
you use the share UI you only can share a link so a person can watch it on your
own instance, this patch adds a checkbox to the share UI to share the origin URL
instead if needed.

origin-url

### DIFF
--- a/client/src/app/shared/shared-share-modal/video-share.component.html
+++ b/client/src/app/shared/shared-share-modal/video-share.component.html
@@ -173,6 +173,13 @@
                 i18n-labelText labelText="Loop"
               ></my-peertube-checkbox>
             </div>
+
+            <div *ngIf="!isLocalVideo() && !isVideoInEmbedTab()" class="form-group">
+              <my-peertube-checkbox
+                inputName="originUrl" [(ngModel)]="customizations.originUrl"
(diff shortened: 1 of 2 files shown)


client/src/app/shared/shared-share-modal/video-share.component.html
```diff
@@ -173,6 +173,13 @@ <h4 i18n class="modal-title">Share</h4>
                 i18n-labelText labelText="Loop"
               ></my-peertube-checkbox>
             </div>
+
+            <div *ngIf="!isLocalVideo() && !isVideoInEmbedTab()" class="form-group">
+              <my-peertube-checkbox
+                inputName="originUrl" [(ngModel)]="customizations.originUrl"
+                i18n-labelText labelText="Origin URL"
```
Chocobozzz, line 180: Maybe `Use origin instance URL` instead?

jeena: Good point, I changed it.

Chocobozzz:
Thanks @jeena

PR #3247, opened 2020-11-03
Update guide: plugin development

kontrollanten:


Chocobozzz:
Why do you need `npm link`?

kontrollanten:
I'm sorry. I was running `node dist/server/tools/peertube.js plugins install --path /home/kontrollanten/Repos/peertube-plugin-s3-storage` after `npm run dev` and got a "module not found" error, but now I found that it was the "dist/server/tools/peertube.js" module that wasn't found.

It seems like `npm run dev` removes the server tools, it would be great if we didn't have to rerun `npm run setup:cli` again after starting the dev task. But that's another story.

Issue #3245, opened 2020-11-02
markdown emoji list- one long list


test2a:
![Screenshot_20201102_153421](https://user-images.githubusercontent.com/39449563/97855596-f7e3ea00-1d20-11eb-95bc-c52f9f6f11b7.png)
markdown emoji list button "view complete list" shows the entire emojis in one LONG list. We can easily have 2-3 columns and the vertical scroll would be greatly reduced. the current one makes going back up very difficult as there are 151 entries. 2 columns would get us 1/3 scroll

**Additional information**
* PeerTube version and URL:
* Browser name and version:
* NodeJS version:

* Link to browser console log if useful:
* Link to server log if useful (`journalctl` or `/var/www/peertube/storage/logs/`):

rigelk:
@kimsible would you be up for that?

kimsible:
@rigelk thanks for notifying me.

@test2a thanks for your this remark ! 

Sure I'm up for that =).

PR #3253, opened 2020-11-05
Improve markdown-it emoji list column display

kimsible:
Fix https://github.com/Chocobozzz/PeerTube/issues/3245

- Use CSS flex instead of a classic table to extend the columns on the full modal width and make it responsive ;
- Center the columns ;
- Populate the the markdown-it emoji list from object to array to avoid an alphabetical order of the keys.

emoji-list
image

### DIFF
--- a/client/src/app/+videos/+video-watch/comment/video-comment-add.component.scss
+++ b/client/src/app/+videos/+video-watch/comment/video-comment-add.component.scss
@@ -107,7 +107,8 @@ form {
   .emoji-flex-item {
     text-align: left;
     margin: auto;
-    width: 227px;
+    min-width: 227px;
+    flex: 1;
 
     code {
       display: inline-block;
--- a/client/src/app/+videos/+video-watch/comment/video-comment-add.component.scss
+++ b/client/src/app/+videos/+video-watch/comment/video-comment-add.component.scss
@@ -99,13 +99,21 @@ form {
   }
 }
 
-.table-emoji {
-  td {
-    &:first-child {
-      width: 20px;
-    }
+.emoji-flex {
+  display: flex;
+  flex-flow: row wrap;
+  align-items: center;
 
-    vertical-align: top;
+  .emoji-flex-item {
+    text-align: left;
+    margin: auto;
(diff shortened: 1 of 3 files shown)


client/src/app/+videos/+video-watch/comment/video-comment-add.component.scss
```diff
@@ -99,13 +99,21 @@ form {
...
-      width: 20px;
+.emoji-flex {
+  display: flex;
+  flex-flow: row wrap;
+  align-items: center;
+
+  .emoji-flex-item {
+    text-align: left;
+    margin: auto;
+    width: 227px;
```
test2a, line 110: here @kimsible can i suggest using % instead of fixing it at 227px? wide screens, narrow screens throw a fit and this

kimsible: Thanks @test2a for your comment. Fixing it at 227px allow to fit the longest markup (stuck_out_tongue...) What percent value would you use ?

test2a: how about this https://jsfiddle.net/kdr9b1vg/
i am not at all an expert on anything, :worried:
i checked this on my phone using codepen and it worked somewhat.

kimsible: thanks @test2a for the snippet, I understand what you mean now.
In my case we need to update the width flex-item if any new markup is longer than current longest one.
The `margin: auto` on flex-item replaces the flex percent width I think. 
Also I'm not an expert too btw. :) I'll check see what I can do.

rigelk:
@kimsible is this ready for review?

kimsible:
> @kimsible is this ready for review?

I think yes, because the flex percent solution looks to not be adapted to the longest markup names of emojis.

rigelk (approved):

rigelk:
@kimsible thanks for your quick response to the issue and sensible fix! Also thank you @test2a for your review.

Issue #3260, opened 2020-11-06
When editing a comment break line are displayed as <br/> html.


kimsible:
When we delete and re-draft a comment, the break lines are still displayed as <br /> html element.

![Peek 06-11-2020 16-10](https://user-images.githubusercontent.com/1877318/98382041-f3823e80-204a-11eb-9646-cd87891fcebd.gif)

PR #3261, opened 2020-11-06
Correct break line display for re-draft comments

kimsible:
Fix https://github.com/Chocobozzz/PeerTube/issues/3260


rigelk (approved):

rigelk:
@kimsible thanks for the quick PR!

kimsible:
@rigelk btw after testing lists with comment, it looks like somewhere we replace break line to <br /> before to store any data. So, it breaks markdown lists.

image

Do we have to disable support of lists in comments ?

rigelk:
> Do we have to disable support of lists in comments ?

We don't have to disable it, why do you ask?

PR #3266, opened 2020-11-07
adjust registration form width on mobile

test2a:
fixed box size was forcing more than 100% screen on mobile devices. this should help. i could be wrong though

### DIFF
--- a/client/src/app/+signup/+register/register.component.scss
+++ b/client/src/app/+signup/+register/register.component.scss
@@ -63,8 +63,8 @@
 }
 
 input:not([type=submit]) {
-  @include peertube-input-text(100%);
-
+  @include peertube-input-text(400px);
+width:100%;
   display: block;
 
   &#username,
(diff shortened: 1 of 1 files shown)


test2a:
Screen Shot 2020-11-07 at 14 52 06

test2a:
@rigelk can you please check if this actually works or  am i just imagining things here?

rigelk:
@test2a it works at that screen size, but then it changes it for _all display sizes too_…

test2a:
@rigelk in that case i think i will switch this over to draft because it would need some time

test2a:
@rigelk would this work now?

rigelk:
@test2a no. specifically, you need [media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries) as is done [here](https://github.com/Chocobozzz/PeerTube//blob/583eb04b541175035d6d452ca626a96ebf2b7437/client/src/app/app.component.scss#L73).

rigelk:
@test2a please open a new PR once you are done with your implementation

PR #3284, opened 2020-11-10
Dependencies versions

frankstrater:
## Description

The PeerTube dependencies guide doesn't check versions after install and this leads to non-descriptive errors post-install if the dependencies versions were wrong.

There have been countless times has there been frustration and an issue posted in the PeerTube community, because for example a wrong NodeJS version was already installed on the target server.

## Solution

Checking for installed dependencies versions is the responsibility of the one who's installing, but the minimal versions of the dependencies should be referenced in the dependencies guide. These minimal versions are mentioned in the main readme, so the simplest addition to the dependencies guide is a link to that content. 

See also the issue on this project repo: https://github.com/beeldengeluid/extending-peertube/issues/8

### DIFF
--- a/support/doc/dependencies.md
+++ b/support/doc/dependencies.md
@@ -15,7 +15,7 @@
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-Make sure you install the correct versions of the dependencies as mentioned in the [PeerTube README.md](https://github.com/Chocobozzz/PeerTube#dependencies)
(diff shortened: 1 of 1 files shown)


rigelk:
@frankstrater I'll merge manually with some adaptations. In the future, please allow maintainers to push to your remote :slightly_smiling_face:

rigelk:
I adapted it in 867ed6514. Thanks for the idea @frankstrater!

PR #3288, opened 2020-11-11
fix link in dependencies version guide

frankstrater:
@rigelk Not sure about the absolute url (master or develop branch), so adapt at will


rigelk:
@frankstrater why do you need `sudo`? please keep the line reference.

frankstrater:
@rigelk (as commented previously) Without it I get 'Executable not found` for each engine, which seems logical to me (on Fedora 33 and not elevated to sudo) as they should not be accessible to any user...

rigelk (approved):

rigelk:
thanks @frankstrater!

PR #3289, opened 2020-11-11
remove duplicate file checks in video model

Arman92:
## Description

video file picking was duplicated, I removed the duplicate code!

## Has this been tested?


rigelk:
@Arman92 nice catch, thanks for the PR!

Issue #3292, opened 2020-11-11
Client overrides (logo, favicon, PWA icons) does't work behind Nginx


kimsible:
Client overrides logo, favicon, PWA icons seem to not work behind Nginx

https://github.com/Chocobozzz/PeerTube/pull/2897#issuecomment-672963723

kimsible:
After testing the new docker config with nginx it is caused by the assets optimization of Nginx that does not consider client-overrides : https://github.com/Chocobozzz/PeerTube/blob/b8fa24491c37fd7c0cca562a9469dfb254171919/support/nginx/peertube#L61

PR #3297, opened 2020-11-12
add client overrides to nginx configuration

kimsible:
## Description
In production client-overrides don't work with nginx because client assets were bypassed without considering client-overrides handled in PeerTube. 

## Related issues
- Fix https://github.com/Chocobozzz/PeerTube/issues/3292
- Also fixes a the doc for a docker-compose command with multiples compose files.

## Has this been tested?
I tested it with production scripts, it may need more tests ? @rigelk ?

### DIFF
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -58,18 +58,8 @@ server {
   }
 
   # Bypass PeerTube for performance reasons. Could be removed
-  # Must be consistent with static client-overrides list in /server/controllers/client.ts
-  location ~ ^/client/(assets/images/(logo\.svg|favicon\.png)) {
-    add_header Cache-Control "public, max-age=31536000, immutable";
-
-    root /var/www/peertube;
-
-    try_files /storage/client-overrides/$1 /peertube-latest/client/dist/$1 $uri;
-  }
-
-  # Bypass PeerTube for performance reasons. Could be removed
-  # Must be consistent with static client-overrides list in /server/controllers/client.ts
-  location ~ ^/client/(assets/images/icons/icon-(36x36|48x48|72x72|96x96|144x144|192x192|512x512)\.png)$ {
+  # Should be consistent with client-overrides assets list in /server/controllers/client.ts
+  location ~ ^/client/(assets/images/(icons/icon-36x36\.png|icons/icon-48x48\.png|icons/icon-72x72\.png|icons/icon-96x96\.png|icons/icon-144x144\.png|icons/icon-192x192\.png|icons/icon-512x512\.png|logo\.svg|favicon\.png))$ {
     add_header Cache-Control "public, max-age=31536000, immutable";
 
     root /var/www/peertube;
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -62,9 +62,9 @@ server {
   location ~ ^/client/(assets/images/(icons/icon-36x36\.png|icons/icon-48x48\.png|icons/icon-72x72\.png|icons/icon-96x96\.png|icons/icon-144x144\.png|icons/icon-192x192\.png|icons/icon-512x512\.png|logo\.svg|favicon\.png))$ {
     add_header Cache-Control "public, max-age=31536000, immutable";
 
-    root /var/www/peertube;
+    root /var/www/peertube/storage/client-overrides;
 
-    try_files /storage/client-overrides/$1 /peertube-latest/client/dist/$1 $uri;
+    try_files /$1 $uri;
   }
 
   # Bypass PeerTube for performance reasons. Could be removed
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -58,6 +58,26 @@ server {
   }
 
   # Bypass PeerTube for performance reasons. Could be removed
+  # Must be consistent with static client-overrides list in /server/controllers/client.ts
+  location ~ ^/client/(assets/images/(logo\.svg|favicon\.png)) {
+    add_header Cache-Control "public, max-age=31536000, immutable";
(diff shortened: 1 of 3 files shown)


kimsible:
Hi @JohnXLivingston, this configuration of nginx works for you ?

support/nginx/peertube
```diff
@@ -57,6 +57,26 @@ server {
     root /var/www/certbot;
   }
 
+  # Bypass PeerTube for performance reasons. Could be removed
+  # Must be consistent with static client-overrides list in /server/controllers/client.ts
+  location ~ ^/client/(assets/images/(logo\.svg|favicon\.png)) {
```
rigelk, line 62:
```suggestion
  location ~ ^/client/assets/images/(logo\.svg|favicon\.png)$ {
```

support/nginx/peertube
```diff
@@ -57,6 +57,26 @@ server {
...
+    add_header Cache-Control "public, max-age=31536000, immutable";
+
+    root /var/www/peertube;
+
+    try_files /storage/client-overrides/$1 /peertube-latest/client/dist/$1 $uri;
+  }
+
+  # Bypass PeerTube for performance reasons. Could be removed
+  # Must be consistent with static client-overrides list in /server/controllers/client.ts
+  location ~ ^/client/(assets/images/icons/icon-(36x36|48x48|72x72|96x96|144x144|192x192|512x512)\.png)$ {
```
rigelk, line 72:
```suggestion
  location ~ ^/client/assets/images/icons/icon-(36x36|48x48|72x72|96x96|144x144|192x192|512x512)\.png$ {
```

rigelk (changes requested):
Why not merge the two with:

```
location ~ ^/client/assets/images/(logo\.svg|favicon\.png|icons/icon-(36x36|48x48|72x72|96x96|144x144|192x192|512x512)\.png)$
```

support/nginx/peertube
kimsible: @rigelk are your sure your suggestion would also catch `logo.svg?569beefcd2eea8ed70010c70310317f33b5fade2` ?

support/nginx/peertube
kimsible: same here about the parenthesis from `assets`

support/nginx/peertube
```diff
@@ -57,6 +57,26 @@ server {
...
+    add_header Cache-Control "public, max-age=31536000, immutable";
+
+    root /var/www/peertube;
+
+    try_files /storage/client-overrides/$1 /peertube-latest/client/dist/$1 $uri;
```
kimsible, line 77: same here about the parenthesis from `assets`

support/nginx/peertube
rigelk: Ah I see, my bad, I forgot the regex group was used

kimsible: btw it's working event with the `$`, nginx seems to not see query strings at this point.

kimsible:
@rigelk I've merged the rules with more explicit file names in the regexp, it's long but better, I think, to maintain.

JohnXLivingston:
> Hi @JohnXLivingston, this configuration of nginx works for you ?

It works... But it is no more dynamic... If files don't exist, it fails and there are missing icons. Or I'm I missing something? (I don't really know nginx syntax)

kimsible:
Thanks @JohnXLivingston 

> It works... But it is no more dynamic..

Sorry I don't see what you mean by "dynamic", you mean nginx ? 

> If files don't exist, it fails and there are missing icons. Or I'm I missing something? (I don't really know nginx syntax)

What kind of failure ? The icons should not be missing if the dist client assets exists, would make a screenshot or a copy of your error ?

JohnXLivingston:
> What kind of failure ? The icons should not be missing if the dist client directory exists, would you make a screenshot or a copy of your error ?

As I said, I'm not familiar with nginx configuration, perhaps I'm wrong. What happens if the try_files directive don't find the file? If it continues to the next «location ~»? In this case, my bad, it's ok.

support/nginx/peertube
```diff
@@ -57,6 +57,16 @@ server {
...
   }
 
+  # Bypass PeerTube for performance reasons. Could be removed
+  # Should be consistent with client-overrides assets list in /server/controllers/client.ts
+  location ~ ^/client/(assets/images/(icons/icon-36x36\.png|icons/icon-48x48\.png|icons/icon-72x72\.png|icons/icon-96x96\.png|icons/icon-144x144\.png|icons/icon-192x192\.png|icons/icon-512x512\.png|logo\.svg|favicon\.png))$ {
+    add_header Cache-Control "public, max-age=31536000, immutable";
+
+    root /var/www/peertube;
+
+    try_files /storage/client-overrides/$1 /peertube-latest/client/dist/$1 $uri;
```
rigelk, line 67:
```suggestion
    root /var/www/peertube/storage/client-overrides;

    try_files /$1 $uri;
```

The next location block to match will be the client assets location block, which already does this.

rigelk (approved):

rigelk:
@kimsible thanks a lot for spotting and fixing the issue!

JohnXLivingston:
Are the config key storage.client_overrides, the constant CLIENT_OVERRIDES_DIR and the associated code in server/controllers/client.ts still needed?

kimsible:
@JohnXLivingston 

> Are the config key storage.client_overrides, the constant CLIENT_OVERRIDES_DIR and the associated code in server/controllers/client.ts still needed?

Yes, as not everybody uses nginx as a reverse proxy.

PR #3313, opened 2020-11-16
refresh nginx config and optimize delivery

rigelk:
## Description

This PR refactors the Nginx configuration for the following points:
- update tls version to include 1.3 by default. so far it was not included by default to make room for previous versions of Nginx, but since 2018 Debian stable has included Nginx in version 1.14.1, and tls 1.3 is available since Nginx 1.13.0.
- clearly indicate that new minimum required version.
- update outdated ssl_ciphers to remove cipher required to support android 4.4, since that version is unsupported since March 2020.
- reordered configuration in sections for easier maintenance: performance optimizations are separated from the vital application/websocket parts.
- move parts that always require manual configuration at the top: peertube host and server name, use server_name 
- move peertube host to a more flexible upstream block: it allows to configure it in one place instead of 3, and is future-proof regarding load-balancing.
- simplified port 80 block: Let’s Encrypt supports 301 redirects.
- group certificate-related config together.
- remove reslover config: it defaults to /etc/resolv.conf which is more than enough.
- align values with their neighbors for easier reading
- always specify units
- always specify default values when they differ from the values set
- use ’m’ for minutes, ’M’ for megabytes
- add consensual optimizations wrt file serving:
  - add timeout optimizations
  - add file descriptor cache optimizations
  - enable sendfile with chunk size > rate limit
  - enable threading
  - tcp optimizations
  - point to further, more system-specific optimizations in the section description
- CDN configuration reduced to one line change

## Has this been tested?

I checked the configuration validates/passes the nginx config checker.

I checked sections reordering doesn’t impact response (locations match not in order but in best match order, and we don’t use return in the server context, so this doesn’t change the current behaviour).

I checked the configuration is working via the docker-compose setup and the production setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,10 @@ Be part of a network of multiple small federated, interoperable video hosting pr
   <a href="https://automate.browserstack.com/public-build/T2g0SVQyMTJtYVRWaEZQTE9LdU9UU1ZDSkY5dTJ6dGlNRUVIajd4QlIwST0tLVdTQ1NUZ0pUTzRZT3lyZzkvWGNSZ1E9PQ==--278d3c68ae0827ce57e245e9f5a3b8c35d6e2278%">
     <img src="https://automate.browserstack.com/badge.svg?badge_key=T2g0SVQyMTJtYVRWaEZQTE9LdU9UU1ZDSkY5dTJ6dGlNRUVIajd4QlIwST0tLVdTQ1NUZ0pUTzRZT3lyZzkvWGNSZ1E9PQ==--278d3c68ae0827ce57e245e9f5a3b8c35d6e2278%"/>
   </a>
+
+  <a href="https://weblate.framasoft.org/projects/peertube/angular/">
+    <img src="https://weblate.framasoft.org/widgets/peertube/-/angular/svg-badge.svg"/>
+  </a>
 </p>
 
 <p align="center">
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -1,3 +1,5 @@
+# Minimum Nginx version required:  1.13.0 (released Apr 25, 2017)
+
 # Uncomment in production to redirect HTTP to HTTPS. Leave commented for docker-compose.
 #server {
 #  listen 80;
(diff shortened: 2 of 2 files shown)


support/nginx/peertube
```diff
@@ -1,202 +1,239 @@
...
+
+  ##
+  # Security hardening (as of Nov 15, 2020)
+  ##
 
-  # Security hardening (as of 11/02/2018)
-  ssl_protocols TLSv1.2; # TLSv1.3, TLSv1.2 if nginx >= 1.13.0
+  ssl_protocols             TLSv1.2 TLSv1.3;
   ssl_prefer_server_ciphers on;
-  # Remove ECDHE-RSA-AES256-SHA if you don't want compatibility with Android 4
```
Chocobozzz, line 26: So you removed `ECDHE-RSA-AES256-SHA` by default?

support/nginx/peertube
```diff
@@ -1,202 +1,239 @@
+# Minimum Nginx version required:  1.13.0 (released Apr 25, 2017)
+
 # Uncomment in production to redirect HTTP to HTTPS. Leave commented for docker-compose.
 #server {
 #  listen 80;
 #  listen [::]:80;
 #  server_name ${WEBSERVER_HOST};
-#
-#  location /.well-known/acme-challenge/ {
```
Chocobozzz, line 7: I think we need this block when we fetch our first cert if we don't use the standalone webserver of certbot

README.md
```diff
@@ -47,8 +47,8 @@ Be part of a network of multiple small federated, interoperable video hosting pr
     <img src="https://david-dm.org/Chocobozzz/PeerTube/dev-status.svg?path=client" alt="devDependency Status" />
   </a>
 
-  <a href="https://automate.browserstack.com/public-build/T2g0SVQyMTJtYVRWaEZQTE9LdU9UU1ZDSkY5dTJ6dGlNRUVIajd4QlIwST0tLVdTQ1NUZ0pUTzRZT3lyZzkvWGNSZ1E9PQ==--278d3c68ae0827ce57e245e9f5a3b8c35d6e2278%">
-    <img src="https://automate.browserstack.com/badge.svg?badge_key=T2g0SVQyMTJtYVRWaEZQTE9LdU9UU1ZDSkY5dTJ6dGlNRUVIajd4QlIwST0tLVdTQ1NUZ0pUTzRZT3lyZzkvWGNSZ1E9PQ==--278d3c68ae0827ce57e245e9f5a3b8c35d6e2278%"/>
+  <a href="https://weblate.framasoft.org/projects/peertube/angular/">
```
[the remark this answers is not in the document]
rigelk: It’s always failing to load though

support/nginx/peertube
rigelk: Yes

support/nginx/peertube
rigelk: As mentioned in the PR description, Let’s Encrypt supports 301 redirects, so this block is superfluous.

support/nginx/peertube
Chocobozzz: I mean could you add a comment like *Add ECDHE-RSA-AES256-SHA if you want compatibility with Android 4*?

README.md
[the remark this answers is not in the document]
Chocobozzz: Yes I know, because it's been a long time since my nice E2E test but it's part of the deal with Browser Stack steam

support/nginx/peertube
Chocobozzz: Yes but it will be redirected on the https version, but if it's for the first cert it will fail no?

rigelk: Admittedly I didn’t test that part’s impact. Scaling back, I don’t want to break things.

rigelk:
@Chocobozzz tested on sherri, I modified a few things further and left the config on the server.

Issue #3324, opened 2020-11-18
Different default avatar for channels


kimsible:
Currently, the default avatar is the same for the profile account and a channel.
It would be nice to have a different one for each other, it would be a first step to help to understand better the concept of a channel.

![image](https://user-images.githubusercontent.com/1877318/99527634-68377000-299d-11eb-8291-290229e88703.png)

kimsible:
We could use something closer than the icons in the menu left for each default avatar : 

![image](https://user-images.githubusercontent.com/1877318/99528690-f52ef900-299e-11eb-8451-68b57b17c6dd.png)

rigelk:
Please don't tag your issue title yourself.

Chocobozzz:
This icon fits well for the account. Don't hesitate to make a PR with the channel icon (or another "channel" icon) in 192x192

PR #3326, opened 2020-11-18
Add new default different avatar for channel and account

kimsible:
## Description

Add a new default avatar based on Featured icons and for each Actor type : VideoChannel and Account.
This way we can easily see when the default avatar referring to a channel or an account. 

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/3324

## Has this been tested?

Default avatar is only a client side display, so the only tests I made are displaying the changed view as the screenshots show bellow.

## Screenshots

Account menu
avatar1

Comments 
avatar2

Admin tables
avatar3

Account page profile
avatar4

Channel page profile
avatar5

Channel list
avatar6

Video description
avatar7

Search
avatar8


Chocobozzz:
Hello,

Why removing the old default avatar? I think it's appropriate to define an account, and it's more cheerful than the new avatar.

kimsible:
> Hello,
> 
> Why removing the old default avatar? I think it's appropriate to define an account, and it's more cheerful than the new avatar.

@Chocobozzz sorry if you're attached to it, but it's to be consistent with the rest of the design, the image takes the user / account icon of featured icons.

And sorry again if you did that image, it's not against you, but it's not cheerful for me and some other PeerTube users, it's more creepy and weird to use a big smiley to represent a default profile / account avatar, it's not very common. For example, in QueerMotion, no member understood it was a default avatar at the beginning despite the remark that it's creepy :/.

In any case, I don't say it will fix the fact that a user does not fill its profile and set its avatar but it's really better to be consistent with common design « codes » and the rest of the PeerTube design (features icons).

Chocobozzz:
> @Chocobozzz sorry if you're attached to it, but it's to be consistent with the rest of the design, the image takes the user / account icon of featured icons.

I'm not attached to anything. I just argued that the new one is too austere compared to the current one.

> And sorry again if you did that image, it's not against you

I did not. Please let's leave my *potential* personal feelings out of this discussion :)

> it's more creepy and weird to use a big smiley to represent a default profile

Then we can search for a better alternative. Your edit seems a good idea. I'll see what I can do but it will take time.

kimsible:
> I did not. Please let's leave my potential personal feelings out of this discussion :)

Ok ok =). You put them in the discussion yourself. I understand that a project like PeerTube asks lots of attention and a community (as every community) is not always as cheerful as it should be when it comes to request features or describe an issue, that's why we have to try to not let our personal feelings out too quickly.

> Then we can search for a better alternative. Your edit seems a good idea. I'll see what I can do but it will take time.

Maybe we can ask to a designer ? I'll try to find one interested, we have couple of ones in QueerMotion.

Issue #1184, opened 2018-10-03
Replace References to 'Mastodon'


zcdunn:
PeerTube version or URL: PeerTube v1.0.0-beta.16
Browser name/version: Firefox
NodeJS version: ?

* **What is the expected behaviour?**
UI says 'Subscribe with a fediverse account:'
* **What do you see instead?**
UI says 'Subscribe with a Mastodon account:'

PeerTube federates with ActivityPub so it should use 'ActivityPub' or 'fediverse' to reference non PeerTube accounts, since users can interact from software other than PeerTube and Mastodon.

![image](https://user-images.githubusercontent.com/7121969/46431098-dfa3da00-c718-11e8-9b31-1da0dbfa152d.png)

rigelk:
Actually the first version was very platform-neutral :) but user tests showed most people didn't understand what ActivityPub was, or tried to use their PeerTube handle, which isn't (yet) possible...

headache-booth:
Maybe "Subscribe with other fediverse software" might be a good option then?

rigelk:
As soon as PeerTube supports it, we'll try again

rosbeef:
follow with mastodon is not clear. 
follow with activitypub was better maybe somthing like follow with activitypub (mastodon,pleroma, ...)

or simpler follow with and a list of socials activitypub logo

-----------------------
| follow with :       |
| Mas | Ple | PTB |
| textbox              |
______________

It could be interesting to be able to follow with an other peertube instance account.

kingemir:
So the reviewer of the proposed pull request https://github.com/Chocobozzz/PeerTube/pull/3334 wishes that the redirection to subscription is tested with all ActivityPub platforms before he can acknowledge such a change in the wording. 
I can confirm that the redirection to subscription works for following platforms: **Hubzilla** (Versions 5.0.1 and 5.0.2)

moussieri:
Redirection to subscription to peertube a channel works also for following platforms: Zap (Zot protocol) Version 20.11.23

Chocobozzz:
Done in https://github.com/Chocobozzz/PeerTube/pull/3598

PR #3334, opened 2020-11-20
replace Mastodon by ActivityPub

kingemir:
To solve issue #1184

## Has this been tested?


rigelk:
@kingemir thank you for your interest in this issue, but I don't think you solve the problem, nor that the discussion has reached a satisfying conclusion in the linked issue for you to submit a PR. For instance, I disagree with your assumption that ActivityPub is more recognisable than Mastodon. I also think it is less accurate: the function was tested with Mastodon only, and putting "ActivityPub" might lead users to think any platform advertised as an ActivityPub platform could subscribe to this account by just entering their handle ; the reality is more limited: only Pleroma and Mastodon support that kind of redirection to subscription.

Please also respect reviewers and fill the PR template appropriately.

PR #3339, opened 2020-11-20
Refactor ffmpeg utils

Chocobozzz:
## Goal

Cleaner ffmpeg utils, so we can have multiple transcoding profiles (hight quality, default, etc) in the future. We could also imagine plugins injecting a new transcoding profile in PeerTube.

## TODO

- [x] Split ffmpeg utils in two files: ffmpeg utils and ffprobe utils
- [x] Reuse existing probe data if we can
- [x] Extract transcoding options in a profile object

## What's new

 * `ffmpeg-utils.ts` is a little bit more complex, because it's designed to be encoders agnostic: https://github.com/Chocobozzz/PeerTube/blob/feature/refactor--fmpeg/server/helpers/ffmpeg-utils.ts
 * Extract encoder options in a *profile* module: https://github.com/Chocobozzz/PeerTube/blob/feature/refactor--fmpeg/server/lib/video-transcoding-profiles.ts
 * PeerTube will try encoders in this order, if available in ffmpeg and in the profiles: https://github.com/Chocobozzz/PeerTube/blob/feature/refactor--fmpeg/server/initializers/constants.ts#L339
 * To use the transcoding functions, you just send as param the encoders objects and the profile you want to use  (if the profile is not available for the chosen encoder, PeerTube will use the default profile): 
    * https://github.com/Chocobozzz/PeerTube/blob/feature/refactor--fmpeg/server/lib/live-manager.ts#L268
    * https://github.com/Chocobozzz/PeerTube/blob/feature/refactor--fmpeg/server/lib/video-transcoding.ts#L39

For now we only have a `default` profile. But I would like to add an "high quality" profile, or a profile dedicated to audio content (better bitrate, less compression etc). If we're happy with this system, we could imagine that plugins can add their own profile. And maybe their own encoders.


Chocobozzz:
PR description updated.

server/lib/live-manager.ts
```diff
@@ -269,8 +265,17 @@ class LiveManager {
     const deleteSegments = videoLive.saveReplay === false
 
     const ffmpegExec = CONFIG.LIVE.TRANSCODING.ENABLED
-      ? runLiveTranscoding(rtmpUrl, outPath, allResolutions, fps, deleteSegments)
-      : runLiveMuxing(rtmpUrl, outPath, deleteSegments)
+      ? await getLiveTranscodingCommand({
+        rtmpUrl,
+        outPath,
+        resolutions:
+        allResolutions,
```
rigelk, line 272:
```suggestion
        resolutions: allResolutions,
```

Chocobozzz: :+1:

Issue #324, opened 2018-03-06
Allow uploads to be resumed


conferences-gesticulees:
Uploading videos take hours with most connections available today. Waiting for hours to get nothing because the connection breaks for a couple minutes, the server gets restarted, or a power failure at 5am is highly frustrating.

Users should be able to resume their uploads. Ideally, it should be possible to resume a partial upload. At the very least, it should be possible to complete the publishing process after an upload is done. For example if an error happens during the processing stage, users should not have to upload the whole video again.

A checksum of the bytes already uploaded could be compared to ensure the local file has not been modified in the meantime.

Partial uploads should not be kept on the server indefinitely. I guess keeping them for 10 days would be long enough.

MayeulC:
How about uploading the video with webtorrent as well (with the current way as a fallback)? It could even open up some interesting functionalities down the road, like allowing the video to be watched before it's completely transferred, and maybe helping pave the road to livestreaming :)

rigelk:
@MayeulC actually uploading the video with WebTorrent would mean uploading the video with a WebRTC data channel to the server and using BitTorrent in it to declare (creating a torrent) and then transfer the file. WebRTC acts as a mere transport layer for a BitTorrent session whose goal is to share a file. The first problem is the fact that torrents are immutable: it prevents livestreaming (see issue #151). In our case the file is a video, but what prevents a client from declaring a torrent for other file types? The server should be able to check the file conforms to what a video is, thus downloading the video in its entirety anyway before forwarding the declared torrent. So I'm afraid using WebTorrent for the upload wouldn't bring anything, except added complexity.

JigmeDatse:
@rigelk I've been thinking about this feature.  The underlying .torrent could potentially be uploaded, but to do so would require the local generation of the .torrent file.  Now I'm not sure if this is a possibility or not with how things are currently designed.  

From my understanding the only way that would allow that, without someone manually creating the torrent locally than uploading it (and seeding it) would be to try to do it with JavaScript in the browser.  

I don't know if that is a feasible way to do things or not either.  But if there were the option of using the torrent protocol to upload, that could allow for the creation of clusters.  Without that, I think the issues I've had with being able to watch videos will probably continue, as I suspect that the limited numbers of peers is a significant factor.  

It is also possible that the peers that do exist for a given video with performance issues are behind a "too slow connection."

Chocobozzz:
[...]

PR #3347, opened 2020-11-23
more explicit error messages for file uploads

### LINKED ISSUE #3050: Obvious error message to replace "unknown error"
Hello,

Sometimes, when i upload a video, it fails because of an "unknown error". After some testing, i discovered that it was because of another video was being transcoded, so my CPU was constantly at 100%. So i think the so-called unknown error is hiding the fact that the server doesn't have enough resources for now and that we simply have to wait until transcoding is done, it would be clearer than "unknown error".

rigelk:
## Description

File uploads are often met with network errors. They can also be met with classic HTTP errors as defined in our spec. Both are _dealt with_ (reset + notification), but notification messages only last 5 seconds, and their messages are not always describing the code correctly nor left in the console for later investigation. Since users uploading large files or with unleliable connections are not always monitoring their upload page, they might miss a temporary notification.

This PR brings:
- sticky notification option
- notifications log to the console
- _some_ upload errors now trigger sticky notifications
- RP sets a header showing maximum file size for errors

@Chocobozzz are you interested by the maximum file size information being given by the reverse-proxy, or is it superfluous?

## Related issues

This issue is related to https://github.com/Chocobozzz/PeerTube/issues/324#issuecomment-731694361, and
fixes #3050

## Has this been tested?


rigelk:
@Chocobozzz do you know how I could test that in the CI?

Chocobozzz:
> @Chocobozzz are you interested by the maximum file size information being given by the reverse-proxy, or is it superfluous?

This is a good idea

> do you know how I could test that in the CI?

E2E CI?

Chocobozzz:
What's the status of this PR? I would like to ship it in v3 since upload errors are very frustrating at the moment. But I'm not sure you should use the `sticky` attribute. Instead, I think we should display the error on top of the upload form, with a retry button.

rigelk:
@Chocobozzz I’ll keep the notifier refactoring, as sticky messages might come in handy elsewhere.

rigelk:
I changed the error display so that it both notifies and keep the message in the progressbar, which now has a retry and cancel button. Maybe some errors could disable the retry button (like error 413 which a retry will not help solve).

Chocobozzz:
Nice!

PR #3348, opened 2020-11-23
Add about link icon in header right on bigger screens width than mobile

kimsible:
## Description

Add about link icon in header right to the publish button for screens width bigger than 500px (mobile).
This way, the about pages are well promoted. **On mobile screens the about link stays in the menu**.

## Screenshots
Capture d’écran du 2020-11-23 15-38-52

Peek 23-11-2020 15-37


Chocobozzz:
Sorry but adding another button in this area is not something that we want.

kimsible:
@Chocobozzz maybe you could have wait for other point of view of the community.

As I said « This way, the about pages are well promoted. »

Could you just consider that there is more than chocobozzz than using PeerTube ?

PR #3363, opened 2020-11-25
add My History API to openapi

Arman92:
## Description

`/users/me/history/videos` and `/users/me/history/videos/remove` were missing in openapi docs, I added them!

## Related issues

No related issue. just found these API endpoints were missing.

## Has this been tested?

### DIFF
--- a/support/doc/api/openapi.yaml
+++ b/support/doc/api/openapi.yaml
@@ -70,6 +70,9 @@ tags:
     description: >
       Operations related to your subscriptions to video channels, their
       new videos, and how to keep up to date with their latest publications!
+  - name: My History
+    description: >
+      Operations related to your watch history. 
   - name: My Notifications
     description: >
       Notifications following new videos, follows or reports. They allow you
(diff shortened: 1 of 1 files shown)


support/doc/api/openapi.yaml
```diff
@@ -917,6 +920,43 @@ paths:
...
+      tags:
+        - My History
+      requestBody:
+        content:
+          multipart/form-data:
+            schema:
+              type: object
+              properties:
+                beforeDate:
+                  description: If passed, history before this date will be deleted
```
rigelk, line 955: The fact the impact only applies if the argument is passed is implicit, since you didn't mark the argument as required

```suggestion
                  description: history before this date will be deleted
```

rigelk (changes requested):
Thanks for the addition @Arman92, it looks good to me (modulo the following change).

_note_: I added a commit to your PR since you added a "My History" category and it needs to be explicitly marked as being listed, as every new category in our API spec.

Arman92:
> Thanks for the addition @Arman92, it looks good to me (modulo the following change).
> 
> _note_: I added a commit to your PR since you added a "My History" category and it needs to be explicitly marked as being listed, as every new category in our API spec.

Cool, Thanks I didn't know about this.

PR #3377, opened 2020-11-27
Add input-password type

### LINKED ISSUE #3375: Add input-password type

A plugin like [auth-ldap](https://framagit.org/framasoft/peertube/official-plugins/-/blob/master/peertube-plugin-auth-ldap/main.js#L46) displays the password in clear text in the settings form.

I would like plugins with sensitive information to allow edition without displaying their content.

Add a `input-password` that plugins can use to set sensitive information.

None that I can think of.

I think I can provide a PR for that with roughly this content:
*  From https://github.com/Chocobozzz/PeerTube/blob/781ba981263ee6524fea1a95836108d252a124f4/shared/models/plugins/register-client-form-field.model.ts#L1-L4 to 
```typescript
 type: 'input' | 'input-checkbox' | 'input-password' | 'input-textarea' | 'markdown-text' | 'markdown-enhanced'
```

* From https://github.com/Chocobozzz/PeerTube/blob/781ba981263ee6524fea1a95836108d252a124f4/client/src/app/shared/shared-forms/dynamic-form-field.component.html#L6-L8 to
```html
  <input *ngIf="setting.type === 'input'" type="text" [id]="setting.name" [formControlName]="setting.name" />

  <input *ngIf="setting.type === 'input-password'" type="password" [id]="setting.name" [formControlName]="setting.name" />

  <textarea *ngIf="setting.type === 'input-textarea'" type="text" [id]="setting.name" [formControlName]="setting.name"></textarea>

```

PS: I hope I'm not saying anything wrong that already exists :sweat_smile: (did not find anything in docs about it).

madmath03:
## Description

:sparkles: Add input-password to edit sensitive information without displaying content.

Funded by @Monogramm.

## Related issues

* Fix #3375: Add input-password type 

## Has this been tested?

### DIFF
--- a/client/src/app/shared/shared-forms/dynamic-form-field.component.html
+++ b/client/src/app/shared/shared-forms/dynamic-form-field.component.html
@@ -5,6 +5,8 @@
 
   <input *ngIf="setting.type === 'input'" type="text" [id]="setting.name" [formControlName]="setting.name" />
 
+  <input *ngIf="setting.type === 'input-password'" type="password" [id]="setting.name" [formControlName]="setting.name" />
+
   <textarea *ngIf="setting.type === 'input-textarea'" type="text" [id]="setting.name" [formControlName]="setting.name"></textarea>
 
   <my-help *ngIf="setting.type === 'markdown-text'" helpType="markdownText"></my-help>
--- a/shared/models/plugins/register-client-form-field.model.ts
+++ b/shared/models/plugins/register-client-form-field.model.ts
@@ -1,7 +1,7 @@
 export interface RegisterClientFormFieldOptions {
   name: string
   label: string
-  type: 'input' | 'input-checkbox' | 'input-textarea' | 'markdown-text' | 'markdown-enhanced'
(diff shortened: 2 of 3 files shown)


rigelk:
@madmath03 thanks for the PR! I have an improvement suggestion though: having a button in the input to show the text in clear upon user request. And since we already have that kind of input in `client/src/app/+admin/users/user-edit/user-password.component.html`, it would be nice to factorize that input logic in a shared component. Could you do that?

madmath03:
> @madmath03 thanks for the PR! I have an improvement suggestion though: having a button in the input to show the text in clear upon user request. And since we already have that kind of input in `client/src/app/+admin/users/user-edit/user-password.component.html`, it would be nice to factorize that input logic in a shared component. Could you do that?

@rigelk I need to take a look. I'm really new to PeerTube so I do not know if I will be able (or will have the time) to do this, but it sure is a good idea.

Chocobozzz:
Thanks @madmath03, don't hesitate do make another PR with @rigelk proposal

Issue #2240, opened 2019-11-13
Default channel selection when owner change (if unique channel)


klorydryk:
PeerTube v1.4.1

In /my-account/ownership , when I receive an owner change request, there is a pop-up to select the video channel. By default, the selector is empty.
If there is only 1 channel available, this channel should be selected by default, as there is no alternative.

Poslovitch:
Working on this.

PR #3382, opened 2020-11-29
Select first available channel when accepting ownership change

Poslovitch:
# Description

**This is a draft as the implementation it introduces must be discussed.**

Per request in #2240, I've slightly changed the select form in `my-accept-ownership.component.html` so that the first element of `videoChannels` is automatically selected. This, however, **occurs regardless of the amount of channels an account has**. This is the behavior I want to discuss with you. Should it remain as is, or should I refactor this PR to follow more closely the request of #2240?

Secondly, I had to change the `disable` condition of the "Accept" button. While the aforementioned change somehow worked fine if an account had multiple channels, the "Accept" button would remain greyed out if the account only had one channel.

## Related issues

Resolves https://github.com/Chocobozzz/PeerTube/issues/2240

## Has this been tested?

I don't know if tests should be added to the tests suite and/or modified.

Manual testing showed that the changes introduced by this PR are working as expected on my end.

## Screenshots

These screenshots are taken just right after clicking on the "checkmark" button to open the popup.

**When the account only has one channel**
image

**When the account has 2 or more channels**
image

### DIFF
--- a/client/src/app/+my-library/my-ownership/my-accept-ownership/my-accept-ownership.component.html
+++ b/client/src/app/+my-library/my-ownership/my-accept-ownership/my-accept-ownership.component.html
@@ -11,7 +11,7 @@
       <div class="peertube-select-container">
         <select formControlName="channel" id="channel" class="form-control">
           <option i18n value="undefined" disabled>Channel that will receive the video</option>
-          <option *ngFor="let channel of videoChannels" [value]="channel.id">{{ channel.displayName }}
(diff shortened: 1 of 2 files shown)


Chocobozzz:
> I've slightly changed the select form in my-accept-ownership.component.html so that the first element of videoChannels is automatically selected. This, however, occurs regardless of the amount of channels an account has.

Auto select a channel is fine, we do this in many other places. The user can also change the video channel after the ownership change.

Regarding the code, use `this.form.patchValue` in the component instead of `[selected]="i == 0"` in the template

Chocobozzz:
Thanks @Poslovitch

Poslovitch:
Thanks for merging 🙂

Issue #3372, opened 2020-11-27
multi server support


test2a:
**Describe the problem to be solved**
#56 was closed in 2017 and i would like to request this once again
today we expect the single server to do all the jobs. be it transcoding, serving files, management and others. there is a limit, both cost wise and physically a single VM or a physical desktop/server machine can serve. What i am proposing is allowing multiple servers in the backend structure with a load balancer so that we can build better more resilient instances.
that said, i do want to request allowing GPU transcoding because apparently that is not allowed yet.
let this be a 2021 roadmap goal, but this is something peertube would need to grow horizontally instead of going vertically up with a single server.

We would want storage on multiple servers but because s3 is already available, this is solved already.

**Describe the solution you would like:**
* allow load balancing and multiple servers doing separate jobs of serving users, transcoding, storage management, users management
* GPU transcoding

**Describe alternatives you have considered**

test2a:
now with LIVE streaming a reality, how many live streams can a single CPU handle before collapsing? if we have GPU transcoding, we could theoretically have any number of live streams because increasing live would just mean adding more GPUs

test2a:
plex remote transcoder does this job pretty nicely. 
would it be possible to integrate this software into peertube because doing that would solve this issue almost in entirety

https://github.com/UnicornTranscoder/UnicornTranscoder
https://github.com/wnielson/Plex-Remote-Transcoder

rigelk:
@test2a we cannot integrate it, but it's architecture is interesting.

test2a:
@rigelk yes. if this is actively being developed and is working as expected for them, this is a feature that is essential for a project like peertube. my concern is if some instance wants to host say 50-100 live streams, would they be able to do that on a single instance setup aka single cpu?

rigelk:
> my concern is if some instance wants to host say 50-100 live streams, would they be able to do that on a single instance setup aka single cpu?

no, most probably not at the moment.

rigelk:
@test2a the issue you mention has been closed because supporting multiple backends is a very difficult task. Hardware transcoding is a much more reasonable goal, but don't misunderstand: GPU encoding is not easy to configure at all, as you need to choose which device will transcode, which specific codec and driver applies to this device, and which specific options apply.

Anyway, thanks for the links, they are very inspiring!

Issue #947, opened 2018-08-22
Support for transcoding by remote workers


scanlime:
I'm trying to run a peertube instance with a large volume of 1080p60 content and the transcode bottleneck is significant. One potentially easy way to improve this when additional CPU power is available on the local network could be to scale out the transcode with a  worker daemon.

I'm opening this in the hope of a discussion for a future enhancement, and to keep track of that work.

In my case I'm imagining the workers being located in the same data center, but it seems like this could be extended to geographically diverse or even crowdsourced transcode sites.

Personally I'd be happy with a solution where the remote workers run a userspace daemon that manages ffmpeg with little to no local storage, accessing the peertube server's disks either over an existing network storage protocol or existing media streaming protocol.

I'm curious what takes other folks have. I know this is unlikely to be interesting for people running peertube on a small VPS, unless the transcode workers can be elsewhere on cheaper hosting (like at your house, on DSL).

Jorropo:
This is planned but in [desktop version](https://github.com/Jorropo/peertube-desktop) (but not yet (desktop version is in infdev (before alpha)))

rigelk:
A simpler way to implement is for clients to transcode first and then let the server decide if the video is properly transcoded.

But adding workers is another level of implementation. It's also something we need to define, because we have multiple possibilities: parallel transcoding of chunks of the same video, or transcoding of different videos. It also means the workers have to register at the PeerTube instance, and the instance has to monitor their progress with the trancoding tasks it gives them. Arguably, that's way more work than just making the client and server agree on whether or not the video needs additional trancoding.

rigelk:
Well, they are two different kind of "hard" :stuck_out_tongue: 

I'm not sure we want to check everything either ; checking for faststart and proper codec use via ffprobe should be enough.

utack:
Maybe this helps you  
https://github.com/tdaede/dve
Might however not be a method that is usable, as it includes usage of ssh

roipoussiere:
I totally support this issue.

The is a real use-case where a federation of video producers wants to create many PeerTube instances, but use a unique transcoding server in order to reduce instance hosting costs.

Nutomic:
This should actually be pretty easy to implement with ssh. I'm sure there is some ssh library that could be used. Basically, Peertube would do this:

```
scp video-to-transcode.mp4 transcoding-host:~
ssh transcoding-host ffmpeg ...
scp transcoding-host:transcoded-video.mp4 .
```

[...]

PR #3383, opened 2020-11-30
add support for alternative ffmpeg wrappers

rigelk:
## Description

This PR adds support for the `REMOTE_FFMPEG_PATH` environment variable (to differentiate from the `FFMPEG_PATH` that is already natively supported by `fluent-ffmpeg`), allowing simple (non-live, producing a video and using only one input file) transcoding jobs to be executed by another binary than a typical ffmpeg. It also adds a REST endpoint to allow such a program to retrieve the input file for its job, and injects the job id as a variable when calling ffmpeg.

## Related issues

The current solution is inspired from https://github.com/Chocobozzz/PeerTube/issues/947 and more recently https://github.com/Chocobozzz/PeerTube/issues/3372, which pointed me towards https://github.com/UnicornTranscoder/UnicornFFMPEG and more generally https://github.com/UnicornTranscoder whose architecture is more or less reliant on the same principles used in this PR, but implemented using Plex-specific endpoints. The goal is not to reuse their code but rather take from their design principles:

Point the `REMOTE_FFMPEG_PATH` environment variable to a script that sends arguments to a cluster manager. The cluster assigns the task to a worker which will retrieve the video at the PeerTube REST API and begin processing it with the transmitted arguments. The aforementioned script polls it until the task is completed, to then download the resulting file from the cluster worker via the returned url.

## Has this been tested?


Chocobozzz:
Closing due to inactivity. Feel free to reopen

PR #3385, opened 2020-12-01
Fix sort fields  on Video Playlists

Arman92:
## Description

displayName does not exist in VideoPlaylistModel, the correct field is `name`

## Has this been tested?


Chocobozzz:
Thanks @Arman92

PR #3388, opened 2020-12-02
Trigger a Gitpod dev image rebuild (in order to get upstream's Node.js 12 upgrade)

jankeromnes:
## Description

PeerTube's Gitpod dev image still uses Node.js 10, while [upstream]() was updated to Node.js 12.

The reason is that no change has been made to the Dockerfile in a while.

Adding a trivial change invalidates & rebuilds the dev image, and I've also added a comment to explain why.

## Has this been tested?


Chocobozzz:
Thanks @jankeromnes :)

Issue #2805, opened 2020-05-29
Allow creating a user without creating a channel


MCMic:
**What happened?**
When creating my account on a peertube instance, I’m asked a channel name and id after having chosen my user name and user id.

**What do you expect to happen instead?**
I expected this step to be optional and to be able to create my user account without creating a channel yet.

I understand that creating a channel is proposed when creating an account, but not why it is mandatory. I expect a lot of people will create an account not to upload videos but to watch and comment.
I may upload videos someday but I do not know yet on which subject so I cannot chose a channel name now.

After user creation, it seems I can actually delete the channel I was forced to create, but it warns me that I will never be able to use the same channel name again, which is frightening.
I should have been able to create my account without creating a channel, and without reserving a name that no-one can use again.
Even for inscances and federation this situation is bad because it results in lots of empty channels everywhere that will pop-up in search results and databases for no actual use.

**Additional information**
* PeerTube version or URL:
https://video.antopie.org/signup

MCMic:
Related to #782 but not a duplicate.

Chocobozzz:
Hi,

The channel step is optional when you register on PeerTube so I think I can close this issue.

Issue #782, opened 2018-07-08
Allow creating user accounts to post comments only


fflorent:
* **PeerTube version or URL:** 1.0.0b9

On Youtube, posting comments is a key feature. On Peertube, currently it is hard to post a new comment for a new comer when the instance is close (i.e. does not allow anyone to upload videos on it). A way to achieve it would consist in creating an account in another instance or posting on Mastodon. That's hard to understand for people not aware on how Peertube works :).

What would you think of allowing creating accounts just to post comments on a video?

Chocobozzz:
You can create a user with a video quota of 0 :)

mcb2003:
An idea could be then: to replace the upload quota with a message stating that this account can only be used for commenting, if the quota is 0.

Chocobozzz:
It's a good idea @mcb2003. Maybe a "Cannot upload video" would be better since a user can like/dislike videos too :)

PR welcome!

fflorent:
Yeah, probably a user with a video quota of 0 is the idea, some thoughts about discoverability:
 - people managing a Peertube instance are not aware of this (at least, I was not ¯\_(ツ)_/¯);
 - when they set this quota, not sure whether it is clear for the users they cannot upload a video, their account can just comment;

For the first point, @mcb2003's idea is good. For the second, replacing this message in the create account panel

> Initial video quota: 5 GB

With something explaining that the account is for commenting exclusively would solve the issue :) (if that's not the case yet).

Also a question: if the quota is set to 0, is the upload button available?

Florent

tcitworld:
In my opinion, this is a really bad idea from a user story point of view : people will want to create accounts everywhere they want to comment, which will lead to lots of useless duplicates across the fediverse.

My solution would be linked to https://github.com/Chocobozzz/PeerTube/issues/470, allowing users to view videos (and their comments) from any source - even if not allowed by the admin - into their own PeerTube instance. Users would be redirected from the external instance to their own through web protocol handlers (like Mastodon has `web+mastodon://`), and the link would carry video and maybe comment information for the own instance to find if necessary.

Therefore people would write comments with their own account on the instance they registered in the first place.

fflorent:
@Chocobozzz I think that allowing the creation of accounts with a 0Mb quota is confusing (v 1.4.1): 

![image](https://user-images.githubusercontent.com/371705/69030501-0fbf4280-09d8-11ea-9ea2-dbbf048af2f7.png)

[...]

PR #3393, opened 2020-12-03
register: hide channel step if upload quota is 0

kontrollanten:
## Description

When default upload quota is set to 0, the channel step in the registration will be hidden, as proposed in #782 

## Related issues

closing #782 
relates #2805 

## Has this been tested?

I couldn't find any appropriate way to add tests since there's no unit tests for the client and the tests in `server/tests/client.ts` seems to be on a higher lever than this feature.

## Screenshots

videodisabled


Chocobozzz:
Thanks @kontrollanten

PR #3397, opened 2020-12-03
server accounts/videos: trigger filter hook

kontrollanten:
## Description
`filter:api.videos.list.*` hooks weren't triggered when listing an accounts videos. It's still not working for `/me/videos` since I don't know how to solve it there. I

## Has this been tested?

1) Visit an accounts videos page (ie /accounts/root/videos) and verify it's not broken
1) Install a plugin that hooks `filter:api.videos.list.result` and verify it works (for example https://www.npmjs.com/package/peertube-plugin-sort-originally-published-at)

### DIFF
--- a/server/controllers/api/accounts.ts
+++ b/server/controllers/api/accounts.ts
@@ -176,12 +176,12 @@ async function listAccountVideos (req: express.Request, res: express.Response) {
     accountId: account.id,
     user: res.locals.oauth ? res.locals.oauth.token.User : undefined,
     countVideos
-  }, 'filter:api.videos.list.params')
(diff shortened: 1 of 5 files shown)


rigelk:
@kontrollanten you erronously filled the template, marking that you did tests without explaining which.

kontrollanten:
Sorry, now I've added that

Chocobozzz:
Thanks, I just added a commit to use dedicated hook name for accounts/channels videos endpoints.

kontrollanten:
Thanks! Don't hesitate to ask me to do this kind of changes in the future.

PR #3400, opened 2020-12-04
add optional authentication for plugin routes

kontrollanten:
## Description
* When creating custom routes with plugins, let the plugin author check if the user is authenticated or not.
* `authenticate` middleware now sets `res.locals.authenticated` to `true` upon successful authentication.

## Has this been tested?


Chocobozzz:
Thanks @kontrollanten

PR #3401, opened 2020-12-05
Fix PostreSQL initialization for Gitpod

jankeromnes:
## Description

Fixes https://github.com/gitpod-io/gitpod/issues/2429

> Unable to boot a Workspace : fails at "Building Workspace Image"

---

Solution: https://github.com/gitpod-io/gitpod/issues/2429#issuecomment-739238805

> I think I know what the problem is. PeerTube's Gitpod Dockerfile attempts to initialize a PostgreSQL database:
> 
> https://github.com/Chocobozzz/PeerTube/blob/95bd9515b86e577033d9929c829c77c0df761acf/support/docker/gitpod/Dockerfile#L15
> 
> For some reason, this now fails with "Permission denied", when `pg_start` tries to create `/workspace/.pgsql/data` with `mkdir -p`:
> 
> https://github.com/gitpod-io/workspace-images/blob/c9896cde76e0698b1ed1b93ad464d77f87dd96e6/postgres/Dockerfile#L13
> 
> Probably user `gitpod` is not authorized to create directories under `/`.
> 
> However, probably setting up the PostgreSQL database doesn't even make sense in the Dockerfile, because everything under `/workspace` will get replaced anyway when you start a Gitpod workspace (because `/workspace` is replaced by a mounted partition where the repository then gets cloned into -- i.e. everything you put in `/workspace` in the Dockerfile will be gone at runtime).
> 
> So, instead, I think the proper fix is to set up the PostgreSQL database not in the Dockerfile but in the `.gitpod.yml` tasks once the workspace is running.

## Has this been tested?

Can be tested by opening my fork in Gitpod: https://gitpod.io/#https://github.com/jankeromnes/PeerTube

## Screenshots
Screenshot 2020-12-05 at 12 37 01

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -10,7 +10,11 @@ ports:
 - port: 9000
   onOpen: ignore
 tasks:
-- command: redis-server
-- before: export NODE_CONFIG="{\"import\":{\"videos\":{\"torrent\":{\"enabled\":false}}},\"webserver\":{\"hostname\":\"$(gp url 3000 | cut -d/ -f3)\",\"port\":\"443\",\"https\":true}}"
-  init: yarn install --pure-lockfile
+- name: Redis
+  command: redis-server
+- name: PeerTube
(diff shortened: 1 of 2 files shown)


rigelk:
Thanks a bunch @jankeromnes & @Poslovitch for spotting the issue!

Poslovitch:
Thanks for merging @rigelk!

Issue #1539, opened 2019-01-08
Canonical link injected into HTML is invalid


tcitworld:
It links to the current URL of the video whereas it should link to the URL of the video on the original instance.
https://github.com/Chocobozzz/PeerTube/blob/e76d578414d2f2f853bbb1d4bbe180da05aa4eab/server/lib/client-html.ts#L119
https://github.com/Chocobozzz/PeerTube/blob/e76d578414d2f2f853bbb1d4bbe180da05aa4eab/server/lib/client-html.ts#L191

While it should be something like
```ts
tagsString += `<link rel="canonical" href="${video.url}" />`
```

Should improve SEO a bit.

Chocobozzz:
Fixed in https://github.com/Chocobozzz/PeerTube/commit/c04eb647db4f543a31a8100c1ec9a86c700bca6a

Thanks @tcitworld 

> Should improve SEO a bit.

Unfortunately, Google bot uses a very old chrome engine so it cannot parse our javascript and it internally renders a blank page. And Angular CLI (developed by google) does not support Babel so...

![office-blased](https://user-images.githubusercontent.com/5180488/50977288-43952600-14f2-11e9-8ed4-ca4383c768b1.gif)

Chocobozzz:
@tcitworld No, I'll try to find the dependency that ship ES2015 code and pin it to a previous version instead.

PR #3406, opened 2020-12-06
Set canonical link to original video/playlist url

tcitworld:
The local URL was used instead of the original url

## Description

The local URL was used instead of the original url

## Related issues

Previously fixed in https://github.com/Chocobozzz/PeerTube/issues/1539

Added a test to check but didn't manage to run them, so they're all yours.

### DIFF
--- a/server/tests/client.ts
+++ b/server/tests/client.ts
@@ -8,19 +8,20 @@ import {
   addVideoInPlaylist,
   cleanupTests,
   createVideoPlaylist,
-  flushAndRunServer,
   getAccount,
   getCustomConfig,
   getVideosList,
   makeHTMLRequest,
   ServerInfo,
-  serverLogin,
   setDefaultVideoChannel,
   updateCustomConfig,
   updateCustomSubConfig,
   uploadVideo,
   updateMyUser,
-  updateVideoChannel
+  updateVideoChannel,
+  doubleFollow,
+  flushAndRunMultipleServers,
+  setAccessTokensToServers
 } from '../../shared/extra-utils'
 
 const expect = chai.expect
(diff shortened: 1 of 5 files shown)


server/tests/client.ts
```diff
@@ -47,8 +51,19 @@ describe('Test a client controllers', function () {
   before(async function () {
     this.timeout(120000)
 
-    server = await flushAndRunServer(1)
-    server.accessToken = await serverLogin(server)
+    servers = await flushAndRunMultipleServers(2)
```
kontrollanten, line 54: A fancy way would be `[server, secondServer] = await flushAndRunMultipleServers(2)`

server/tests/client.ts
```diff
@@ -47,8 +51,19 @@ describe('Test a client controllers', function () {
...
+    const server = servers[0]
+
+    await setAccessTokensToServers(servers)
+
+    {
+      const res = await uploadVideo(servers[0].url, servers[0].accessToken, { name: 'video' })
```
kontrollanten, line 60: Do we really need to upload a video here? According to line 78 there's already a video on the server.

If we need this line, then line 78 has to be updated (that's why tests are failing).

server/tests/client.ts
```diff
@@ -318,6 +333,11 @@ describe('Test a client controllers', function () {
 
       checkIndexTags(res.text, 'PeerTube updated', 'my short description', 'body { background-color: red; }')
     })
+
+    it('Should use the original video URL for the canonical tag', async function () {
+      const res = await makeHTMLRequest(servers[1].url, '/videos/watch/' + videoUUID)
+      expect(res.text).to.contain(`<link rel="canonical" href="${videoOriginalUrl}" />`)
```
kontrollanten, line 339: By using `server.video` we could change the test to:

```
      const res = await makeHTMLRequest(servers[1].url, '/videos/watch/' + server.video.uuid)
      expect(res.text).to.contain(`<link rel="canonical" href="${server.video.url}" />`)
```

kontrollanten (changes requested):

Chocobozzz:
Thanks @tcitworld  @kontrollanten

PR #3408, opened 2020-12-06
hls-plugin: destroy hls upon third err

kontrollanten:
## Description
According to hls.js docs `hls.destroy()` should be called
https://github.com/video-dev/hls.js/blob/master/docs/API.md#final-step-destroying-switching-between-streams

## Has this been tested?

I think it's a bit fussy to create a unit test just to verify this, but if you want to I'll add one.


Chocobozzz:
Thanks @kontrollanten

PR #3409, opened 2020-12-06
replace numbers with typed http status codes

rigelk:
## Description

In a previous PR (#3347) I introduced typed http codes, and planned to integrated them at large later due to their being unrelated to said PR. This PR pursues that initial goal of having typed http codes, to allow backtracking their use and properly documenting where and when they should be used via JS documentation and intelligent code completion that usually displays such comments.

A slight change: I made errors in the client html generation serve-side to throw a 500 error instead of a 400 error.

## Has this been tested?

While not technically introducing any change, I replaced occurences manually and checked them again before comitting.


Chocobozzz:
Seems `api-1` test fails

Issue #3412, opened 2020-12-07
WebTorrent plugin: Unable to get property 'magnetUri' of undefined or null reference


kontrollanten:
**Describe the current behavior**
Some of our users are currently getting the following errors. It does seems to happen randomly; different OS, different browsers and different videos. It's only a few users that are experiencing this.

```
TypeError: Unable to get property 'magnetUri' of undefined or null reference
  at Wn.prototype.updateVideoFile(/client/sv-SE/6-es2015.1930b8c591ce0c935166.js:1:180221)
  at Wn.prototype.initializePlayer(/client/sv-SE/6-es2015.1930b8c591ce0c935166.js:1:184371)
  at Anonymous function(/client/sv-SE/6-es2015.1930b8c591ce0c935166.js:1:179101)
  at Anonymous function(/client/sv-SE/17-es2015.01b27a81d2ff17aae33f.js:1:246071)
  at Array.prototype.forEach(native code)
  at Anonymous function(/client/sv-SE/17-es2015.01b27a81d2ff17aae33f.js:1:246013)
  at Anonymous function(/client/sv-SE/17-es2015.01b27a81d2ff17aae33f.js:1:248993)
  at u.prototype.invokeTask(/client/sv-SE/polyfills-es2015.a056ef761beb7a20be01.js:1:28421)
  at a.prototype.runTask(/client/sv-SE/polyfills-es2015.a056ef761beb7a20be01.js:1:23928)
  at f.invokeTask(/client/sv-SE/polyfills-es2015.a056ef761beb7a20be01.js:1:29568)
```

Sentry also provides this information:

```
{snip} is.fallbackToHttp(t,()=>(this.player.playbackRate(r),n()));this.addTorrent(this.currentVideoFile.magnetUri,i,t,()=>(this.player.playbackRate {snip}
{snip} }initializePlayer(){if(this.buildQualities(),this.autoplay)return this.player.posterImage.hide(),this.updateVideoFile(void 0,{forcePlay:!0,s {snip}
```

**Steps to reproduce:**
I'm not able to reproduce the errors, I've just seen them in Sentry.

**Describe the expected behavior**
1) Clearer error message
2) Gently handle those error / detect the error earlier

**Additional information**

* PeerTube instance:
  * version: 2.4.0
  * NodeJS version: 12.20.0
  * Ffmpeg version: 4.2.4

Chocobozzz:
[...]

PR #3413, opened 2020-12-07
WebTorrent Plugin: Clearify error message

kontrollanten:
## Description
This PR shouldn't change any functionality, but just give a more clear error message when `videoFiles`

## Related issues
#3412 

## Has this been tested?

1) Start the dev server
1) Start a video
1) Verify nothing is broken

### DIFF
--- a/client/src/assets/player/webtorrent/webtorrent-plugin.ts
+++ b/client/src/assets/player/webtorrent/webtorrent-plugin.ts
@@ -139,6 +139,10 @@ class WebTorrentPlugin extends Plugin {
         : this.pickAverageVideoFile()
     }
 
+    if (videoFile === undefined) {
+      throw Error(`Can't update video file since videoFile is undefined.`)
+    }
+
     // Don't add the same video file once again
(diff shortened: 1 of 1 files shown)


kontrollanten:
Updated description with test instructions.

Chocobozzz:
Hi,

Thanks for the PR but I would rather to not "refactor" the webtorrent plugin since it's in maintenance mode. I think the root cause is that `videoFile` could be `null` [here](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/assets/player/webtorrent/webtorrent-plugin.ts#L126)

Then, we could also throw [here](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/assets/player/webtorrent/webtorrent-plugin.ts#L141) is `!!videoFile` is `false`

kontrollanten:
I see. Now I've reverted the refactor commits and changed the check to `!!videoFile`.

Chocobozzz:
Thanks @kontrollanten

PR #3423, opened 2020-12-08
emit more specific status codes on video upload

rigelk:
## Description

- reduce http status codes list to potentially useful codes
- document codes and their preferred use
- convert more codes to typed ones

## Has this been tested?

It relies on the test suite statuses.

## TODO

- [x] refactor extname checker back to mimetype checker

### DIFF
--- a/server/middlewares/validators/videos/videos.ts
+++ b/server/middlewares/validators/videos/videos.ts
@@ -8,6 +8,7 @@ import { VideoChangeOwnershipAccept } from '../../../../shared/models/videos/vid
 import {
   isBooleanValid,
   isDateValid,
+  isFileFieldValid,
   isIdOrUUIDValid,
   isIdValid,
   isUUIDValid,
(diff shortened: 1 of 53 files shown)


server/middlewares/validators/videos/videos.ts
```diff
@@ -55,11 +57,11 @@ import { HttpStatusCode } from '../../../../shared/core-utils/miscs/http-error-c
...
-    .custom((value, { req }) => isVideoFile(req.files)).withMessage(
-      'This file is not supported or too large. Please, make sure it is of the following type: ' +
-      CONSTRAINTS_FIELDS.VIDEOS.EXTNAME.join(', ')
-    ),
-  body('name').custom(isVideoNameValid).withMessage('Should have a valid name'),
+    .custom((value, { req }) => isFileFieldValid(req.files, 'videofile'))
+    .withMessage('Should have a file'),
+  body('name')
+    .custom(isVideoNameValid)
+    .withMessage('Should have a valid name'),
```
rigelk, line 64: @Chocobozzz I think downgrading the check here provided I do the rest of the check below is okay, but is this fine by you?

PR #3427, opened 2020-12-08
Video embed page: Inject custom CSS and set page title

### LINKED ISSUE #3420: Custom CSS not injected to embedded player
When embedding a video the custom CSS isn't injected.

**Steps to reproduce:**

1. Define `body#custom-css #video-wrapper { display: none; }` as custom CSS.
1. Embed a video.
1. See that it's not hidden.

The body element has the `custom-css` class name, so I assume the intent is to inject the custom CSS. I'd also like to add a more explicit context class to the embed page, like CSS class `embedded-player` so that it's easier to write rules targeting only embedded player.

kontrollanten:
## Description
* Add the custom CSS to video embed HTML page.
* Add a CSS class to the body element to make it easier to target.
* Set HTML title at embed page to the instance name.

## Related issues
closes #3420

## Has this been tested?

1) Set `.standalone-video-embed { background: green; } .standalone-video-embed #video-wrapper { display: none; }` as custom CSS.
1) Set a custom name for your PeerTube instance.
1) Copy the iframe target for an embed tag and visit the page.
1) Watch green.

## Screenshots
image

### DIFF
--- a/client/src/standalone/videos/embed.html
+++ b/client/src/standalone/videos/embed.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>PeerTube</title>
+    <!-- title tag -->
 
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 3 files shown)


Chocobozzz:
Thanks

PR #3447, opened 2020-12-11
Fix typ in admin config

kontrollanten:
## Description

>  e.g. is short for exempli gratia, a Latin phrase that means “for the sake of example.” 
> ...
>  I.e.’s Latin origin is the phrase id est, which translates to English as “that is to say” or “in other words.”

https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/

## Screenshots

image


Chocobozzz:
Thanks

Issue #3134, opened 2020-08-29
Discussion for better managing release / scheduled date


quentinDupont:
Hello and thanks for all the amazing work :tada: 
I don't raise a bug but propose a discussion for improvement

**What I would like :**

Upload my video on Peertube in private mode.
Make my changes for the thumbnail, description, etc.
Think about a release date
Schedule the public release date of the video

**What Peertube allows me to do**

Schedule the release date as soon as uploaded

**Why it's not perfect for my use :**

When I upload my video, I still don't know the desired release date of the video.
Uploading it allows me to share it with my colleague who manages video descriptions, subtitles etc. as well as posts on social networks announcing the release of the video.
Once the release date is chosen, we would like to be able to schedule it on Peertube.

Thanks for reading and sharing you opinion :slightly_smiling_face:

test2a:
you can use something like change video ownership as a band-aid solution for sending videos to your colleagues as you said. You upload the video, set it to private, then give ownership to someone else, they will change something and then they will give ownership to someone else. Unless peertube has better user organization as there are some issues tending to that, this should be helpful for you

kontrollanten:
Great proposal! It could be implemented like this

<img src="https://user-images.githubusercontent.com/6680299/101342679-340bec80-3883-11eb-9c9d-8dedf7bb2971.png" width="300" />

The scheduling can be added to `JobQueue` just as `videos-views`:

https://github.com/Chocobozzz/PeerTube/blob/adc1f09c0dbd997f34028c1c82d1c118dc8ead80/server/lib/job-queue/job-queue.ts#L218-L222

kontrollanten:
It looks like it partially implemented;

https://github.com/Chocobozzz/PeerTube/search?q=scheduleUpdate

kontrollanten:
After digging into this I found out that this feature already exists and works 😄 I don't know how I can have missed this. Either I'm dumb or the UI can be improved. Either way I think this issue can be closed?

@quentinDupont If you save your video as private then you can schedule the publishing when you're updating it (and not only when creating it).

quentinDupont:
> After digging into this I found out that this feature already exists and works smile I don't know how I can have missed this. Either I'm dumb or the UI can be improved. Either way I think this issue can be closed?
> 
> @quentinDupont If you save your video as private then you can schedule the publishing when you're updating it (and not only when creating it).

Ok i will check that and close the issue if it's what I had in mind :) 
Thanks for the work anyway !

[...]

PR #3455, opened 2020-12-11
Enable scheduled publishing for imports

kontrollanten:
## Description
* Enable scheduled publishing for import URL/torrent.
* Disable instead of hide the schedule publishing option in cases where it's not enabled.

## Reflections
* Shouldn't it be possible to schedule the publishing even if the video is marked as internal or unlisted?

## Related issues
#3134 

## Has this been tested?

1) Import a URL
1) Check that it's possible to schedule the publishing.

1) Import a torrent
1) Check that it's possible to schedule the publishing.

1) Set a videos privacy setting to _Internal_
1) Edit the video and see that _Scheduled_ is disabled.

## Screenshots

image


kontrollanten:
I just found out that I didn't add the schedule option in the first step of import. What's the reason that the privacy and channel fields are both in step 1 and 2 of the import process? Can we remove the fields in the first step to keep it simpler?

Chocobozzz:
Hello,

We disabled the schedule privacy from import because we're not sure when the file will be imported, so the feature could not work as expected. And we did not have too many users asking this feature for imports. I think most of them will import the file in private, and schedule a privacy change afterwards.

kontrollanten:
I see. What about just disable the option instead of hide it (so the user at least knows that this functionality exists)? Shall I remove the privacy option in import step 1 (since it's redundant)?

Chocobozzz:
> I see. What about just disable the option instead of hide it (so the user at least knows that this functionality exists)?

I don't think we should display a disabled thing that cannot be enabled in this screen. It would just be confusing for users.

> Shall I remove the privacy option in import step 1 (since it's redundant)?

We added the channel and privacy in the first step to highlight these settings, even if you can update them in the second screen

kontrollanten:
> > I see. What about just disable the option instead of hide it (so the user at least knows that this functionality exists)?
> 
> I don't think we should display a disabled thing that cannot be enabled in this screen. It would just be confusing for users.

I see your point, but I don't agree. It cannot be enabled in this screen, but on this video. Me myself thought PeerTube lacked the functionality of scheduling publishing during my first month using it, since I just imported videos and never saw that alternative (and since there's an issue for this I'm not alone about that). Can we add a question mark with a popover that describes the different privacy options (and why _Scheduled_ is disabled)?


> > Shall I remove the privacy option in import step 1 (since it's redundant)?
> 
> We added the channel and privacy in the first step to highlight these settings, even if you can update them in the second screen

Sorry maybe I'm missing something, but I don't see why it should be highlighted. It'd make sense if it'd be a quick way to just skip the second step, but the user has to go through the second step and see those fields again anyway.

Chocobozzz:
> I see your point, but I don't agree. It cannot be enabled in this screen, but on this video. Me myself thought PeerTube lacked the functionality of scheduling publishing during my first month using it, since I just imported videos and never saw that alternative (and since there's an issue for this I'm not alone about that). 

Okay then.

>  Can we add a question mark with a popover that describes the different privacy options (and why Scheduled is disabled)?

Adding in the schedule description "this option will be can be enabled after the import" could be interesting

>  but I don't see why it should be highlighted.

The multi channel concept is quite different from other video platforms so we want the user to see it can upload the video in a particular channel. The privacy option is also essential, so the user understands at the first step it will upload this video in this particular privacy.

kontrollanten:
I've updated with the following changes:
* I reverted the commit which made it possible to schedule an import.
* Privacy options are now shown as a helper (which makes it possible to read the whole description for the alternatives).
* The schedule option is always shown, but disabled on those places where it previously wasn't shown.
* A short description is shown why it's disabled (e.g. "only available after import")

Maybe I did a more complex solution than necessary (it could have been done with just writing a general description beside the Privacy label), but I think it's a good step forward anyway.

Upload/import step 1:
upload-import-step-1

Import step 2:
import-step2
import-step2-options

Upload step 2:
upload-step2

Edit a public video:
edit-public-video

Edit a private video:
edit-pivate-video

shared/models/videos/video-privacy.enum.ts
```diff
@@ -1,4 +1,5 @@
 export const enum VideoPrivacy {
+  SCHEDULED = -1,
```
Chocobozzz, line 2: This is not a video privacy, then should not be added to this enum. It's the reason why we used `VideoEdit.SPECIAL_SCHEDULED_PRIVACY`

Chocobozzz (changes requested):

shared/models/videos/video-privacy.enum.ts
kontrollanten: True, but the form field name is "privacy". Can we rename the enum to `VideoPrivacyOption` or `PublishOption`?

Chocobozzz: This enum? No because it's used by the server, that only accept these values. But you can create another enum in the client named `PublishOption`.

kontrollanten:
I'll close this since I don't have time to finish it.

PR #3464, opened 2020-12-13
Register service worker

### LINKED ISSUE #296: Service Worker isn't registered
* **PeerTube version or commit:**
Peertube 0.0.24
* **Browser name/version:**
Google Chrome

Browsers should detect the service worker
SW isn't detected. Probably it's not registered.

_Lighthouse audit:_

kontrollanten:
## Description
Registers the ng service worker. I guess the asset names has changed since `ngsw-config.json` was written, therefore I just switched just to wildcard. 

## Related issues
Closes #296

## Has this been tested?

Open up DevTools in Brave and check in Network tab to see that JS and CSS assets are fetched via service worker.

## Screenshots
image

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -11,20 +11,9 @@ if (environment.production) {
 const bootstrap = () => platformBrowserDynamic()
   .bootstrapModule(AppModule)
   .then(bootstrapModule => {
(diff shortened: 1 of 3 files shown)


Chocobozzz:
Thanks

PR #3489, opened 2020-12-19
add `proxy_ignore_client_abort` flag to Nginx conf

### LINKED ISSUE #3484: HTTP 499 error during upload, and possible fix

A user was reporting upload errors. The first problem here is that the client gives basically no useful debug information, but with the user's help I traced it to a particular server-side log on nginx, showing an HTTP 499 code for the POST to the actual upload endpoint.

**Steps to reproduce:**

I don't have specific steps yet, it seems to occur only under some network configurations. The same evening this user was reporting trouble, I had been uploading successfully from elsewhere.

The first necessary improvement would seem to be better error reporting when an upload fails. But of course uploads should also not fail without good reason.

**Additional information**

While looking for possible causes of the HTTP 499, I hit on this post with a possible solution that seemed like it might apply to this use case: https://stackoverflow.com/a/52987394

I tried setting `proxy_ignore_client_abort on` and asked the user to retry, and this time it worked. It's just one data point, not a definitive explanation, but it seemed worth reporting. I suspected timeouts as well, but my timeout and max body size limits are all quite large, as the instance is configured for large video files.

```
location / {
    proxy_pass http://127.0.0.1:9000;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header Host $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;

    # Hard limit on uploads
    client_max_body_size 200G;
    proxy_connect_timeout       100m;
    proxy_send_timeout          100m;
    proxy_read_timeout          100m;
    send_timeout                100m;
    proxy_ignore_client_abort on;
  }
```

* PeerTube instance:
  * URL: https://diode.zone
  * version: 2.4.0
  * NodeJS version: 10.23.0
  * Ffmpeg version:

rigelk:
The `proxy_ignore_client_abort` flag specifies whether nginx will
monitor possible connection close while waiting for an upstream
server response. If an error occurs while sending a response, the
connection will be closed regardless of the flag, much like if
there where no nginx at all.

fixes #3484

### DIFF
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -158,6 +158,7 @@ server {
   keepalive_timeout         10s; # default is 75
   resolver_timeout          10s; # default is 30
   reset_timedout_connection on;
+  proxy_ignore_client_abort on;
 
   tcp_nopush                on; # send headers in one piece
   tcp_nodelay               on; # don't buffer data sent, good for small data bursts in real time


Chocobozzz:
Would `proxy_request_buffering` to `off` could actually fix https://github.com/Chocobozzz/PeerTube/issues/3484? I'm not against `proxy_ignore_client_abort`, but honestly I don't understand what it really do :thinking:

rigelk:
@Chocobozzz as mentioned in the issue, https://stackoverflow.com/a/52987394/12767986 explains it better than I could.

> HTTP 499 is a custom Nginx code used when the client's API closed the connection before it gets any response.

see https://www.belugacdn.com/499-error-code/, which indicates other cases could lead to a 499 than what `proxy_request_buffering` covers.

PR #3507, opened 2020-12-23
add script printing command to generate a resolution for a given file

rigelk:
having hints as to what PeerTube would do for a file/resolution could be interesting, especially for debugging and to explain to admins what the resolution transcoding profile is (potentially integrating the commands as examples in the administration configuration interface later).


Chocobozzz:
Please add a little unit test :)

PR #3508, opened 2020-12-23
authenticate upload requests before they transmit video

rigelk:
## Description

Adds a simple route to check for proper authentification before any traffic reaches the backend. The route is used internally and prevents requests to DDoS a PeerTube server with a large client body size.

## Has this been tested?

Ideally the app should be tested on a live instance running with the new route.

### DIFF
--- a/server/controllers/api/users/token.ts
+++ b/server/controllers/api/users/token.ts
@@ -35,6 +35,10 @@ tokensRouter.post('/scoped-tokens',
   asyncMiddleware(renewScopedTokens)
 )
 
+tokensRouter.get('/check-auth',
+  authenticate
+)
+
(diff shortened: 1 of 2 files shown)


Chocobozzz:
What's the use case? To prevent upload failure with the web client due to bad auth? Or another thing?

rigelk:
@Chocobozzz to prevent DDoS transferring up to 8Gb before the request is even authenticated.

Chocobozzz:
Okay then `proxy_request_buffering` to `off` could actually fix this issue in a better way?

Chocobozzz:
Closing due to inactivity. Feel free to reopen

Issue #3520, opened 2020-12-26
Accept trailing slash with `peertube auth add`


hopeseekr:
## The Problem:

When adding an authorization for the `peertube-cli`, one must not have anything after the domain_name:port.

    prompt: instance url:  https://www.example.com
    prompt: username:  hopeseekr
    prompt: password:  ************

You MUST NOT have a trailing `/` in your `instance url`, or it will always say `expected 200 "OK", got 404 "Not Found"`. It took me over 20 minutes to figure out that this was the problem.

See Issue #3091.

## The Solution

It'd be very very helpful for PeerTube to autoremove everything after and including the domain name:port.

I am going to submit a PR for this.

Issue #3091, opened 2020-08-15
Peertube CLI auth got forbidden


gai00:
Environment: docker - chocobozzz/peertube:production-buster
Version: 2.3.0

config:
```yaml
listen:
  hostname: '0.0.0.0'
  port: 9000
webserver:
  https: true
  hostname: my.host.name
  port: 443
# ...
```

Problem:
I try to upload my video via CLI, but got forbidden.
exec command in container. 
(`docker exec -it {container_id} bash`)
```
$ peertube auth add
prompt: instance url:  http://localhost:9000
prompt: username:  root
prompt: password:  ************

expected 200 "OK", got 403 "Forbidden"

curl localhost:9000
>> peertube html content
```

I tried `https://my.host.name`, lost certs in container...
I think https is no problem. but I want to upload without ssl. (maybe faster)

Chocobozzz:
Please fill the template. We need server logs

hopeseekr:
It took me FOREVER to figure out:

    prompt: instance url:  https://www.example.com
    prompt: username:  hopeseekr
    prompt: password:  ************

You MUST NOT have a trailing `/` in your `instance url`, or it will always say `expected 200 "OK", got 404 "Not Found"`.

It'd be very very helpful for PeerTube to autoremove everything after and including the domain name:port.

I am going to submit a PR for this.

PR #3521, opened 2020-12-27
cli `auth add`: do not fail with extraneous characters

hopeseekr:
## Description
**The Solution:**
I have hardened `auth add` by stripping out everything from the third '/' to the end of the instance URL.

**The Problem:**
When adding an authorization for the peertube-cli, before this commit you could not have anything after the domain_name:port.

For instance, if there was a trailing / in your instance URL, before this commit it will always fail with

    expected 200 "OK", got 404 "Not Found".

It took me over 20 minutes to figure out that this was the problem.

## Related issues

* Issue #3091: Peertube CLI auth got forbidden
* Issue #3520: Accept trailing slash with `peertube auth add`

## Has this been tested?

Please describe the tests that you ran to verify your changes.

I attempted to run the entire test suite, but after many failed starts and one really long one, I cancelled the run after it had been going 37+ minutes. I then attempted to run the CLI tests alone, but that, too, I cancelled after 27 minutes. I then commented out everything inside `server/tests/cli/index.ts` except the `peertube.ts`. I then attempted to run it via the command found in the `CONTRIBUTING.md`:

    TS_NODE_FILES=true npm run mocha -- --exit -r ts-node/register -r tsconfig-paths/register --bail server/tests/cli/index.ts 

but all I ended up with was:

     TypeError: Cannot read property 'app' of undefined

So yeah, I was unable to run the tests, but I did create tests for my own feature.

This is the first time I've ever attempted to run Mocha tests before, and I found it very confusing. I am used to the downright simplicity of `phpunit tests/MyTest.php` or even `phpunit --filter beginningOfMyTestCaseName`. I stumbled around worse than a drunken pirate fresh off a won parlay.

### DIFF
--- a/server/tests/cli/peertube.ts
+++ b/server/tests/cli/peertube.ts
@@ -66,6 +66,18 @@ describe('Test CLI wrapper', function () {
       await execCLI(`${env} ${cmd} auth add -u ${server.url} -U user_1 -p super_password`)
     })
 
+    it('Should not fail to add a user if there is a slash at the end of the instance URL', async function () {
+      this.timeout(60000)
+
+      const env = getEnvCli(server)
+      let fullServerURL
+      fullServerURL = server.url + '/'
+      await execCLI(`${env} ${cmd} auth add -u ${fullServerURL} -U user_1 -p super_password`)
+
+      fullServerURL = server.url + '/asdfasdf'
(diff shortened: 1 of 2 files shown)


rigelk:
@hopeseekr could you remove all semicolons to please the linter?

Chocobozzz:
Thanks @hopeseekr

Issue #534, opened 2018-05-04
Add ability to create friendly url for video links


dedesite:
I created this issue after reading [this french comment](https://framacolibri.org/t/ma-premiere-video-mise-sur-le-net-trop-bien-inclues-qq-infos-utiles-ou-pas/2855)  about peertube on framacolibri.

I think it would be create to add the ability to create friendly url for an uploaded.

Something similar as you could find on a CMS like wordpress.

* **PeerTube URL:**  https://peertube.video

* **What is the expected behaviour?**

Be able to create link like  https://peertube.video/videos/watch/my-super-video

Of course, if an url is already taken, then the upload interface tell the user to choose another url.
* **What do you see instead?**

Only link like https://peertube.video/videos/watch/10c1c52f-82f4-42b7-918b-93e3ad8760fc

henke37:
Might need some permission checking to prevent abuse, but this is a good idea. If there is a risk of conflict, simply require that the path is prefixed with the username.

DavidLibeau:
We should add friendly url for accounts to (http://peertube.video/@username and/or http://peertube.video/accounts/username) !

Chocobozzz:
@DavidLibeau It's already the case (in develop for now) https://github.com/Chocobozzz/PeerTube/issues/584#issuecomment-391985410

albjeremias:
its really a very big URL for sharing a link from peertube.. and ugly.. can't we just create an automatic slug from the title of the video? or have a smaller identifier? this hash is huge and ugly

magicbrothers:
@Chocobozzz at least my MR/PR was not useless if this 'bad practice' will be fixed now ;)
Would it be an option to use (the first x characters of) a salted hash of that interal ID in the URL? We'd need a reverse lookup table for that, not pretty, but maybe required. Or, as @rigelk suggested:
> But if you add another field for videos, a very short random string could do (but would require adding it to the API, which is a bit more work)

magicbrothers:
Or another, maybe better idea:
Use the internal ID in the link but append a 'crc32(id + secret)', so that a guessing of unlisted videos isn't possible.
What about that, @Chocobozzz , may I try to implement that?

magicbrothers:
Thanks, @kontrollanten, for your support!
Unfortunately I am not only new to PeerTube development, but also to NodeJS :D (But not to programming in general)
So probably the best way for me is to wait for the implementation of #3680, so I can look through those commits and learn about how all that stuff works.
[...]

PR #3543, opened 2021-01-07
Add short url option to share menu, enhances #534

magicbrothers:
## Description

Creates a new tab in the share menu that displays a link using the instance's interal video ID instead of the UUID.
It's useful if you need shorter links.

## Related issues

#534 Add ability to create friendly url for video links

## Has this been tested?

Link creation works and creates valid urls. I can't test it for federated videos, but that should work, because I saw (using an API call like /videos/uuid-of-remote-video), that even a remote video has an id on the viewer's server.

### DIFF
--- a/client/src/app/shared/shared-share-modal/video-share.component.html
+++ b/client/src/app/shared/shared-share-modal/video-share.component.html
@@ -77,6 +77,16 @@
           </ng-template>
         </ng-container>
 
+        <ng-container ngbNavItem="shortUrl">
+          <a ngbNavLink i18n>Short URL</a>
+
+          <ng-template ngbNavContent>
+            <div class="nav-content">
+              <my-input-toggle-hidden [value]="getVideoShortUrl()" [withToggle]="false" [withCopy]="true" [show]="true" [readonly]="true"></my-input-toggle-hidden>
+            </div>
+          </ng-template>
+        </ng-container>
+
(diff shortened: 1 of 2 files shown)


kontrollanten:
Nice feature! Why isn't this link always used (instead of the UUID one)?

rigelk:
@kontrollanten you can search by UUID, and it's bad practice to expose internal database IDs.

kontrollanten:
> it's bad practice to expose internal database IDs.

That's what I expected. Then I guess creating short links with internal db id is bad practice as well? I guess most users feel safe to publish unlisted videos and be sure that probably no one will guess the link to the video. But if we start expose the internal db id then it will be really easy to "brute force" for new links (for example if someone sees that 1001 and 1003 is published, then they can guess 1002 which is unlisted).

rigelk:
@kontrollanten exactly. Sloving #534 cannot be done this way ; beyond chosen urls, an automatic shortened url could be done via much simpler means. https://github.com/oculus42/short-uuid comes to mind since it is fully translatable to/from our UUID naming scheme. But if you add another field for videos, a very short random string could do (but would require adding it to the API, which is a bit more work).

Chocobozzz:
Hello,

Sorry @magicbrothers but unfortunately it's not a good idea to expose this internal ID in the URL. We also may remove its support in the future.

Issue #3544, opened 2021-01-07
Allow to not log ping requests


johanfleury:
**Describe the problem to be solved**

It can happen that a process crashes in a way where it gets stuck but doesn’t exits. For this reason Docker (and Kubernetes) has a feature called [health check](https://docs.docker.com/engine/reference/builder/#healthcheck) that allows to check that the process is still working as expected (and restart the container if not).

This can be done by a script doing some hard work, but most of the time it’s just about curl’ing a simple API endpoint that replies with HTTP status code 200.

PeerTube doesn’t have such an health check endpoint.

**Describe the solution you would like:**

This endpoint should do the following:

* Reply with status code 200
* (_Optional_) Reply with a JSON body describing the current status (it’s often just `{"status": "ok"}`)
* Avoid logging request either at the discretion of the administrator (with a specific setting in the configuration files) or at the discretion of the client (with a query parameter e.g. `<endpoint>?silent=true`)

The path of the endpoint could be something like `/health/live` (in line with how Kubernetes names this feature, and would allow for an eventual future implementation of a “readiness” use this feature and allow a `/health/ready`

**Describe alternatives you have considered**

I am currently using `/` to check if PeerTube is still alive, but the main issue is that it logs every single request:

```
$ k logs -n peertube peertube-0 peertube | grep -m 1 kube-probe
[***:443] 2021-01-07 14:38:57.755 info: 10.2.0.1 - - [07/Jan/2021:14:38:57 +0000] "GET / HTTP/1.1" 200 3997 "-" "kube-probe/1.20"
$ k logs -n peertube peertube-0 peertube | grep -c kube-probe
1048
```

kontrollanten:
Hi! Have you seen that `/api/v1/ping` exists? But I think it still logs the requests.

johanfleury:
Hey, no I didn’t saw it.

Regarding logging, I’ve looked at `morgan`’s documentation and I think I can come up with a PR.

PR #3550, opened 2021-01-07
Allow administrator to disable logging of ping requests

johanfleury:
## Description

Allow administrator to disable logging of ping requests on `/api/v1/ping`.

This is disabled by default, but can be enabled by setting `log.log_ping_requests` to `true` (or the `PEERTUBE_LOG_PING_REQUESTS` environment variable in Docker).

## Related issues

Implements #3544

## Has this been tested?

Not sure how I can implement unit test for that.

### DIFF
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -100,6 +100,7 @@ log:
     maxFileSize: 12MB
     maxFiles: 20
   anonymizeIP: false
+  log_ping_requests: true
 
 trending:
   videos:
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -100,6 +100,7 @@ log:
     maxFileSize: 12MB
     maxFiles: 20
   anonymizeIP: false
+  log_ping_requests: true
 
 trending:
   videos:
--- a/server.ts
+++ b/server.ts
@@ -158,7 +158,10 @@ morgan.token('user-agent', (req: express.Request) => {
   return req.get('user-agent')
 })
 app.use(morgan('combined', {
-  stream: { write: logger.info.bind(logger) }
+  stream: { write: logger.info.bind(logger) },
+  skip: function (req, res) {
+    return (req.path === '/api/v1/ping' && CONFIG.LOG.LOG_PING_REQUESTS === false)
+  },
 }))
 
 // For body requests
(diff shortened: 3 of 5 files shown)


johanfleury:
Ok, so I’ve opened this PR as a draft because I’m not sure if this is the correct way to do this. It should work, but it feels a bit messy (like what if we add a new path?)

Also we let the administrator configure this feature?

kontrollanten:
I think it looks like a good start! If we want to add a new path it can be `/api/v1/ping/email` and then just check for the beginning of the path to determine.

But maybe some people use the request logging to monitor potential attacks, then I guess this can be a bad solution.

Chocobozzz:
Hello,

It's should not be the responsibility of the client to decide to log or not the request. So please instead of using a query parameter, add a config key below https://github.com/Chocobozzz/PeerTube/blob/develop/config/production.yaml.example#L102 `logPingRequest: true`

rigelk:
Why not just whitelist the IP emitting the health check?

johanfleury:
@Chocobozzz:
>  So please instead of using a query parameter, add a config key below https://github.com/Chocobozzz/PeerTube/blob/develop/config/production.yaml.example#L102 logPingRequest: true

I chose to use snake case (i.e. `log_ping_requests`) instead of camel case to be more consistent with other configuration options.

@kontrollanten:
> But maybe some people use the request logging to monitor potential attacks, then I guess this can be a bad solution.

I don’t see how having logs disabled on that specific endpoint allows any attack. I also doubt many PeerTube instance in production are facing the internet directly. Most of them are probably deployed behind a reverse proxy with access logs enabled.

Anyway this is now at the administrator discretion.

@rigelk
> Why not just whitelist the IP emitting the health check?

I feel like this would be a bit too much for just disabling logs from this endpoint.

Also, in some environments (e.g. Docker and Kubernetes) the source IP address is dynamic so you would end up allowing a subnet for the whole cluster which would have the same effect as just disabling logging on that endpoint

johanfleury:
The test failing doesn’t seem to be related with my changes.

config/production.yaml.example
```diff
@@ -100,6 +100,7 @@ log:
     maxFileSize: 12MB
     maxFiles: 20
   anonymizeIP: false
+  log_ping_requests: false
```
Chocobozzz, line 103: Let this to `true`, and if the admin uses it for health check it can disable it

Chocobozzz (changes requested):

config/production.yaml.example
johanfleury: Fixed

Chocobozzz:
Thanks @johanfleury

PR #3561, opened 2021-01-08
Warn/exit on unavailable port for RTPM/HTTP

### LINKED ISSUE #3558: Lack of errors when the port 1935 is already in use

If there is already another RTMP server listening on port 1935, Peertube start without any error. The only log is «Running RTMP server on port 1935», and peertube seems happy. But in fact peertube will not receive any connection.

**Steps to reproduce:**

1. Stop peertube
2. Setup a RTMP server on port 1935 (see below for a nginx snippet to do this). 
3. Start peertube.

For setting up a RTMP server with nginx, you can add this in your /etc/nginx/nginx.conf:

```
rtmp {
  server {
        listen 1935;
        chunk_size 4096;
        
        application live {
                live on;
                interleave on;
                record all;
                record_path /a/path/where/to/record/;
                record_unique on;
        }
    }
}
```

With this config, nginx will record any RTMP stream in the chosen folder. So you can easily check if it is nginx or peertube that receive the stream (if there is something recorded, it is nginx, otherwise it is peertube).

Peertube should detect that the port is already in use, log an error, and crash.

**Additional information**

* PeerTube instance:
  * version: 3.0.0

rigelk:
## Related issues

closes #3558

## Has this been tested?

I tested port messages via `ffmpeg -listen 1 -i rtmp://localhost:1935`.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -123,6 +123,7 @@ import { Hooks } from './server/lib/plugins/hooks'
 import { PluginManager } from './server/lib/plugins/plugin-manager'
 import { LiveManager } from './server/lib/live-manager'
 import { HttpStatusCode } from './shared/core-utils/miscs/http-error-codes'
+import { isPortTaken } from '@server/helpers/utils'
 
 // ----------- Command line -----------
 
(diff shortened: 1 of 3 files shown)


Chocobozzz:
Thanks, but implemented in https://github.com/Chocobozzz/PeerTube/commit/c2b823826db167594f8157b0b324b6fb1559caeb in a simpler way

PR #3563, opened 2021-01-08
Mark 'These settings apply only to' string as i18n-able

mkljczk:
## Description

Marked a string as i18n-able, spotted in production.

## Has this been tested?


rigelk:
Thanks @mkljczk!

PR #3564, opened 2021-01-09
show first decimal for views above a thousand

### LINKED ISSUE #3549: Display total views count
As soon as views count reaches 1 000 it will get shortened to 1K. The users won't see any updated view count until there's 1 000 more views. Since the view count is an important point for the users (and publishers) I think it's also important to let them see how it's changing.

There's a lot of space after the views count, so it shouldn't be any problem to show the exact view count up to 99 999.

rigelk:
Shows the decimal above 1000 views (which is then formated as 1K and otherwise was truncated). 

## Related issues

closes #3549

## Screenshots

Screenshot_2021-01-09 Trending videos - PeerTube

### DIFF
--- a/client/src/app/shared/shared-main/angular/number-formatter.pipe.ts
+++ b/client/src/app/shared/shared-main/angular/number-formatter.pipe.ts
@@ -10,10 +10,25 @@ export class NumberFormatterPipe implements PipeTransform {
     { max: 1000000000, type: 'M' }
   ]
 
+  /**
+   * @param x number
+   * @param n number of decimals to get (defaults to 1, needs to be >= 1)
+   */
+  static getDecimalForNumber (x: number, n = 1) {
+    const v = x.toString().split('.')
+    const f = v[1] || ''
+    if (f.length > n) return +f.substr(0,n)
+    return +f
+  }
+
   transform (value: number) {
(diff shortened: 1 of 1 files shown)


tcitworld:
My idea was only when the number is < 10K, I find it too long in your screenshot.

rigelk:
Screenshot_2021-01-09 Trending videos - PeerTube(1)

client/src/app/shared/shared-main/angular/number-formatter.pipe.ts
```diff
@@ -10,10 +10,25 @@ export class NumberFormatterPipe implements PipeTransform {
...
   ]
 
+  /**
+   * @param x number
+   * @param n number of decimals to get (defaults to 1, needs to be >= 1)
+   */
+  static getDecimalForNumber (x: number, n = 1) {
+    const v = x.toString().split('.')
+    const f = v[1] || ''
+    if (f.length > n) return +f.substr(0,n)
```
Chocobozzz, line 20: small missing space before `n)`

rigelk:
```suggestion
    if (f.length > n) return +f.substr(0, n)
```

Chocobozzz:
Thanks

Issue #3486, opened 2020-12-19
Adding More Search Functionality


King-Noctis:
Can you...

### Add Search (Bar) for Playlist within Playlists (example within Watch Later...)
![SharedScreenshot6](https://user-images.githubusercontent.com/76251188/102689663-dee4b900-423a-11eb-827d-3b90ae2cc7e6.jpg)

### Add Search (Bar) for History
![SharedScreenshot7](https://user-images.githubusercontent.com/76251188/102689782-c7f29680-423b-11eb-81e4-b6fced48f4a6.jpg)

### Add Search (Bar) for Subscriptions (Videos from your subscriptions)
![SharedScreenshot3](https://user-images.githubusercontent.com/76251188/102689706-35ea8e00-423b-11eb-89c1-93e8a4f04ab0.jpg)

### Add Search (Bar) for Account/Channel Page (VIDEOS CHANNELS ; VIDEOS)
- ![SharedScreenshot](https://user-images.githubusercontent.com/76251188/102689787-d5a81c00-423b-11eb-9a7f-1ce2055c9665.jpg)
- ![SharedScreenshot2](https://user-images.githubusercontent.com/76251188/102689790-dc369380-423b-11eb-91c9-095514600204.jpg)

Thanks.

PR #3576, opened 2021-01-10
Allow user to search through their watch history

rigelk:
## Description

Adds support for a `search` query parameter and adds a corresponding search field within the history interface.
Some interface corrections applied to the history interface.

## Related issues

related to meta #3486

## Has this been tested?

## Screenshots

Screenshot_2021-01-10 My video history - PeerTube

### DIFF
--- a/client/src/app/shared/shared-main/users/user-history.service.ts
+++ b/client/src/app/shared/shared-main/users/user-history.service.ts
@@ -23,7 +23,7 @@ export class UserHistoryService {
 
     let params = new HttpParams()
     params = this.restService.addRestGetParams(params, pagination)
-    params = this.restService.addObjectParams(params, { search })
+    params = params.append('search', search)
 
     return this.authHttp
                .get<ResultList<Video>>(UserHistoryService.BASE_USER_VIDEOS_HISTORY_URL, { params })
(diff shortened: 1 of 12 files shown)


client/src/app/shared/shared-main/users/user-history.service.ts
```diff
@@ -18,11 +18,12 @@ export class UserHistoryService {
     private videoService: VideoService
   ) {}
 
-  getUserVideosHistory (historyPagination: ComponentPaginationLight) {
+  getUserVideosHistory (historyPagination: ComponentPaginationLight, search?: string) {
     const pagination = this.restService.componentPaginationToRestPagination(historyPagination)
 
     let params = new HttpParams()
     params = this.restService.addRestGetParams(params, pagination)
+    params = this.restService.addObjectParams(params, { search })
```
Chocobozzz, line 26: `params = params.append('search', search)`

Chocobozzz (changes requested):

client/src/app/shared/shared-main/users/user-history.service.ts
rigelk:
```suggestion
    params = params.append('search', search)
```

PR #3594, opened 2021-01-13
Use correct English (About page)

mkljczk:
## Has this been tested?

### DIFF
--- a/client/src/app/+about/about-follows/about-follows.component.html
+++ b/client/src/app/+about/about-follows/about-follows.component.html
@@ -13,7 +13,7 @@
   </div>
 
   <div class="col-xl-6 col-md-12">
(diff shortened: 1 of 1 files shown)


client/src/app/+about/about-follows/about-follows.component.html
```diff
@@ -13,7 +13,7 @@ <h2 i18n class="subtitle">Followers instances ({{ followersPagination.totalItems
   </div>
 
   <div class="col-xl-6 col-md-12">
-    <h2 i18n class="subtitle">Followings instances ({{ followingsPagination.totalItems }})</h2>
```
Chocobozzz, line 16: Just use Followings please

Chocobozzz (changes requested):

Issue #3593, opened 2021-01-13
Implement remote interaction


toby3d:
PeerTube instances support the ActivityPub protocol, allowing you to post comments to videos from Mastodon/Pleroma. But for some reason I can't publish comments in the PeerTube B instance from my account in the PeerTube A instance. For example: post comment [here](https://video.pony.gallery/videos/watch/a018eade-c00e-4914-b88c-4a0516922b67) (video posted on `video.pony.gallery`) via `toby3d@video.toby3d.me` [actor](https://video.toby3d.me/accounts/toby3d) on `video.toby3d.me`. I'm not quite sure if this is a bug or a feature?

Chocobozzz:
Hello

I was able to comment this video using my mastodon account (framapiaf.org) and my peertube account (peertube2.cpy.re). Maybe the admin/uploader deleted your comment, or muted your instance or your account? Or maybe there is a bug, but we would need you to contact the admin of video.pony.gallery to be sure

toby3d:
After entering the PeerTube username and clicking on the "remote interaction" button, I find the following error log in the console. The funny thing is that when using the Mastodon username, everything works fine.
<details>
  <summary>Firefox console error message</summary>

  ```javascript
undefined 3-es2015.3ade6778933a1a900303.js:1:73446
    formValidated https://video.pony.gallery/client/ru-RU/3-es2015.3ade6778933a1a900303.js:1
    invoke https://video.pony.gallery/client/ru-RU/polyfills-es2015.6eeee87a2e05401a2cbd.js:1
    onInvoke https://video.pony.gallery/client/ru-RU/main-es2015.6e2be889cf2324c0ce50.js:3
    invoke https://video.pony.gallery/client/ru-RU/polyfills-es2015.6eeee87a2e05401a2cbd.js:1
    run https://video.pony.gallery/client/ru-RU/polyfills-es2015.6eeee87a2e05401a2cbd.js:1
    O https://video.pony.gallery/client/ru-RU/polyfills-es2015.6eeee87a2e05401a2cbd.js:1
    invokeTask https://video.pony.gallery/client/ru-RU/polyfills-es2015.6eeee87a2e05401a2cbd.js:1
    onInvokeTask https://video.pony.gallery/client/ru-RU/main-es2015.6e2be889cf2324c0ce50.js:3
    invokeTask https://video.pony.gallery/client/ru-RU/polyfills-es2015.6eeee87a2e05401a2cbd.js:1
[...]

Issue #2888, opened 2020-06-20
Remote Subscribe with Any PeerTube Account


Malsasa:
I am a user and not a developer. I want an ability to subscribe to remote channel with user's PeerTube account regardless their instance instead of "an account on this instane" which leads him to making a new account instead of subscribing. If Mastodon's supported, I believe PeerTube should be supported too. 

**What happened?**
Subscribe options look like screenshot below.![1](https://user-images.githubusercontent.com/3586695/85209452-615eef80-b362-11ea-8c4b-da994cc0fd61.png)

**What do you expect to happen instead?**
It should be "Subscribe with a PeerTube account" instead of "...account on this instance" that supports any user's instance. This is my simple mockup of my feature request:
![sub](https://user-images.githubusercontent.com/3586695/85210004-b997f080-b366-11ea-865f-e8c313927e4c.png)

**Steps to reproduce:**
In my opinion, it should be:
1. User clicks Subscribe button.
2. User types his PeerTube username.
3. User sees authentication with **Remote Subscribe** button (similar to Mastodon's one).
4. User clicks Subscribe button (similar to Mastodon's one).
5. User subscribed.

Thank you very much for PeerTube and this is my first issue report.

jeena:
This is one is also one of my biggest hurdles when it comes to recommending PeerTube to my friends. I have explained how to do that right now in https://tube.jeena.net/videos/watch/b9da3790-6710-415d-b610-45762c1a007b

r4dh4l:
> This is one is also one of my biggest hurdles when it comes to recommending PeerTube to my friends.

I would even say it is a showstopper. I took me months to convince some minor institutions to give PeerTube a try (as alternative to YouTube). Recently they asked me: "Ok, we have a PeerTube channel now - how we can subscribe to yours?"

I'm very thankful for projects like PeerTube but sometimes I wonder if people developing these projects are too "far away" from how "average people" are used to use web services (or the other way around). I have problems to understand how something like "subscibing accounts on other servers" is an issue *in 2020* (please explain me, I would really like to understand).

However: Thanks for developing PeerTube!

rigelk:
@r4dh4l : what are your UI suggestions?

r4dh4l:
> what are your UI suggestions?

A subscribe button which is *not* a dropdown menu that behaves in the following way:

- if you are logged in on the same server the account/channel you want to subscribe it will be directly subscribed
[...]

PR #3598, opened 2021-01-14
Feature/remote interaction

Chocobozzz:
Implement remote interaction (comment & subscribe):

 * Added a `http://ostatus.org/schema/1.0/subscribe` `rel` in webfinger for actors
 * Added a page to receive these interactions (`/remote-interaction` page):
   * If the user is not logged in, the instance will redirect to the login page (and then re redirect to the remote interaction page)
   * If the user is logged in, the page will fetch the remote object and redirect the user on it (channel page/video page)

The subscribe button redesign will be discussed later

Closes https://github.com/Chocobozzz/PeerTube/issues/3593
Closes https://github.com/Chocobozzz/PeerTube/issues/2888

### DIFF
--- a/client/src/app/+remote-interaction/remote-interaction-routing.module.ts
+++ b/client/src/app/+remote-interaction/remote-interaction-routing.module.ts
@@ -1,0 +1,23 @@
+import { NgModule } from '@angular/core'
+import { RouterModule, Routes } from '@angular/router'
+import { LoginGuard } from '@app/core'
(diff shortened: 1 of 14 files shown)


rigelk:
Inputing a valid username (for the input validator anyway) doesn’t mean the remote supports webfinger. If it returns an error (for instance a CORS deny) it won't report it to the user but just in the console.

PR #3616, opened 2021-01-18
Patch 1

hspeijer:
## Description

Small documentation change to ensure correct version of ffmpeg on ubuntu 18

## Has this been tested?

### DIFF
--- a/support/doc/dependencies.md
+++ b/support/doc/dependencies.md
@@ -39,7 +39,7 @@ Follow the below guides, and check their versions match [required external depen
 sudo snap install ffmpeg
 ```
 
-5. Run:
+6. Run:
 
 ```
 sudo apt update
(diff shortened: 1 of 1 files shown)


rigelk:
hi @hspeijer,

I don't think we should be reliant on snap, unless this is the way Ubuntu advises to install a recent enough ffmpeg version.

Chocobozzz:
@rigelk AFAIK ubuntu 18 only has ffmpeg 3.x, so not compatible with our HLS option. Snap seems the recommended way to install packages on ubuntu no?

tcitworld:
Ubuntu 18.04 is not the latest LTS though.

hspeijer:
I am not sure if ubuntu 20 does come with a newer version of ffmpeg. In any case I managed to make it work using this page (https://launchpad.net/~savoury1/+archive/ubuntu/ffmpeg4) which might be worthy to be included in the documentation as a side note for ubuntu 18 which is still being used a lot.

Issue #3488, opened 2020-12-19
Make instances feel diffrent


Openmedianetwork:
Using simple view counts over time across an instance will bring the same content to the top on every instance - we already have a VERY bad spam problem due to the backend tools. Let's come up with creative ideas for mediating this.

Issue: currently local content loses out to popular external content in the most liked and trending.

Solution: add a UI or plugin admin tick option to give this a say 50% boost in likes/views when calculating display order so that it more easily pushes up into the federated popular content on a local instance

Outcome: this will add local "flaver" to instances they often lack, due to being swamped by generic content from across the network.

Have this as a UI option or plugin.

test2a:
What you are describing is what Reddit mods used to do. I used to read how they could rebalance or whatever. Dunno what tangible affect will they  have because you are intentionally gaming users into interaction

test2a:
https://www.theverge.com/2016/12/6/13862042/reddit-upvote-downvote-scoring-system-recalculation


Read up on this.

Chocobozzz:
Hello,

> we already have a VERY bad spam problem due to the backend tools

Could you elaborate? I'm not sure to understand what it means.

Openmedianetwork:
As lots of people talk about peertube instances have very bad spam problems in video content flows.
The implementation has suffered from the #geekproblem in that they have a #closed mod workflow rather than a #open trust based workflow. This is a normal issue and a clear #block  on the rollout of #openweb projects.

Openmedianetwork:
> 
> 
> https://www.theverge.com/2016/12/6/13862042/reddit-upvote-downvote-scoring-system-recalculation
> 
> Read up on this.

What am suggesting is different to this, not adjusting the votes or view counts, `simply` Appling a mod option to make local content show up with popular external content to make different instances feel individual rather than them all looking the same, with the same content displayed in popular and liked views - if you are worried about the wording then change it to "on this instance" if the mod is applied. have this as a UI or a plugin option.

test2a:
There is a tab called "local" that only shows local content. Wouldnt that solve your issue?

Openmedianetwork:
[...]

PR #3625, opened 2021-01-22
add variant of the "hot" trending algorithm

rigelk:
## Description

The current changes don't replace the "Trending" page or behaviour, but merely add a "Hot" variant inspired by [Reddit's own algorithm](https://github.com/iangreenleaf/reddit/blob/45e8209d8d4236367a6f7247068c13ab2307afb4/r2/r2/lib/db/_sorts.pyx#L45), albeit extended for video to integrate views and comments.

Another change is the modularization of abstract video list titles. Simply setting a string is not enough for more involved and varying interfaces.

## Related issues

No issue focused on changing the trending algorithm, just on "customizing" the landing page.
It could be related to  #3488 as in standardizing scores will lead up to allowing options to how they are computed: potentially pushing forward local videos if need be.

## Has this been tested?

- [x] 👍 only light tests as follows for now. any other required?

I tested queries on sherri.

## Screenshots

Screenshot_2021-01-20 Trending videos - PeerTube

## Questions

Should I create a dedicated route for the algorithm, much like "trending" has one?


test2a:
Can we have a live separate like this? Instances are starting to get v3 but no way to specify livestreams yet.
Same for search in the instance and sepia search. I think there is already an open issue for that one.

server/models/video/video-query-builder.ts
```diff
@@ -239,16 +241,65 @@ function buildListQuery (model: typeof Model, options: BuildVideosQueryOptions)
...
+     * see https://github.com/reddit-archive/reddit/blob/master/r2/r2/lib/db/_sorts.pyx#L47-L58
+     */
+    const weights = {
+      like: 3,
+      dislike: 3,
+      view: 1 / 12,
+      comment: 2 // a comment takes more time than a like to do, but can be done multiple times
+    }
+
+    cte.push( // TODO: exclude blocklisted comments
```
rigelk, line 274: @Chocobozzz I might need your advice on how to include the blocklist here. I took inspiration from `server/models/video/video-comment.ts` but I fall short on how to adapt the use of user blocklists as in https://github.com/Chocobozzz/PeerTube//blob/b49f22d8f9a52ab75fd38db2d377249eb58fa678/server/models/video/video-comment.ts#L70

rigelk: Unless you want to only use the server blocklist, in order to keep a cacheable version of the route?

client/src/app/+videos/video-list/trending/video-trending-header.component.ts
```diff
@@ -0,0 +1,59 @@
...
+  selector: 'video-trending-title-page',
+  host: { 'class': 'title-page title-page-single' },
+  styleUrls: [ './video-trending-header.component.scss' ],
+  templateUrl: './video-trending-header.component.html'
+})
+export class VideoTrendingHeaderComponent extends VideoListHeaderComponent {
+  buttons: VideoTrendingHeaderItem[]
+
+  constructor (
+    @Inject('data') public data: any,
```
Chocobozzz, line 25: You cannot type `data`?

client/src/app/shared/shared-video-miniature/video-list-header.component.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export abstract class GenericHeaderComponent {
+  constructor (@Inject('data') public data: any) {}
+}
+
+@Component({
+  selector: 'h1',
+  host: { 'class': 'title-page title-page-single' },
+  template: `
+<div placement="bottom" [ngbTooltip]="data.titleTooltip" container="body">
```
Chocobozzz, line 11: Please create an `html` template

client/src/app/shared/shared-video-miniature/video-list-header.component.ts
```diff
@@ -0,0 +1,20 @@
+import { Component, Inject } from '@angular/core'
+
+export abstract class GenericHeaderComponent {
+  constructor (@Inject('data') public data: any) {}
+}
+
+@Component({
+  selector: 'h1',
```
Chocobozzz, line 8: Not sure to understand this `h1`

server/models/video/video-query-builder.ts
```diff
@@ -239,16 +241,65 @@ function buildListQuery (model: typeof Model, options: BuildVideosQueryOptions)
...
     const viewsGteDate = new Date(new Date().getTime() - (24 * 3600 * 1000) * options.trendingDays)
 
     joins.push('LEFT JOIN "videoView" ON "video"."id" = "videoView"."videoId" AND "videoView"."startDate" >= :viewsGteDate')
     replacements.viewsGteDate = viewsGteDate
 
-    attributes.push('COALESCE(SUM("videoView"."views"), 0) AS "videoViewsSum"')
+    attributes.push('COALESCE(SUM("videoView"."views"), 0) AS "score"')
 
     group = 'GROUP BY "video"."id"'
+  } else if (options.hot && options.isCount !== true) {
```
Chocobozzz, line 254: Wrap this `if` in a `if (options.isCount !== true)`...

server/models/video/video-query-builder.ts
Chocobozzz: create the blocklist ids in the cte, and then use it here to exclude author id that are in the created table

server/models/video/video-query-builder.ts
```diff
@@ -239,16 +241,65 @@ function buildListQuery (model: typeof Model, options: BuildVideosQueryOptions)
...
+     */
+    const weights = {
+      like: 3,
+      dislike: 3,
+      view: 1 / 12,
+      comment: 2 // a comment takes more time than a like to do, but can be done multiple times
+    }
+
+    cte.push( // TODO: exclude blocklisted comments
+      '"totalCommentsWithoutVideoAuthor" AS (' +
```
Chocobozzz, line 275: I don't think it will scale: you're building a temp' table for every videos and comments in the DB.

Chocobozzz (changes requested):

Chocobozzz:
Please add ability for admin to enable/disable trending algorithm + default one

client/src/app/shared/shared-video-miniature/video-list-header.component.ts
rigelk: it is just the outer tag for the component

server/models/video/video-query-builder.ts
rigelk: You're right. Maybe as a computed field on video?

client/src/app/shared/shared-video-miniature/video-list-header.component.ts
Chocobozzz: `h1`? But it's already an existing html tag. We should create dedicated selector with `my-` prefix

server/models/video/video-query-builder.ts
Chocobozzz: No sorry, we can't store and manage another field just for 1 specific sort of 1 query

rigelk: Ok, I'll simplify then.

rigelk: Should be better.

client/src/app/+videos/video-list/trending/video-trending-header.component.ts
Chocobozzz: ping

config/default.yaml
```diff
@@ -365,6 +365,10 @@ instance:
 #    - 18 # Food
 
   default_client_route: '/videos/trending'
+  default_trending_route: '/videos/most-viewed'
+  pages:
```
Chocobozzz, line 369:
```
algorithms:
  enabled: 
    - 'views-interval' # Default one, used by PeerTube since the beginning
    - 'reddit-hot' # Takes into account...
  default: 'views-interval'
```

This way we can easily add algorithms in the future. Then, you just build the default trending route yourself depending on the default algorithm. You could just use a query parameter in the client to define the chosen algo: `/videos/trending?algorithm=hot`

client/src/app/+videos/video-list/trending/video-trending-header.component.ts
```diff
@@ -0,0 +1,78 @@
...
+          // don't filter if auto-blacklist is not enabled as this will be the only list
+          if (config.instance.pages.hot.enabled) {
+            const index = this.buttons.findIndex(b => b.path === 'hot')
+            this.buttons[index].hidden = false
+          }
+        })
+  }
+
+  get visibleButtons () {
+    return this.buttons.filter(b => !b.hidden)
```
Chocobozzz, line 71: **don't use a getter that build an array**

Chocobozzz (changes requested):

Chocobozzz:
Please add tests

PR #3631, opened 2021-01-24
use yarn instead of npm

pietsch:
The current documentation instructs instance admins to install yarn, not npm.

## Has this been tested?

### DIFF
--- a/support/systemd/peertube.service
+++ b/support/systemd/peertube.service
@@ -8,7 +8,7 @@ Environment=NODE_ENV=production
 Environment=NODE_CONFIG_DIR=/var/www/peertube/config
 User=peertube
 Group=peertube
-ExecStart=/usr/bin/npm start
+ExecStart=/usr/bin/yarn start
 WorkingDirectory=/var/www/peertube/peertube-latest
 StandardOutput=syslog
 StandardError=syslog


pietsch:
You probably forgot to update the systemd service file when you switched from npm to yarn.

I just installed PeerTube 3.0.1 on a fresh Debian 10 VM and had to make this change to make the peertube service start.

rigelk:
`npm` is installed along with node, and while we use `yarn` for package management, it isn't required for running scripts, which `start` is part of.

pietsch:
In this case, you should fix the documentation to make it clear that `npm` is required.

When I installed `yarn`, I did not install `npm`. I did what is recommended on the [yarn site](https://classic.yarnpkg.com/en/docs/install#debian-stable):
```
sudo apt update && sudo apt install --no-install-recommends yarn
```

rigelk:
@pietsch how did you install node on you machine? Package managers install npm alongside node, but YMMV.

pietsch:
I simply did `apt install nodejs` because Debian 10 happens to ship node version 10.x as requested by the PeerTube documentaion.

pietsch:
At least on Debian, `npm` is just a package `nodejs` suggests, so it is not installed by default when you install `nodejs`.

PR #3632, opened 2021-01-24
redirect to login on 401, display 403 variant

rigelk:
## Description

Outside of components using `LoginGuard`, we can detect which objects require to be logged in based on their return code, which `AuthInterceptor` can help us redirect. This is especially useful for internal videos, as mentioned in https://framacolibri.org/t/avoir-le-bouton-de-login-sur-lerreur-404/10761.

## Has this been tested?

This is done client-side only, relying on existing server codes, so no tests were added to the CI harness. However, I tested manually for internal, private and blocked videos.

## Screenshots

Screenshot_2021-01-24 PeerTube

### DIFF
--- a/client/src/app/+page-not-found/page-not-found.component.html
+++ b/client/src/app/+page-not-found/page-not-found.component.html
@@ -1,20 +1,29 @@
 <div class="root">
-  <div *ngIf="status === 404" class="box">
+  <div *ngIf="status !== 403 && status !== 418" class="box">
     <strong>{{ status }}.</strong>
     <span class="ml-1 text-muted" i18n>That's an error.</span>
 
     <div class="text mt-4" i18n>
-      We couldn't find any ressource tied to the URL {{ pathname }} you were looking for.
+      We couldn't find any {{ getRessourceName() }} tied to the URL {{ pathname }} you were looking for.
     </div>
 
     <div class="text-muted mt-4">
       <span i18n="Possible reasons preceding a list of reasons a `Not Found` error page may occur">Possible reasons:</span>
 
       <ul>
-        <li i18n>The page may have been moved or deleted</li>
         <li i18n>You may have used an outdated or broken link</li>
+        <li i18n>The {{ getRessourceName() }} may have been moved or deleted</li>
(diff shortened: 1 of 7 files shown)


client/src/app/+page-not-found/page-not-found.component.html
```diff
@@ -1,23 +1,32 @@
...
         <li i18n>You may have used an outdated or broken link</li>
+        <li i18n>The {{ getRessourceName() }} may have been moved or deleted</li>
         <li i18n>You may have typed the address or URL incorrectly</li>
       </ul>
     </div>
   </div>
 
+  <div *ngIf="status === 403" class="box">
+    <strong>{{ status }}.</strong>
+    <span class="ml-1 text-muted">You are not authorized here.</span>
```
Chocobozzz, line 23: i18n

client/src/app/+page-not-found/page-not-found.component.html
```diff
@@ -1,23 +1,32 @@
...
+
+    <div class="text mt-4" i18n="">
```
Chocobozzz, line 25: just i18n without =""

client/src/app/+page-not-found/page-not-found.component.ts
```diff
@@ -9,10 +10,16 @@ import { HttpStatusCode } from '@shared/core-utils/miscs/http-error-codes'
...
   status = HttpStatusCode.NOT_FOUND_404
+  type: string
 
   public constructor (
-    private titleService: Title
-  ) {}
+    private titleService: Title,
+    private router: Router
+  ) {
+    const state = this.router.getCurrentNavigation()?.extras.state
```
Chocobozzz, line 19: Put this in `ngOnInit`

Chocobozzz (changes requested):

client/src/app/+page-not-found/page-not-found.component.ts
rigelk: `getCurrentNavigation` cannot be called outside the constructor.

PR #3635, opened 2021-01-24
prettify SQL queries during debug

rigelk:
## Description

Debugging SQL queries is never easy, so easing their read would be welcome 🙂 

## Screenshots

Screenshot from 2021-01-24 20-35-36

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,6 +147,7 @@
     "sequelize-typescript": "^2.0.0-beta.1",
     "sitemap": "^6.1.0",
     "socket.io": "^3.0.2",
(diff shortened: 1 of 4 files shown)


Chocobozzz:
LGTM. Weird the tests fail :thinking: (I re ran them)

rigelk:
@Chocobozzz timeouts seem to be impacted?

Chocobozzz:
I'll debug that

rigelk:
Thanks

Chocobozzz:
Your prettifier is too slow, I replaced it by the origin prettifier (that does not have colors but it's better than nothing).

PR #3643, opened 2021-01-26
localize decimal separator in video miniatures

kontrollanten:
## Description
[Some countries](https://en.wikipedia.org/wiki/Decimal_separator#Countries_using_decimal_comma) are using comma as decimal separator.
https://caniuse.com/internationalization

## Has this been tested?

1) Switch language to Swedish.
1) See that decimal points for video views are comma.

## Screenshots

image

### DIFF
--- a/client/src/app/shared/shared-main/angular/number-formatter.pipe.ts
+++ b/client/src/app/shared/shared-main/angular/number-formatter.pipe.ts
@@ -1,14 +1,9 @@
-import { Pipe, PipeTransform } from '@angular/core'
+import { Inject, LOCALE_ID, Pipe, PipeTransform } from '@angular/core'
 
 // Thanks: https://github.com/danrevah/ngx-pipes/blob/master/src/ng-pipes/pipes/math/bytes.ts
 
 @Pipe({ name: 'myNumberFormatter' })
 export class NumberFormatterPipe implements PipeTransform {
-  private dictionary: Array<{max: number, type: string}> = [
-    { max: 1000, type: '' },
-    { max: 1000000, type: 'K' },
-    { max: 1000000000, type: 'M' }
-  ]
 
   /**
    * @param x number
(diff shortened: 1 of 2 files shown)


client/src/app/shared/shared-main/shared-main.module.ts
```diff
@@ -150,6 +150,8 @@ import { VideoChannelService } from './video-channel'
   providers: [
     DatePipe,
 
+    DecimalPipe,
```
Chocobozzz, line 153: Where is used?

Chocobozzz (commented):
Using https://angular.io/api/common/formatNumber would be simpler?

client/src/app/shared/shared-main/shared-main.module.ts
kontrollanten: Oh, sorry, that's a relic.

kontrollanten:
One thing that just hit me. Shall I put the Intl call in a try/catch, so it'll be handled gently in unsupported browsers?

client/src/app/shared/shared-main/angular/number-formatter.pipe.ts
```diff
@@ -21,14 +16,25 @@ export class NumberFormatterPipe implements PipeTransform {
...
+    { max: 1000000000, type: 'M' }
+  ]
+  constructor (@Inject(LOCALE_ID) private localeId: string) {}
+
   transform (value: number) {
     const format = this.dictionary.find(d => value < d.max) || this.dictionary[this.dictionary.length - 1]
     const calc = value / (format.max / 1000)
     const integralPart = Math.floor(calc)
     const decimalPart = NumberFormatterPipe.getDecimalForNumber(calc)
+    const decimalSeparator = Intl.NumberFormat(this.localeId)
```
kontrollanten, line 31: Shall I put this in a try/catch, so it'll be handled gently in unsupported browsers?

PR #3645, opened 2021-01-26
videos-views: make update interval configurable

### LINKED ISSUE #3548: View count update interval
Today the video view count updates once in a hour, which means that newly uploaded videos has low view counts in the first hours. I think many people think that low view counts means that other people aren't interested in this video, and therefore I'm probably neither interested.

View counts updated more often.

It should be a lightweight operation to run each fifth minute. Otherwise:
* Make it configurable how often to run it, with at least four times a hour as default.
* Create different levels of updates, which can mean that when a new video is published it will update more often the upcoming hours.

kontrollanten:
## Description
Make it configurable how often the views are calculated.

## Related issues
closes #3548

## Has this been tested?

1) Set views.videos.update_interval to some intensive cron expression.
1) Start the server and watch the jobs are getting created.


Chocobozzz:
Views updates require CPU at the same time in the federation because we have to send `Video` `Update` activities which is not a trivial task to process:
screen_2021-01-27-09:16:38

Until views federation is optimized, we can't afford to run this process multiple times per hour

kontrollanten:
Okay, can we add a disclaimer about that? Is this only an issue when using federation? Will the following instances also be affected?

kontrollanten:
I found out that the video views processing function also needs to be updated. I'll be back.

kontrollanten:
Any comments?

Chocobozzz:
Nothing more than what I said in https://github.com/Chocobozzz/PeerTube/pull/3645#issuecomment-768117283

Until we don't optimize views federation, we can't afford to merge this PR.

kontrollanten:
Do you have any idea where I can start dig to figure out what's causing the CPU usage?

scanlime:
re impact of this change, I am personally more worried about the activitypub broadcast backlog than the CPU usage. servers with many followers can already get into situations where the outgoing messages don't send as fast as they're queued (see bug #4249), and this would increase the number of messages going out if i understand correctly.

PR #3654, opened 2021-01-28
client(video-watch): add share btn tooltip

kontrollanten:
## Description
Adds a tooltip to the share button below the video player.

## Has this been tested?

## Screenshots

image

### DIFF
--- a/client/src/app/+videos/+video-watch/video-watch.component.html
+++ b/client/src/app/+videos/+video-watch/video-watch.component.html
@@ -104,7 +104,11 @@
                     <span class="icon-text" i18n>SUPPORT</span>
                   </button>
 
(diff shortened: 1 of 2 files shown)


rigelk:
I don't think this tooltip is necessary, contrary to its neighbors, which add information that might not be guessed accurately from the text/icons.

kontrollanten:
Maybe you're right. The idea came from that we'd like to have a script that focuses on the share button after the video has finished (to encourage users to share), and if it's has a tooltip it'll get more attention.

rigelk:
> we'd like to have a script that focuses on the share button after the video has finished […] [and] if it has a tooltip it'll get more attention

I understand but that's not a valid reason to modify the UI for everyone. Your plugin/script should add its own tooltip/modal/etc.

PR #3672, opened 2021-02-01
Added s3 edge signing

june07:
## Description

Using digitalocean spaces as per the docs required files to be public (per s3), otherwise private acl files did not work.  For greater control setting files as private is needed and for access to private files, pre-signing is required.  This change uses the aws4 npm library to pre-sign files using 3 additional configuration keys:

S3: {
    HOST: config.get<string>('s3.host'),
    KEY_ID: config.get<string>('s3.key_id'),
    KEY_SECRET: config.get<string>('s3.key_secret')
  }

Because the PeerTube docs recommend using redirects for S3, pre-signing the query is done so that it will be passed thru to the nginx CDN redirect.  I tested the digitalocean spaces edge endpoint and it works.  Pre-signing is done using the origin host but as per the response I received from Digitalocean support folks, the signature works with the edge servers, and I confirmed as much.

image

## Has this been tested?

- [x] yes, tested in a running instance but no unit tests added. 

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     "apicache": "^1.4.0",
     "async": "^3.0.1",
     "async-lru": "^1.1.1",
+    "aws4": "^1.11.0",
     "bcrypt": "5.0.0",
     "bittorrent-tracker": "^9.0.0",
     "bluebird": "^3.5.0",
--- a/server/initializers/config.ts
+++ b/server/initializers/config.ts
@@ -72,6 +72,11 @@ const CONFIG = {
     PLUGINS_DIR: buildPath(config.get<string>('storage.plugins')),
     CLIENT_OVERRIDES_DIR: buildPath(config.get<string>('storage.client_overrides'))
   },
+  S3: {
+    HOST: config.get<string>('s3.host'),
+    KEY_ID: config.get<string>('s3.key_id'),
+    KEY_SECRET: config.get<string>('s3.key_secret')
+  },
   WEBSERVER: {
     SCHEME: config.get<boolean>('webserver.https') === true ? 'https' : 'http',
     WS: config.get<boolean>('webserver.https') === true ? 'wss' : 'ws',
--- a/server/models/video/video.ts
+++ b/server/models/video/video.ts
@@ -130,6 +130,7 @@ import { VideoShareModel } from './video-share'
 import { VideoStreamingPlaylistModel } from './video-streaming-playlist'
 import { VideoTagModel } from './video-tag'
 import { VideoViewModel } from './video-view'
+import * as aws4 from 'aws4'
 
 export enum ScopeNames {
   AVAILABLE_FOR_LIST_IDS = 'AVAILABLE_FOR_LIST_IDS',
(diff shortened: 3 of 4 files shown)


rigelk:
Hi and thanks for this interesting PR!

Before making a review though, I wonder what you mean by "For greater control setting files as private is needed". Your bucket is going to be dedicated to videos anyway, so what does making it private bring you?

june07:
The project I am working on requires video files that are not
publicly available but rather limited to subscribers. Given that, having
file access wide open isn't an option. Users must be authenticated for
video access, which is what signing gives in this case. Otherwise anyone
(user or otherwise) with a static uuid ("private") link will be able to
access files.

Bandwidth control is another related issue.

Basically only authenticated users should have access to files and even
that access should be time limited based on the last signing time.
Essentially there should never be a non expiring URL to be used for file
access...

june07:
By the way, thanks for the great project.

Also, I noticed the failed tests and I see it's due to the new configuration options that were added and yet not added to the test configs.  Before attempting to resolve those, I'll be awaiting feedback on the PR.

Chocobozzz:
Hi,

Instead of updating peertube core for this specific use case, please develop a plugin: https://docs.joinpeertube.org/api-plugins?id=hooks

You can use the `filter:api.video.get.result` hook to update the fields you want

june07:
Maybe you could help point me in the right direction as far as how to update the video object...

I've tried changing the video.url and other things but nothing seems to work and the client side API /api/v1/videos/ endpoint doesn't reflect any change.  The output of that endpoint is a files array:

files[0]:
fileDownloadUrl: ...
fileUrl: ...
magnetUri: ...
metadataUrl: ...
torrentDownloadUrl: ...
torrentUrl: ...

My code looks like...

```
registerHook({
    target: 'filter:api.video.get.result',
    handler: async video => {
      const settings = await settingsManager.getSettings([
        's3-endpoint',
        's3-key_id',
        's3-key_secret'
      ])

      let cdnObject = `/videos/${video.dataValues.uuid}-${video.VideoFiles[0].resolution}${video.VideoFiles[0].extname}`
      let hostname = new URL(settings['s3-endpoint']).hostname
      let opts = {
        hostname,
        path: cdnObject,
        service: 's3',
        signQuery: true // query must be signed since redirects are used
      }
      aws4.sign(opts, {
        accessKeyId: settings['s3-key_id'],
        secretAccessKey: settings['s3-key_secret']
      })
  
      // THIS IS THE PART I'M NOT SURE ABOUT...
      video.VideoFiles[0].fileUrl = `${settings['s3-endpoint']}${cdnObject}?${opts.path.split('?')[1]}`
      return video
    }
  })
```

Chocobozzz:
The `fileUrl` attribute is generated on the fly by https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/video/video.ts#L2063. You can try to redefine the function.

Or try to use `filter:api.video-watch.video.get.result` client hook to update the API result

june07:
> You can try to redefine the function.
This is exactly what I did in the PR?  And while this PR's solution works fine, I'm still unable to get either

>filter:api.video.get.result
or
> filter:api.video-watch.video.get.result

working via the plugin route, presumably because of the following fact
> The `fileUrl` attribute is generated on the fly by https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/video/video.ts#L2063.

So since "redefine the function" as I did in this PR is a working solution, and that part of the code is in core, and given
> Instead of updating peertube core for this specific use case, please develop a plugin: https://docs.joinpeertube.org/api-plugins?id=hooks

Is the answer that this functionality is not wanted in core?  Seems like a pretty key component to using S3 as per the documentation.  Of course I can simply use my own fork and continue pulling from upstream, just curious about the reasoning behind not wanting S3 edge pre-signing in core when we have S3 in core and pre-signing is a significant part of the S3 feature set.  Also, per this PR, when no configuration exists, the pre-signing code is simply skipped...
https://github.com/june07/PeerTube/blob/91b76b99ab0a723b3b1bbb20100ccbae0c0fc493/server/models/video/video.ts#L2065-L2067

PR #3679, opened 2021-02-02
generate unique thumbnail/preview filenames

### LINKED ISSUE #3678: Change thumbnail filename upon image update
When a video thumbnail/preview image is changed the URI doesn't change, which means that the old image is still cached in browsers and CDN.

When a new image is uploaded, change the filename so that it's always safe to cache the pictures forever (which will lead to a better user experience).

Instead of updating existing row in `thumbnail` table, create a new row with a new filename.

kontrollanten:
## Description
Each time a new preview image is uploaded a new filename is generated. This makes it possible to use aggressive caching for the previews and new file uploads will give immediate effect.

## Related issues
closes #3678

## Has this been tested?

* Change video preview image and verify that the preview and thumbnail image is getting updated without doing a reload in the browser.

### DIFF
--- a/server/controllers/lazy-static.ts
+++ b/server/controllers/lazy-static.ts
@@ -18,7 +18,7 @@ lazyStaticRouter.use(
 )
 
 lazyStaticRouter.use(
-  LAZY_STATIC_PATHS.PREVIEWS + ':uuid.jpg',
+  LAZY_STATIC_PATHS.PREVIEWS + ':filename',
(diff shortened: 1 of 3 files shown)


kontrollanten:
I haven't figured out why the tests are failing and now I don't have more time to put on this issue. Maybe I'll get back to this in the future.

Chocobozzz:
@kontrollanten I'll implement this issue

PR #3681, opened 2021-02-02
add best trending strategy based on Reddit's best

rigelk:
## Description

inspired from https://www.reddit.com/r/changelog/comments/7spgg0/best_is_the_new_hotness/
this implementation only adds freshness, and doesn't personalize based
on subscribed communities yet.

## Related issues

None.

## Has this been tested?

## Screenshots

Screenshot_2021-02-02 Trending videos - PeerTube

## Questions

Should I merge this with `hot` sort?
Should I only display `best` sort for logged-in users?

### DIFF
--- a/server/models/video/video-query-builder.ts
+++ b/server/models/video/video-query-builder.ts
@@ -31,8 +31,8 @@ export type BuildVideosQueryOptions = {
 
   videoPlaylistId?: number
 
+  trendingAlgorithm?: string // best, hot, or any other algorithm implemented
   trendingDays?: number
-  hot?: boolean
 
   user?: MUserAccountId
   historyOfUser?: MUserId
@@ -252,7 +252,7 @@ function buildListQuery (model: typeof Model, options: BuildVideosQueryOptions) 
       attributes.push('COALESCE(SUM("videoView"."views"), 0) AS "score"')
 
       group = 'GROUP BY "video"."id"'
-    } else if (options.hot) {
+    } else if ([ 'best', 'hot' ].includes(options.trendingAlgorithm)) {
       /**
        * "Hotness" is a measure based on absolute view/comment/like/dislike numbers,
        * with fixed weights only applied to their log values.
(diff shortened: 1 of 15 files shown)


server/models/video/video-query-builder.ts
```diff
@@ -33,6 +33,7 @@ export type BuildVideosQueryOptions = {
 
   trendingDays?: number
   hot?: boolean
```
Chocobozzz, line 35: Maybe we could create a param like `trendingAlgorithm`?

server/models/video/video-query-builder.ts
```diff
@@ -510,8 +522,6 @@ function wrapForAPIResults (baseQuery: string, replacements: any, options: Build
       '"VideoFiles"."videoId"': '"VideoFiles.videoId"',
 
       '"VideoStreamingPlaylists"."id"': '"VideoStreamingPlaylists.id"',
-      '"VideoStreamingPlaylists"."playlistUrl"': '"VideoStreamingPlaylists.playlistUrl"',
```
Chocobozzz, line 513: Please keep them

rigelk: Wow I didn't mean to delete that.

test2a:
At this rate, isnt it better to also add a "new" so that "recently added" can be shown on that page as well?
It would declutter the sidebar and on trending page the experience would be unified.

Best, hot, likes, views, new.

rigelk:
@test2a I don't think age alone can be a measure of trendiness.

test2a:
On Reddit, the stream is sortable between all these fields. Im basing on that because you imported best and hot from reddit in the first place. Im saying make the experience simpler for users who are new.

Oh, you are saying the tab is "trending" and hence the sorts. Hmm.

rigelk:
@test2a we might need the space in the menu in the near future ; in the meantime there's no rush changing things IMHO.

test2a:
@rigelk while we are at this, views and hot and hopefully best also works on "last 24 hours". is that correct? if yes then why do likes show historical stats like from last year? shouldnt sort by likes too show "most liked in recent videos"?

PR #3684, opened 2021-02-03
Rewrite prose for JavaScript disabled message

cloudrac3r:
## Description

This should improve the clarity of the message and make it easy to
understand what the problem is at a glance.

- Clearly show that the page is PeerTube, without having to read the whole prose
- Clearly show that the problem is JavaScript, without having to read the whole prose
- Reorganise suggestions into a bullet list

I figured this was trivial enough that I didn't need to open an issue for discussion first. Just let me know if you have comments :)

## Has this been tested?

Yes, manually, by loading the HTML file in a browser with scripts disabled, and checking that the page looked like what I wanted.

### DIFF
--- a/client/src/index.html
+++ b/client/src/index.html
@@ -18,6 +18,12 @@
       .icon-logo {
         background-image: url(/client/assets/images/logo.svg?[logoContentHash]);
       }
+      .script-disclaimer {
+        display: block;
+        font-size: 20px;
+        max-width: 600px;
+        margin: 0 auto;
+      }
     </style>
 
     <!-- base url -->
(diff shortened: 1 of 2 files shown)


rigelk (approved):
I changed a few things along the way to simplify further maintenance.

rigelk:
@cloudrac3r thanks! This is indeed simpler to read and act upon.

cloudrac3r:
Did you test your changes? I found that `noscript` would be replaced with a `span`. So it couldn't be styled directly, and `max-width` wouldn't apply anyway since the element was inline. That's why I added the nested div.

rigelk:
yes, I tested them the same way you did.

cloudrac3r:
I can't get it to apply the styles, but I trust that you know more than me, so you're probably doing it right and I'm not. :) Thanks!

rigelk:
what browser are you testing this with? Are you running the instance in a specific mode (dev? prod?)?

PR #3724, opened 2021-02-09
Add libvips to OpenBSD dependencies list

Gopherslol:
## Description
Forgot to include libvips as a dependency. This change fixes that

## Has this been tested?

### DIFF
--- a/support/doc/dependencies.md
+++ b/support/doc/dependencies.md
@@ -347,7 +347,7 @@ rc-service postgresql-11 start
 1. Install Packages:
 
 ```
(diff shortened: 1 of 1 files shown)


rigelk:
What makes libvips required? I remember having removed sharp explicitly to avoid depending on libvips

Gopherslol:
I included it because it is listed as a dependency on other operating systems. I can create an instance without it to see what happens if that helps

Gopherslol:
Specifically, vips is listed under the freebsd instructions

rigelk:
Right, so this is something that is not required anymore and can be removed from the FreeBSD instructions too

Gopherslol:
Copy that. I can take care of that

PR #3726, opened 2021-02-09
Remove vips dependency from FreeBSD instructions

Gopherslol:
## Description
Removes outdated dependency libvips from freebsd instructions

## Has this been tested?


rigelk:
Thanks @Gopherslol!

Issue #3707, opened 2021-02-06
InvalidStateError: RTCPeerConnection is gone (did you enter Offline mode?)


kontrollanten:
**Describe the current behavior**
Our Sentry account got depleted due to thousands of `InvalidStateError: RTCPeerConnection is gone (did you enter Offline mode?)` errors from our clients. I guess the error itself is due to firewall or browser restrictions, but in average the users received ~80 errors each which looks like something is unhandled. 3,79% of the errors appears at /recently-added, the rest at video watch page.

**Steps to reproduce:**
I don't know how to reproduce.

**Describe the expected behavior**
The error should only be thrown once, if at all.

* PeerTube instance:
  * version: 3.0.1

* Browser name, version and platforms on which you could reproduce the bug:
All of the browsers are Firefox. Versions 81, 82, 84 and 85.
Platforms Ubuntu, Linux, Windows, MacOS.

Chocobozzz:
Do you have the stacktrace of this error?

kontrollanten:
Unfortunately not

Chocobozzz:
See https://github.com/feross/simple-peer/pull/783

Chocobozzz:
PR merged, fixed by https://github.com/Chocobozzz/PeerTube/commit/f9450247a8265033152412a1b98c86b48b802311

cacious7:
I am still receiving this error on sentry. How do I resolve it, please? The error description is the same as this one.

PR #3745, opened 2021-02-16
Show error modal upon uncatched JS error

### LINKED ISSUE #3736: Display decent message to legacy browsers
Some of our users have tried to visit our instance from their LG TV and reported that it's more or less a blank page. I got some information from a user and it appeared that the TV he bought last year has Chrome 38 installed. 

**Steps to reproduce:**

1. Visit a PeerTube instance with Chrome 38 (I can recommend BrowserStack, since you're an open source project i think you can get a free account).

I know Chrome 38 isn't supported, but I think that since some users still have it should be a decent error message displayed explaining the issue.

* PeerTube instance:
  * URL: https://videos.lukesmith.xyz
  * version: 3.0.1

Chrome 40

image

kontrollanten:
## Description
Handle uncatched JS errors by logging the error and show a modal.

## Related issues
closes #3736

## Has this been tested?

1. Add `throw Error();` in AppComponents `ngOnInit`.
2. Load the page.

1. Load PeerTube in Chrome 40.
2. See "Incompatible browser" screen.

## Screenshots

image

image

### DIFF
--- a/client/src/app/app.module.ts
+++ b/client/src/app/app.module.ts
@@ -1,7 +1,7 @@
 import 'focus-visible'
 import { environment } from 'src/environments/environment'
 import { APP_BASE_HREF, registerLocaleData } from '@angular/common'
-import { NgModule } from '@angular/core'
+import { ErrorHandler, NgModule } from '@angular/core'
 import { BrowserModule } from '@angular/platform-browser'
 import { ServiceWorkerModule } from '@angular/service-worker'
 import { ServerService } from '@app/core'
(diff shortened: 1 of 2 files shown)


Chocobozzz:
See https://github.com/Chocobozzz/PeerTube/issues/3736#issuecomment-781345696

Also, we don't want to display a modal everytime the client throws an unexpected error.

kontrollanten:
> Also, we don't want to display a modal everytime the client throws an unexpected error.

Shouldn't it be a message displayed for the user if the application crashes?

Chocobozzz:
Yes but how to know if the application crashed? An unhandled error may not crash the entire application. For example this kind of error https://github.com/Chocobozzz/PeerTube/issues/3707

kontrollanten:
Sure, but that's bugs that we should get rid of. Isn't it better that the users always get an error message, and in few cases when it isn't necessary, than that the app just isn't working without any message?

rigelk:
@kontrollanten you assume errors are always explainable and understandable by the average user. I'd wager that there are more errors than you can explain, or wish to display.

There is a difference between what the interface should be and what it can be, provided all bugs are rid of.

PR #3756, opened 2021-02-18
Add missing dependency to Debian / Ubuntu section

janeymunoz:
## Description

The "Production guide" states to use `wget` to download the latest version of the Peertube client, and it was not included in the `apt install` dependency list

## Has this been tested?


Chocobozzz:
Thanks

PR #3770, opened 2021-02-23
Fix titles do not trim leading and trailing spaces

### LINKED ISSUE #3623: Titles do not trim leading and trailing spaces.

**Steps to reproduce:**

1. Upload/Edit video properties
2. Type two characters as name (note that peertube will not allow it)
3. Add a space at the end of the title (peertube title length now satisfied)

Leading and trailing blank spaces should be removed/shouldn't count for minimum title length.

Ramiellll:
## Description

Fix titles do not trim leading and trailing spaces

## Related issues

fixes #3623 

## Has this been tested?

- [x ] 🙅 no, because they aren't needed

1. Upload/Edit video properties
2. Type a leading space or a trailing space


Chocobozzz:
Thanks @Ramiellll

Issue #3776, opened 2021-02-25
Prevent users from changing the theme


1992Sofia:
**Describe the problem to be solved**
I want to prevent the user from changing the theme.

**Describe the solution you would like:**
_I have already implemented a solution._
![theme_change](https://user-images.githubusercontent.com/79608996/109089981-8934fd00-7712-11eb-995a-8d36432a0c92.png)
There is now a checkbox in the settings that can be used to prevent the user from changing the theme.

Would you allow this solution to be merged?

Chocobozzz:
Hello,

Since it does not seem to be a frequently asked feature by admins, I'd rather it to be implemented in a plugin

1992Sofia:
Hey,

thanks for your quick reply.
For an implementation via a plugin, I would need something like a settings filter and an action like "action.usersettings.open" or is there already a way to implement this?

if a plugin solution is too complicated please take a look at:
https://github.com/Chocobozzz/PeerTube/pull/3782

Chocobozzz:
> For an implementation via a plugin, I would need something like a settings filter and an action like "action.usersettings.open" or is there already a way to implement this?

Just create a plugin that hides the field using CSS :)

1992Sofia:
Hey,

that was my first idea at the beginning, but somehow the css solution seemed very messy 😁
But okay i solved it in my theme. 

Thanks for your help 🙂!

PR #3782, opened 2021-02-26
Prevent users from changing the theme 

1992Sofia:
## Description
if a plugin solution is too complicated please merge. 

## Related issues
https://github.com/Chocobozzz/PeerTube/issues/3776

## Has this been tested?


Chocobozzz:
Hello,

Thanks for the PR But see https://github.com/Chocobozzz/PeerTube/issues/3776#issuecomment-785722167

PR #3787, opened 2021-02-26
add sort select to my videos, improve publishedAt sort for videos

### LINKED ISSUE #3779: Order my videos on publish order

When planning a lot of videos (think uploading on weekend for 1 video a day for the next 2 weeks), I upload videos by playlist (game A, then game B, then…).
The "my videos" tab show me the videos ordered by last action. Action meaning uploaded, published, …
So to me it can be a bit confusing to see a "planning" of videos that will go out:
 1. I upload a video to be published in a week
 2. An older video gets published
 3. The video at the top of the list is not the video that will be published last, but the one that just has been published.

This way it is hard to see if I already planned a video a day or if I may did a mistake.

Order the videos from most recent to oldest publication date (future and past). Not ordered by last action.

Youtube do this way already so here have my example: the video in "importing" state is currently being upload (planned publish on the 3rd of march) The "Basement" videos where uploaded in the same session some days ago.
image

If I had the same on peertube the "importing" video would be at the top and to compare the dates I would have to do it by hand instead of seeing them chronologically.

Having a real calendar view/an ics/caldav being automatically generated? (I know that it’s much more work and surely out of scope.)

rigelk:
## Related issues

fixes #3779

## Has this been tested?

- [x] 👍 yes, light UI tests are enough

## Screenshots

Screenshot_2021-02-26 My videos - PeerTube(1)

### DIFF
--- a/client/src/app/+my-library/my-videos/my-videos.component.ts
+++ b/client/src/app/+my-library/my-videos/my-videos.component.ts
@@ -44,6 +44,7 @@ export class MyVideosComponent implements OnInit, DisableForReuseHook {
   videosSearch: string
   videosSearchChanged = new Subject<string>()
   getVideosObservableFunction = this.getVideosObservable.bind(this)
+  sort: VideoSortField = '-publishedAt'
 
   user: User
 
@@ -81,6 +82,10 @@ export class MyVideosComponent implements OnInit, DisableForReuseHook {
     this.videosSearchChanged.next()
   }
 
+  onChangeSortColumn () {
+    this.videosSelection.reloadVideos()
+  }
+
   disableForReuse () {
     this.videosSelection.disableForReuse()
   }
@@ -92,7 +97,7 @@ export class MyVideosComponent implements OnInit, DisableForReuseHook {
   getVideosObservable (page: number, sort: VideoSortField) {
     const newPagination = immutableAssign(this.pagination, { currentPage: page })
 
-    return this.videoService.getMyVideos(newPagination, sort, this.videosSearch)
+    return this.videoService.getMyVideos(newPagination, this.sort, this.videosSearch)
       .pipe(
         tap(res => this.pagination.totalItems = res.total)
       )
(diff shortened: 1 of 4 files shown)


client/src/app/+my-library/my-videos/my-videos.component.ts
```diff
@@ -92,7 +97,7 @@ export class MyVideosComponent implements OnInit, DisableForReuseHook {
   getVideosObservable (page: number, sort: VideoSortField) {
     const newPagination = immutableAssign(this.pagination, { currentPage: page })
 
-    return this.videoService.getMyVideos(newPagination, sort, this.videosSearch)
+    return this.videoService.getMyVideos(newPagination, this.sort, this.videosSearch)
```
Chocobozzz, line 100: What about the `sort` parameter?

server/models/utils.ts
```diff
@@ -54,6 +54,14 @@ function getVideoSort (value: string, lastSort: OrderItem = [ 'id', 'ASC' ]): Or
 
       [ Sequelize.col('VideoModel.views'), direction ],
 
+      lastSort
+    ]
+  } else if (field === 'publishedAt') {
+    return [
+      [ 'ScheduleVideoUpdate', 'updateAt', direction + ' NULLS LAST' ],
```
Chocobozzz, line 61: Did you test the repercussions on the query?

rigelk: Yes. I can't find anywhere else that would have unpublished videos shown, though.

client/src/app/+my-library/my-videos/my-videos.component.ts
rigelk: It doesn't get updated in my experience, unfortunately.

Chocobozzz: Because you don't update the URL when you sort videos

rigelk: I tried updating the URL, no to avail. It seems the scope gets frozen when `.bind(this)` is called, which means, with the component initialization.

rigelk: I think there is no way using the frozen `sort` value passed at `.bind` time can do here.

Chocobozzz:
Merged manually

PR #3808, opened 2021-03-02
Hide publish button when video publishing is unavailable

### LINKED ISSUE #3807: Hide publish button when video quota is 0
The publish button in the right corner is always shown.

A) When a user is not logged in it should be hidden if the video quota is 0 for new users.
B) When a user is logged in it should be hidden if the video quota is 0 for the current user.

The button is one of few buttons and therefore takes a lot of user focus. And when clicked on the button (when video quota is 0) it looks like something isn't working properly.

kontrollanten:
## Description
Hides publish button, top right corner, when uploads are disabled.

## Related issues
Closes #3807 

## Has this been tested?

1) Set video quota for new users to unlimited.
1) Logout.
1) See the upload button.

1) Set video quota for new users to 0.
1) Logout.
1) Don't see the upload button.

1) Set the video quota for new users to 0.
1) Set the video quota for a particular user to unlimited.
1) Login as that user.
1) See the publish button.

## Screenshots

image

### DIFF
--- a/client/src/app/app.component.ts
+++ b/client/src/app/app.component.ts
@@ -116,11 +116,9 @@ export class AppComponent implements OnInit, AfterViewInit {
   }
 
   isUploadEnabled () {
-    if (this.authService.isLoggedIn()) {
-      return this.authService.getUser().videoQuota !== 0
-    } else {
-      return this.serverConfig.user.videoQuota !== 0
-    }
+    return this.authService.isLoggedIn()
+      ? this.authService.getUser().videoQuota !== 0
+      : this.serverConfig.user.videoQuota !== 0
   }
 
   hideBroadcastMessage () {
--- a/client/src/app/app.component.ts
+++ b/client/src/app/app.component.ts
@@ -115,6 +115,14 @@ export class AppComponent implements OnInit, AfterViewInit {
     return this.authService.isLoggedIn()
   }
 
+  isUploadEnabled () {
+    if (this.authService.isLoggedIn()) {
+      return this.authService.getUser().videoQuota !== 0
+    } else {
+      return this.serverConfig.user.videoQuota !== 0
+    }
+  }
+
   hideBroadcastMessage () {
     peertubeLocalStorage.setItem(AppComponent.BROADCAST_MESSAGE_KEY, this.serverConfig.broadcastMessage.message)
 
--- a/client/src/app/header/header.component.ts
+++ b/client/src/app/header/header.component.ts
@@ -7,5 +7,5 @@ import { Component, Input } from '@angular/core'
 })
 
 export class HeaderComponent {
-  @Input() showPublishButton: boolean
+  @Input() showPublishButton = true
 }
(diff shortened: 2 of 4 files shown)


client/src/app/header/header.component.ts
```diff
@@ -1,9 +1,11 @@
...
 
 @Component({
   selector: 'my-header',
   templateUrl: './header.component.html',
   styleUrls: [ './header.component.scss' ]
 })
 
-export class HeaderComponent {}
+export class HeaderComponent {
+  @Input() showPublishButton: boolean
```
rigelk, line 10:
```suggestion
  @Input() showPublishButton = true
```

client/src/app/app.component.ts
```diff
@@ -115,6 +115,14 @@ export class AppComponent implements OnInit, AfterViewInit {
     return this.authService.isLoggedIn()
   }
 
+  isUploadEnabled () {
+    if (this.authService.isLoggedIn()) {
+      return this.authService.getUser().videoQuota !== 0
+    } else {
+      return this.serverConfig.user.videoQuota !== 0
+    }
```
rigelk, line 123:
```suggestion
    return this.authService.isLoggedIn()
      ? this.authService.getUser().videoQuota !== 0
      : this.serverConfig.user.videoQuota !== 0
```

rigelk (changes requested):

kimsible:
Hmm hiding the publish is now in the user scenario @rigelk ?

https://github.com/Chocobozzz/PeerTube/pull/3357#issuecomment-733666215

rigelk:
The user is logged in and doesn't have any quota, that's not the same situation.

kontrollanten:
What's the status of this? Is there anything I can do?

Chocobozzz:
Sorry for the late review/answer @kontrollanten

Hiding the upload button has already been discussed in https://github.com/Chocobozzz/PeerTube/pull/3357 so it's not a feature we want to implement.

What I can do is to add a `userInformationLoaded` plugin hook, and a `getServerConfig` helper on the client so admins can create/install a plugin that hides this upload button depending on their need.

kimsible:
> Sorry for the late review/answer @kontrollanten
> 
> Hiding the upload button has already been discussed in #3357 so it's not a feature we would like to implement.
> 
> What I can do is to add a `userInformationLoaded` plugin hook, and a `getServerConfig` helper on the client so admins can create/install a plugin that hides this upload button depending on their need.

Thanks @Chocobozzz, sounds a very good compromise.

>> In my opinion the real problem is that publish button is confusing with the search box at left, we may click on the publish button to run a search.

> This is exactly the problem we're facing. But that's a separate issue then I guess.

Yep @kontrollanten it has been reported by some of our members too. If it can help we tried to hack it with a simple CSS override but you'll have a problem with the main container search and the auto-complete will not follow. So yes, in my opinion, the only way to resolve properly this issue (or center the search box ? many things could actually be proposed) is to edit the template angular source file.

Chocobozzz:
> This is exactly the problem we're facing. But that's a separate issue then I guess.

A UX work on the header is planned

> What I can do is to add a userInformationLoaded plugin hook, and a getServerConfig helper on the client so admins can create/install a plugin that hides this upload button depending on their need.

Done in https://github.com/Chocobozzz/PeerTube/commit/7bc572c7531580600d02b4fc5f1b6cf54df139f0

PR #3815, opened 2021-03-03
unzip silently on production guide

pedro-nonfree:
## Description

added argument `-q` to unzip to it is silent output such as previous curl and wget commands in this guide


rigelk:
Why?

PR #3818, opened 2021-03-03
Patch 1

ChrystianRubio:
## Has this been tested?

- [ x ] 🙅 no, because they aren't needed


Chocobozzz:
Thanks but we don't really need this, and this breaks the CI.

Issue #3810, opened 2021-03-02
Simplify download modal


kontrollanten:
**Describe the problem to be solved**
Today there's a lot of information in the download video modal which may be confusing for novice users. It's unclear what happens when selecting one of the tab (Format, Video stream and Audio stream), we've had some users that selected audio stream and thought that would download the audio only. Non-technical users gets confused when the torrent option is selected and don't really know what that means. If you select "direct download" and choose Download the browser will open the file, instead of start downloading it.

**Describe the solution you would like:**
- [ ] Less information, at least by default.
- [x] ~Direct download by default.~ Solved in #3880 
- [x] ~The copy icon is [uncommon](https://www.google.com/search?q=copy+icon&source=lnms&tbm=isch&sa=X&ved=2ahUKEwiI1OS1rpLvAhVpkosKHfrTBAQQ_AUoAXoECBQQAw&biw=1650&bih=819), I'd propose the `duplicate` icon and/or a text "Copy" text instead.~ Solved in #3867

**Describe alternatives you have considered**  

![image](https://user-images.githubusercontent.com/6680299/109705100-aa28a280-7b97-11eb-9aff-c2fca0f41a8e.png)

PR #3880, opened 2021-03-23
Set direct download as default download method

kontrollanten:
## Description
By setting direct download as default method users with low tech skills can just click "Download" and it works as they expect. Whilst users that want to use torrent probably also are more technological skilled and therefore more comfortable in the UX and can easily switch to torrent if they'd like.

## Related issues
Relates to #3810 

## Has this been tested?

1) Go to a video watch page.
1) Open download modal.
1) See that direct dl is checked as default.


rigelk (approved):
Judging by the number of desktop torrent clients that support WebTorrent as of 2021, we probably don't need to push torrent as the favored download option anymore.

PR #3890, opened 2021-03-26
Show less information in download modal

### LINKED ISSUE #3810: Simplify download modal

kontrollanten:
## Description
* Resolution is switched with tabs instead of select list.
* Metadata is collapsed by default.
* Solves a bug: previously the metadata wasn't updated when choosing a resolution for the second time (check the last commit).

## Related issues
Closes #3810 

## Has this been tested?

1) Visit a video that has subtitles and multiple resolutions.
1) Open Download modal.
1) Switch between video and subtitle.
1) Switch between resolutions.

## Screenshots

Peek 2021-03-26 09-58

### DIFF
--- a/client/src/app/shared/shared-video-miniature/video-download.component.scss
+++ b/client/src/app/shared/shared-video-miniature/video-download.component.scss
@@ -17,6 +17,10 @@
   .nav-tabs {
     margin-top: 10x;
   }
+  
+  .glyphicon {
+    margin-right: 5px;
+  }
 }
 
 .peertube-select-container {
--- a/client/src/app/shared/shared-video-miniature/video-download.component.scss
+++ b/client/src/app/shared/shared-video-miniature/video-download.component.scss
@@ -1,6 +1,24 @@
 @import 'variables';
 @import 'mixins';
 
+.nav-content {
+  margin-top: 30px;
+}
+
+.advanced-filters-button {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin-top: 20px;
+  font-size: 16px;
+  font-weight: 600;
+  cursor: pointer;
+
+  .nav-tabs {
+    margin-top: 10x;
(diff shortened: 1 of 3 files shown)


client/src/app/shared/shared-video-miniature/video-download.component.scss
```diff
@@ -1,6 +1,24 @@
...
+  justify-content: center;
+  align-items: center;
+  margin-top: 20px;
+  font-size: 16px;
+  font-weight: 600;
+  cursor: pointer;
+
+  .nav-tabs {
+    margin-top: 10x;
+  }
```
rigelk, line 19:
```suggestion
  }
  
  .glyphicon {
    margin-right: 5px;
  }
```

rigelk (approved):
LGTM. I think we can do better than this advanced/simple toggle button, like with a "peertube-link style" toggle _in_ the modal action row (bottom) left-hand corner.

rigelk:
Thanks @kontrollanten for taking the refactorization!

Issue #3864, opened 2021-03-17
Kubernetes installation for "Unofficial installations"


arthurlogilab:
**Parts of the documentation concerned:**

https://docs.joinpeertube.org/install-unofficial describes a number of ways to deploy and install peertube. Congrats for the approach, one size doesn't fit all, so it's nice to see one can choose from a number of options. 

After having deployed a few peertube instances using yunohost (kudos to all maintainers, it works great!), we've decided to deploy an instance to kubernetes with s3-type storage (https://docs.joinpeertube.org/admin-remote-storage). We're obviously using the "official" docker images from https://github.com/Chocobozzz/PeerTube/tree/develop/support/docker/production 

**Describe the desired information/correction**

We would like to know if the project or other users are interested in a contribution to the documentation that describes how we've done it with kubernetes and the caveats of this approach. (We could reference some open issues that might improve this type of deployment in the future).

rigelk:
Hi @arthurlogilab ; I think adding that kind of information here is what the page exists for.

Are you aware of https://git.lecygnenoir.info/LecygneNoir/peertube-helm or other examples using Kubernetes to deploy PeerTube?

arthurlogilab:
I was aware of a 9 months old gist https://gist.github.com/cclloyd/5f023168309ffa905957f3dae8c15557 that describes a deployment (we used part of that). 

No I wasn't aware of the helm chart. Thanks! I'll take a look into it see if we can learn some things from it. 

Some people prefer not to use helm charts, so we might be able to fit two sections, one for "kubernetes" and another for "kubernetes with helm" if you think that helm chart is worth mentioning. 

We'll submit the documentation and configuration files as soon as we can. Thanks for you feedback.

arthurlogilab:
It seems @LecygneNoir 's helm chart is inspired/forked from https://github.com/MikaXII/helm-charts/tree/master/peertube which might be a good source of example too.

LecygneNoir:
Hello!

I confirm I forked from this repo, as it was nearly abandoned at the time I begin to work on it.

I think the work from @pulpo may be interesting too, especially because it supports helm3, which mine does not unfortunately. (I have absolutely no idea of the work needed to migrate from helm 2 to helm 3)

They have created the MR #2758 some times ago, that cannot be integrated directly in code, but the unofficial installation documentation could be a great way to add it for people interested!

[...]

PR #3891, opened 2021-03-26
[support/docker] expose webtorrent setting in os variable settings

arthurlogilab:
## Description

When enabling hls, one might want to disable webtorrent, so exposing it in the OS variable config

## Related issues

Related to our work on #3864 

## Has this been tested?


Chocobozzz:
Thanks @arthurlogilab

PR #3898, opened 2021-03-29
Account and channel pages redesign

Chocobozzz:
~~Still a WIP, but most parts are OK.~~

We are redesigning the account and the channel pages. The main goal is to better differentiate them.

Main points:
 * Different headers so it's easier to know if we are on the account page or on the channel page
 * Different avatar shapes: square for the channel, circle for the account
 * Removed the about page, and display these information (description, creation date...) in the header
 * Changed the way we display channels in the account page: orange background, truncated description...
 * Better responsive
 * Prefer channel display over the account in miniatures

**Account desktop**
account-ipad

**Account ipad**
account-full

**Account mobile**
account-mobile

**Channel desktop**

channel-full

**Channel ipad**

channel-ipad

**Channel mobile**

channel-mobile


rigelk:
@Chocobozzz could you keep the avatars consistent? (square or round)

Chocobozzz:
> could you keep the avatars consistent? (square or round)

Done in https://github.com/Chocobozzz/PeerTube/pull/3898/commits/6f9f598928d5bd205a51e766624862da93eb674e

> I would also remove the "view owner account" button honestly ;

I'll discuss about this with our designer :+1:

rigelk:
@Chocobozzz I noticed the magnifying glass is also not of the same line-height, nor of the same color as before (black). Was it changed for a specific purpose?

Chocobozzz:
> not of the same line-height

In fact it's the same line-height if we take into account the sub title bottom border. I'll check if I can align it with the text only.

> nor of the same color as before

I think we chose the main color because it's a clickable button, but I agree it's inconsistent with the other sub title links.

EricGuic:
Hi, nice redesign. A consistent square/round avatar for the channel/account is a good idea to better distinguish them.

On your last screencapture (channel on mobile), I think you can put the two icons (rss radar and gear) on the same line that the title "published videos" (maybe the title justified to the left side, and the icons to the right side).

Chocobozzz:
Hi,

Yes these icons are quite broken on mobile. I did not work on them in this PR because we plan a redesign of the videos list header (to allow filters etc).

PR #3901, opened 2021-03-30
fix missing title attribute on <iframe> tag suggested for embedding

pthavarasa:
fixed this issue. Now there is a title in the iframe tag.
share video -> video title
share playlist -> playlist title
share playlist video(right click in the video to copy embed) -> video title


client/src/assets/player/utils.ts
```diff
@@ -170,9 +170,10 @@ function secondsToTime (seconds: number, full = false, symbol?: string) {
   return time
 }
 
-function buildVideoOrPlaylistEmbed (embedUrl: string) {
+function buildVideoOrPlaylistEmbed (embedUrl: string, embedTitle: string) {
```
Chocobozzz, line 173: I think you need to escape the `embedTitle`. You can use the `escapeHTML` from `server/helpers/core-utils` after moving the function in `shared/core-utils/renderer.html.ts`

server/controllers/services.ts
```diff
@@ -79,6 +79,7 @@ function buildOEmbed (options: {
   const embedUrl = webserverUrl + embedPath
   let embedWidth = EMBED_SIZE.width
   let embedHeight = EMBED_SIZE.height
+  let embedtitle = title
```
Chocobozzz, line 82: Use `const` and escape it

Also it should be `embedTitle`

Chocobozzz (changes requested):

client/src/assets/player/utils.ts
pthavarasa: hi @Chocobozzz can you check if i made correct changes ?

server/controllers/services.ts
```diff
@@ -79,6 +79,7 @@ function buildOEmbed (options: {
   const embedUrl = webserverUrl + embedPath
   let embedWidth = EMBED_SIZE.width
   let embedHeight = EMBED_SIZE.height
+  const embedTitle = title
```
Chocobozzz, line 82: Please `escapeHTML` here too

server/tests/api/server/services.ts
```diff
@@ -88,7 +88,7 @@ describe('Test services', function () {
 
     const res = await getOEmbed(server.url, oembedUrl)
     const expectedHtml = '<iframe width="560" height="315" sandbox="allow-same-origin allow-scripts" ' +
-      `src="http://localhost:${server.port}/video-playlists/embed/${playlistUUID}" ` +
+      `title="${video.name}" src="http://localhost:${server.port}/video-playlists/embed/${playlistUUID}" ` +
```
Chocobozzz, line 91: Should be `title="The Life and Times of Scrooge McDuck" since it's the playlist embed

pthavarasa: can i assign directly the string ?

Chocobozzz:
Thanks @pthavarasa!

lunar-debian:
:+1:  Thanks @pthavarasa and @Chocobozzz !

PR #3909, opened 2021-04-01
Don't pause video upon modal open

### LINKED ISSUE #3816: Don't pause video when modal is opened
When opening a modal (Share, Support or Download) upon watching a video, the video gets paused. As a content creator you'd like the user  to open both Share and Support modals and act on them, I think users avoid to open the modals since it will pause their video. As a user you'd like to share and support a video while watching it.

I'd like our users to be able to open the dialogs while watching a video (without it getting paused).

kontrollanten:
## Description
Removes the pausing of video upon modal open.

## Related issues
Closes #3816 

## Has this been tested?

1) Start a video.
1) Open share modal.
1) Open download modal.
1) Open some other modal from the action menu.

### DIFF
--- a/client/src/app/+videos/+video-watch/video-watch.component.ts
+++ b/client/src/app/+videos/+video-watch/video-watch.component.ts
@@ -282,8 +282,6 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
     // Check video was playing before opening support modal
     const isVideoPlaying = this.isPlaying()
 
-    this.pausePlayer()
-
     const modalRef = this.supportModal.show()
 
     modalRef.result.then(() => {
@@ -294,8 +292,6 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
   }
 
   showShareModal () {
-    this.pausePlayer()
-
     this.videoShareModal.show(this.currentTime, this.videoWatchPlaylist.currentPlaylistPosition)
   }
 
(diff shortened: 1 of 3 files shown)


Chocobozzz:
Hello,

You remove a behaviour added on purpose because the community asked for this feature (but I can't find the issue :/). So I'm not sure what we should do.

kontrollanten:
I couldn't find the issue either. An option could be to make it configurable, but I can't really see any big benefit with auto pausing upon modal open. It may be in some particular cases convenient to have the auto pause, but to me it seems really narrow. I think users expect the video to be playing in as many cases as possible, i.e. in iOS and Android most videos continue to play even when the user closes the browser.

rigelk:
@kontrollanten could you remove the video pause logic for the share modal too?

rigelk (approved):

client/src/app/+videos/+video-watch/video-watch.component.ts
```diff
@@ -282,8 +282,6 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
     // Check video was playing before opening support modal
     const isVideoPlaying = this.isPlaying()
 
-    this.pausePlayer()
-
     const modalRef = this.supportModal.show()
 
     modalRef.result.then(() => {
```
Chocobozzz, line 287: You don't need this anymore if we don't pause the player

PR #3917, opened 2021-04-03
add support for inputOptions and videoFilters in transcode plugin

### LINKED ISSUE #3846: Support inputOptions in ffmpeg plugins
```PeerTube v3.1.0-rc.1```

I'm trying to transcoding ```vp9+opus``` or ```vp9+flac``` in one video,  but peertube default not use ```-strict experimental```, it's cause error exception on ```ffmpeg v4.2.x```

Add ```-strict experimental``` when peertube make m3u8 file command.

### PR DESCRIPTION
**Pull request by TheoLeCalvar**
## Description

This PR adds support for inputOptions and videoFilters in transcode plugins.
Transcode plugins can be used to add support for new encoders.
However, some require inputOptions and thus cannot be used with current transcode plugin API.

For instance, vaapi hardware accelerated encoders require these input options `-hwaccel vaapi -hwaccel_output_format vaapi -vaapi_device /dev/dri/renderD128`

## Related issues

Closes #3846

## Has this been tested?

### DIFF
--- a/server/helpers/ffmpeg-utils.ts
+++ b/server/helpers/ffmpeg-utils.ts
@@ -277,6 +277,7 @@ async function getLiveTranscodingCommand (options: {
       logger.debug('Apply ffmpeg live video params from %s using %s profile.', builderResult.encoder, profile, builderResult)
 
       command.outputOption(`${buildStreamSuffix('-c:v', i)} ${builderResult.encoder}`)
+      command.addInputOptions(builderResult.result.inputOptions)
       command.addOutputOptions(builderResult.result.outputOptions)
     }
 
@@ -294,6 +295,7 @@ async function getLiveTranscodingCommand (options: {
       logger.debug('Apply ffmpeg live audio params from %s using %s profile.', builderResult.encoder, profile, builderResult)
 
       command.outputOption(`${buildStreamSuffix('-c:a', i)} ${builderResult.encoder}`)
+      command.addInputOptions(builderResult.result.inputOptions)
       command.addOutputOptions(builderResult.result.outputOptions)
     }
 
@@ -605,6 +607,7 @@ async function presetVideo (
       localCommand.audioCodec(builderResult.encoder)
     }
 
+    command.addInputOptions(builderResult.result.inputOptions)
     command.addOutputOptions(builderResult.result.outputOptions)
     addDefaultEncoderParams({ command: localCommand, encoder: builderResult.encoder, fps })
   }
--- a/server/helpers/ffmpeg-utils.ts
+++ b/server/helpers/ffmpeg-utils.ts
@@ -396,8 +396,8 @@ async function buildx264VODCommand (command: ffmpeg.FfmpegCommand, options: Tran
 
   if (options.resolution !== undefined) {
     scaleFilterValue = options.isPortraitMode === true
-      ? `${options.resolution}:-2`
-      : `-2:${options.resolution}`
+      ? `w=${options.resolution}:h=-2`
+      : `w=-2:h=${options.resolution}`
   }
 
   command = await presetVideo({ command, input: options.inputPath, transcodeOptions: options, fps, scaleFilterValue })
--- a/server/helpers/ffmpeg-utils.ts
+++ b/server/helpers/ffmpeg-utils.ts
@@ -9,6 +9,7 @@ import { execPromise, promisify0 } from './core-utils'
 import { computeFPS, getAudioStream, getVideoFileFPS } from './ffprobe-utils'
 import { processImage } from './image-utils'
 import { logger } from './logger'
+import { FilterSpecification } from 'fluent-ffmpeg'
 
 /**
  *
@@ -226,21 +227,14 @@ async function getLiveTranscodingCommand (options: {
 
   const varStreamMap: string[] = []
 
-  command.complexFilter([
+  const complexFilter: FilterSpecification[] = [
     {
       inputs: '[v:0]',
       filter: 'split',
       options: resolutions.length,
       outputs: resolutions.map(r => `vtemp${r}`)
-    },
-
-    ...resolutions.map(r => ({
-      inputs: `vtemp${r}`,
-      filter: 'scale',
-      options: `w=-2:h=${r}`,
-      outputs: `vout${r}`
-    }))
-  ])
+    }
+  ]
 
   command.outputOption('-preset superfast')
   command.outputOption('-sc_threshold 0')
@@ -278,6 +272,13 @@ async function getLiveTranscodingCommand (options: {
 
       command.outputOption(`${buildStreamSuffix('-c:v', i)} ${builderResult.encoder}`)
       applyEncoderOptions(command, builderResult.result)
+
+      complexFilter.push({
+        inputs: `vtemp${resolution}`,
+        filter: getScaleFilter(builderResult.result),
+        options: `w=-2:h=${resolution}`,
+        outputs: `vout${resolution}`
+      })
     }
 
     {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: shared/models/videos/video-transcoding.model.ts [hunk 1]
```diff
@@ -12,7 +12,9 @@ export type EncoderOptionsBuilder = (params: {
 export interface EncoderOptions {
   copy?: boolean // Copy stream? Default to false
 
-  outputOptions: string[]
+  inputOptions?: string[]
+  videoFilters?: string[]
```
Chocobozzz, line 16: Please remove this, to not couple too much with `ffmpeg-fluent`. We can already use video filters using `outputOptions`

TheoLeCalvar: I agree that less coupling with `ffmeg-fluent` is better.
However, specifying filters directly in `outputOptions` entails that there is no use of `scale()` or `videoFilters()` anywhere else in peertube code.
`ffmpeg-fluent` keeps track of all filters and generate a single `-filter:v` argument with the entire filter chain because `ffmpeg` will only use the latest `-filter:v` argument and ignore previous ones.

To illustrate the issue:
```js
command .input('in.mkv')
        .size('?x360')
        .outputOptions('-vf vflip')
        .output('out.mp4')
        .on('start', (cmd) => {
            console.log(cmd)
        })
        .run()
```

Generates this command line `ffmpeg -i in.mkv -y -filter:v scale=w=trunc(oh*a/2)*2:h=360 -vf vflip out.mp4` which will ignore the scale filter and only apply the vflip.

support/doc/plugins/guide.md
```diff
@@ -324,6 +324,7 @@ async function register ({
 
       // You can also return a promise
       return {
+        inputOptions: [],
```
Chocobozzz, line 327: Then please add doc for the `videoFilters` param :)

Chocobozzz (commented):
Please could you add tests in https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/plugins/plugin-transcoding.ts using https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/fixtures/peertube-plugin-test-transcoding-one/main.js?

See https://docs.joinpeertube.org/contribute-getting-started?id=unit-tests for the documentation

support/doc/plugins/guide.md
TheoLeCalvar: During my testing with live transcoding profile I noticed that profiles that specify `videoFilters`  do not generate working `ffmpeg` commands because of the complex filter used for scaling all resolutions.

Should I put a warning in the documentation stating that `videoFilters` must not be used for live transcode filters or should I change the `applyEncoderOptions` to ignore `videoFilters` if it is building a live transcode command ?

TheoLeCalvar:
I'm wondering if this PR should also include a mechanism to mitigate #3919 .
Plugins could provide the name of the filter to use for scaling, with `scale` as default value.

With this info we could use the correct filter name here
https://github.com/Chocobozzz/PeerTube/blob/b70025bfdc5f3d7aa3e827f6cadfa1e21483a038/server/helpers/ffmpeg-utils.ts#L239

And we could replace the call to `size()` here https://github.com/Chocobozzz/PeerTube/blob/b70025bfdc5f3d7aa3e827f6cadfa1e21483a038/server/helpers/ffmpeg-utils.ts#L400
by a manually written scale filter that would use the filter name provided by the transcode-plugin.

Chocobozzz:
Okay thanks, so to summary if we do the following things, does it fix all the issues?

  * Don't use `.size()` but a custom `.outputOption` in PeerTube core
  * Add ability for plugins to register a custom scale function that should return `{ filterName: string, value: string }`. The default in PeerTube would be something like `{ filterName: 'scale', value: 'w=trunc(oh*a/2)*2:h=360' }` or `{ filterName: 'scale', value: 'w=-2:h=360' }` (not sure what is the best)
  * Don't provide `.videoFilters` param for plugins

TheoLeCalvar:
That looks good for me, it enough for my use case (adding hardware transcode).

> Add ability for plugins to register a custom scale function that should return { filterName: string, value: string }. The default in PeerTube would be something like { filterName: 'scale', value: 'w=trunc(oh*a/2)*2:h=360' } or { filterName: 'scale', value: 'w=-2:h=360' } (not sure what is the best)

I did a quick search in ffmpeg documentation and it looks like all `scale` filters support the `w=-2:h=XX`  syntax.
I'm not sure why `fluent-ffmpeg` uses `trunc` because `-2` does the same.
The ffmpeg documentation explains it here https://trac.ffmpeg.org/wiki/Scaling

So I think it is not strictly necessary for plugins to provide the arguments to the filter since they are always the same in our use case.

Chocobozzz:
Okay thanks for the feedback. I'll update this PR.

TheoLeCalvar:
Ok, thanks.

Chocobozzz:
I updated the PR accordingly. @TheoLeCalvar Could you review & test please?

server/helpers/ffmpeg-utils.ts
```diff
@@ -277,7 +271,14 @@ async function getLiveTranscodingCommand (options: {
       logger.debug('Apply ffmpeg live video params from %s using %s profile.', builderResult.encoder, profile, builderResult)
 
       command.outputOption(`${buildStreamSuffix('-c:v', i)} ${builderResult.encoder}`)
-      command.addOutputOptions(builderResult.result.outputOptions)
+      applyEncoderOptions(command, builderResult.result)
+
+      complexFilter.push({
+        inputs: `vtemp${resolution}`,
+        filter: getScaleFilter(builderResult.result),
+        options: `w=-2:h=${resolution}`,
```
TheoLeCalvar, line 279: Here you use the `w=${width}:h=${height}`  format whereas you use the `${width}:${height}` format everywhere else.
I think you should stick to one format.

server/lib/video-transcoding-profiles.ts
```diff
@@ -24,9 +24,10 @@ import { VIDEO_TRANSCODING_FPS } from '../initializers/constants'
 
 const defaultX264VODOptionsBuilder: EncoderOptionsBuilder = async ({ input, resolution, fps }) => {
   const targetBitrate = await buildTargetBitrate({ input, resolution, fps })
-  if (!targetBitrate) return { outputOptions: [ ] }
+  if (!targetBitrate) return { inputOptions: [ ], outputOptions: [ ] }
```
TheoLeCalvar, line 27: This changes could be reverted now that `inputOptions` is optional.

TheoLeCalvar: This comment applies for all changes in this file.

support/doc/plugins/guide.md
```diff
@@ -323,7 +323,9 @@ async function register ({
       const streamString = streamNum ? ':' + streamNum : ''
 
       // You can also return a promise
+      // All these options are optional and defaults to []
```
TheoLeCalvar, line 326: You should add the new `scaleFilter` to the documentation and specify its default value.

server/tests/plugins/plugin-transcoding.ts
```diff
@@ -15,9 +15,11 @@ import {
   sendRTMPStreamInVideo,
   setAccessTokensToServers,
   setDefaultVideoChannel,
+  testFfmpegStreamError,
   uninstallPlugin,
   updateCustomSubConfig,
   uploadVideoAndGetId,
+  waitFfmpegUntilError,
```
TheoLeCalvar, line 22: `waitFfmpegUntilError` is not used and should be removed.

TheoLeCalvar (commented):
Except these small problems it looks good to me.
Tests look good.
I also ported my plugin to this version and it works great.

Chocobozzz:
Thanks!

PR #3920, opened 2021-04-03
remove unnecessary black border above and below video in player

violoncelloCH:
## Description
This removes the unnecessary black border/space above and below the video in normal player mode (if browser window is wide). It's achieved by further precising the aspect ratio 16/9 which is not just 1.7, but ~1.777777778. Therefore the video control take a little less vertical space and a 16/9 video fits perfectly into the same space. (see Screenshots below)

## Related issues
none AFAIK

## Has this been tested?

visual difference can be seen on Screenshots below

## Screenshots

| before | after |
| --- | --- |
| image | image |

### DIFF
--- a/client/src/app/+videos/+video-watch/video-watch.component.scss
+++ b/client/src/app/+videos/+video-watch/video-watch.component.scss
@@ -3,7 +3,7 @@
 @import '_bootstrap-variables';
 @import '_miniature';
 
-$player-factor: 1.7; // 16/9
+$player-factor: #{16/9};
 $video-info-margin-left: 44px;
 
 @function getPlayerHeight($width){


violoncelloCH:
ci failure seems unrelated (docker pull error - probably a temporary network issue)

client/src/app/+videos/+video-watch/video-watch.component.scss
```diff
@@ -3,7 +3,7 @@
 @import '_bootstrap-variables';
 @import '_miniature';
 
-$player-factor: 1.7; // 16/9
+$player-factor: 1.7777; // 16/9
```
rigelk, line 6:
```suggestion
$player-factor: #{16/9}
```

rigelk (changes requested):
SASS allows expression evaluation and interpolation ; maybe use it instead?

client/src/app/+videos/+video-watch/video-watch.component.scss
violoncelloCH: ohh, of course... I'm not used to scss and didn't know that was possible... thank you!

Chocobozzz:
Nice! Thanks @violoncelloCH

violoncelloCH:
Thanks for merging!

PR #3923, opened 2021-04-05
Minimal PeertubeModalService to open settings from "can be redefined..."

### LINKED ISSUE #2576: Make the "can be redefined by the users" text to be a hyperlink and link to the settings option where the user can redefine that
Make the "can be redefined by the users" text (from the About page) to be a hyperlink and link to the settings option where the user can redefine that. We can make it just open the settings page or we can make it open the settings page and flash two times (with a somewhat bright color) the option in question. 

This issue concerns both the registered and unregistered users.

aschaap:
## Description

Fixes #2576.

## Has this been tested?

- [X] 🙅 no, because they aren't needed

In the dev environment, open the About page and click on "can be redefined by the users". The expected result is that the settings dialog opens.

### DIFF
--- a/client/src/app/menu/menu.component.ts
+++ b/client/src/app/menu/menu.component.ts
@@ -134,7 +134,7 @@ export class MenuComponent implements OnInit {
       })
 
     this.modalService.openQuickSettingsSubject
-      .subscribe((msg) => this.openQuickSettings());
+      .subscribe((msg) => this.openQuickSettings())
   }
 
   isRegistrationAllowed () {
--- a/client/src/app/menu/menu.component.ts
+++ b/client/src/app/menu/menu.component.ts
@@ -134,7 +134,7 @@ export class MenuComponent implements OnInit {
       })
 
     this.modalService.openQuickSettingsSubject
-      .subscribe((msg) => this.openQuickSettings())
+      .subscribe(() => this.openQuickSettings())
   }
 
   isRegistrationAllowed () {
--- a/client/src/app/menu/menu.component.ts
+++ b/client/src/app/menu/menu.component.ts
@@ -10,6 +10,7 @@ import { LanguageChooserComponent } from '@app/menu/language-chooser.component'
 import { QuickSettingsModalComponent } from '@app/modal/quick-settings-modal.component'
 import { ServerConfig, UserRight, VideoConstant } from '@shared/models'
 import { NgbDropdown, NgbDropdownConfig } from '@ng-bootstrap/ng-bootstrap'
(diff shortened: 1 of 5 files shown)


client/src/app/menu/menu.component.ts
```diff
@@ -130,6 +132,9 @@ export class MenuComponent implements OnInit {
         this.authService.userInformationLoaded
           .subscribe(() => this.buildUserLanguages())
       })
+
+    this.modalService.openQuickSettingsSubject
+      .subscribe((msg) => this.openQuickSettings())
```
Chocobozzz, line 137: `() => this.openQuickSettings`

client/src/app/shared/shared-instance/instance-features-table.component.ts
```diff
@@ -11,7 +12,10 @@ export class InstanceFeaturesTableComponent implements OnInit {
   quotaHelpIndication = ''
   serverConfig: ServerConfig
 
-  constructor (private serverService: ServerService) { }
+  constructor (
+    private serverService: ServerService,
+    private modalService: PeertubeModalService
+    ) { }
```
Chocobozzz, line 18: Align `) { }` with `constructor`

client/src/app/shared/shared-main/peertube-modal/peertube-modal.service.ts
```diff
@@ -0,0 +1,7 @@
+import { Injectable } from '@angular/core'
+import { Subject } from 'rxjs'
+
+@Injectable({ providedIn: 'root' })
+export class PeertubeModalService {
+  openQuickSettingsSubject = new Subject<any[]>()
```
Chocobozzz, line 6: Should be `new Subject<void>()`

Chocobozzz (changes requested):

client/src/app/menu/menu.component.ts
aschaap: Thank you for your prompt review!

aschaap:
Failing api-2 test:
1) Should have a valid playlist oEmbed response

Failing api-3 tests:
1) Should have the correct thumbnails
1) Should automatically update the thumbnail when adding an element
1) Should create a playlist on server 2 and have the playlist on server 1 but not on server 3

@Chocobozzz, I'm not confident these failing tests are related to the changes you suggested; perhaps there is an issue somewhere in the testing infrastructure?

Chocobozzz:
Thanks for the PR and sorry for the tests, my bad.

PR #3925, opened 2021-04-05
Switching from libxmljs to libxmljs2; fixes #3924

### LINKED ISSUE #3924: Switch to libxmljs2 for Node.js 14.x

When setting up the dev environment with Node.js v14.16.0, `yarn install --pure-lockfile` fails on the libxmljs dependency, due to a 404 error when trying to fetch the precompiled binary (and a subsequent 404 when trying to fetch the source). Libxmljs v0.19.7 is ~2 years old, and doesn't appear to be maintained. (Additionally, Node.js v12.x is in maintenance mode.)

Libxmljs2 is an actively maintained fork, and switching to it would have to happen at some point to move on from Node.js v12.x.

I briefly looked for other forks/alternatives, but this seems to be the path of the least resistance.

aschaap:
## Description

Switching from libxmljs to libxmljs2; fixes #3924 .

## Has this been tested?

Running `yarn install --pure-lockfile` succeeds.


Chocobozzz:
See https://github.com/Chocobozzz/PeerTube/issues/3924#issuecomment-813252302

PR #3933, opened 2021-04-07
Resumable video uploads

### LINKED ISSUE #324: Allow uploads to be resumed

kontrollanten:
## Description
Make audio/video uploads resumable when the internet connection is flaky. This is done by using [node-uploadx](https://github.com/kukhariev/node-uploadx) and [ngx-upload](https://github.com/kukhariev/ngx-uploadx). This family of libraries supports multiple operations modes ([TUS v1.0](https://github.com/tus/tus-resumable-upload-protocol/blob/master/protocol.md), a [Google Resumable v3.0](https://developers.google.com/drive/api/v3/manage-uploads#resumable) variant and multipart). We selected the default mode, which is using a variant of Google Resumable (used in Drive and YouTube upload endpoints) and is summarized here: https://github.com/kukhariev/node-uploadx/blob/master/proto.md

`/api/v1/videos/upload` was left untouched.
`/api/v1/videos/upload-resumable` was created and dedicated to this new upload method.

The OpenAPI spec was modified to include the new upload API.

## Related issues
closes #324
https://github.com/kukhariev/node-uploadx/issues/318

## Has this been tested?

  - [x] video upload
  - [x] audio upload

## Screenshots

Resume video upload


server/controllers/api/videos/index.ts
```diff
@@ -112,18 +103,30 @@ videosRouter.get('/',
   commonVideosFiltersValidator,
   asyncMiddleware(listVideos)
 )
+
+const getTmpPath = (fileId: string) => `/tmp/peertube-${fileId}`
+
+videosRouter.use('/x-upload', uploadx({
```
rigelk, line 109: I would rather use a more descriptive route name, like `resumable-upload`.

rigelk (changes requested):
@kontrollanten thanks for taking care of this issue! I didn't review much the backend, but will check more thoroughly once the PR leaves its draft state.

rigelk:
@kontrollanten btw if you need help for the implementation (like tests or tmp cleanup as job), I can give you a hand.

kontrollanten:
> @kontrollanten btw if you need help for the implementation (like tests or tmp cleanup as job), I can give you a hand.

Thanks. I'm thinking about what the best solution is for the server tests that are using the `shared/extra-utils/videos/uploadVideo` function. In one way the new upload process is an implementation detail that are out of scope for our testing purpose, but in the other way we need an easy way to upload videos in the tests. But to rewrite the logic from `ngx-uploadx` just for testing purpose feels overkill.

rigelk:
It seems node-uploadx has been updated with the required changes to further this PR: https://github.com/kukhariev/node-uploadx/releases/tag/v4.3.0

server/controllers/api/videos/index.ts
```diff
@@ -112,18 +107,66 @@ videosRouter.get('/',
   commonVideosFiltersValidator,
   asyncMiddleware(listVideos)
 )
+
+const getTmpPath = (fileId: string) => `/tmp/peertube-${fileId}`
+
+const clearUploadFile = filePath =>
+  existsSync(filePath) && unlinkSync(filePath)
```
rigelk, line 114:
```suggestion
function clearUploadFile (filePath: string) {
  return Promise.all(
    exists(filePath),
    unlink(filePath)
  ).catch(err => logger.warn('Cannot clear upload file %s.', filePath, { err }))
}
```

rigelk (changes requested):

server/controllers/api/videos/index.ts
kontrollanten: Beautiful

client/src/app/+videos/+video-edit/video-add-components/video-upload.component.ts
```diff
@@ -78,16 +139,17 @@ export class VideoUploadComponent extends VideoSend implements OnInit, AfterView
...
+      )
   }
 
   ngAfterViewInit () {
     this.hooks.runAction('action:video-upload.init', 'video-edit')
   }
 
-  ngOnDestroy () {
-    if (this.videoUploadObservable) this.videoUploadObservable.unsubscribe()
-  }
```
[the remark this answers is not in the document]
kontrollanten: By not providing any uploadId all uploads are paused, and that's what we want?

https://github.com/kukhariev/ngx-uploadx/blob/637e258fe366b8095203f387a6101a230ee4f8e6/src/uploadx/lib/uploadx.service.ts#L112-L113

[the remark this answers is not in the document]
kontrollanten: Shouldn't we cancel the upload upon component deletion?

client/src/app/+videos/+video-edit/video-add-components/video-upload.component.ts
```diff
@@ -147,95 +196,42 @@ export class VideoUploadComponent extends VideoSend implements OnInit, AfterView
     this.notifier.info($localize`Upload cancelled`)
```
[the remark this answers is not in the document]
kontrollanten: True!

client/src/app/+videos/+video-edit/video-add-components/video-upload.component.ts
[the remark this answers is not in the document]
rigelk: Oh I missed that comment! Thanks.

Yes, for now let's keep things simple and cancel the upload upon component deletion (like a route change, etc.).

server/middlewares/validators/videos/videos.ts
```diff
@@ -70,58 +65,48 @@ const videosAddValidator = getCommonVideoEditAttributes().concat([
...
 
-    if (!isVideoFileSizeValid(videoFile.size.toString())) {
-      res.status(HttpStatusCode.PAYLOAD_TOO_LARGE_413)
+    if (!isVideoFileSizeValid(videoFileMetadata.size.toString())) {
+      return res.status(HttpStatusCode.PAYLOAD_TOO_LARGE_413)
          .json({
            error: 'This file is too large.'
          })
-
-      return cleanUpReqFiles(req)
```
rigelk, line 99: You don't clean the stored file in the middleware anymore ; is there a reason?

rigelk:
@kontrollanten I rebased and moved your code to `upload-resumable`, and restored the `upload` route. This prevents removing from the API blueprint, and also allows us to not break tests at the same time as we add new ones.

kontrollanten:
Thanks @rigelk ! What do we got left to be able to merge this? Tests and a cleanup job for /tmp? If you haven't started on the cleanup job I think I can do that. I that it should be a repeat job that runs once an hour.

rigelk:
> Tests and a cleanup job for /tmp?

Yes, and I'd like to cleanup the object passing (file, metadata) which is brittle at the moment. Having well-typed object passing there is important, and especially regarding the plugin hook which receives the object.

As for tests, we need to try not as much as the legacy upload, but at least the audio-only resumable upload, which is handled very differently from the rest within the middleware.

server/controllers/api/videos/index.ts
```diff
@@ -184,18 +200,43 @@ async function addVideo (req: express.Request, res: express.Response) {
...
+
+  return addVideo(req, res, { videoPhysicalFile, videoInfo, files })
+}
+
+async function addVideoResumable (req: express.Request, res: express.Response) {
+  const videoPhysicalFile = res.locals.videoFile as any
+  const videoInfo: VideoCreate & { size: number} = videoPhysicalFile.metadata
+  const files = {
+    bg: {
+      path: getTmpPath(videoPhysicalFile.metadata.audioBg)
```
rigelk, line 213: @kontrollanten why did you name it `bg` and `audioBg`?

kontrollanten: As far as I remember the `bg` name isn't used anywhere, it was just a necessary key to be able to pass to `buildVideoThumbnailsFromReq`. But of course it'd be better to name it to `audioBg` to decrease confusion risk.

package.json
```diff
@@ -127,6 +127,7 @@
     "morgan": "^1.5.3",
     "multer": "^1.1.0",
     "node-media-server": "^2.1.4",
+    "node-uploadx": "^4.3.0",
```
kukhariev, line 130: To avoid including aws and google packages, it is better to import only `@uploadx/core`

server/controllers/api/videos/index.ts
```diff
@@ -66,10 +67,15 @@ import { liveRouter } from './live'
 import { ownershipVideoRouter } from './ownership'
 import { rateVideoRouter } from './rate'
 import { watchingRouter } from './watching'
+import { DiskStorageOptions, uploadx } from 'node-uploadx'
```
kukhariev, line 70:
```ts
import { DiskStorageOptions, uploadx } from '@uploadx/core'
```

package.json
rigelk: Thanks for the heads up!

client/src/app/+videos/+video-edit/video-add-components/video-upload.component.ts
```diff
@@ -205,7 +205,7 @@ export class VideoUploadComponent extends VideoSend implements OnInit, OnDestroy
       this.resumableUploadService.handleFiles(this.previewfileUpload, {
         ...this.options,
         metadata: {
-          isAudioBg: true
+          isAudio: true
```
kontrollanten, line 208: Why rename this? We're sending the background image here, not the audio file.

rigelk: You're right. I will rename it to `previewfileUploadForAudio` or something more explicit, because `Bg` is not clear enough ; it got me confused at least --'

kontrollanten:
It looks like the cleanup job fails to run in the tests, but I've no idea how to debug.

server/helpers/utils.ts
```diff
@@ -1,3 +1,5 @@
+import { promises as fsp } from 'fs'
```
rigelk, line 1: don't use that pattern, and rather directly use promisable functions from `fs-extra`

package.json
Chocobozzz: Is is resolved?

rigelk: Yes, in [`db16ad7` (#3933)](https://github.com/Chocobozzz/PeerTube/pull/3933/commits/db16ad7420ca3f3d1456e13c723c40843b05875c)

rigelk: My bad, I had misunderstood. Solved in the [`0379f07` (#3933)](https://github.com/Chocobozzz/PeerTube/pull/3933/commits/0379f0762d2002b191253f689153864b45baafa7).

server/helpers/utils.ts
```diff
@@ -11,6 +11,13 @@ function deleteFileAsync (path: string) {
     .catch(err => logger.error('Cannot delete the file %s asynchronously.', path, { err }))
 }
 
+function clearFile (filePath: string) {
```
Chocobozzz, line 14: Doesn't seem to be used

server/helpers/utils.ts
```diff
@@ -80,14 +87,29 @@ function getUUIDFromFilename (filename: string) {
   return result[0]
 }
 
+let tmpPath
+async function getTmpPath (filename?: string) {
+  if (!tmpPath) {
+    try {
+      tmpPath = join(CONFIG.STORAGE.TMP_DIR, 'resumable-uploads')
```
Chocobozzz, line 94: Create the directory in the installer function so you don't need an async function here: https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/installer.ts#L80

And please use a better name than `getTmpPath`

server/initializers/constants.ts
```diff
@@ -211,7 +217,8 @@ const SCHEDULER_INTERVALS_MS = {
   autoFollowIndexInstances: 60000 * 60 * 24, // 1 day
   removeOldViews: 60000 * 60 * 24, // 1 day
   removeOldHistory: 60000 * 60 * 24, // 1 day
-  updateInboxStats: 1000 * 60// 1 minute
+  updateInboxStats: 1000 * 60, // 1 minute
+  removeTmpVideo: 60000 * 60 // 1 hour
```
Chocobozzz, line 221: Either use a scheduler class or a repeatable job, but don't mix them. The scheduler is more simple but don't know when it will run (depends when the server was started)

server/lib/job-queue/handlers/video-upload-tmp-folder-cleaner.ts
```diff
@@ -0,0 +1,53 @@
...
+import { SCHEDULER_INTERVALS_MS } from '@server/initializers/constants'
+
+async function processVideoUploadTmpFolderCleaner () {
+  const tmpPath = await getTmpPath()
+  let files: string[]
+
+  logger.debug('processVideoUploadTmpFolderCleaner: Reading files from path %s', tmpPath)
+
+  try {
+    files = await readdir(tmpPath)
```
Chocobozzz, line 15: If you created the folder at peertube startup, this function should not fail so you don't need to try catch this case

server/lib/job-queue/handlers/video-upload-tmp-folder-cleaner.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as path from 'path'
+import { logger } from '@server/helpers/logger'
+import { getTmpPath } from '@server/helpers/utils'
+import { SCHEDULER_INTERVALS_MS } from '@server/initializers/constants'
+
+async function processVideoUploadTmpFolderCleaner () {
+  const tmpPath = await getTmpPath()
+  let files: string[]
+
+  logger.debug('processVideoUploadTmpFolderCleaner: Reading files from path %s', tmpPath)
```
Chocobozzz, line 12: Merge this debug with the one line 17

server/lib/job-queue/handlers/video-upload-tmp-folder-cleaner.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  try {
+    files = await readdir(tmpPath)
+  } catch (error) {
+    logger.error('Failed to read resumable video upload tmp folder.', { error })
+    return
+  }
+
+  logger.debug('processVideoUploadTmpFolderCleaner: Found %d files', files.length)
```
Chocobozzz, line 21: Don't use a prefix, use logger tags instead: https://github.com/Chocobozzz/PeerTube/blob/develop/server/controllers/api/videos/index.ts#L70 & https://github.com/Chocobozzz/PeerTube/blob/develop/server/controllers/api/videos/index.ts#L261

server/lib/job-queue/handlers/video-upload-tmp-folder-cleaner.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  await Promise.all(files.map(async function (file) {
+    const filePath = path.join(tmpPath, file)
+    let statResult: Stats
+
+    try {
```
Chocobozzz, line 27: Wrap the entire map function in a try/catch since you don't recover from the error

server/lib/job-queue/handlers/video-upload-tmp-folder-cleaner.ts
```diff
@@ -0,0 +1,53 @@
...
+      statResult = await stat(filePath)
+    } catch (error) {
+      logger.error(`Failed to run stat for ${filePath}`, { error })
+      return
+    }
+
+    // current time (in miliseconds):
+    const now = new Date().getTime()
```
Chocobozzz, line 35: Move it outside from the map

server/lib/job-queue/handlers/video-upload-tmp-folder-cleaner.ts
```diff
@@ -0,0 +1,53 @@
...
+    // time of last status change, plus 1h (in miliseconds):
+    const endTime = new Date(statResult.ctime).getTime() + SCHEDULER_INTERVALS_MS.removeTmpVideo
```
Chocobozzz, line 37: I think you can use `statResult.ctimeMs` directly?

Chocobozzz, line 37: I don't think you should use `SCHEDULER_INTERVALS_MS.removeTmpVideo`

Instead, use a private variable to store the last run time (so do nothing for the first run), and delete files < `lastRunTime`

Also, I don't think 1 hour is enough since uploads can be very long

server/lib/job-queue/handlers/video-upload-tmp-folder-cleaner.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    if (now > endTime) {
+      try {
+        await unlink(filePath)
```
Chocobozzz, line 41: Use `remove` from fs-extra

server/middlewares/validators/videos/videos.ts
```diff
@@ -126,6 +129,92 @@ const videosAddValidator = getCommonVideoEditAttributes().concat([
...
+    .custom(isIdValid).withMessage('Should have correct video channel id'),
+
+  async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    const videoFileMetadata = {
+      mimetype: req.body.mimeType,
+      size: req.body.size,
+      originalname: req.body.name
+    } as express.FileUploadMetadata
+    const user = res.locals.oauth.token.User
+    const file: Express.Multer.File & { id: string, metadata: any } = req.body
```
Chocobozzz, line 148: What is `id`? And is it possible to type `metadata`?

server/middlewares/validators/videos/videos.ts
```diff
@@ -126,6 +129,92 @@ const videosAddValidator = getCommonVideoEditAttributes().concat([
...
+    file.path = `${CONFIG.STORAGE.VIDEOS_DIR}${file.id}`
```
Chocobozzz, line 149: Is it a directory/file name or a path? If it's a path, use `join()`

server/middlewares/validators/videos/videos.ts
```diff
@@ -126,6 +129,92 @@ const videosAddValidator = getCommonVideoEditAttributes().concat([
...
+
+    if (
+      !file.metadata.isPreviewForAudio &&
+      !await doesVideoChannelOfAccountExist(file.metadata.channelId, user, res)
+    ) return
+
+    if (!await isVideoAccepted(req, res, file)) {
+      await clearUploadFile(file.path)
+      return res.status(HttpStatusCode.BAD_REQUEST_400).json({ error: 'Video was refused.' })
```
Chocobozzz, line 158: Yo don't need to send a status, `isVideoAccepted` already does the job

server/middlewares/validators/videos/videos.ts
```diff
@@ -126,6 +129,92 @@ const videosAddValidator = getCommonVideoEditAttributes().concat([
...
+      originalname: req.body.name
+    } as express.FileUploadMetadata
+    const user = res.locals.oauth.token.User
+    const file: Express.Multer.File & { id: string, metadata: any } = req.body
+    file.path = `${CONFIG.STORAGE.VIDEOS_DIR}${file.id}`
+
+    if (
+      !file.metadata.isPreviewForAudio &&
+      !await doesVideoChannelOfAccountExist(file.metadata.channelId, user, res)
+    ) return
```
Chocobozzz, line 154: You don't need to cleanup the files?

server/middlewares/validators/videos/videos.ts
```diff
@@ -126,6 +129,92 @@ const videosAddValidator = getCommonVideoEditAttributes().concat([
...
+
+    if (!await isVideoAccepted(req, res, file)) {
+      await clearUploadFile(file.path)
+      return res.status(HttpStatusCode.BAD_REQUEST_400).json({ error: 'Video was refused.' })
+    }
+
+    if (file.metadata.isPreviewForAudio) {
+      const filename = `${file.id}-${uuidv4()}`
+      file.filename = filename
+      try {
```
Chocobozzz, line 164: I don't think you should catch the error, the middleware should handle the error to automatically returns 500 and log the error

server/middlewares/validators/videos/videos.ts
```diff
@@ -126,6 +129,92 @@ const videosAddValidator = getCommonVideoEditAttributes().concat([
...
+
+    res.locals.videoFileResumable = file
+
+    if (req.method !== 'POST' || req.body.isPreviewForAudio) {
+      return next()
+    } /** The middleware applies what follows only to non-audio POST */
+
+    logger.debug('Checking videosAddResumable parameters', { parameters: req.body, files: req.files })
+
+    if (areValidationErrors(req, res)) return
```
Chocobozzz, line 189: Every `return` should cleanup the file no?

server/middlewares/validators/videos/videos.ts
```diff
@@ -534,3 +628,11 @@ async function isVideoAccepted (req: express.Request, res: express.Response, vid
 
   return true
 }
+
+async function addDurationToVideo (videoFile: Express.Multer.File & { duration?: number, metadata?: any }) {
+  const duration: number = await getDurationFromVideoFile(videoFile.path)
+
+  if (!isNaN(duration)) {
```
Chocobozzz, line 635: If it's a `NaN`, shouldn't we fail?

shared/extra-utils/videos/videos.ts
```diff
@@ -378,61 +384,107 @@ async function uploadVideo (url: string, accessToken: string, videoAttributesArg
...
-              .field('privacy', attributes.privacy.toString())
-              .field('channelId', attributes.channelId)
-
-  if (attributes.support !== undefined) {
-    req.field('support', attributes.support)
-  }
+  /* START part varying depending on selected mode */
+  let res
+  if (mode === 'legacy') {
+    const path = '/api/v1/videos/upload'
```
Chocobozzz, line 390: Please create a dedicated function to build the legacy upload

server/controllers/api/videos/index.ts
```diff
@@ -112,18 +119,27 @@ videosRouter.get('/',
   commonVideosFiltersValidator,
   asyncMiddleware(listVideos)
 )
+
+videosRouter.post('/upload',
+  authenticate,
+  reqVideoFileAdd,
+  asyncMiddleware(videosAddValidator),
```
Chocobozzz, line 126: Please rename to `videosAddLegacyValidator`

server/controllers/api/videos/index.ts
```diff
@@ -184,17 +200,46 @@ async function addVideo (req: express.Request, res: express.Response) {
...
+async function addVideoResumable (req: express.Request, res: express.Response) {
+  interface VideoPhysicalFile {
+    duration: number
+    filename: string
+    size: number
+    path: string
+    metadata: VideoCreate & express.FileUploadMetadata
+  }
+
+  const videoPhysicalFile = res.locals.videoFileResumable as VideoPhysicalFile
```
Chocobozzz, line 217: This is not typed in express typings?

server/controllers/api/videos/index.ts
```diff
@@ -184,17 +200,46 @@ async function addVideo (req: express.Request, res: express.Response) {
...
+  const videoInfo = videoPhysicalFile.metadata
+  const files = { bg: { path: await getTmpPath(videoPhysicalFile.filename) } }
```
Chocobozzz, line 219: What is/means `bg`?

server/controllers/api/videos/index.ts
```diff
@@ -184,17 +200,46 @@ async function addVideo (req: express.Request, res: express.Response) {
...
+
+  return addVideo(req, res, { videoPhysicalFile, videoInfo, files })
+}
+
+async function addVideo (req: express.Request, res: express.Response, parameters: {
+  videoPhysicalFile: { duration: number, filename: string, size: number, path: string }
+  videoInfo: VideoCreate | VideoCreate & express.FileUploadMetadata
```
Chocobozzz, line 226: Please create a dedicated type for `VideoCreate & express.FileUploadMetadata`

server/controllers/api/videos/index.ts
```diff
@@ -184,17 +200,46 @@ async function addVideo (req: express.Request, res: express.Response) {
...
   videoData.duration = videoPhysicalFile['duration'] // duration was added by a previous middleware
 
   const video = new VideoModel(videoData) as MVideoFullLight
-  video.VideoChannel = res.locals.videoChannel
+  video.VideoChannel = videoChannel
   video.url = getLocalVideoActivityPubUrl(video) // We use the UUID, so set the URL after building the object
 
   const videoFile = new VideoFileModel({
-    extname: extname(videoPhysicalFile.filename),
+    extname: videoPhysicalFile.filename ? extname(videoPhysicalFile.filename) : extname(videoInfo.name),
```
Chocobozzz, line 242: You don't have `.filename` for a resumable upload? If no, could we add a `extension` field to the resumable upload metadata? Because I'm not confident in guessing the extension using `videoInfo.name`

server/controllers/api/videos/index.ts
```diff
@@ -112,18 +119,27 @@ videosRouter.get('/',
...
 )
+
+videosRouter.post('/upload',
+  authenticate,
+  reqVideoFileAdd,
+  asyncMiddleware(videosAddValidator),
+  asyncRetryTransactionMiddleware(addVideoLegacy)
+)
+
+videosRouter.use('/upload-resumable',
```
Chocobozzz, line 130: `/resumable-upload`?

Chocobozzz (changes requested):
Great work!

server/controllers/api/videos/index.ts
rigelk: Me neither! That's the main cause of errors, as we use names without extensions in the tests (except for `resumable` upload ones, ofc).

server/controllers/api/videos/index.ts
rigelk: I meant `-resumable` as a suffix to convey the notion this is a variant.

server/helpers/utils.ts
rigelk: It is, its import is just renamed: https://github.com/Chocobozzz/PeerTube//blob/0379f0762d2002b191253f689153864b45baafa7/server/middlewares/validators/videos/videos.ts#L62

Chocobozzz: Okay thanks, then why don't you use `remove()` from fs extr?

rigelk: I though unlink had a special meaning, but it seems we can use just use `deleteFileAsync` from the existing utils…

server/middlewares/validators/videos/videos.ts
rigelk: `id` is the upload id, AFAIK. `metadata` has been typed to the interface provided by uploadx.

server/middlewares/validators/videos/videos.ts
```diff
@@ -56,8 +58,10 @@ import { AccountModel } from '../../../models/account/account'
 import { VideoModel } from '../../../models/video/video'
 import { authenticatePromiseIfNeeded } from '../../auth'
 import { areValidationErrors } from '../utils'
+import { getResumableUploadPath, deleteFileAsync as clearUploadFile } from '../../../helpers/utils'
+import { File } from '@uploadx/core'
```
Chocobozzz, line 62: `File as UploadXFile`

server/middlewares/validators/videos/videos.ts
```diff
@@ -1,5 +1,7 @@
+/* eslint-disable max-len */
```
Chocobozzz, line 1: Just use this comment for the line where you need it

server/controllers/api/videos/index.ts
```diff
@@ -184,17 +203,38 @@ async function addVideo (req: express.Request, res: express.Response) {
...
+  const videoInfo = videoPhysicalFile.metadata
+  const files = { image: { path: getResumableUploadPath(videoPhysicalFile.filename) } }
+
+  return addVideo(req, res, { videoPhysicalFile, videoInfo, files })
+}
+
+async function addVideo (req: express.Request, res: express.Response, parameters: {
+  videoPhysicalFile: { duration: number, filename: string, size: number, path: string }
+  videoInfo: VideoCreate | Metadata
+  files
```
Chocobozzz, line 222: Could you type this?

server/controllers/api/videos/index.ts
```diff
@@ -184,17 +203,38 @@ async function addVideo (req: express.Request, res: express.Response) {
...
+  const files = req.files
+
+  return addVideo(req, res, { videoPhysicalFile, videoInfo, files })
+}
 
-  const videoData = buildLocalVideoFromReq(videoInfo, res.locals.videoChannel.id)
+async function addVideoResumable (req: express.Request, res: express.Response) {
+  const videoPhysicalFile = res.locals.videoFileResumable
+  const videoInfo = videoPhysicalFile.metadata
+  const files = { image: { path: getResumableUploadPath(videoPhysicalFile.filename) } }
```
Chocobozzz, line 214: I don't understand what is `image`, and why we assign the path of the video

Chocobozzz (commented):
Please add tests for:
 * RemoveDanglingResumableUploadsScheduler
 * Same check params than legacy upload (I think you can easily automate these tests using a function)
 * Add more upload tests, especially when the user specifies an audio background, a preview/thumbnail
 * Quota tests

Also, how do you ensure a user is not able to upload many resumable uploads during 12 hours (cleaner schedule) filling server disk?

client/src/app/+videos/+video-edit/video-add-components/video-upload.component.ts
```diff
@@ -275,6 +273,50 @@ export class VideoUploadComponent extends VideoSend implements OnInit, AfterView
         )
   }
 
+  private uploadVideoFile (audioBg?: string) {
+    this.resumableUploadService.handleFiles(this.getVideoFile(), {
+      ...this.options,
+      metadata: {
+        ...this.options.metadata,
+        audioBg
```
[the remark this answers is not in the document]
kontrollanten: It was used by the BE to add the background to the video, but now I can't find the implementation in the backend.

server/controllers/api/videos/index.ts
kontrollanten: It was hard to find a good name. I went for `UploadVideoFiles`, but feel free to propose something better.

server/lib/schedulers/remove-dangling-resumable-uploads-scheduler.ts
```diff
@@ -0,0 +1,55 @@
...
+  let statResult: Stats
+
+  try {
+    statResult = await stat(filePath)
+  } catch (error) {
+    logger.error('Failed to run stat for %s', filePath, { error, ...lTags() })
+    return
+  }
+
+  if (statResult.ctimeMs < limit) await deleteFileAsync(filePath)
```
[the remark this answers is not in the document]
kontrollanten: Isn't it better to just fix the `deleFileAsync` function and use the error logging in there?

server/middlewares/validators/videos/videos.ts
```diff
@@ -106,26 +108,130 @@ const videosAddValidator = getCommonVideoEditAttributes().concat([
...
+    .isNumeric()
+    .exists()
+    .withMessage('Should specify the file length'),
+  header('x-upload-content-type')
+    .isString()
+    .exists()
+    .withMessage('Should specify the file mimetype'),
+
+  async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    if (req.method !== 'POST') return next()
```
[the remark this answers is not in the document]
kontrollanten: I think so, cause this is called also upon `PATCH` and `DELETE`.

server/middlewares/validators/videos/videos.ts
```diff
@@ -534,3 +646,11 @@ async function isVideoAccepted (req: express.Request, res: express.Response, vid
 
   return true
 }
+
+async function addDurationToVideo (videoFile: any) {
+  const duration: number = await getDurationFromVideoFile(videoFile.path)
+
+  if (isNaN(duration)) throw new Error("Couldn't get video duration")
```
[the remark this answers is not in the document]
kontrollanten: Is there a purpose that the `typescript-eslint/quotes` rule is set to `off`?

server/middlewares/async.ts
```diff
@@ -31,9 +32,31 @@ function asyncRetryTransactionMiddleware (fun: (req: Request, res: Response, nex
...
+      ? Promise.resolve((fun as RequestHandler)(req, res, next))
+      : next()
+  }
+}
+
+function executeIfPOST (fun) {
+  return executeIfMethod(fun, 'POST')
+}
+
+function onlyAllowMethods (methods: string[]) {
```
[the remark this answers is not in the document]
kontrollanten: I wasn't able to solve this, I got type errors in the method since `req.method` is typed as `string`. I'm novice to typescript so maybe that's possible to solve.

server/middlewares/validators/videos/videos.ts
```diff
@@ -106,26 +108,130 @@ const videosAddValidator = getCommonVideoEditAttributes().concat([
...
+    }
+
+    if (!isVideoFileSizeValid(videoFileMetadata.size.toString())) {
+      res.status(HttpStatusCode.PAYLOAD_TOO_LARGE_413)
+         .json({ error: 'This file is too large.' })
+
+      return clearUploadFile(file.path)
+    }
+
+    if (await isAbleToUploadVideo(user.id, videoFileMetadata.size) === false) {
```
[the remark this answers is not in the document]
kontrollanten: What do you say about refactoring those validators to middlewares (instead of helpers)? Then it would look like something below (I think the function name should be renamed though). That makes it easier to get an overview about whats validated.

```
videosRouter.post('/upload',
  authenticate,
  reqVideoFileAdd,
  isAbleToUploadVideo,
  areErrorsInScheduleUpdate,
  doesVideoChannelOfAccountExist,
  asyncMiddleware(videosAddLegacyValidator),
  asyncRetryTransactionMiddleware(addVideoLegacy)
)

videosRouter.all('/upload-resumable',
  onlyAllowMethods([ 'POST', 'PUT', 'DELETE' ]), // uploadx also allows GET and PATCH
  authenticate,
  executeIfPOST(asyncMiddleware(videosAddResumableInitValidator)),
  uploadx.upload(uploadxOptions), // uploadx doesn't use call next() before the file upload completes
  isAbleToUploadVideo,
  areErrorsInScheduleUpdate,
  doesVideoChannelOfAccountExist,
  asyncMiddleware(videosAddResumableValidator),
  asyncMiddleware(addVideoResumable)
)
```

server/middlewares/async.ts
```diff
@@ -32,7 +32,8 @@ function asyncRetryTransactionMiddleware (fun: (req: Request, res: Response, nex
   }
 }
 
-function executeIfMethod (fun: (req: Request, res: Response, next: NextFunction) => void | Promise<void>, method: string) {
+function executeIfMethod (fun: (req: Request, res: Response, next: NextFunction)
+=> void | Promise<void>, method: 'GET' | 'POST' | 'PUT' | 'PATCH') {
```
rigelk, line 36:
```suggestion
=> void | Promise<void>, method: 'CONNECT' | 'DELETE' | 'GET' | 'HEAD' | 'OPTIONS' |  'PATCH' | 'POST' | 'PUT' | 'TRACE') {
```

rigelk (dismissed):

server/lib/schedulers/remove-dangling-resumable-uploads-scheduler.ts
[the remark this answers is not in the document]
Chocobozzz: I meant `deleteFileAsync` is not `async` (so not awaitable) which is its purpose: delete in async way and log potential errors so the caller doesn't need to take care about the `.catch`. But I agree its name is not very good

server/middlewares/validators/videos/videos.ts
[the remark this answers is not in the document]
Chocobozzz: We already check the post method in https://github.com/Chocobozzz/PeerTube/pull/3933/files/507c8fd6ae7745dec4facb0b9287191235101438#diff-2a52ca834ee742d4858fbac2097a6454ad1cb41079473d23a86d5f17163f1499R134

server/middlewares/validators/videos/videos.ts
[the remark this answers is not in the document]
Chocobozzz: I think I disabled this rule because it was too strict: we want to use single quotes first but allow backticks when the string contains single quotes (to avoid escaping)

server/middlewares/validators/videos/videos.ts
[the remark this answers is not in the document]
Chocobozzz: Both methods seems fine but I prefer to use helpers for consistency (so we don't have to also update the update endpoint etc).

server/middlewares/async.ts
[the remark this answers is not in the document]
rigelk: I'll handle it @kontrollanten 😉

server/middlewares/validators/videos/videos.ts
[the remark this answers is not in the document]
rigelk: Yes I should have removed it.

client/src/app/+videos/+video-edit/video-add-components/video-upload.component.ts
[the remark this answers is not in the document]
rigelk: @kontrollanten could you please check why that is, and if the background is handled any other way?

server/controllers/api/videos/index.ts
rigelk: I'm pretty sure this has to do with the audio background image being set for _some_ requests, but that should be clarified by @kontrollanten (who named it 'audioBg' initially).

client/src/app/+videos/+video-edit/video-add-components/video-upload.component.ts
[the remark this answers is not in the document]
kontrollanten: The background for audio files seems to have been broken since this was implemented. I'll try to reimplement it.

server/controllers/api/videos/index.ts
kontrollanten: A lot has changed since I implemented this. As far as I remember `files` was sent to a function that ran `Object.keys(files).forEach(...` on it. And it was used only when doing audio uploads (with a background image).

I tried to investigate it now and it doesn't seem to have any effect by setting `const files = undefined`. But I can't understand how the backend is able to pair the image file with the audio file anymore. Before the only link between them was the audioBg filename sent by the client, but this doesn't seem to be the case anymore. @rigelk do you know if this has changed? Otherwise I'll investigate further tomorrow.

server/middlewares/validators/videos/videos.ts
```diff
@@ -204,31 +182,10 @@ const videosAddResumableInitValidator = getCommonVideoEditAttributes().concat([
...
-      return clearUploadFile(file.path)
-    }
-
-    if (await isAbleToUploadVideo(user.id, videoFileMetadata.size) === false) {
-      res.status(HttpStatusCode.PAYLOAD_TOO_LARGE_413)
-         .json({ error: 'The user video quota is exceeded with this video.' })
-
-      return clearUploadFile(file.path)
-    }
+    /* eslint-disable @typescript-eslint/no-floating-promises */
```
kontrollanten, line 187:
```suggestion
    // eslint-disable-next-line @typescript-eslint/no-floating-promise

File: server/middlewares/validators/videos/videos.ts [hunk 47]
```diff
@@ -77,50 +80,123 @@ const videosAddValidator = getCommonVideoEditAttributes().concat([
...
+    try {
+      const storage = new DiskStorage({ directory: getResumableUploadPath() })
+      const userPartiallyUploadedFiles = await storage.get(user.id) // uploadx's default naming relies on req.user.id to prefix file names
+      const userPartiallyUploadedFilesSize = await Promise.all(userPartiallyUploadedFiles.map(f => stat(f.name).then(s => s.size)))
+      const userPartiallyUploadedFilesTotalSize = userPartiallyUploadedFilesSize.reduce((a, b) => a + b, 0) + file.size
+      if (userPartiallyUploadedFilesTotalSize > CONSTRAINTS_FIELDS.VIDEOS.PARTIAL_UPLOAD_SIZE.max) throw new Error()
+    } catch (err) {
+      res.status(HttpStatusCode.PAYLOAD_TOO_LARGE_413)
+         .json({ error: 'This file is too large. It exceeds the partial uploads combined file size authorized.' })
+    }
```
rigelk, line 199: @Chocobozzz is it enough of a countermeasure for partial uploads filling the disk?

server/controllers/api/videos/index.ts
rigelk: Nevermind @kontrollanten I got it ; my initial refactorings had lost the meaning of your code regarding that part. My next commit will adress that.

client/src/app/+videos/+video-edit/video-add-components/video-upload.component.ts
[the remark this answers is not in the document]
rigelk: As mentioned in https://github.com/Chocobozzz/PeerTube/pull/3933#discussion_r621166089

server/controllers/api/videos/index.ts
rigelk: I managed to fix it by properly passing the `filename` body parameter, which was properly injected by `ngx-upload` but was missing from my test helper.

server/controllers/api/videos/index.ts
```diff
@@ -212,7 +212,9 @@ async function addVideoLegacy (req: express.Request, res: express.Response) {
 async function addVideoResumable (req: express.Request, res: express.Response) {
   const videoPhysicalFile = res.locals.videoFileResumable
   const videoInfo = videoPhysicalFile.metadata
-  const files = { image: { path: getResumableUploadPath(videoPhysicalFile.filename) } }
+  const files = { // only used for audio background generation
+    audioBg: { path: getResumableUploadPath(videoPhysicalFile.metadata.audioBg) }
+  }
 
   return addVideo(req, res, { videoPhysicalFile, videoInfo, files })
```
kontrollanten, line 219: May it be more clear if the files property is only set when `videoPhysicalFile.metadata.audioBg` is truthy?

```
  return addVideo(req, res, {
    videoPhysicalFile,
    videoInfo,
    files: videoPhysicalFile.metadata.audioBg ? {
      audioBg: { path: getResumableUploadPath(videoPhysicalFile.metadata.audioBg) }
    } : null,
  })
```

rigelk: @kontrollanten I might be old school, but IMHO comment > code here.

rigelk: I ended up rewriting the initial request, so hopefully this should be more clear anyway 😉

Chocobozzz:
PR is ready, waiting for https://github.com/kukhariev/node-uploadx/pull/337

Chocobozzz:
Thanks @rigelk @kontrollanten and @kukhariev for your work and reactivity

PR #3942, opened 2021-04-08
Fix #3940: unload plugin modules on updates

### LINKED ISSUE #3940: Plugins modules are not correctly unloaded on uninstall (or updates)

While working on new plugins, I have sometimes weird behaviors that requires a peertube restart.
I thinks that it is because of these lines: https://github.com/Chocobozzz/PeerTube/blob/0564fc09d3ea46d78cc4e454f52aefeca68b73e2/server/lib/plugins/plugin-manager.ts#L413-L415

Plugins are unloaded using:
```javascript
delete require.cache[modulePath]
```
But if they are requires in the module main file, it will not be unloaded.

So, when you update your module code, it will mix old and new modules.

Plugin imported files should be unloaded too.
I think this module can help: https://www.npmjs.com/package/decache

**Additional information**

* PeerTube instance:
  * version: 3.1.0
  * NodeJS version: 12

JohnXLivingston:
## Description

Properly unload modules loaded by plugins when updating them.

## Related issues

This is a fix for https://github.com/Chocobozzz/PeerTube/issues/3940

## Has this been tested?


JohnXLivingston:
I replaced the call to ```delete require.cache[...]``` by a call to ```decache```. It is done in registerPlugin when upgrading a plugin.
Maybe we should also do it when unregistring plugins?

Chocobozzz:
Thanks, I'll merge this PR manually

JohnXLivingston:
> Thanks, I'll merge this PR manually

Can you wait a moment? I'm going to add decache calls in config reloading code (cf comment on the issue)

Chocobozzz:
I added it in https://github.com/Chocobozzz/PeerTube/commit/bae930372678692f94601d503eaf7deabaa82a3f

Thanks @JohnXLivingston for the PR and the tests!

PR #3945, opened 2021-04-10
add link to open video on origin instance

### LINKED ISSUE #3624: Open videos in original instance
When looking underneath a video you get the information for the original instance but if you click that link it just opens the homepage of the instance instead of the current video on that instance.
I would like to be able to get a direct link to the video on that instance. This could be useful when the video for whatever reason if not playing on the instance you are on but would work on the original instance.

rigelk:
## Description

Adds a "open video there" link next to the origin instance value. It is inconspicuous enough to not clutter the interface, yet communicates the action clearly in the context of the "origin instance" label and value, which is renamed as "origin" for this new dual usage.

## Related issues

closes #3624

## Has this been tested?

### DIFF
--- a/client/src/app/+videos/+video-watch/video-watch.component.html
+++ b/client/src/app/+videos/+video-watch/video-watch.component.html
@@ -230,8 +230,8 @@
         </div>
 
         <div *ngIf="video.isLocal === false" class="video-attribute">
-          <span i18n class="video-attribute-label">Origin instance</span>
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Please add the link in the main `a` element with the appropriate tooltip and without the icon. We can also put the complete link instead of the origin so it's easier to understand that we'll go on a remote instance.

rigelk:
You want me te add the link in the existing link? How so?

PR #3946, opened 2021-04-10
Add  `i18n` prop where it is missing

mkljczk:
## Has this been tested?


Chocobozzz:
Thanks @mkljczk

PR #3950, opened 2021-04-11
fix localization on page not found

### LINKED ISSUE #3700: Improve localization on page not found component
Have I understood the code correctly that "{{ getRessourceName() }}" in [client/src/app/+page-not-found/page-not-found.component.html](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/+page-not-found/page-not-found.component.html) will be replaced by the translation of either "video" or "resource" from [client/src/app/+page-not-found/page-not-found.component.ts](https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/%2Bpage-not-found/page-not-found.component.ts)?

Those words would be translated into Swedish as "video" and "resurs". The issue is that "the video" translates as "videon" and "the resourse" as "resursen". Note the different suffixes -n and -en. (And it could be even worse if the inserted strings were not from the same grammatical gender.)

On top of this, the string on line 15 begin with "The {{ getRessourceName() }}" and would need to have its first letter capitalized.

The usual workaround would be to add a general descriptor like "resursen" (meaning _the resource_) before the inserted string everywhere to side step this issue, but here that would result in "the resource resource".

If there is not a plan to have more options, I would suggest to hardcode the two possible values for all the strings where the function call is preceeded with "The" or "the".

rigelk:
## Description

Move video/resource switch to template, allowing sentence-wide translations that respect context and capitalization when needed.

## Related issues

fixes #3700

## Has this been tested?

### DIFF
--- a/client/src/app/+page-not-found/page-not-found.component.html
+++ b/client/src/app/+page-not-found/page-not-found.component.html
@@ -3,8 +3,9 @@
     <strong>{{ status }}.</strong>
     <span class="ml-1 text-muted" i18n>That's an error.</span>
 
-    <div class="text mt-4" i18n>
-      We couldn't find any {{ getRessourceName() }} tied to the URL {{ pathname }} you were looking for.
+    <div class="text mt-4">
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Thanks!

PR #3970, opened 2021-04-16
Add docker guide for running locally without SSL

### LINKED ISSUE #3957: Allows to use docker in a private network
**Parts of the documentation concerned:**

https://docs.joinpeertube.org/install-docker

**Problems**

Following the steps to setup and start a Docker Compose based instance of PeerTube, I ran into three errors:

_Let's Encrypt_

...did not like being on a private, unreachable server on a local network. I could work around this two different ways:

* Approach 1: Access the IP address and port of the PeerTube app directly, skipping the proxy, and using http (no ssl), potentially to include setting up a DNS entry in a local DNS server or in /etc/hosts to have a domain name matching the one I chose while following the initial instructions. This approach also requires removing the port 443 server section of the Nginx config so it doesn't crash due to missing certificate files (or removing Nginx altogether from the docker-compose)
* Approach 2: Omitting Cert Bot and Lets Encrypt and configuring the Nginx proxy to use my own private key and certificate chain

I tried Approach 1 above but not Approach 2.

_Cannot retrieve OAuth Client credentials: undefined_

This is a popup on the home page that appears as soon as the home page opens.

_Invalid client: client is invalid_

If you change the `.env` file to set `PEERTUBE_SIGNUP_ENABLED=true` and go through the `Create Login` forms, and then try to login with your correct credentials, the above error appears in red at the top, and you remain not logged in.

**Describe the desired information/correction**

* Clarifications/corrections to existing instructions on how to setup a local instance without getting the OAuth and login errors
* Instructions which describe how to setup a local instance without SSL. It would be acceptable that this local instance not be able to connect to other public instances

**Are you up to submit a pull request?**

Yes. I'd be willing to do "documentation testing" and submit a PR for a documentation change, once I have a working process for a local instance, or be involved in other ways that would help advance this.

**Other**

I'd also like to upvote #3754.

c0llab0rat0r:
## Description

Adds a second, simpler Docker guide for local-only use that omits SSL, and having `.env` and `docker-compose.yml` files which for most users will not need any modification at all and work "out of the box" excepting a DNS or hosts file change.

The existing guide is a bit overwhelming for a first attempt, due to the complexity of its closer-to-production approach.

This new guide will be more approachable for people new to PeerTube.

## Related issues

Resolves #3957.

## Has this been tested?

I tested it by following the steps in the guide after writing the guide.

## Screenshots

Not relevant; no changes to appearance.


Chocobozzz:
Hello,

I'd rather not maintain another docker-compose and docker guide. Could you try to merge the files instead?

c0llab0rat0r:
> Hello,
> 
> I'd rather not maintain another docker-compose and docker guide. Could you try to merge the files instead?

There's several scenarios where PeerTube could be used with docker-compose (or another orchestration), and each scenario requires a different combination of Docker images:

_Production + Let's Encrypt_
postfix + postgres + redis + peertube + nginx + certbot

_Production + Purchased Certificates_
postfix + postgres + redis + peertube + nginx
Omits certbot

_Local + Self-Signed Certificates_
postfix + postgres + redis + peertube + nginx
Omits nginx, certbot

_Local Without SSL_
postfix + postgres + redis + peertube
Omits nginx, certbot

Since docker-compose doesn't support conditionally starting or disabling containers, a separate docker-compose file is required. It's a function of how docker containers are designed to be composed together to create a larger unit.

Having the simpler, non-SSL example makes it easier for people new to PeerTube to understand what the minimal required configuration is for a functional instance, and to understand which configurations are add-ons.

c0llab0rat0r:
There's a progression of features.

While the compose files can't be merged, I think the guide could. The guide could be rearranged to start with the simple, local-only, non-SSL setup, and have a "Part 2" that adds certificates and SSL.

There's also some room for improvement in the existing "production" docker-compose file. There's some entries in it's related `.env` file that specify host names that are affixed (unchanging) in the docker-compose file, so those entries should be moved from the .env file (which has values that vary by deployment) and into the docker-compose file (which has values that don't vary less or not at all by deployment).

rigelk:
> While the compose files can't be merged, I think the guide could. The guide could be rearranged to start with the simple, local-only, non-SSL setup, and have a "Part 2" that adds certificates and SSL.

Could you make a PR with this more incremental change?

c0llab0rat0r:
> > While the compose files can't be merged, I think the guide could. The guide could be rearranged to start with the simple, local-only, non-SSL setup, and have a "Part 2" that adds certificates and SSL.
> 
> Could you make a PR with this more incremental change?

Updated.

c0llab0rat0r:
@rigelk,

What's your feedback on the revised guide?

fengwang:
This is very important for self-hosted servers that are only accessible from a LAN.

And, is port 9000 hard-coded in peertube? I have another container occupied this port, and I found changing it to another port will break the login oath.

Chocobozzz:
Hi and thanks for your work. Unfortunately we won't have time to maintain a local guide/docker-compose. We could add a section in the [unofficial install doc]https://docs.joinpeertube.org/install-unofficial) (https://framagit.org/framasoft/peertube/documentation/)

Issue #3897, opened 2021-03-29
Add "last video" option to playlist url


pprw:
A playlist url is a convenient way to share videos frequently updated.

However, it focuses the first (and older) video which induces a manual action to watch the last (newer) video uploaded.

You can point to a particular video (with `?playlistPosition=4` for example) but not to the last/newer uploaded video (afaik).

An option `?playlistPosition=last` would be nice.

hroussel:
I think we just need to add a route param in [my-video-playlist-elements.component.ts](https://app-int.quod.ai/github/Chocobozzz/PeerTube/answers/details?open_file=client%2Fsrc%2Fapp%2F%2Bmy-library%2Fmy-video-playlists%2Fmy-video-playlist-elements.component.ts&state=%7B%22client%2Fsrc%2Fapp%2F%2Bmy-library%2Fmy-video-playlists%2Fmy-video-playlist-elements.component.ts%22%3A%7B%22range%22%3A%7B%22lineFrom%22%3A59%2C%22lineTo%22%3A64%7D%2C%22activeQuestionId%22%3A4132%7D%7D). Is that right @Chocobozzz ?

Poslovitch:
I'm working on this.

PR #3974, opened 2021-04-17
Added "last" and a reversed ?playlistPosition in playlist URL

Poslovitch:
## Description

I implemented the `?playlistPosition=last` alongside a "new" feature: the ability to provide a negative playlistPosition, which acts like a "reversed" array index. `?playlistPosition=-1` selects the first video **from the bottom of the playlist**, `?playlistPosition=-2` the second, etc.

This was done as to avoid modifying the prototype of the `updatePlaylistIndex (position: number)` method. 

## Related issues

Resolves https://github.com/Chocobozzz/PeerTube/issues/3897.

## Has this been tested?

- [x] ð yes, light tests as follows are enough

1. Create a playlist and add a dozen videos to it.
2. Check that `?playlistPosition=last` works.
3. Check that `?playlistPosition=-1`, `?playlistPosition=-3`, etc. works.
4. Check that `?playlistPosition=1`, `?playlistPosition=3`, etc. works.
5. Check that `?playlistPosition=-100` is defaulted to 1.
6. Check that `?playlistPosition=saladedefruit` is defaulted to 1.

### DIFF
--- a/client/src/app/+videos/+video-watch/video-watch-playlist.component.ts
+++ b/client/src/app/+videos/+video-watch/video-watch-playlist.component.ts
@@ -117,6 +117,7 @@ export class VideoWatchPlaylistComponent {
 
   updatePlaylistIndex (position: number) {
     if (this.playlistElements.length === 0 || !position) return
+    if (position < 0) position = (this.playlistPagination.totalItems + 1) + position // Handle the "reverse" index
 
     for (const playlistElement of this.playlistElements) {
       // >= if the previous videos were not valid
--- a/client/src/app/+videos/+video-watch/video-watch-playlist.component.ts
+++ b/client/src/app/+videos/+video-watch/video-watch-playlist.component.ts
@@ -117,8 +117,11 @@ export class VideoWatchPlaylistComponent {
 
   updatePlaylistIndex (position: number) {
     if (this.playlistElements.length === 0 || !position) return
-    if (position < 0) position = (this.playlistPagination.totalItems + 1) + position // Handle the "reverse" index
 
+    // Handle the reverse index
+    if (position < 0) position = this.playlist.videosLength + position + 1;
+
+    console.log("updatePlaylistIndex: " + position)
     for (const playlistElement of this.playlistElements) {
       // >= if the previous videos were not valid
       if (playlistElement.video && playlistElement.position >= position) {
--- a/client/src/app/+videos/+video-watch/video-watch-playlist.component.ts
+++ b/client/src/app/+videos/+video-watch/video-watch-playlist.component.ts
@@ -121,7 +121,6 @@ export class VideoWatchPlaylistComponent {
     // Handle the reverse index
     if (position < 0) position = this.playlist.videosLength + position + 1;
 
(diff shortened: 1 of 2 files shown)


client/src/app/+videos/+video-watch/video-watch.component.ts
```diff
@@ -188,7 +188,14 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
     })
 
     this.queryParamsSub = this.route.queryParams.subscribe(queryParams => {
-      this.playlistPosition = queryParams[ 'playlistPosition' ]
+      // Handle the ?playlistPosition
+      let position = queryParams[ 'playlistPosition' ]
```
Chocobozzz, line 192:
```suggestion
      const positionParam = queryParams[ 'playlistPosition' ]
      let position: number
      if (positionParam === 'last) position = -1
      ...
```

client/src/app/+videos/+video-watch/video-watch.component.ts
```diff
@@ -188,7 +188,14 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
...
+      if (position === 'last') position = -1 // Handle the "last" index
+      if (typeof position === 'string') position = parseInt(position, 10) // Sanitization needed for the "reverse" index
+      if (isNaN(position)) position = 1
```
Chocobozzz, line 195: Add also a `console.error` because we should not have this case

client/src/app/+videos/+video-watch/video-watch-playlist.component.ts
```diff
@@ -117,6 +117,7 @@ export class VideoWatchPlaylistComponent {
 
   updatePlaylistIndex (position: number) {
     if (this.playlistElements.length === 0 || !position) return
+    if (position < 0) position = (this.playlistPagination.totalItems + 1) + position // Handle the "reverse" index
```
Chocobozzz, line 120: I don't think you need to calculate the position here since you already built it in the video watch component

client/src/app/+videos/+video-watch/video-watch.component.ts
```diff
@@ -188,7 +188,14 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
...
+
+      if (this.playlist && position < 0) this.playlistPosition = this.playlist.videosLength - position + 1
```
Chocobozzz, line 197: `this.playlist.videosLength + position + 1` since `position < 0` and try to use a ternary:

```
this.playlistPosition = this.playlist && position < 0
  ? ...
  : ...
```

Chocobozzz (changes requested):

client/src/app/+videos/+video-watch/video-watch-playlist.component.ts
Poslovitch: Actually, the position is not properly calculated in the video watch component, because this.playlist is undefined at that time. In the commit I just pushed, I got rid of it. In the video watch component, `this.playlistPosition` can be negative, while `updatePlaylistIndex()` does the job of translating the actually negative position into an index it can use.

Issue #3612, opened 2021-01-17
Configurable minimum sign-up age


JasXSL:
Currently when you sign up, you have to agree that you're at least 16 years old. In order to change this, you'll have to edit all the translation files in the deployment.

It would be nice if we could change this value in the settings, since we're working on a shard where you need to be 18+ to join.

I know you can alter it with the custom JS injections, but it feels like a hacky workaround.

rigelk:
Please don't tag your issue title yourself, and use the issue template.

Poslovitch:
I'm working on this.

cdauth:
It's nice that the age can be configured now. However, the minimum age that can be configured now is 1, and I would like to disable any minimum age. This could probably be achieved by allowing setting the minimum age to 0 and editing the checkbox message so that it doesn't mention any minimum age if it is 0.

PR #3984, opened 2021-04-18
rephrase signup limit validator error

Poslovitch:
## Description

While working on #3612, I noticed there was a typo in the 'min' error message of the SIGNUP_LIMIT_VALIDATOR.

## Related issues

None.

## Has this been tested?

None.


rigelk:
hi @Poslovitch,

The contrived logic behind this voluntary typo lies in the use of `-1` as a special value disabling the signup limit. I would suggest leaving the message as is, and adding a comment along the code explaining the situation.

Poslovitch:
> The contrived logic behind this voluntary typo lies in the use of -1 as a special value disabling the signup limit. I would suggest leaving the message as is, and adding a comment along the code explaining the situation.

Alright. Then I think there should also be a "text" under the field in the form that explains that. When I did some light tests, I messed around with the config editor and figured out that -1=unlimited only by trying that value. It's not explained in the UI at all.

rigelk:
@Poslovitch instead of adding text to the UI, we could just add that information (regarding the special value) to the error message of the validator.

Issue #212, opened 2018-01-17
Live stream of comment, mastodon-like?


ldidry:
I wrote a comment on mastodon on a video, but I had to refresh the video page to see it. It could be cool if comments shows as thex arrived. Maybe with a websocket channel?

test2a:
This can be a good tool for moderation. Wonder why no work has been done on this for so long....

rigelk:
@test2a it's known to be one of the most ressource-intensive parts of Mastodon

test2a:
@rigelk interesting. Scarce resources need to be managed judiciously. Still, this could be a nice addition if we have enough computing resources. Is the tech not developed or can.be made available but not enabled due to said scarcity ?

PR #4000, opened 2021-04-22
Auto load new comments

kontrollanten:
## Description
When watching a video new comments will automatically be loaded with a notification above the comments field.

Possible improvements:
- [ ] Don't load when video is in fullscreen.
- [ ] Don't load when page isn't visible.
- [ ] Make the update interval configurable and possible to disable in the config file.
- [ ] Notify new comments via web socket instead.

## Related issues
#212

## Has this been tested?

## Screenshots
image

### DIFF
--- a/client/src/app/+videos/+video-watch/comment/video-comments.component.html
+++ b/client/src/app/+videos/+video-watch/comment/video-comments.component.html
@@ -25,6 +25,11 @@
       [textValue]="commentThreadRedraftValue"
     ></my-video-comment-add>
 
+    <div *ngIf="newComments.length > 0" class="alert alert-primary">
+      <a class="alert-link" (click)="onClickShowNewComments($event)" href="">
+        {newComments.length, plural, =0 {Comments} =1 {1 new comment} other {{{newComments.length}} new comments}}
+      </a>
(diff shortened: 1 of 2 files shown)


kontrollanten:
@Chocobozzz and @rigelk I'd like to here your thoughts before I go further with this. Is this something that can be implemented? We can set this to disabled by default, so it won't be resource intensive for those who doesn't choose to opt in.

rigelk:
Before thinking of auto-loading new comments, an intermediary step should be to have a button to trigger checks manually. Then, IMHO your implementation _should_ be disabled by default (but then we need an UX scenario to activate it).

Either way, we could further save requests via an exponential backoff strategy.

Chocobozzz:
Hello,

I think it can be implemented using socket io instead of a periodic refresh.

kontrollanten:
> Hello,
> 
> I think it can be implemented using socket io instead of a periodic refresh.

Yes, that's probably a more efficient way of solving it. But what about use this as an intermediate step?

Chocobozzz:
> But what about use this as an intermediate step?

I'd rather not have to maintain intermediate steps :)

PR #4002, opened 2021-04-22
Use account initial as default avatar

### LINKED ISSUE #3935: User initials as default avatar
The majority of our users doesn't have any custom avatar. This means that there's a lot of comments with the default gray avatar, which looks a bit "boring".

The default avatar can be a colored circle with the user initial within, instead of the grey smiley.

kontrollanten:
## Description
Change default avatar to account name initial.

## Related issues
closes #3935

## Has this been tested?

1) Check avatar on account page.
1) Check avatar on channel page.
1) Check avatar in notifications.
1) Check avatar in comments.
1) Check avatar in comment add component.
1) Check avatar in video watch page.
1) Check avatar in my subscriptions.

## Screenshots
image

### DIFF
--- a/client/src/app/shared/shared-account-avatar/account-avatar.component.html
+++ b/client/src/app/shared/shared-account-avatar/account-avatar.component.html
@@ -1,5 +1,8 @@
 <ng-template #img>
-  <img [class]="class" [src]="avatarUrl" i18n-alt alt="Account avatar" />
+  <img *ngIf="avatarUrl || !initial" [class]="class" [src]="avatarUrl || defaultAvatarUrl" i18n-alt alt="Account avatar" />
+  <div *ngIf="!avatarUrl && initial" [class]="class">
+    <span>{{ initial }}</span>
+  </div>
 </ng-template>
 
 <a *ngIf="account && href" [href]="href" target="_blank" rel="noopener noreferrer" [title]="title">
--- a/client/src/app/shared/shared-account-avatar/account-avatar.component.ts
+++ b/client/src/app/shared/shared-account-avatar/account-avatar.component.ts
@@ -12,7 +12,7 @@ export class AccountAvatarComponent {
     avatar?: { url?: string, path: string }
     url: string
   }
-  @Input() size: '25' | '34' | '36' | '40' | '120' = '36'
+  @Input() size: '25' | '32' | '34' | '36' | '40' | '120' = '36'
 
   // Use an external link
   @Input() href: string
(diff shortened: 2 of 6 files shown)


kontrollanten:
The failed test seems unrelated.

client/src/app/shared/shared-account-avatar/account-avatar.component.ts
```diff
@@ -23,17 +23,40 @@ export class AccountAvatarComponent {
...
+      mno: 'yellow',
+      pqr: 'orange',
+      stv: 'red',
+      wxyz: 'dark-blue'
+    }
+
+    const theme = Object.keys(themes).find(chars => chars.includes(this.initial))
+
+    return themes[theme]
+  }
```
rigelk, line 61: Please use a simple hash-based colour generation function, which allows more diversity even for two names beginning by the same letter.
i.e. https://stackoverflow.com/questions/3426404/create-a-hexadecimal-colour-based-on-a-string-with-javascript

rigelk (changes requested):
LGTM, but I would opt for a different colour generation method based on the name.

client/src/app/shared/shared-account-avatar/account-avatar.component.ts
kontrollanten: Looks nice, but how do we ensure the contrast between text and bg is good? And can it be an ugly UI if we just generate random colors?

rigelk: You can generate colors in a restricted color space.

kontrollanten: I'm sorry but I don't understand how this should be implemented. Shall we define a set of color spaces and match those with a text color? It sounds tricky to me. And by using color spaces I'm not confident with that the contrast between bg and text is good enough, but maybe there's an easy way to ensure that?

client/src/app/shared/shared-account-avatar/account-avatar.component.html
```diff
@@ -1,5 +1,8 @@
 <ng-template #img>
-  <img [class]="class" [src]="avatarUrl" i18n-alt alt="Account avatar" />
+  <img *ngIf="avatarUrl || !initial" [class]="class" [src]="avatarUrl || defaultAvatarUrl" i18n-alt alt="Account avatar" />
```
Chocobozzz, line 2: Why would we still need the default avatar url?

kontrollanten: I saw in some places where the avatar was used that it was conditional checks if the `account` was set, so I assumed there's cases when that's applicable. For example here https://github.com/Chocobozzz/PeerTube/blob/e771e82dfac5f8cd5462fecf6618f9b1fb314e30/client/src/app/%2Bvideos/%2Bvideo-watch/comment/video-comment.component.ts#L108

Can it be if a an user/account has been deleted but the comments still exists?

PR #4010, opened 2021-04-24
Implemented configurable minimum signup age

Poslovitch:
## Description

I added a config option to set the minimum age required in order to signup to the instance. It defaults to 16 (which has been the "default" age on PeerTube up to now).
I therefore modified the RegisterComponent so that it can display this age.

## Related issues

Resolves https://github.com/Chocobozzz/PeerTube/issues/3612.

## Has this been tested?

I believe light tests were enough, but should tests be added to the test suite?
If so:
* On which test suit?
* Which tests scenarios?

## Screenshots

In the sign up page:
image

In the admin "Basic configuration" editor :
image

### DIFF
--- a/client/src/app/+admin/config/edit-custom-config/edit-basic-configuration.component.html
+++ b/client/src/app/+admin/config/edit-custom-config/edit-basic-configuration.component.html
@@ -167,6 +167,20 @@
 
                 <small i18n *ngIf="hasUnlimitedSignup()" class="text-muted">Signup won't be limited to a fixed number of users.</small>
               </div>
+              
+              <div [ngClass]="getDisabledSignupClass()" class="mt-3">
+                <label i18n for="signupMinimumAge">Minimum required age to create an account</label>
+
+                <div class="number-with-unit">
+                    <input
+                      type="number" min="1" id="signupMinimumAge" class="form-control"
+                      formControlName="minimumAge" [ngClass]="{ 'input-error': formErrors['signup.minimumAge'] }"
+                    >
+                    <span i18n>{form.value['signup']['minimumAge'], plural, =1 {year old} other {years old}}</span>
+                </div>
+
+                <div *ngIf="formErrors.signup.minimumAge" class="form-error">{{ formErrors.signup.minimumAge }}</div>
+              </div>
             </ng-container>
           </my-peertube-checkbox>
         </div>
@@ -478,7 +492,7 @@
         <ng-container formGroupName="twitter">
 
           <div class="form-group">
-            <label i18n for="signupLimit">Your Twitter username</label>
+            <label i18n>Your Twitter username</label>
 
             <input
               type="text" id="servicesTwitterUsername" class="form-control"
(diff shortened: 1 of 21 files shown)


client/src/app/+admin/config/edit-custom-config/edit-basic-configuration.component.html
```diff
@@ -478,7 +492,7 @@
         <ng-container formGroupName="twitter">
 
           <div class="form-group">
-            <label i18n for="signupLimit">Your Twitter username</label>
```
Chocobozzz, line 481: Oup's, thanks. It should be `for="servicesTwitterUsername"` instead

client/src/app/+signup/+register/register-step-terms.component.html
```diff
@@ -2,8 +2,8 @@
   <div class="form-group form-group-terms">
     <my-peertube-checkbox inputName="terms" formControlName="terms">
       <ng-template ptTemplate="label">
-        <ng-container i18n>
-          I am at least 16 years old and agree
+        <ng-container i18n> 
+          I am at least {{ minimumAge }} years old and agree
```
Chocobozzz, line 6: Please remove added spaces at the end of these lines

client/src/app/+signup/+register/register.component.html
```diff
@@ -17,6 +17,7 @@
 
           <my-register-step-terms
             [hasCodeOfConduct]="!!aboutHtml.codeOfConduct"
+            [minimumAge]="this.minimumAge"
```
Chocobozzz, line 20:
```suggestion
            [minimumAge]="minimumAge"
```

config/default.yaml
```diff
@@ -229,6 +229,7 @@ contact_form:
 signup:
   enabled: false
   limit: 10 # When the limit is reached, registrations are disabled. -1 == unlimited
+  minimum_age: 16
```
Chocobozzz, line 232: Please add a comment to explain this is used in the signup form

shared/extra-utils/server/config.ts
```diff
@@ -98,7 +98,8 @@ function updateCustomSubConfig (url: string, token: string, newConfig: DeepParti
     signup: {
       enabled: false,
       limit: 5,
-      requiresEmailVerification: false
+      requiresEmailVerification: false,
+      minimumAge: 16
```
Chocobozzz, line 102: Please use another value and check it is correctly set with the default in https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/server/config.ts#L63 and correctly updated in https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/server/config.ts#L153

Chocobozzz (changes requested):
Thanks!

client/src/app/+signup/+register/register-step-terms.component.html
rigelk:
```suggestion
          I am at least {{ minimumAge }} years old and agree
```

config/default.yaml
rigelk:
```suggestion
  minimum_age: 16 # Used to configure the signup form
```

Poslovitch:
Quick heads-up : I'll provide the requested changes this week.

Chocobozzz:
Merged manually in https://github.com/Chocobozzz/PeerTube/commit/1f256e7d3cf056c2d999260155cdba58ae1b878b

Issue #3725, opened 2021-02-09
Answering to comments via Mastodon / Pleroma / another PeerTube instance…


fflorent:
**Describe the current behavior**

When logged out, one can post a comment by entering our own Activity Pub identifier through the "remote interact" feature:

![Remote interact](https://user-images.githubusercontent.com/371705/107408674-92677c80-6b0b-11eb-9166-7a8dd496c404.png)

However we miss the same thing for answering to comments: 

- While being logged in:
![While being connected](https://user-images.githubusercontent.com/371705/107408827-c2af1b00-6b0b-11eb-8f7c-9143cf03294e.png)

- Otherwise:
![image](https://user-images.githubusercontent.com/371705/107408947-e96d5180-6b0b-11eb-9ab4-532a40326a13.png)

**Describe the expected behavior**

We should be able to click the answer button and get the same popup for remote interact in order to be consistent (at least if possible).

**Additional information**

* PeerTube instance:
  * URL: https://skeptikon.fr
  * version: 3.0.1

* Browser name, version and platforms on which you could reproduce the bug: Firefox Nightly 87.0a1 (2021-02-07) (64 bits)
* Link to browser console log if relevant: N/A
* Link to server log if relevant (`journalctl` or `/var/www/peertube/storage/logs/`): N/A

---

I filled it as a bug as I see that as an UX inconsistency. You may reclassify it as a feature if you prefer.
Thanks for your work :pray:

PR #4012, opened 2021-04-24
Added ability to reply to comments with remote interaction

Poslovitch:
## Description

I got rid of the conditions that required a user to be logged in in order to reply to comments.

## Related issues

Resolves https://github.com/Chocobozzz/PeerTube/issues/3725

## Has this been tested?

While non-logged-in users now have access to the reply button (see screenshot), I could not test if the reply worked properly (I expect it should, though) because my pleroma instance wouldn't let me do a remote interaction (no clue why that is).

## Screenshots

image


Chocobozzz:
Thanks @Poslovitch

PR #4048, opened 2021-05-06
throw on failed parameter checks instead of pure boolean checks

### LINKED ISSUE #4026: API does not yield detailed error messages
posting content via API the validation rules are not document and the errors often hard to understand without any indication what the lmit is. i.e.:

- 120 character limit for name
- limit of 5 tags, limit of 30 character tag limit.

would be great if those limits would be documented in the video/put call at:
https://docs.joinpeertube.org/api-rest-reference.html#tag/Video/paths/~1videos~1{id}/put
in addition the error should indicate what the limit is, right now it just responds with:

{'errors': {'name': {'value': '...', 'msg': 'Should have a valid name', 'param': 'name', 'location': 'body'}}}

rigelk:
## Description

The goal is to have more descriptive error messages on malformed parameters, by replacing all-encompassing parameter checks that just produce a boolean, by an error thrown as close as possible to the failing check, where each check has an associated message:

```diff
{
  "errors": {
    "name": {
      "value": "<video name>",
-     "msg": "Should have a valid name",
+     "msg": "Should have a video name between 3 and 120 characters long",
      "param": "name",
      "location": "body"
    }
  }
}
```

I didn't change all checks, but focused on parameters which error messages might need reading the documentation in detail.

### Implementation details

Error being thrown in validation middlewares are okay so long as they are used within a [validator function](https://express-validator.github.io/docs/validation-chain-api.html#customvalidator). However some validation middlewares are also currently used as pure boolean checks, i.e. within another validation middleware or within a full-request validation function. The error there needs to be catched, for which we transform the potentially-returning function by wrapping it with `EtoB`:

```typescript
/**
 * Handle errors thrown by non-pure boolean validators as boolean
 */
function EtoB (fun: Function) (...args: any[]) => boolean | Error {
  return (...args) => {
    try {
      fun(...args)
    } catch (error) {
      return error
    }
    return true
  }
}
```

We also need to distinguish functions that might throw an `Error` from that which only return booleans. Sadly, [there is still no way to express that yet in Typescript itself](https://github.com/microsoft/TypeScript/issues/13219) ; so I marked such functions with JSDoc comments for now, i.e.:

```typescript
/**
 * @throws {Error}
 */
function isDateValid (value: string) {
  if (!exists(value)) throw new Error('Should have a date')
  if (!validator.isISO8601(value)) throw new Error('Should have a date conforming to ISO8601')
  return true
}
```

Within a pure-boolean check, this translates to `if (!EtoB(isDateValid)(value)) { …`. If you want to programmatically get the error message, you can switch the return mode by using `const error = EtoB(isDateValid, true)(value)`.

## Related issues

fixes #4026

## Has this been tested?

### DIFF
--- a/scripts/create-import-video-file-job.ts
+++ b/scripts/create-import-video-file-job.ts
@@ -6,7 +6,7 @@ import { resolve } from 'path'
 import { VideoModel } from '../server/models/video/video'
 import { initDatabaseModels } from '../server/initializers/database'
 import { JobQueue } from '../server/lib/job-queue'
-import { isUUIDValid } from '@server/helpers/custom-validators/misc'
+import { catchErrorAsBoolean, checkUUID } from '@server/helpers/custom-validators/misc'
 
 program
   .option('-v, --video [videoUUID]', 'Video UUID')
@@ -31,7 +31,7 @@ run()
 async function run () {
   await initDatabaseModels(true)
 
-  if (isUUIDValid(options.video) === false) {
+  if (catchErrorAsBoolean(checkUUID)(options.video) === false) {
     console.error('%s is not a valid video UUID.', options.video)
     return
   }
--- a/scripts/create-transcoding-job.ts
+++ b/scripts/create-transcoding-job.ts
@@ -8,7 +8,7 @@ import { JobQueue } from '../server/lib/job-queue'
 import { computeResolutionsToTranscode } from '@server/helpers/ffprobe-utils'
 import { VideoTranscodingPayload } from '@shared/models'
 import { CONFIG } from '@server/initializers/config'
-import { isUUIDValid } from '@server/helpers/custom-validators/misc'
+import { catchErrorAsBoolean, checkUUID } from '@server/helpers/custom-validators/misc'
 
 program
   .option('-v, --video [videoUUID]', 'Video UUID')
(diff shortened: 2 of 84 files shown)


Chocobozzz:
A function named `isDateValid` must not throw. It must return a boolean as started in its name. A `checkDate` would be a more appropriate name.

Mixing validators that return a boolean, with validators that throw an error, with validators that return an array `[ boolean, string ]` is really not convenient. We should stick to one method that fits all our use cases (middlewares validators, database validator and AP checks)

`EtoB` is not a good function name and should not start with a capital letter.

rigelk:
@Chocobozzz should I put the renamed functions in a dedicated file to separate them from boolean validators? Something like `videos.throwable.ts` for a `videos.ts` validator sister file?

Chocobozzz:
> should I put the renamed functions in a dedicated file to separate them from boolean validators? Something like videos.throwable.ts for a videos.ts validator sister file?

I don't think so because of my second point:

> Mixing validators that return a boolean, with validators that throw an error, with validators that return an array [ boolean, string ] is really not convenient. We should stick to one method that fits all our use cases (middlewares validators, database validator and AP checks)

I think we should migrate all validator to a throw workflow

rigelk:
Closing since the PR adds complexity when messages are arguably better improved on their own.

PR #4055, opened 2021-05-07
Fix CLI plugins list command: use command parameters.

JohnXLivingston:
## Description

The `plugins list` command was not using --url, --username and --password parameters, and was always requesting on the first server in the .netrc file.

## Has this been tested?


Chocobozzz:
Thanks!

PR #4072, opened 2021-05-10
Fix type.

JohnXLivingston:
## Description

The new isSettingHidden option was wrongly typed (it is a frontend function, but was using backend type).

## Has this been tested?

I ran «npm run build» with success.

### DIFF
--- a/shared/models/plugins/register-client-settings-script.model.ts
+++ b/shared/models/plugins/register-client-settings-script.model.ts
@@ -1,8 +1,8 @@
-import { RegisterServerSettingOptions } from "./register-server-setting.model"
+import { RegisterClientFormFieldOptions } from "./register-client-form-field.model"
 
 export interface RegisterClientSettingsScript {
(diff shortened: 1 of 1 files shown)


rigelk:
@JohnXLivingston please use descriptive titles.

Chocobozzz:
Hello,

This is the correct type. `RegisterServerSettingOptions` is a shared type, because the server sends it in thought the API so the client also uses it. `RegisterServerSettingOptions` extends `RegisterClientFormFieldOptions` adding `private` property.

JohnXLivingston:
> @JohnXLivingston please use descriptive titles.
Sorry, will do next time.




> Hello,
> 
> This is the correct type. `RegisterServerSettingOptions` is a shared type, because the server sends it in thought the API so the client also uses it. `RegisterServerSettingOptions` extends `RegisterClientFormFieldOptions` adding `private` property.

Ok, I see.

PR #4105, opened 2021-05-18
Show logo in mobile view

kontrollanten:
## Description
Changes the mobile CSS so that only the instance name is hidden in mobile view. Since there's a lot of 

## Has this been tested?

1) Open the site with a screen res with max-width < 500 px.

## Screenshots

image


Chocobozzz:
I think the hamburger icon and the logo are too close for a touch screen

Chocobozzz:
Seems the PR is broken

PR #4114, opened 2021-05-20
Allow SSL database parameter

stakovicz:
Allow SSL database parameter via environment variable

## Has this been tested?


support/docker/production/config/custom-environment-variables.yaml
```diff
@@ -19,6 +19,7 @@ database:
   suffix: "PEERTUBE_DB_SUFFIX"
   username: "PEERTUBE_DB_USERNAME"
   password: "PEERTUBE_DB_PASSWORD"
+  ssl: "PEERTUBE_DB_SSL"
```
Chocobozzz, line 22:
```suggestion
  ssl:
  __name: "PEERTUBE_DB_SSL"
  __format: "json"
```

So it's interpreted as a boolean

Issue #3830, opened 2021-03-05
Send notifications via Push and Notification API


kontrollanten:
**Describe the problem to be solved**
Add support for using [Push API](https://developer.mozilla.org/en-US/docs/Web/API/Push_API) and [Notification API](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API) for notifications. This will make it easier to notify users about an upcoming live stream and also help keep comment discussions active.

**Describe the solution you would like:**
As a user I'd like to be able to get all notifications pushed with Notification API.

**Describe alternatives you have considered**  
Maybe this can be done via a plugin. Then we would need some kind of `action:notification-created` hook and also som ability inject code running in the service worker.

A basic implementation guide: https://flaviocopes.com/push-api/#overview

kontrollanten:
It seems that it's not much needed to be able to implement this via a plugin. When a ServiceWorker is registered it's registered within a scope. By default the scope is the same path as the service worker file is within (/plugins/[pluginName]/[version]/static/serviceWorker.js in this case). This will lead to a problem upon updates, since the ServiceWorker scope changes for each version. I see two different solutions:

A) Add a plugin path without version; /plugins/[pluginName]/static/serviceWorker.js
B) Add [Service-Worker-Allowed header](https://medium.com/dev-channel/two-http-headers-related-to-service-workers-you-never-may-have-heard-of-c8862f76cc60) to the ServiceWorker asset. This mean that either this header should be added to all plugins JS files or it should be optionally added in the plugin `package.json`.

Any opinions?

kontrollanten:
This can be done with current PeerTube API; https://framagit.org/kontrollanten/peertube-plugin-one-signal

PR #4126, opened 2021-05-28
add client hook filter:videojs.options

### LINKED ISSUE #4086: Plugin filter to alter videojs options
We'd like to use videojs-chromecast plugin to add Chromecast support.

Relates to #2430 

We've created a proof of concept here https://github.com/andreasgan/PeerTube/pull/1/files

I think this can be solved with a PeerTube plugin with the following API.

```
  registerHook({
    target: 'filter:videojs.options',
    handler: ({ options }) => ({
      ...options,
      techOrder: [ 'chromecast', 'html5' ],
      chromecast: {
        modifyLoadRequestFn: function (loadRequest: any) {
          loadRequest.media.hlsSegmentFormat = 'fmp4';
          loadRequest.media.hlsVideoSegmentFormat = 'fmp4';
          return loadRequest;
        }
      },
    }),
  })

```

kontrollanten:
## Description
Add possibility to alter videojs options with a hook.

## Related issues
closes #4086

## Has this been tested?

1) Install `peertube-plugin-tv-streaming`.
1) Watch a HLS video.
1) Connect to Chromecast by clicking on icon in the right corner of video.


client/src/assets/player/peertube-player-manager.ts
```diff
@@ -144,7 +145,7 @@ export class PeertubePlayerManager {
       ])
     }
 
-    const videojsOptions = this.getVideojsOptions(mode, options, p2pMediaLoader)
+    const videojsOptions = await this.getVideojsOptions(mode, options, p2pMediaLoader)
```
Chocobozzz, line 148: Instead of changing `client/src/root-helpers/plugins.ts` file, just add a `videojsOptionsFilter` parameter in `PeertubePlayerManagerOptions`.

embed and watch page will create a filter function that send the object to plugins

kontrollanten: Isn't it more simple to have the filter function on one place instead of duplicating the code?

Chocobozzz: Yes you're right but I'd prefer to not couple the peertube plugin system to the player manager (if we can avoid it)

kontrollanten: I see. But maybe it opens up for more possibilities to make the plugin system agnostic, I'm thinking about running hooks/filter in ServiceWorker (caching, push notifications), etc,

kontrollanten: This should open up for #3830 by allowing a hook to be called in the ServiceWorker.

kontrollanten: What do you think?

Chocobozzz: this is a good idea, I'll update and merge the PR

Chocobozzz:
I renamed the hook to `filter:internal.player.videojs.options.result`

PR #4143, opened 2021-06-01
refactor API errors to standard error format

### LINKED ISSUE #4088: Error responses as HTTP Problem details/RFC7807

Currently we have 3+ ways of responding to errors within PeerTube:

Basic errors:
```http
HTTP/1.1 404 Not Found
Content-Type: application/json

{
  "error": "Account not found"
}
```

Basic errors with a code (a PeerTube-specific server error code, here corresponding to DOES_NOT_RESPECT_FOLLOW_CONSTRAINTS):
```http
HTTP/1.1 403 Forbidden
Content-Type: application/json

{
  "errorCode": 1,
  "error": "Could not follow server"
}
```

Validation errors:
```http
HTTP/1.1 400 Bad Request
Content-Type: application/json

{
  "errors": {
    "id": {
      "value": "a117eb-c6a9-4756-bb09-2a956239f",
      "msg": "Should have an valid id",
      "param": "id",
      "location": "params"
    }
  }
}
```

While _the current situation is fine_, I think at least __producing the JSON response of errors via a unique factory__ would allow mapping them/processing them as needed, and achieve a bit of yak shaving: translatable error messages, ubiquitous error codes, etc.

I think we can take inspiration from the HTTP Problem details specification in [RFC7807](https://tools.ietf.org/html/rfc7807). It has seen a few implementations that provide examples of how it can be handled:
- https://www.npmjs.com/package/express-http-problem-details
- https://github.com/kamranahmedse/express-api-problem
- https://www.npmjs.com/package/problem-json

They generally provide different factories/handlers that produce such responses:

```http
HTTP/1.1 403 Forbidden
Content-Type: application/problem+json
Content-Language: en

{
  "title": "Could not get video",
  "detail": "Could not get the video since it originates from an actor not followed by your server, which doesn't allow fetching unrelated actors' videos",
  "instance": "/api/v1/videos/9c9de5e8-0a1e-484a-b099-e80766180a6d",
  "type": "https://docs.joinpeertube.org/errors/DOES_NOT_RESPECT_FOLLOW_CONSTRAINTS",
  "docs": "https://docs.joinpeertube.org/api-rest-reference.html#tag/Video/paths/~1videos~1{id}/get"
}
```

We could further document some errors with a `type` parameter that links to a given error code. As for validation errors, we could provide them as follows:

```http
HTTP/1.1 400 Bad Request
Content-Type: application/problem+json
Content-Language: en

{
  "title": "Invalid parameter",
  "instance": "/api/v1/videos/a117eb-c6a9-4756-bb09-2a956239f",
  "invalid-params": {
    "id": {
      "value": "a117eb-c6a9-4756-bb09-2a956239f",
      "msg": "Should have a valid id",
      "param": "id",
      "location": "params"
    }
  },
  "type": "about:blank",
  "docs": "https://docs.joinpeertube.org/api-rest-reference.html#tag/Video/paths/~1videos~1{id}/get"
}
```

Having a single factory for these errors would also allow producing an associated log entry, with an request uid that we could provide in the response for easier debugging/finding in the logs.

rigelk:
## Description

This PR introduces a factory for errors, under the form of an express helper, to format error messages following a more standard error format.

The goal is to unify response formats as described in #4088, especially for client-facing APIs (I left ActivityPub endpoints as they were). In that regard, I removed any use of `sendStatus` since if it sends a needed message, it should go through the factory. If not, then better use an empty body.

## Related issues

fixes #4088

## Has this been tested?


server/helpers/express-utils.ts
```diff
@@ -125,6 +126,34 @@ function getCountVideos (req: express.Request) {
   return req.query.skipCount !== true
 }
 
+// helpers added in server.ts and used in subsequent controllers used
+const apiResponseHelpers = (req, res: express.Response, next = null) => {
+  res.fail = (options) => {
```
Chocobozzz, line 131: Please type `options`

server/helpers/express-utils.ts
```diff
@@ -125,6 +126,34 @@ function getCountVideos (req: express.Request) {
...
+    res.json(new ProblemDocument({
+      status,
+      title,
+      instance,
+      type: type && '' + type,
+      detail: message
+    }, extension))
+  }
+
+  if (next !== null) next()
```
Chocobozzz, line 154: `if (next) next()`

server/middlewares/validators/activitypub/activity.ts
```diff
@@ -10,7 +10,7 @@ async function activityPubValidator (req: express.Request, res: express.Response
   if (!isRootActivityValid(req.body)) {
     logger.warn('Incorrect activity parameters.', { activity: req.body })
     return res.status(HttpStatusCode.BAD_REQUEST_400)
-              .json({ error: 'Incorrect activity.' })
```
Chocobozzz, line 13: Why did you remove the message?

server/helpers/express-utils.ts
```diff
@@ -125,6 +126,34 @@ function getCountVideos (req: express.Request) {
   return req.query.skipCount !== true
 }
 
+// helpers added in server.ts and used in subsequent controllers used
+const apiResponseHelpers = (req, res: express.Response, next = null) => {
```
Chocobozzz, line 130: I don't think it should be a middleware. Why don't you create an helper function `sendError (res: express.Response)` instead?

server/helpers/express-utils.ts
```diff
@@ -125,6 +126,34 @@ function getCountVideos (req: express.Request) {
...
+  res.fail = (options) => {
+    const { data, status, message, title, type, docs, instance } = {
+      data: null,
```
Chocobozzz, line 133: I don't think you need this + check https://zaiste.net/posts/javascript-destructuring-assignment-default-values/ for default values

server/helpers/express-utils.ts
```diff
@@ -125,6 +126,34 @@ function getCountVideos (req: express.Request) {
...
+      ...options,
+      status: options.status || HttpStatusCode.BAD_REQUEST_400
+    }
+
+    const extension = new ProblemDocumentExtension({
+      ...data,
+      docs: docs || res.docs
```
Chocobozzz, line 140: I don't really like to put the API doc URL in `res.docs`. I think your can build a function that automatically assign the doc depending on `req.originalUrl` and `req.method` if `docs` is not specified in `options`.

server/helpers/express-utils.ts
```diff
@@ -125,6 +126,34 @@ function getCountVideos (req: express.Request) {
...
+    })
+
+    res.status(status)
+    res.setHeader('Content-Type', 'application/problem+json')
+    res.json(new ProblemDocument({
+      status,
+      title,
+      instance,
+      type: type && '' + type,
```
Chocobozzz, line 149: `type` should be an URI

server/helpers/express-utils.ts
```diff
@@ -125,6 +126,34 @@ function getCountVideos (req: express.Request) {
...
+      detail: message
+    }, extension))
```
Chocobozzz, line 151: Please keep `error` and `code` fields for compatibility

Chocobozzz (changes requested):
Please keep the `expect(res.body.code)` and `expect(res.body.error)` tests for compatibility + add a few tests regarding your new error implementation

server/helpers/express-utils.ts
rigelk: > I think your can build a function that automatically assign the doc depending on `req.originalUrl` and `req.method` if `docs` is not specified in `options`.

`req` is often not passed down to where the response is put to fail.

server/helpers/express-utils.ts
rigelk: Why? It seems more logical to replace response middlewares with a response middleware.

server/helpers/express-utils.ts
rigelk: Options are already typed through https://github.com/rigelk/PeerTube/blob/3bd73a228179fe99b12a3f917b32c741764d485c/server/typings/express/index.d.ts#L90-L94

Issue #4071, opened 2021-05-10
404 for remote avatar


johanfleury:
I suffered a data loss recently and restored PeerTube with an empty storage. Since then, some avatars are not showing and I couldn’t find any way to make PeerTube reload them (the files are not present in the data directory).

I had a similar issue with thumbnails, but unsubscribing from a channel then subscribing again fixed that (I had no luck trying to restore avatars with this technique).

Here is a screenshot that shows the issue:

![Missing avatars on a user and their channel](https://user-images.githubusercontent.com/7230353/117690938-6189f880-b189-11eb-866c-d3fd90fde4cb.png)

The avatar URL returns a 404 status code:

```
$ curl -I -X GET https://peertube.br0.fr/lazy-static/avatars/2b8ff6f3-b4fd-45d0-9238-4dc28148c292.png
HTTP/2 404
date: Mon, 10 May 2021 16:16:27 GMT
content-length: 0
tk: N
access-control-allow-origin: *
strict-transport-security: max-age=15724800; includeSubDomains
permissions-policy: interest-cohort=()
```

And the avatar is not there:

```
$ ll avatars/2b8ff6f3-b4fd-45d0-9238-4dc28148c292.png
ls: cannot access 'avatars/2b8ff6f3-b4fd-45d0-9238-4dc28148c292.png': No such file or directory
```

The question is, is there a way to make PeerTube download a channel/user avatar when it’s not there?

If not, what would be the best solution? I’m thinking of two ways of doing this:

* Detect when a remote user/channel avatar is missing on disk and fetch it then (for instance when PeerTube returns a 404 error).
* Have a background job that checks avatars and fetch them when they are missing.

I could probably work on a PR if you want to.

AnonymousWebHacker:
I think, you can regenerate thist avatar. See https://docs.joinpeertube.org/maintain-tools
https://docs.joinpeertube.org/maintain-tools?id=regenerate-thumbnailsjs

johanfleury:
Thanks for the suggestion, but unfortunately this script was added in 3.2-rc1 and I’m still running version 3.1 (the latest stable release).

Also, after having a quick look at the code, it seems that it’s not doing anything related to avatars, so I don’t think it would actually solve my issue.

AnonymousWebHacker:
That was what I was going to tell you, I confused avatar with preview of the videos [thumbnails] :)
[...]

PR #4149, opened 2021-06-01
Allow setting log level using PEERTUBE_LOG_LEVEL envvar

johanfleury:
## Description

Add an environment variable for `log.level`.

## Related issues

I was asked in #4071 to run PeerTube with log level set to debug and figured it was not possible to set it using an environment variable.

## Has this been tested?

I tested on my own instance by modifying the `custom-environment-variables.yaml` file and setting `LOG_LEVEL=debug`.


Chocobozzz:
Thanks

PR #4151, opened 2021-06-02
Remove annoying superfluous slash

frankstrater:
## Description

After running the automatic upgrade script a lot of paths get an extra slash (like the symbolic link for peertube-latest), because PEERTUBE_PATH has an unnecessary trailing slash. Although this doesn't lead to bugs, it's annoying the hell out of me for a couple of years now.

## Has this been tested?

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -2,7 +2,7 @@
 
 set -eu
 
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks

PR #4163, opened 2021-06-05
allow public video privacy to be deleted in the web client

### LINKED ISSUE #4159: Deleting privacy options breaks video uploading
I've created a very basic plugin that all it does is this:
```
  registerHook({
    target: 'action:application.listening',
    handler: () => {
      videoPrivacyManager.deletePrivacy(1);
      videoPrivacyManager.deletePrivacy(2);
      videoPrivacyManager.deletePrivacy(3);
    }
  })
```

Context: I'd like a 'family' peertube which only Internal (4) videos can be uploaded.

In doing the above, the video upload page breaks in multiple ways, most commonly, I cannot publish videos as the Privacy dropdown isn't filled out/cannot be selected properly/results in blank dropdown.

**Steps to reproduce:**

1. Create a basic plugin, with just the above handler
2. Naviate to the 'Publish Video' page and observe the Privacy

Screen Shot 2021-06-04 at 11 08 35 pm

I'd expect only 'Internal' is available to be selected, and the Dropdown box just contains that.

**Additional information**

* PeerTube instance:
  * URL: Private/internal
  * version: 3.2.1
  * NodeJS version: docker buster-production
  * Ffmpeg version: docker buster-production

* Browser name, version and platforms on which you could reproduce the bug: Firefox 89 Mac

rigelk:
## Description

A small fix that factorises the static default video privacy to the function returning filtered video privacies, and computes the default video privacy based on those returned by the API.

## Related issues

fixes #4159

## Has this been tested?

- [x] 👍 yes, light tests using the feature

### DIFF
--- a/client/src/app/shared/shared-main/video/video.service.ts
+++ b/client/src/app/shared/shared-main/video/video.service.ts
@@ -1,8 +1,8 @@
-import { Observable, of, throwError } from 'rxjs'
-import { catchError, map, mergeMap, switchMap } from 'rxjs/operators'
-import { HttpClient, HttpErrorResponse, HttpParams, HttpRequest } from '@angular/common/http'
+import { Observable } from 'rxjs'
+import { catchError, map, switchMap } from 'rxjs/operators'
+import { HttpClient, HttpParams, HttpRequest } from '@angular/common/http'
 import { Injectable } from '@angular/core'
-import { ComponentPaginationLight, RestExtractor, RestService, ServerService, UserService, AuthService } from '@app/core'
+import { ComponentPaginationLight, RestExtractor, RestService, ServerService, UserService } from '@app/core'
 import { objectToFormData } from '@app/helpers'
 import {
   FeedFormat,
@@ -378,7 +378,7 @@ export class VideoService implements VideosProvider {
                )
   }
 
-  explainedPrivacyLabels (privacies: VideoConstant<VideoPrivacy>[]) {
+  explainedPrivacyLabels (privacies: VideoConstant<VideoPrivacy>[], defaultPrivacyId = VideoPrivacy.PUBLIC) {
     const base = [
       {
         id: VideoPrivacy.PRIVATE,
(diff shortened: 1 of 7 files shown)


client/src/app/shared/shared-main/video/video.service.ts
```diff
@@ -398,9 +398,14 @@ export class VideoService implements VideosProvider {
       }
     ]
 
-    return base
+    const videoPrivacies = base
       .filter(o => !!privacies.find(p => p.id === o.id)) // filter down to privacies that where in the input
       .map(o => ({ ...privacies[o.id - 1], ...o })) // merge the input privacies that contain a label, and extend them with a description
+
+    return {
+      firstStepPrivacyId: videoPrivacies.find(p => p.id === VideoPrivacy.PUBLIC).id || videoPrivacies[0].id,
```
Chocobozzz, line 406: Put `VideoPrivacy.PUBLIC` as a default arg so we explicitly choose the default privacy in components

Chocobozzz: `firstStepPrivacyId` -> `defaultPrivacyId`

Chocobozzz (changes requested):

client/src/app/shared/shared-main/video/video.service.ts
```diff
@@ -398,9 +398,14 @@ export class VideoService implements VideosProvider {
       }
     ]
 
-    return base
+    const videoPrivacies = base
       .filter(o => !!privacies.find(p => p.id === o.id)) // filter down to privacies that where in the input
       .map(o => ({ ...privacies[o.id - 1], ...o })) // merge the input privacies that contain a label, and extend them with a description
+
+    return {
+      defaultPrivacyId: videoPrivacies.find(p => p.id === defaultPrivacyId).id || videoPrivacies[0].id,
```
Chocobozzz, line 406: Have you tested this code? Because `videoPrivacies.find` can return null

rigelk: I corrected the attribute access.

PR #4170, opened 2021-06-06
add script to transfer channel between local users

### LINKED ISSUE #4169: Transfer ownership of channel and playlist with CLI
Before implementing a full UX/UI ownership transfer for channel (https://github.com/Chocobozzz/PeerTube/issues/1417) or playlist (https://github.com/Chocobozzz/PeerTube/issues/3468, maybe it could be done via a simple request with the CLI by the admin ?

As it seems ownership transfers are pretty much asked, allowing admins to do it after a request of a user could be a first very welcomed feature.

rigelk:
## Description

A quick script to transfer the ownership of a channel and its actor. CUrrently the channel is transfered quite simply, and an AP Update for the actor is sent. However the videos owned by the channel are not yet updated over AP to reflect that change (that's a TODO).

## Related issues

fixes #4169

## Has this been tested?

No test was added to the suite, but I tested on local development instance accounts back and forth.


Chocobozzz:
Please add unit tests

scripts/transfer-channel.ts
```diff
@@ -0,0 +1,73 @@
...
+    channelActor.Account = to
+
+    console.log('Updating channel…')
+    const updatedChannel = await channel.save({ transaction: t })
+    console.log('Updating channel actor…')
+    await channelActor.save({ transaction: t })
+    console.log('Federating channel update…')
+    await sendUpdateActor(updatedChannel, t)
+    console.log('Federating videos update for the channel…')
+    await federateAllVideosOfChannel({ id: updatedChannel.id } as MChannelId)
```
Chocobozzz, line 71: I don't think you need this

server/tests/cli/transfer-channel.ts
```diff
@@ -0,0 +1,67 @@
...
+
+describe('Test transfer channel scripts', function () {
+  let server: ServerInfo
+  const user1 = { username: 'user_1', password: 'super password' }
+  const user2 = { username: 'user_2', password: 'super password' }
+  let user1Token
+
+  before(async function () {
+    this.timeout(30000)
+    server = await flushAndRunServer(1)
```
Chocobozzz, line 28: Please test with 2 servers to also test federation

Chocobozzz (changes requested):

scripts/transfer-channel.ts
rigelk: Why?

scripts/transfer-channel.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    channel.accountId = to.id
+    channelActor.Account = to
+
+    console.log('Updating channel…')
+    const updatedChannel = await channel.save({ transaction: t }) as MChannelDefault
+    console.log('Updating channel actor…')
+    await channelActor.save({ transaction: t })
+    console.log('Federating channel update…')
+    await sendUpdateActor(updatedChannel, t)
```
rigelk, line 68: This line yields an error I fail to understand.

kimsible: do you need some help ?

Chocobozzz:
I don't think we should use a CLI script for this purpose. I think you should create a specific endpoint '/video-channels/change-owner' or something like that. Running the job queue in a dedicated script is not really safe. Moreover with the API, we can in the future create a dedicated page in the UI.

kimsible:
> I don't think we should use a CLI script for this purpose. I think you should create a specific endpoint /video-channels/change-owner or something like that. Running the job queue in a dedicated script is not really safe. Moreover with the API, we can in the future create a page in the UI.

Ok, I understand the point. Despite @rigelk  has already done a great work, we can take inspiration from this code to move the feature in a specific endpoint ?

The endpoint would be accessible via CURL and will accept a `client_id` and `client_secret` as parameters ? it would also be accessible only for admins ?

hugopeixoto:
I'm trying to understand the code of this PR to figure out what needs to be done to transfer a channel. The important lines seem to be these:

```js
    channel.accountId = to.id
    channelActor.Account = to

    const updatedChannel = await channel.save({ transaction: t }) as MChannelDefault
    await channelActor.save({ transaction: t })
    await sendUpdateActor(updatedChannel, t)
```

I understand that the `channel` table has an `accountId` column, which maps to the `accounts` table. That explains the first line (and respective `save`).

I don't understand what `channelActor.Account = to` is supposed to achieve. I don't see any `Account` related columns in the actor table. Is this necessary? I don't understand what this and the respective `save` lines are doing.

If the actor doesn't keep track of its owner, does `sendUpdateActor` do anything? Is this why Chocobuzz said "I don't think you need this" wrt this line, because there are no changes to propagate?

PR #4176, opened 2021-06-08
Remove reference to non-existent external resource

frankstrater:
I'm not sure if https://packages.joinpeertube.org/ is just temporarily offline or moved, but currently this link is dead.

### DIFF
--- a/support/doc/plugins/guide.md
+++ b/support/doc/plugins/guide.md
@@ -753,7 +753,7 @@ Left menu links can be filtered (add/remove a section or add/remove links) using
 ### Publishing
 
 PeerTube plugins and themes should be published on [NPM](https://www.npmjs.com/) so that PeerTube indexes
-take into account your plugin (after ~ 1 day). An official PeerTube index is available on https://packages.joinpeertube.org/ (it's just a REST API, so don't expect a beautiful website).
+take into account your plugin (after ~ 1 day).
 
 ## Write a plugin/theme
 
--- a/support/doc/plugins/guide.md
+++ b/support/doc/plugins/guide.md
@@ -752,8 +752,7 @@ Left menu links can be filtered (add/remove a section or add/remove links) using
 
 ### Publishing
 
-PeerTube plugins and themes should be published on [NPM](https://www.npmjs.com/) so that PeerTube indexes
-take into account your plugin (after ~ 1 day).
+PeerTube plugins and themes should be published on [NPM](https://www.npmjs.com/) so that PeerTube indexes take into account your plugin (after ~ 1 day).
 
 ## Write a plugin/theme
 


support/doc/plugins/guide.md
```diff
@@ -753,7 +753,7 @@ Left menu links can be filtered (add/remove a section or add/remove links) using
 ### Publishing
 
 PeerTube plugins and themes should be published on [NPM](https://www.npmjs.com/) so that PeerTube indexes
-take into account your plugin (after ~ 1 day). An official PeerTube index is available on https://packages.joinpeertube.org/ (it's just a REST API, so don't expect a beautiful website).
```
Chocobozzz, line 756: Replace by 

` An official PeerTube index is available on https://packages.joinpeertube.org/ (it's just a REST API, so don't expect a website).`

frankstrater: I don't understand.

rigelk: The link is a web API, not a web site that your browser can display.

frankstrater: Sure, but then I expect it to return something on a simple GET. As of now I get a 404, as most users do when clicking on this link on the guide.

rigelk: It's an API, it's not meant for you click on, but for your PeerTube instance to use.

frankstrater: Just saying "it's just a REST API" without any description anywhere, doesn't make me an idiot. Removing one word "beautiful" as a suggestion to improve your own documentation isn't really helpful and rather dismissive. But, hey, it's your project, do whatever you feel like...

rigelk: There is a misunderstanding: I don't have control on what is displayed when you arrive at that page.

In the meantime, you are welcome to suggest a better description in the documentation - which is the point of documenting, right? Removing information just because you can't interpret it is actually what I find most dismissive here.

rigelk: Actually there is a meaningful difference between what suggested Chocobozzz and the original text. Having the adjective meant that the URL still pointed to a website, which isn't the case.

support/doc/plugins/guide.md
```diff
@@ -752,8 +752,7 @@ Left menu links can be filtered (add/remove a section or add/remove links) using
 
 ### Publishing
 
-PeerTube plugins and themes should be published on [NPM](https://www.npmjs.com/) so that PeerTube indexes
-take into account your plugin (after ~ 1 day). An official PeerTube index is available on https://packages.joinpeertube.org/ (it's just a REST API, so don't expect a beautiful website).
+PeerTube plugins and themes should be published on [NPM](https://www.npmjs.com/) so that PeerTube indexes take into account your plugin (after ~ 1 day).
```
rigelk, line 755:
```suggestion
PeerTube plugins and themes should be published on [NPM](https://www.npmjs.com/) so that PeerTube indexes take into account your plugin (after ~ 1 day). An official plugin index is available on [packages.joinpeertube.org](https://packages.joinpeertube.org/api/v1/plugins), with no interface to present packages since its only use is [within the configuration](https://github.com/Chocobozzz/PeerTube/blob/develop/config/default.yaml#L188-L194).
```

rigelk (changes requested):
More explanations as to how the index should be used.

Chocobozzz:
Thanks, I used the doc explanation in https://github.com/Chocobozzz/PeerTube/commit/754c73f32bd4c97e96a4b459e0d9f316074a2176

support/doc/plugins/guide.md
Chocobozzz: > Just saying "it's just a REST API" without any description anywhere, doesn't make me an idiot. 

We never said this PR was useless, an explanation was needed.

> Removing one word "beautiful" as a suggestion to improve your own documentation isn't really helpful and rather dismissive.

I agree my suggestion was not really great

> But, hey, it's your project, do whatever you feel like...

Not sure to understand this reaction. But anyway I think the new sentence is clearer. I'll try to add an HTML index in the indexer when I'll have time.

frankstrater: >  Removing information just because you can't interpret it is actually what I find most dismissive here.

Pff, pretty lame how you're responding, but at least you changed it to an actual working link, which was the point of this pull request: either remove the dead link or provide a working link. 

Next time I'll just post an issue and leave it at that, so you can ignore it at your leisure without insulting me.

Issue #4192, opened 2021-06-16
Sort playlists by latest published


kontrollanten:
**Describe the problem to be solved**
We're using playlist as a way to categorize content for our viewers. Currently the user has to scroll down to the bottom of the playlist to be able to see if there's any new content.

**Describe the solution you would like:**
We'd like to be able to sort our playlists by latest published videos.

**Describe alternatives you have considered**  
A) Youtube has solved this by adding a setting for each playlist that makes it possible to add new videos on top of the playlist. That's a more generic solution than sort by publish date.

B) `filter:api.video-playlist.list.result` hook to make it possible to solve this with a plugin. This will make it harder to have this setting per playlist, but it could be a good start.

kontrollanten:
I created this hook in alternative B and realized that it may be confusing to have a `position` property that will not be respected (in case a plugin is added to change to a custom order). So therefore I'm considering a third, and probably the best, alternative

C) Add a hook upon adding a video to a playlist, to make it possible to change the position property so that it will be on top instead of bottom of the playlist.

AnonymousWebHacker:
I think I like the **A** (something native without plugins). It would be an option on the same list for this.

**C**  Also valid too , for when you want to add a video to a list, and specify if at the beginning or end

PR #4196, opened 2021-06-18
Add video-playlist-element.created hook

kontrollanten:
## Description
Add `video-playlist-element.created` hook which is triggered when a video is added to a playlist. This will make it possible to have a custom order of videos in a playlist.

## Related issues
#4192 

## Has this been tested?

1) Install https://framagit.org/kontrollanten/peertube-plugin-playlist-reverse-order and follow usage instructions.
1) Verify the sort order.


Chocobozzz:
Please add tests in https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/plugins/action-hooks.ts

Chocobozzz:
Thanks

PR #4207, opened 2021-06-25
API doc: downloadEnabled boolean to PUT video

frankstrater:
## Description

When updating a video through the API, you can set the downloadEnabled boolean on PUT, but it's missing from the OpenAPI doc.

## Has this been tested?

Yes, I'm using a script on the API to bulk update downloadEnabled to true for all videos from a channel. Tested on a live production instance with the latest release.

## Addendum

This pull request also contains an addition of the word 'video' which seemed to be accidentally left out in a description for handling live videos.


Chocobozzz:
Thanks

PR #4208, opened 2021-06-26
add new filter hook to /overviews/videos

ahdinosaur:
Hi, Thanks for PeerTube! :smiley_cat:

## Description

This change adds new filter hooks `filter:api.overviews.videos.list.params` and `filter:api.overviews.videos.list.params` to /overviews/videos.

The motivation is because I want to make a plugin to [use `originallyPublishedAt` in all places where `publishedAt` is used](https://github.com/ahdinosaur/peertube-plugin-originally-published-at).

I got every other page working using hooks, (well except for the user history page), but the overview ("Discovery") page seems to use a different API endpoint without any hooks setup.

This is my first time poking in the code, so not sure if this is the best approach, but seemed to follow existing patterns.

## Has this been tested?

What and where should I be testing?

### DIFF
--- a/server/controllers/api/overviews.ts
+++ b/server/controllers/api/overviews.ts
@@ -7,6 +7,7 @@ import { CategoryOverview, ChannelOverview, TagOverview, VideosOverview } from '
 import { MEMOIZE_TTL, OVERVIEWS } from '../../initializers/constants'
 import * as memoizee from 'memoizee'
 import { logger } from '@server/helpers/logger'
+import { Hooks } from '../../lib/plugins/hooks'
 
 const overviewsRouter = express.Router()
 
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Thanks!

> What and where should I be testing?

See https://docs.joinpeertube.org/contribute-getting-started?id=unit-tests

Then:
 * Add your filters in this plugin: https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/fixtures/peertube-plugin-test/main.js
 * In the filters, either update the result/params or log something
 * Test your filter in https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/plugins/filter-hooks.ts in a dedicated `describe()`, that uses https://github.com/Chocobozzz/PeerTube/blob/develop/shared/extra-utils/overviews/overviews.ts#L4 to GET the overview endpoint
 * Either check the response is altered, or a specific string has been logged (`waitUntilLog`)

Chocobozzz:
I added the tests, thanks @ahdinosaur

PR #4214, opened 2021-06-29
Expose to plugin a way to reset categories to their initial state

lutangar:
## Description

Expose a `resetCategories` method on the `PluginCategoriesManager` allowing a plugin developer to reset *customizations* done on the `VIDEO_CATEGORIES` to be clean on demand.

> - add a `resetConstants` method allowing to clean only one `type` of *alteration* for a specific plugin.
> - make `hash` map global for easier reference and dry up constants methods a little bit

This would help simplify some customization workflow when it involves renaming, updating, deleting, etc...

> I hesitated to create another method `getCategories` on the manager because I'm having another other issue where I find myself unable to call the API to retrieve the list of the categories. In the `register` method when the instance restarted the Rest API is not yet available... but I'll probably create a separate issue. 
> Would *reading* infos about the constants from these managers be considered insane? Shall it always go through the Rest API instead?

### Possible evolutions (or further refacto)

- expose such method on other *ConstantsManger*
- move `VideoConstant` interface to shared and rename \
- create a `PluginConstantManager` interface such as:
```
export interface PluginConstantManager {
  add: (key: string, label: string) => boolean
  delete: (key: string) => boolean
  reset: () => void
}
```
> All *managers* could then be created generically and reduce code duplication

## Related issues

> see https://framacolibri.org/t/plugin-system-issues/12417 for some background

## Has this been tested?

### DIFF
--- a/server/lib/plugins/register-helpers.ts
+++ b/server/lib/plugins/register-helpers.ts
@@ -1,5 +1,5 @@
 import * as express from 'express'
-import { logger } from '@server/helpers/logger'
+import {logger} from '@server/helpers/logger'
 import {
   VIDEO_CATEGORIES,
   VIDEO_LANGUAGES,
@@ -7,8 +7,8 @@ import {
   VIDEO_PLAYLIST_PRIVACIES,
   VIDEO_PRIVACIES
 } from '@server/initializers/constants'
-import { onExternalUserAuthenticated } from '@server/lib/auth/external-auth'
-import { PluginModel } from '@server/models/server/plugin'
+import {onExternalUserAuthenticated} from '@server/lib/auth/external-auth'
+import {PluginModel} from '@server/models/server/plugin'
 import {
   RegisterServerAuthExternalOptions,
   RegisterServerAuthExternalResult,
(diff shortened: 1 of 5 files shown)


server/tests/plugins/video-constants.ts
```diff
@@ -173,6 +174,30 @@ describe('Test plugin altering video constants', function () {
     }
   })
 
+  it('Should be able to reset categories', async function () {
+    await installPlugin({
+      url: server.url,
+      accessToken: server.accessToken,
+      path: getPluginTestPath('-video-constants')
+    })
+
```
Chocobozzz, line 183: Add

```
    expect(categories[42]).to.exist
    expect(categories[43]).to.exist
```

server/lib/plugins/register-helpers.ts
```diff
@@ -44,6 +44,14 @@ type UpdatedVideoConstant = {
   }
 }
 
+const hash: { [key in AlterableVideoConstant]: VideoConstant } = {
```
Chocobozzz, line 47: `constantsHash`

Chocobozzz (changes requested):
Thanks!

lutangar:
Thanks for the review!

...but I still wonder:

> Would *reading* infos about the constants from these managers be considered insane? Shall it always go through the Rest API instead?

Also what's your sentiment on these suggestions?

> - expose such method on other *ConstantsManger*
> - move `VideoConstant` interface to shared and rename the client interface which has the same name
> - create a `PluginConstantManager` interface

More details in previous comment :arrow_up:

Chocobozzz:
> Would reading infos about the constants from these managers be considered insane?

No it's definitely the way to do. Altering the object like we do is not very clean. I would just use instead:

```ts
export interface ConstantManager <K> {
  add: (key: k, label: string) => boolean
  delete: (key: k) => boolean
  get: () => { [ id: K ]: string }
  reset: () => void
}
```

Chocobozzz:
Closing in favour of https://github.com/Chocobozzz/PeerTube/pull/4251

PR #4218, opened 2021-06-30
Document a way to update the last available version of a plugin manually

lutangar:
May comes in very handy to quickly test a plugin in a production like environment before it reaches a greater world.

> This magic trick comes from @Chocobozzz himself, who passed it on to @JohnXLivingston, who then passed it down to me.


Chocobozzz:
Thanks @lutangar

PR #4229, opened 2021-07-05
Add Service-Worker-Allowed header to plugin static files

### LINKED ISSUE #3830: Send notifications via Push and Notification API

kontrollanten:
## Description
Plugin files listed under serviceWorkerScripts in package.json will be served with a Service-Worker-Allowed header in the HTTP resp.

## Related issues
Closes #3830

## Has this been tested?

- [x] 🙋 no, because I need help <!-- Detail how we can help you >

This could be tested with the hello-world plugin if we add a service worker file to it. Right now it can be tested with https://framagit.org/kontrollanten/peertube-plugin-one-signal (don't use the version published to npm).

### DIFF
--- a/shared/models/plugins/plugin-package-json.model.ts
+++ b/shared/models/plugins/plugin-package-json.model.ts
@@ -21,6 +21,7 @@ export type PluginPackageJson = {
   library: string
 
   staticDirs: { [ name: string ]: string }
+  serviceWorkerScripts: string[]
   css: string[]
 
   clientScripts: ClientScript[]
--- a/server/controllers/plugins.ts
+++ b/server/controllers/plugins.ts
@@ -38,6 +38,12 @@ pluginsRouter.get('/plugins/:pluginName/:pluginVersion/static/:staticEndpoint(*)
   servePluginStaticDirectory
 )
 
+pluginsRouter.get('/plugins/:pluginName/:pluginVersion/service-worker-scripts/:staticEndpoint(*)',
+  getPluginValidator(PluginType.PLUGIN),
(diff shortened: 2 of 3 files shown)


shared/models/plugins/plugin-package-json.model.ts
```diff
@@ -21,6 +21,7 @@ export type PluginPackageJson = {
   library: string
 
   staticDirs: { [ name: string ]: string }
+  serviceWorkerScripts: string[]
```
Chocobozzz, line 24: Put it below `clientScripts`

Chocobozzz (commented):
Hi,

Do we really need to expose such method? I mean you could easily use the router to serve your service worker script using the appropriate header no?

> This could be tested with the hello-world plugin if we add a service worker file to it

Add a plugin that exposes service workers script. In server tests add a get request that checks the server correctly serves the service worker script and correctly expose the appropriate header

kontrollanten:
Oh, I didn't know about the plugins router API. I think that ServiceWorkers are doable without this PR then. I just need to verify.

Issue #4225, opened 2021-07-02
What is the required python version?


Stopi:
According to the [dependencies](https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/dependencies.md) doc, python seems to be obviously required.
However, for Gentoo it is asked to get python 2.7 when this isn't the case for other distro.

My question: is this still a requirement or can we use python 3?
I couldn't find any mention in the [changelog](https://github.com/Chocobozzz/PeerTube/blob/develop/CHANGELOG.md#v230) either.

Python 2 is deprecated since 1st January 2020.
Recently Gentoo removed all dependencies to python 2 from its stable tree, the currently used version is 3.9.
Peertube 3.2.1 is my last package requiring python 2, so I'd like to know.

Hopefully this is just a matter of updating the doc...

frankstrater:
In my experience Python3 (on Red Hat distros) works fine. I believe the Python 2.7 requirement stems from the youtube-dl binary that is included as a node module (needed for the Video Import feature of PeerTube). The latest version of youtube-dl states:

"youtube-dl is a command-line program to download videos from YouTube.com and a few more sites. It requires the Python interpreter, version 2.6, 2.7, or 3.2+, and it is not platform specific."

The npm package of youtube-dl is only slightly outdated, so I think it's fine for Python 3.2 or above

I have no experience with Gentoo, but keep in mind that youtube-dl expects /usr/bin/python to exist. So you might need to add a symlink like:

```
sudo ln -s /usr/bin/python3 /usr/bin/python
```

Certbot needs Python as well, but certbot is not part of the core of PeerTube (and probably works fine with Python 3 as well) and is out of scope.

So, not sure if I'm missing something, but also hope I don't have to maintain multiple Python versions on the server as well.

Stopi:
Hi frankstrater, thanks for your answer.

I believe it's just a matter of changing the doc then.
requirement `dev-lang/python:2.7` for Gentoo might just be changed in `dev-lang/python` in [https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/dependencies.md](https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/dependencies.md) on line 324.
Btw I see this line was written 3 years ago, so deprecation make sense.

I could do a PR if you want to but I'm hesitating to clone a full repo just for correcting a typo (aka removing 4 characters in this case). So if you can do it that would be awesome, since you already are a contributor.

[...]

PR #4238, opened 2021-07-07
Fedora guide fix

frankstrater:
## Description

After installing on Fedora 32 and 33 a couple of times, I noticed some missing parts when following the production guide (that references the dependencies guide)

- vim isn't installed by default anymore (Fedora switched to nano)
- oidentd is needed to get access to postgres (there's other ways, but installig oidentd just works without config)
- the nginx package on the Red Hat distros doesn't contain the sites-enabled and sites-available folders and they are not loaded through the main nginx conf (by adding these folders and using a symlink from conf.d)

## Related issues

The guide for the other Red Hat family distros (CentOS 7, CentOS 8) also could use an update (CentOS 8 will be EOL sooner than 7). I'll create a separate PR for those after testing them.

There's also a small change for Gentoo related to the Python version. See this issue: https://github.com/Chocobozzz/PeerTube/issues/4225 

## Has this been tested?

Yes, tested on Fedora on bare metal. Should be fine for the current Fedora 34 version. Will test again later when my hardware comes available again.


Chocobozzz:
Thanks!

PR #4245, opened 2021-07-10
Change default order in search results to -views

### LINKED ISSUE #4219: Customize default search sort
When users are searching on our PeerTube site they feel that the search results are presented in a "confusing/unlogical" order.

The proposed solution from our users is that an order based on views count would be more easy to understand. This can be done manually each time a user is searching, but most of the users don't go that far, they just quit using the search functionality if they feel uncomfortable using it. Therefore it would be great to set a default sort order.

A) Add a setting in the configuration file; `search.sort: 'views'`.
B) Add a hook so that the sort can be changed by a plugin.

kontrollanten:
## Description
Change default sort order in search results to -views, since the relevance search order doesn't really makes sense.

## Related issues
Closes #4219

## Has this been tested?

Does this really need a test?

### DIFF
--- a/client/src/app/+search/search-filters.component.html
+++ b/client/src/app/+search/search-filters.component.html
@@ -5,7 +5,7 @@
       <div class="form-group">
         <div class="radio-label label-container">
           <label i18n>Sort</label>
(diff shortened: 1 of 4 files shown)


Chocobozzz:
See https://github.com/Chocobozzz/PeerTube/issues/4219#issuecomment-878082813

PR #4251, opened 2021-07-13
Introduce generic constant registries for plugins

lutangar:
## Description

Allow a plugin developer to get back constants values, and reset constants deletions or additions.

> see https://www.martinfowler.com/eaaCatalog/registry.html for a quick `registry` definition

This would introduce a lot of *deprecations*, so feel free to disagree on the naming changes. 

Initial suggestion was :
> ```ts
> export interface ConstantManager <K> {
>  add: (key: K, label: string) => boolean
>  delete: (key: K) => boolean
>  get: () => { [ id: K ]: string }
>  reset: () => void
>}
>```
> https://github.com/Chocobozzz/PeerTube/pull/4214#issuecomment-871453519

But imho this would convey more meaning:
```ts
export interface ConstantRegistry <K extends string | number> {
  addConstant: (key: K, label: string) => boolean
  deleteConstant: (key: K) => boolean
  getConstantValue: (key: K) => string
  getConstants: () => Record<K, string>
  resetConstants: () => void
}
```
Especially the choice of `getConstants` over `get` which would have been ambiguous.
~Not sure if `getConstantValue` should only return the value tho'~

Let me know what you think!

## Related issues

This would eventually supersede #4214 if agreed upon.

## Has this been tested?

### DIFF
--- a/server/lib/plugins/register-helpers.ts
+++ b/server/lib/plugins/register-helpers.ts
@@ -1,13 +1,7 @@
 import * as express from 'express'
 import { logger } from '@server/helpers/logger'
-import {
-  VIDEO_CATEGORIES,
-  VIDEO_LANGUAGES,
-  VIDEO_LICENCES,
-  VIDEO_PLAYLIST_PRIVACIES,
-  VIDEO_PRIVACIES
-} from '@server/initializers/constants'
 import { onExternalUserAuthenticated } from '@server/lib/auth/external-auth'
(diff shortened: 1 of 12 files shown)


Chocobozzz:
Thanks @lutangar! I'll review/merge this next week. If you don't mind I'd rather not change the manager names. I think it doesn't hurt to keep the new method names (`addConstant`...) while deprecating the old ones (`addLicence`...)

Chocobozzz:
Thanks!

PR #4255, opened 2021-07-16
Allow caching for HLS playlists & videos

### LINKED ISSUE #4210: Allow aggressive client caching
When using the CLI create-import-video-file-job to change an already existing video the previous video version may be cached on the client side since the URL doesn't change when the video change. A similar issue may appear if a video is published before all resolutions are transcoded; the master.m3u8 can be cache without all possible resolutions added. This means that we have to use a short cache time to ensure fresh content is available.

When the video is changed (new video file or new resolution) the URL should change to ensure that all caches (CDN, browser, etc) will be cleared. This will make it possible to use aggressive caching in browsers, web workers, CDN, etc.

* Add resolutions list hash to the URL of the master file; `/static/streaming-playlists/hls/[uuid]/[hash]/master.m3u8`
* Add the file hash to the URL of resolution playlists and video files; `/static/streaming-playlists/hls/[uuid]/[hash]/480.m3u8`

kontrollanten:
## Description
HLS video files gets named by a unique uuid (not the same as video uuid), which gets updated when the files changes. All m3u8 playlists and segments files gets query param with a hash based on the video file names. This ensures that if a video changes also the URL will be changed.

## Related issues
Closes #4210

## Has this been tested?

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,7 @@
     "tsconfig-paths": "^3.9.0",
     "tslib": "^2.0.0",
     "useragent": "^2.3.0",
+    "uuidv4": "^6.2.11",
     "validator": "^13.0.0",
     "webfinger.js": "^2.6.6",
     "webtorrent": "^1.0.0",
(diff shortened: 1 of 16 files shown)


Chocobozzz:
Hi,

I appreciate but it's a little bit complex and will require more work. I'll implement this in a dedicated branch and will ping you.

Issue #4262, opened 2021-07-20
Invalid z-index for notification dropdown and admin menu titles


kimsible:
![Screenshot 2021-07-20 at 14-26-09 Users list - PeerTube](https://user-images.githubusercontent.com/1877318/126367576-8fc4be9a-a67c-40c7-af7b-e58eb0313158.png)

frankstrater:
This problem is visible when you have a 'broadcast message' activated. In the screenshot of @kimsible you can see the light blue part of the info level broadcast message which pushes the subnavbar (with 'Federation') below.

kimsible:
> This problem is visible when you have a 'broadcast message' activated. In the screenshot of @kimsible you can see the light blue part of the info level broadcast message which pushes the subnavbar (with 'Federation') below.

@frankstrater indeed, thanks for the details

Chocobozzz:
Closed by https://github.com/Chocobozzz/PeerTube/pull/4266

Issue #4265, opened 2021-07-21
The left side of the window is cut off on the reports page


Alterak:
**Describe the current behavior**

On the reports page, the left side of the popup is clipped. I noticed that if the window opens at the top, then the left part of the text disappears, and if it opens at the bottom, then everything is displayed correctly. I have selected the Russian interface language.

![Screenshot_1](https://user-images.githubusercontent.com/11453567/126442315-37cc4b7b-345d-43c1-b5b6-c8267547d2da.png)
![Screenshot_2](https://user-images.githubusercontent.com/11453567/126442316-037b5b7d-9a5b-4bcb-a5f1-594780f88584.png)

**Steps to reproduce:**

1. Open the reports page.
2. Open the action menu for one report.
3. If the action window opens at the top, it will be clipped.

**Describe the expected behavior**
Should display correctly. The window should not be hidden behind the left side of the menu.

* PeerTube instance:
  * URL: https://peervideo.ru
  * version: 3.3.0

frankstrater:
Confirmed, but not a new problem. Exists as well for 3.2.1

Technical description of this CSS bug:

2 wrapper divs: 1 for dropdown and 1 for dropup

```html
<div _ngcontent-pci-c93="" style="position: absolute; z-index: 1050; top: 0px; left: 0px; will-change: transform; transform: translate(228px, 373px);" class="dropdown">
```

```html
<div _ngcontent-pci-c93="" style="position: absolute; z-index: 1050; top: 0px; left: 0px; will-change: transform; transform: translate(233px, 557px);" class="dropup">
```

CSS entry to overwrite the z-index for dropdown in bootstap.sccs:

```css
.dropdown {
    z-index: 17000!important;
}
```

@Chocobozzz just add `.dropup` to this? Like:

```css
.dropdown, .dropup {
    z-index: 17000!important;
}
```

kimsible:
Thanks @frankstrater I added this fix to the PR https://github.com/Chocobozzz/PeerTube/pull/4266

Chocobozzz:
Closed by https://github.com/Chocobozzz/PeerTube/pull/4266

PR #4266, opened 2021-07-21
Correct dropdown/dropup z-index

kimsible:
## Description

This PR try to correct all known issues about dropdown z-index especially with :
- user dropdown notification placing hover the sub-menu titles (also as a dropdown-menu)
- dropdowns moving hover the header and privacy message on scroll
- missing `.dropup` z-index rule especially in report admin page

## Related issues

Closes https://github.com/Chocobozzz/PeerTube/issues/4262 and https://github.com/Chocobozzz/PeerTube/issues/4265

Plus floating dropdown hover menu and privacy message on scroll as the following screenshots : 

Capture d’écran du 2021-07-21 13-07-19
Capture d’écran du 2021-07-21 12-54-41

## Has this been tested?

### DIFF
--- a/client/src/app/+videos/+video-watch/shared/information/privacy-concerns.component.scss
+++ b/client/src/app/+videos/+video-watch/shared/information/privacy-concerns.component.scss
@@ -25,6 +25,13 @@
   }
 }
 
+// Avoid higher z-index when overlay on touchscreens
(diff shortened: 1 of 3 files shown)


Chocobozzz:
Thanks

Issue #4147, opened 2021-06-01
Large file uploads results in 500


kontrollanten:
**Describe the current behavior**
When uploading a video file at 3 gb I receive a 500 response in the end (~99.9%). Upon reloading the page the video has been created twice, so the upload did actually work.

**Steps to reproduce:**

1. Upload a 3 gb file to an instance that's using s3fs to store the videos.

* PeerTube instance:
  * version: 3.2.0

* Link to server log if relevant (`journalctl` or `/var/www/peertube/storage/logs/`):
```
Jun 01 14:56:31 test-www2 peertube[970]: [test:443] 2021-06-01 14:56:31.159 error: Error in controller. {
Jun 01 14:56:31 test-www2 peertube[970]:   "err": "Error: ffprobe exited with code 1\nffprobe version 4.2.4-1ubuntu0.1 Copyright (c) 2007-2020 the FFmpeg developers
built with gcc 9 (Ubuntu 9.3.0-10ubuntu2)
configuration: --prefix=/usr --extra-version=1ubuntu0.1 --toolchain=hardened --libdir=/usr/lib/x86_64-linux-gnu --incdir=/usr/include/x86_64-linux-gnu --arch=amd64 --enable-gpl --disable-stripping --enable-avresample --disable-filter=resample --enable-avisynth --enable-gnutls --enable-ladspa --enable-libaom --enable-libass --enable-libbluray --enable-libbs2b --enable-libcaca --enable-libcdio --enable-libcodec2 --enable-libflite --enable-libfontconfig --enable-libfreetype --enable-libfribidi --enable-libgme --enable-libgsm --enable-libjack --enable-libmp3lame --enable-libmysofa --enable-libopenjpeg --enable-libopenmpt --enable-libopus --enable-libpulse --enable-librsvg --enable-librubberband --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libtwolame --enable-libvidstab --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx265 --enable-libxml2 --enable-libxvid --enable-libzmq --enable-libzvbi --enable-lv2 --enable-omx --enable-openal --enable-opencl --enable-opengl --enable-sdl2 --enable-libdc1394 --enable-libdrm --enable-libiec61883 --enable-nvenc --enable-chromaprint --enable-frei0r --enable-libx264 --enable-shared
[...]

Issue #3661, opened 2021-01-29
Support S3 storage to store videos


Chocobozzz:
We already have [a documentation](https://docs.joinpeertube.org/admin-remote-storage) to mount a s3 storage, but there are some limitations and issues with fuse-mounted s3 storage. A proper support of s3 within PeerTube to store videos (webtorrent versions and HLS versions) would fix them.

pierreozoux:
S3 is a product that belongs to AWS, and the common technology is called ObjectStore, so if you want to be agnostic about it, I kindly recommend to name it ObjectStore :) (But the awefull truth is that there is no specification, and S3 is the specification..)

grubeli:
I did try to use Backblaze B2 and PeerTube constantly fails to upload relatively big files (~300 MB) and start uploading again with a different name (I checked backend).
Also while uploading it consumes massive amount of "Download traffic" Bandwidth. 

Definitely we need a better support for S3.

@Chocobozzz Thanks for your work

arthurlogilab:
We are using with OVH S3 https://www.ovhcloud.com/en/public-cloud/object-storage/ (which uses openstack's swift implementation) on kubernetes with s3fs sidecar (see https://docs.joinpeertube.org/install-unofficial?id=kubernetes)  . Happy to help out with testing if this development starts.

manuelviens:
Same problem as @grubeli with big files on my instance using S3FS with Wasabi : I get an error (browser) after the upload and before the Publish button becomes available (more than 5 minutes...). 
Then (several hours later) it result with two videos.

kontrollanten:
@manuelviens If you open up Chrome devtools, can you report the HTTP response code you see when the browser error appears (it should be 500, 504 or something similar)? A screenshot at the _Console_ and _Network_ tabs works as well.

kukhariev:
@kontrollanten, @manuelviens ,
 can you test with the  `s3fs  -o nomultipart` option?

kontrollanten:
@kukhariev For me that doesn't help. I still get a 502 from my load balancer after 4 minutes.

> The default timeout value in a browser is 300 seconds, but most network infrastructures include proxies and servers whose timeout is not that long.

https://tools.ietf.org/id/draft-loreto-http-bidirectional-07.html#timeouts

I can find two possible solutions for this:

A) After the upload is done, let the client do long polling until the file has moved. That would mean that the request is closed each 15 second and a new request is done until the move is done.
[...]

PR #4282, opened 2021-07-27
Handle video file move in a job

kontrollanten:
## Description
This is a proposal that shows how heavy tasks during file upload is handled in a job instead of during the HTTP request. This will make PeerTube scale better when an instance grows big and a lot of file uploads are done simultaneously. It will also create a better user experience since the user can leave the video creation before all the video process are done, which is not the case today. And it will also solve issues when using S3 video storage where it can take 10 minutes for a file move to S3 to finish.

@Chocobozzz Should I go further with this? Do you have any preferences regarding how it should be handled?

## Related issues
#4147 
#3661 

## Has this been tested?

### DIFF
--- a/server/controllers/api/videos/live.ts
+++ b/server/controllers/api/videos/live.ts
@@ -98,6 +98,7 @@ async function addLiveVideo (req: express.Request, res: express.Response) {
 
   const [ thumbnailModel, previewModel ] = await buildVideoThumbnailsFromReq({
     video,
+    // @ts-expect-error
     files: req.files,
     fallback: type => {
       return updateVideoMiniatureFromExisting({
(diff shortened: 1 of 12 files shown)


Chocobozzz:
Would https://github.com/Chocobozzz/PeerTube/pull/4290 fix the underlined issue?

kontrollanten:
Yes, It'll solve a huge part of it. After the upload is done there's still a few kind of heavy operations that's done during the HTTP request

* Move the uploaded file.
* Generate preview image.
* Calculate video length.

Hypothetically that can be an issue, but maybe it's better solve that when it's a real issue.

Chocobozzz:
> Hypothetically that can be an issue, but maybe it's better solve that when it's a real issue.

:+1: Then closing and we'll see when the S3 PR will be integrated

PR #4283, opened 2021-07-28
Adding Rocky Linux 8.4

WingsLikeEagles:
## Description

Adding documentation for Rocky Linux 8.4

## Related issues

Adapted from CentOS 8... but seems like there is stuff missing from that section.
Tested in AWS up until the steps end.

## Has this been tested?

- [X ] 💭 no, because this PR is a draft and still needs work

### DIFF
--- a/support/doc/dependencies.md
+++ b/support/doc/dependencies.md
@@ -151,6 +151,51 @@ sudo systemctl enable --now redis
 sudo systemctl enable --now postgresql
 ```
 
+## Rocky Linux 8.4  
+
+1. Pull the latest updates:  
+```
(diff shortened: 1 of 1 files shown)


WingsLikeEagles:
Seems like there are missing steps... NGINX config, firewall (which would be easy to add), and PostgreSQL config, and maybe others.  I copied from the CentOS 8 steps, which appear to be missing the same information.  The steps I added all work from what I can tell.  NodeJS, npm, yarn, Python3.6, libSDL2, and ffmpeg all show good versions from what I can tell.

Chocobozzz:
Thanks

PR #4290, opened 2021-07-30
Add support for saving video files to object storage

### LINKED ISSUE #3661: Support S3 storage to store videos

pingiun:
## Description

I've reworked this PR a bit so here is the new description. I've kept the old description included below.

This PR adds a job that normally runs after transcoding (unless disabled) which uploads video files to an s3 compatible object store. This improves on the current situation because files are uploaded in a job instead of a request, and they use the native APIs to do this, which makes sure files aren't being rewritten. Some model fields are added to keep track of the moving state and to make sure the last part of the job is only run once.

The `movingJobsRunning` field in the video model is incremented whenever a transcode job is created and only decremented once the uploads for the created videoFile are done. Then, the new job can detect when the last videoFile was uploaded because `movingJobsRunning` will be 0 and the cleanup part of the job can run. This uploads the master HLS playlist, removes the files from the local disk, and sets the video to published.

The `storage` field on videoFile and videoStreamingPlaylist is used to create the correct url to send in the API. The client then correctly loads the files from object storage (als long as CORS is correctly configured on the bucket). The `streaming_playlists_url_template` config field is used for custom URL generation, this is explained in the [comment in default.yaml](https://github.com/Chocobozzz/PeerTube/blob/42fec677a1c1cfe400cdedbfe49096d7316d1abd/config/default.yaml#L105-L112)

<details>
<summary>Old PR text before rework</summary>
This adds a job that runs after all transcodes are done that uploads all the video files to object storage (s3 compatible). New fields are added to the video, videoFile, and videoStreamingPlaylist model to keep track of transcodes and where files are stored. 

The `transcodeJobsRunning` field in the video model is incremented whenever a transcode job is created and decremented whenever one finishes. The job that moves files to object storage is added after each transcode job is finished, but only runs when `transcodeJobsRunning` is 0 so that all files can be uploaded at once.

The `storage` field on videoFile and videoStreamingPlaylist is used to create the correct url to send in the API. The client then correctly loads the files from object storage (als long as CORS is correctly configured on the bucket).

Currently the old files are not removed, because I wanted to ask feedback on how to best do this. In principle the job that moves files to object storage could delete the files right after the files are uploaded, because we are sure that no transcoding jobs are running anymore. However it might be the case that a user is currently watching a video that is streaming from those files, and I think removing the files might cause the stream to drop.

In addition to this I also need to implement a mechanism to get the files back to the local filesystem whenever local operations need to be done. I think this is only the case when transcoding but I'm not sure about that.
</details>

### TODO (need help):
- [x] Remove old files from the filesystem
- [x] Mechanism to retrieve files from object storage whenever new transcodes are needed

## Related issues

Fixes #3661

## Has this been tested?


server/types/models/video/video.ts
```diff
@@ -219,3 +219,8 @@ export type MVideoFormattableDetails =
   Use<'VideoStreamingPlaylists', MStreamingPlaylistRedundanciesOpt[]> &
   Use<'VideoFiles', MVideoFileRedundanciesOpt[]> &
   PickWithOpt<VideoModel, 'Trackers', MTrackerUrl[]>
+
+export enum VideoStorageType {
```
pingiun, line 223: I don't think this is the best location for this enum but I wasn't sure where else to put it

pingiun:
I already started working on the new documentation: https://framagit.org/pingiun/documentation/-/commit/2bbeeaa158d572e2234d1f2d82587bd8bfb39acc

kontrollanten:
Elegant! Impressive work.

> Currently the old files are not removed, because I wanted to ask feedback on how to best do this. In principle the job that moves files to object storage could delete the files right after the files are uploaded, because we are sure that no transcoding jobs are running anymore. However it might be the case that a user is currently watching a video that is streaming from those files, and I think removing the files might cause the stream to drop.

Can we add a video state called "processing" (or something similar) until the upload to s3 is done? That would mean that the video isn't available until the files are in the bucket.

A future request would be the possibility to set the s3 download base URL. That way it would be possible to set the file URL to a CDN, to reduce the costs.

pingiun:
> Can we add a video state called "processing" (or something similar) until the upload to s3 is done? That would mean that the video isn't available until the files are in the bucket.

This might actually be the most elegant solution, good idea.

> A future request would be the possibility to set the s3 download base URL. That way it would be possible to set the file URL to a CDN, to reduce the costs.

Because transcoding still needs to happen I don't think this is very useful.

SimplyCorbett:
> A future request would be the possibility to set the s3 download base URL. That way it would be possible to set the file URL to a CDN, to reduce the costs.

Isn't it already possible to do this by caching the /w/xyz portion of the URL using varnish or nuster or similar?

kontrollanten:
> > A future request would be the possibility to set the s3 download base URL. That way it would be possible to set the file URL to a CDN, to reduce the costs.
> 
> Isn't it already possible to do this by caching the /w/xyz and streaming-playlists portion of the URL using varnish or nuster or similar?

Today that's possible, but as I understand this PR it will change the video file URL so that the browser will download it directly from S3 (correct me if I'm wrong, @pingiun ).

pingiun:
> Today that's possible, but as I understand this PR it will change the video file URL so that the browser will download it directly from S3 (correct me if I'm wrong, @pingiun ).

This is correct

SimplyCorbett:
> > Today that's possible, but as I understand this PR it will change the video file URL so that the browser will download it directly from S3 (correct me if I'm wrong, @pingiun ).
> 
> This is correct

Then there needs to be a way to cache the video. This isn't a good idea. Cache mirrors are cheap. I can high bandwidth servers for a fairly cheap price.

Direct S3 would be many times more expensive.

pingiun:
> > > Today that's possible, but as I understand this PR it will change the video file URL so that the browser will download it directly from S3 (correct me if I'm wrong, @pingiun ).
> 
> > 
> 
> > This is correct
> 
> 
> 
> Then there needs to be a way to cache the video. This isn't a good idea. Cache mirrors are cheap. I can get high bandwidth servers for a fairly cheap price.
> 
> 
> 
> Direct S3 would be many times more expensive.

With these changes it's still possible to use the old method of s3 storage via s3fs.

Adding the suggestion from @kontrollanten should be fairly simple though, I can try to add that to this PR

pingiun:
@SimplyCorbett Can you take a look at https://github.com/Chocobozzz/PeerTube/pull/4290/commits/26cabf28921158a6bd26aeccafccb08e50a517fb and see if this change adds the right support for you?

pingiun:
> I think a base URL should be enough for the most cases.

Changed in https://github.com/Chocobozzz/PeerTube/pull/4290/commits/03c6d12e0ff71c268896b173abd4e6e97e06e5d5

kontrollanten:
>  Mechanism to retrieve files from object storage whenever new transcodes are needed

Have you any ideas about how to solve this? I'm thinking about the following solution

1. In `handleHLSJob`, a check is done if the video file exists in PeerTube tmp folder. If not, it will download the file from s3.
2. When `handleHLSJob` is done, a new job, `cleanupTmpVideo`, is created.
3. `cleanupTmpVideo` job will check if there's any video transcoding jobs in queue, if not it will remove the tmp file.

What do you think about that?

Chocobozzz:
Hello and thanks @pingiun 

Since it's still a draft, here's some code remarks:
 * Try to use `objectStorage` prefix instead of `s3` prefix for your function/variables/config names
 * Prefer to use a tree for the config: `s3.streaming_playlists_bucket` -> `object_storage.streaming_playlists.bucket`
 * Use uppercase for config: `S3.STREAMING_PLAYLISTS_BUCKETINFO.bucket` -> `OBJECT_STORAGE.STREAMING_PLAYLISTS.BUCKET` (maybe `BUCKET_NAME` instead of `BUCKET`)
 * I suggest  to rename `moveJobsRunning` to `pendingMovingJobs` (or better, create a dedicated `videoJobInfo` with a `pendingMove` & `videoId` columns so we could also use this table to track pending transcoding jobs)
 * https://github.com/Chocobozzz/PeerTube/pull/4290/files#diff-3e26d41ca4bda1de8e1747af70ca2af642abcc1e9e0bfb94239ff2165acfbde5R19 uses a string instead of an integer
 * I think we should store the origin object storage URL in `fileUrl`, without `base_url` injection. Instead, inject the `base_url` at "runtime" so admins can easily change this configuration without running a script to update DB URLs

server/models/video/video-job-info.ts
```diff
@@ -0,0 +1,85 @@
...
+    return pendingMove
+  }
+
+  static async decreasePendingMove (videoUUID: string): Promise<number> {
+    const options = { type: QueryTypes.SELECT as QueryTypes.SELECT, bind: { videoUUID } }
+
+    const [ { pendingMove } ] = await VideoJobInfoModel.sequelize.query<{pendingMove: number}>(`
+    UPDATE "videoJobInfo" SET "pendingMove" = "videoJobInfo"."pendingMove" - 1, "updatedAt" = NOW() 
+    RETURNING "pendingMove"
+    `, options)
```
pingiun, line 81: This needs to be a custom query as sequelize-typescripts support for decrement queries is quite bad. In addition to [this issue](https://github.com/sequelize/sequelize/issues/12792) the library doesn't seem to generate the right types for static model functions.

server/models/video/video-job-info.ts
```diff
@@ -0,0 +1,85 @@
...
+  static async increaseOrCreatePendingMove (videoUUID: string): Promise<number> {
+    const options = { type: QueryTypes.SELECT as QueryTypes.SELECT, bind: { videoUUID } }
+
+    const [ { pendingMove } ] = await VideoJobInfoModel.sequelize.query<{pendingMove: number}>(`
+    INSERT INTO "videoJobInfo" ("videoUUID", "pendingMove", "createdAt", "updatedAt") 
+    VALUES ($videoUUID, 1, NOW(), NOW()) 
+    ON CONFLICT ("videoUUID") WHERE "videoUUID" = $videoUUID 
+    DO UPDATE SET "pendingMove" = "videoJobInfo"."pendingMove" + 1, "updatedAt" = NOW() 
+    RETURNING "pendingMove"
+    `, options)
```
pingiun, line 70: This is a custom query because I think it's important that this update happens "atomically".

kukhariev:
@pingiun this was tested with files larger than 5 gigabytes?

pingiun:
> @pingiun this was tested with files larger than 5 gigabytes?

Thanks for telling me about this limit, I didn't know before that the official SDK would not split up when sending large files.

I added basic support for this in https://github.com/Chocobozzz/PeerTube/pull/4290/commits/ea8917bb6c4210ec00c73d2bd2cc71d5b8fbc22a, but it might require some more thinking about aborting multipart requests when something went wrong so parts aren't left in the bucket. Another option would just to request the user to have a lifecycle rule setup in the bucket that deletes old object parts.

As can be seen in the commit I made a hacky addition to the ReadStream to make streaming to the API work. I also created a PR at the [official s3 sdk repo](https://github.com/aws/aws-sdk-js-v3/pull/2637) so that we might change this back in the future.

server/types/models/video/video.ts
Chocobozzz: Put it in `shared/models/videos/video-storage.type.ts`

server/controllers/api/videos/upload.ts
```diff
@@ -210,7 +216,12 @@ async function addVideo (options: {
 
   createTorrentFederate(video, videoFile)
     .then(() => {
-      if (video.state !== VideoState.TO_TRANSCODE) return
+      if (video.state !== VideoState.TO_TRANSCODE) {
```
Chocobozzz, line 219: We could create a `MOVE_TO_EXTERNAL_STORAGE` state and display the appropriate message to end users

server/controllers/api/videos/upload.ts
```diff
@@ -210,7 +216,12 @@ async function addVideo (options: {
...
+        // Video will be published before move is complete which may cause some video connections to drop
+        // But it's recommended to enable transcoding anyway, so this is the tradeoff
+        addMoveToObjectStorageJob(video, videoFile)
```
Chocobozzz, line 222: Could we rename this function to `addMoveToObjectStorageJobIfNeeded`? I know it's a long name but when we read the code we know we did not forget the `if (CONFIG.OBJECT_STORAGE.ENABLED)`  here

server/initializers/checker-after-init.ts
```diff
@@ -153,6 +153,24 @@ function checkConfig () {
     }
   }
 
+  // Object storage
+  if (CONFIG.OBJECT_STORAGE.ENABLED === true) {
+    if (CONFIG.TRANSCODING.WEBTORRENT.ENABLED && !CONFIG.OBJECT_STORAGE.VIDEOS.BUCKET_NAME) {
```
Chocobozzz, line 158: Transcoding can be updated at runtime so I think it would be simpler to force admins to fill all bucket names, even if they don't use webtorrent or hls videos

server/initializers/migrations/0660-object-storage.ts
```diff
@@ -0,0 +1,53 @@
...
+      "updatedAt" timestamp WITH time zone NOT NULL,
+      PRIMARY KEY ("id")
+    );
+    `
+
+    await utils.sequelize.query(query)
+  }
+
+  {
+    await utils.queryInterface.addColumn('videoFile', 'storage', { type: Sequelize.INTEGER, allowNull: false })
```
Chocobozzz, line 26: Does it work? By default the column is `null` but `allowNull` is `false`. I suggest to create the column with `allowNull: true`, update the rows and update the column with `allowNull: false`

server/initializers/migrations/0660-object-storage.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  {
+    await utils.sequelize.query(
+      `UPDATE "videoFile" SET "storage" = ${VideoStorageType.LOCAL}`
+    )
+  }
+
+  {
+    await utils.queryInterface.addColumn('videoStreamingPlaylist', 'storage', { type: Sequelize.INTEGER, allowNull: false })
```
Chocobozzz, line 36: Same here

server/lib/job-queue/handlers/move-to-object-storage.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  const video = await VideoModel.loadWithFiles(payload.videoUUID)
+  // No video, maybe deleted?
+  if (!video) {
+    logger.info('Can\'t process job %d, video does not exist.', job.id)
+    return undefined
+  }
+
+  if (CONFIG.TRANSCODING.WEBTORRENT.ENABLED && video.VideoFiles) {
+    await moveWebTorrentFiles(video, payload.videoFileId)
```
Chocobozzz, line 27: If transcoding is disabled we use webtorrent files by default so you need to call `moveWebTorrentFiles`

server/lib/job-queue/handlers/move-to-object-storage.ts
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+    playlist.playlistUrl = generateObjectStoreUrl(masterPlaylistFilename, CONFIG.OBJECT_STORAGE.STREAMING_PLAYLISTS)
+    playlist.segmentsSha256Url = generateObjectStoreUrl(segmentsFileName, CONFIG.OBJECT_STORAGE.STREAMING_PLAYLISTS)
+    playlist.storage = VideoStorageType.OBJECT_STORAGE
+    await playlist.save()
+  }
+
+  // Remove files that were "moved"
+  const tasks: Promise<any>[] = []
```
Chocobozzz, line 126: Why do you remove files in this function? I think we should remove them just after a `storeObject` success

server/lib/job-queue/handlers/video-transcoding.ts
```diff
@@ -135,7 +146,10 @@ async function onHlsPlaylistGeneration (video: MVideoFullLight, user: MUser, pay
     await createLowerResolutionsJobs(video, user, payload.resolution, payload.isPortraitMode, 'hls')
   }
 
-  return publishAndFederateIfNeeded(video)
+  // Publishing will be done by mvoe-to-object-storage if enabled
```
Chocobozzz, line 149: typo

server/lib/job-queue/handlers/video-transcoding.ts
```diff
@@ -168,23 +182,29 @@ async function onVideoFileOptimizer (
 
   const hasNewResolutions = await createLowerResolutionsJobs(videoDatabase, user, videoFileResolution, isPortraitMode, 'webtorrent')
 
-  if (!hasHls && !hasNewResolutions) {
-    // No transcoding to do, it's now published
-    videoPublished = await videoDatabase.publishIfNeededAndSave(undefined)
-  }
+  // Publishing will be done after the move-to-object-storage-job if enabled
+  if (!CONFIG.OBJECT_STORAGE.ENABLED) {
```
Chocobozzz, line 186: `if (!CONFIG.OBJECT_STORAGE.ENABLED) return` to reduce cyclomatic complexity

server/lib/job-queue/handlers/video-transcoding.ts
```diff
@@ -168,23 +182,29 @@ async function onVideoFileOptimizer (
...
+  }
 }
 
 async function onNewWebTorrentFileResolution (
   video: MVideoUUID,
   user: MUserId,
   payload: NewResolutionTranscodingPayload | MergeAudioTranscodingPayload
 ) {
-  await publishAndFederateIfNeeded(video)
+  // Publishing will be done by mvoe-to-object-storage if enabled
```
Chocobozzz, line 204: typo

server/lib/job-queue/job-queue.ts
```diff
@@ -154,6 +160,15 @@ class JobQueue {
       logger.error('Unknown queue %s: cannot create job.', obj.type)
       return
     }
+    if (obj.type === 'video-transcoding') {
```
Chocobozzz, line 163: Prefer to create a `createTranscodingJob` wrapper in `lib/videos` that increment the counter

server/lib/object-storage.ts
```diff
@@ -0,0 +1,153 @@
...
+  ListObjectsV2Command,
+  PutObjectCommand,
+  S3Client,
+  UploadPartCommand
+} from "@aws-sdk/client-s3"
+import { CONFIG } from "@server/initializers/config"
+import { logger } from '@server/helpers/logger'
+import { createReadStream, createWriteStream, ensureDir, open, close, ReadStream, stat, Stats } from "fs-extra"
+import { Readable } from "stream"
+import { pipeline } from "stream/promises"
```
Chocobozzz, line 17: `stream/promises` is not available in node 14. Use `pipelinePromise` from `server/helpers/core-utils` instead

server/lib/object-storage.ts
```diff
@@ -0,0 +1,153 @@
...
+import { dirname } from "path"
+import { min } from "lodash"
+
+type BucketInfo = {BUCKET_NAME: string, PREFIX?: string, BASE_URL?: string}
+const ONE_MIB = 1024 * 1024
+const PART_SIZE = 100 * ONE_MIB
+const MAX_PUT_SIZE = 100 * ONE_MIB
+
+function getS3Client () {
+  return new S3Client({ endpoint: `https://${CONFIG.OBJECT_STORAGE.ENDPOINT}` })
```
Chocobozzz, line 27: We'll mock s3 for integration tests (`http://` scheme then) so we should also support URL here (if does not start with `https?://` automatically add `https://` prefix)

server/lib/object-storage.ts
```diff
@@ -0,0 +1,153 @@
...
+async function objectStoragePut (options: {filename: string, content: string | ReadStream, bucketInfo: BucketInfo}) {
+  const { filename, content, bucketInfo } = options
+  const key = bucketInfo.PREFIX + filename
+  const s3Client = getS3Client()
+  const command = new PutObjectCommand({
+    Bucket: bucketInfo.BUCKET_NAME,
+    Key: key,
+    Body: content
+  })
+  return await s3Client.send(command)
```
Chocobozzz, line 39: `return s3Client.send(command)` since you don't `try/catch`

server/lib/object-storage.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+  const deleteCommand = new DeleteObjectsCommand(deleteParams)
+  await s3Client.send(deleteCommand)
+
+  // Repeat if not all objects could be listed at once (limit of 1000?)
+  if (listedObjects.IsTruncated) await removePrefix(prefix, bucketInfo)
+}
+
+export function generateObjectStoreUrl (filename: string, bucketInfo: BucketInfo) {
+  return `https://${bucketInfo.BUCKET_NAME}.${CONFIG.OBJECT_STORAGE.ENDPOINT}/${bucketInfo.PREFIX}${filename}`
```
Chocobozzz, line 138: We should also support `http://` (just take the same scheme than endpoint for example)

server/lib/video-paths.ts
```diff
@@ -30,6 +32,28 @@ function getVideoFilePath (videoOrPlaylist: MVideo | MStreamingPlaylistVideo, vi
   return join(baseDir, videoFile.filename)
 }
 
+async function getVideoFilePathMakeAvailable (
+  videoOrPlaylist: MVideo | MStreamingPlaylistVideo,
+  videoFile: MVideoFile
+) {
+  const path = getVideoFilePath(videoOrPlaylist, videoFile)
+  if (existsSync(path)) {
```
Chocobozzz, line 40: Don't use `Sync`, use `pathExists`

server/models/video/video-file.ts
```diff
@@ -450,7 +456,20 @@ export class VideoFileModel extends Model<Partial<AttributesOnly<VideoFileModel>
     return !!this.videoStreamingPlaylistId
   }
 
+  generateObjectUrl (video: MVideo) {
```
Chocobozzz, line 459: `buildObjectUrl`

server/models/video/video-job-info.ts
```diff
@@ -0,0 +1,85 @@
...
+  @AllowNull(false)
+  @Default(0)
+  @IsInt
+  @Column
+  pendingMove: number
+
+  @ForeignKey(() => VideoModel)
+  @IsUUID(4)
+  @Column(DataType.UUID)
+  videoUUID: string
```
Chocobozzz, line 50: For consistency, use `videoId` as foreign key

Chocobozzz:
> Do you have an idea at what places the file should be locally available?

I think you can put the files in https://github.com/Chocobozzz/PeerTube/blob/develop/config/production.yaml.example#L78

pingiun:
I'm not sure why the tests are failing, the same thing is happening when I run the tests locally.

> I think you can put the files in https://github.com/Chocobozzz/PeerTube/blob/develop/config/production.yaml.example#L78

I think you may have misunderstood, I meant which places in the code. So which parts in the code need to have the files stored locally, and need a change to support files that may not exist locally?

Chocobozzz:
> I think you may have misunderstood, I meant which places in the code.

Oh sorry, then I think everywhere we use `getVideoFilePath`.

> I'm not sure why the tests are failing, the same thing is happening when I run the tests locally.

The server build fails because of the `stream/promises` import: https://github.com/Chocobozzz/PeerTube/pull/4290/checks?check_run_id=3251005770#step:10:3480

pingiun:
> > I'm not sure why the tests are failing, the same thing is happening when I run the tests locally.
> 
> The server build fails because of the `stream/promises` import: https://github.com/Chocobozzz/PeerTube/pull/4290/checks?check_run_id=3251005770#step:10:3480

Thanks, I didn't catch that. I see that module was only added in Node 15.0 so I will try to remove the usage

server/controllers/api/videos/upload.ts
pingiun: We could also collapse the two things to "processing" if object storage is enabled. The distinction doesn't matter for watching the video, if it's not transcoded or moved you're not getting the optimal experience. In code it already means "processing" so I can rename the state in the enum

server/lib/job-queue/handlers/move-to-object-storage.ts
pingiun: I think the highest resolution transcode is used as input for the other transcodes, so that file is still needed even after it's uploaded to object storage. This function is only run after all the uploades are done so then we don't have that problem

pingiun: I changed this in aa6894eeb, but I'm not sure if it's gonna work due to my above comment

server/lib/object-storage.ts
```diff
@@ -0,0 +1,176 @@
...
+
+function getS3Client () {
+  return new S3Client({ endpoint: CONFIG.OBJECT_STORAGE.ENDPOINT.toString() })
+}
+
+function getPartSize (stats: Stats) {
+  if (process.env.NODE_ENV.includes("test")) {
+    return 10 * ONE_MIB
+  }
+  // Use parts of 1 GiB if the file is very large (it would take more than 1000 requests at 100 MiB per request)
```
kukhariev, line 33: I don't think this is needed. 
The part size should be as large as possible, but not causing timeout errors. It is better to leave the size at 2GB or add it to the config as `мultipartTriggerSize`, for example.

pingiun: What do you mean "leave the size at 2GB"? 

I could add a configuration option, but I think these are fine defaults. Backblaze b2 recommends 100MB in `recommendedPartSize` (see [docs](https://www.backblaze.com/b2/docs/large_files.html), so I think that's a good default (although I use 100MiB here, instead of MB). The only reason this function uses 1GiB sometimes is because otherwise it isn't possible to upload files larger than 1TB.

kukhariev: > We recommend a part size of 100MB, which strikes a good balance between upload throughput and the ability to **upload parts in parallel**.
No paralel parts uploading is used here, so a larger size is preferable for better performance with less pay.

pingiun: That's a good point

server/controllers/api/videos/upload.ts
Chocobozzz: Yes you're right but I think we should keep 2 different states so it's easier to debug things for admins if there have issues with a specific video

Chocobozzz:
I'll update this PR, thanks @pingiun

config/default.yaml
```diff
@@ -97,20 +97,33 @@ storage:
...
   endpoint: 's3.amazonaws.com'
+
+  region: 'us-east-1'
+
+  credentials:
+    access_key_id: 'access-key'
+    secret_access_key: 'secret-access-key'
+
+  # Maximum amount to upload in one request to object storage
+  max_upload_part: 2MB
```
kukhariev, line 111: Is it a typo? Part must be at least 5 MB in size

Chocobozzz: Oup's, should be 2GB

config/default.yaml
```diff
@@ -97,20 +97,33 @@ storage:
...
-  max_upload_part: 2048
-  # Will always use https with default URL generation (see below)
+
+  # Without protocol, will default to HTTPS
   endpoint: 's3.amazonaws.com'
+
+  region: 'us-east-1'
+
+  credentials:
+    access_key_id: 'access-key'
```
pingiun, line 107: I think it's a bad idea to encourage people to store secrets in a configuration file. The official s3 client supports loading from environment variables, which is more secure: https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-credentials-node.html

server/lib/object-storage/shared/object-storage-helpers.ts
```diff
@@ -0,0 +1,221 @@
...
+    logger.debug(
+      'Uploading part %d of file to %s%s in bucket %s',
+      partNumber, bucketInfo.PREFIX, objectStorageKey, bucketInfo.BUCKET_NAME, lTags()
+    )
+
+    // The s3 sdk needs to know the length of the http body beforehand, but doesn't support
+    // streams with start and end set, so it just tries to stat the file in stream.path.
+    // This fails for us because we only want to send part of the file. The stream type
+    // is modified so we can set the byteLength here, which s3 detects because array buffers
+    // have this field set
```
pingiun, line 183: Maybe add a note that this hack can be removed once this is released: https://github.com/aws/aws-sdk-js-v3/pull/2637

server/lib/video.ts
```diff
@@ -133,24 +160,16 @@ async function addOptimizeOrMergeAudioJob (video: MVideoUUID, videoFile: MVideoF
 }
 
 async function addTranscodingJob (payload: VideoTranscodingPayload, options: CreateJobOptions) {
-  // This value is decreased when the move job is finished in ./handlers/move-to-object-storage.ts
-  // Because every transcode job starts a move job for the transcoded file, the value will only reach
-  // 0 again when all transcode jobs are finished and the last move job is running
-  // If object storage support is not enabled all the pendingMove values stay at the amount of transcode
-  // jobs that were started for that video.
-  await VideoJobInfoModel.increaseOrCreatePendingMove(payload.videoUUID)
+  await VideoJobInfoModel.increaseOrCreate(payload.videoUUID, 'pendingTranscoding')
```
pingiun, line 109:
```suggestion
  await VideoJobInfoModel.increaseOrCreate(payload.videoUUID, 'pendingTranscode')
```

server/models/video/video-job-info.ts
```diff
@@ -42,6 +30,12 @@ export class VideoJobInfoModel extends Model<Partial<AttributesOnly<VideoJobInfo
   @Column
   pendingMove: number
 
+  @AllowNull(false)
+  @Default(0)
+  @IsInt
+  @Column
+  pendingTranscoding: number
```
pingiun, line 37:
```suggestion
  pendingTranscode: number
```

config/default.yaml
Chocobozzz: I'm not sure to understand why env variables are more secure since in the peertube context you would have to set them using a file to automatically load them no?

I agree we should find a way to set passwords for peertube in a more secure way (not only for object storage but also for smtp, redis, postgres...), but not sure what we could do.

I'll add a note about the env variable for object storage, I did not know this *auto load from env* thing

Chocobozzz:
PR is ready

@pingiun Could you open a MR for the documentation?

server/lib/object-storage/shared/object-storage-helpers.ts
```diff
@@ -0,0 +1,229 @@
...
+  const { objectStorageKey, inputPath, bucketInfo } = options
+
+  const key = buildKey(objectStorageKey, bucketInfo)
+  const s3Client = getClient()
+
+  const statResult = await stat(inputPath)
+
+  const createMultipartCommand = new CreateMultipartUploadCommand({
+    Bucket: bucketInfo.BUCKET_NAME,
+    Key: key
```
hugomd, line 171: We should be setting `ACL` here, as uploads are private by default for a lot of S3 providers.

Something like this:
```suggestion
    Key: key,
    ACL: "public-read"
```

I'm happy to follow this up in a separate PR. For now I'm just setting the bucket ACL to public.

Chocobozzz: You can review this PR and I'll resolve the conversations directly on develop :)

pingiun: I only tested with backblaze b2, this does indeed seem to be necessary for AWS S3

pingiun:
@Chocobozzz it appears through my testing that you changed something where local files are needed after deleting them. I now see the following in my logs:

```
[kubernetes.docker.internal:9000] 2021-08-17 17:21:46.663 info: Moving video 011c85e3-0c5e-459f-9f8b-4870b746ca41 in job 8.
[kubernetes.docker.internal:9000] 2021-08-17 17:21:48.332 info: Fetching HLS file hls_011c85e3-0c5e-459f-9f8b-4870b746ca41/7c6c9838-4d4f-4280-bf82-435b96d53465-360-fragmented.mp4 from object storage to /data/tmp/ecf8ea4b-217e-417f-9ed1-864c34dc47e1.mp4.
[kubernetes.docker.internal:9000] 2021-08-17 17:21:48.631 info: Creating torrent /data/torrents/e44ecb3d-b22f-4aad-a848-f4fa19cde296-360-hls.torrent.
[kubernetes.docker.internal:9000] 2021-08-17 17:21:48.639 info: Running cleanup after moving files to object storage (video 011c85e3-0c5e-459f-9f8b-4870b746ca41 in job 8)
[kubernetes.docker.internal:9000] 2021-08-17 17:21:48.842 info: Publishing video 011c85e3-0c5e-459f-9f8b-4870b746ca41.
```

So files are uploaded, deleted and then downloaded again. This wasn't happening in my code and doesn't seem very efficient (especially for large uploads)

pingiun:
Another thing that now doesn't work anymore is updating the endpoint url, video's will still be loaded from the old url because it's saved in the database. My code would always generate the URL from scratch (following your [suggestion](https://github.com/Chocobozzz/PeerTube/pull/4290#issuecomment-891670478))

Chocobozzz:
> Another thing that now doesn't work anymore is updating the endpoint url, video's will still be loaded from the old url because it's saved in the database.

In my comment I only talked about `base_url` and not `endpoint`. IMHO when the admin change the `endpoint`, they do not expect that peertube automatically consider that objects previously uploaded use the new endpoint. Because admins would need to migrate old objects, and so it's part of a migration process where peertube should be included. It's the same if admins disable object storage at some point, we should still serving old files from object storage.
However, the `base_url` should stay dynamic since a proxy that just cache files in front of a s3 storage could be easily changed

> So files are uploaded, deleted and then downloaded again. This wasn't happening in my code and doesn't seem very efficient (especially for large uploads)

Thanks, it should be fixed in https://github.com/Chocobozzz/PeerTube/commit/1f6125be8b6306ba34b5ad9df985df462ef9759c

@hugomd I fixed your comment in https://github.com/Chocobozzz/PeerTube/commit/0d4a3c62ab3036508ba1f38653580af80a4b631e

server/lib/object-storage/shared/object-storage-helpers.ts
hugomd: Reflecting on this a bit more, the ACL can likely be handled by the bucket policy.

We could follow up to specifically handle private buckets, pre-signing each URL before handing it off to the client.

hugomd:
> @hugomd I fixed your comment in [0d4a3c6](https://github.com/Chocobozzz/PeerTube/commit/0d4a3c62ab3036508ba1f38653580af80a4b631e)

Thanks for that, I'll pull it in and see how I go with a private bucket!

hugomd:
A nice extension to this could be pre-signing the upload URL and passing it to the client, so they can upload directly to object storage without needing to proxy via Peertube.

Peertube could wait until the upload's finished, and then pull it from object storage for transcoding.

pingiun:
> In my comment I only talked about `base_url` and not `endpoint`. IMHO when the admin change the `endpoint`, they do not expect that peertube automatically considers that objects previously uploaded use the new endpoint. Because admins would need to migrate old objects, and so it's part of a migration process where peertube should be included. It's the same if admins disable object storage at some point, we should still serve old files from object storage.
> However, the `base_url` should stay dynamic because a proxy that just cache files in front of a s3 storage could be easily changed.

Of course, this makes sense.

I created the MR for documentation here: https://framagit.org/framasoft/peertube/documentation/-/merge_requests/98

Chocobozzz:
Perfect, I'll also update the MR when I'll time. Thanks @pingiun

PR #4300, opened 2021-08-03
Standard output type syslog is obsolete in systemd

ROBERT-MCDOWELL:
## Description

removing the setting
StandardOutput
StandardError

## Related issues
Standard output type syslog is obsolete in systemd

## Has this been tested?
no, because this PR does not update server code


Chocobozzz:
Thanks

PR #4315, opened 2021-08-07
Allow configuration to be static/readonly

### LINKED ISSUE #3754: Please consider allowing immutable config

Running peertube in many situations, it is useful for the config to be immutable.  This means read-only.

For example, it's not possible to sensibly package peertube as a helm chart for kubernetes because configmaps are read-only to the OS.

The same would be true of running peertube in Heroku and OpenShift.

Simple: do not fail to run if the configuration is readonly.  This is a choice of the administrator of the system, and can be a security concern.  If you refuse to see the advantage of allowing immutable configuration by default then at least allow it to be turned from an error into a warning by a flag or environment variable.

The only solution at present is to copy the config to somewhere writable on startup which is a messy and error-prone solution.

pingiun:
## Description

Adds support for multiple configuration directories and adds multiple configuration directories in the Dockerfile to support the same default.yaml functionality as before. Also adds the `PEERTUBE_ALLOW_WEBADMIN_CONFIG` environment variable to `custom-environment-variables.yaml` so it can be easily set via Docker. This configuration option disables the admin view to change the configuration and makes it read only (see screenshot below).

This functionality is useful for people who do all configuration statically and mount it at /config in the container. This folder is now only used for user configuration because the files that were previously copied there now stay in the app directories.

## Related issues

Fixes #3754

## Has this been tested?

## Screenshots

image

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -1,5 +1,7 @@
 # /!\ YOU SHOULD NOT UPDATE THIS FILE, USE production.yaml instead /!\ #
 
+allow_webadmin_config: true
+
 listen:
   hostname: '127.0.0.1'
   port: 9000
--- a/client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.html
+++ b/client/src/app/+admin/config/edit-custom-config/edit-custom-config.component.html
@@ -63,7 +63,7 @@
     <div class="col-md-7 col-xl-5"></div>
     <div class="col-md-5 col-xl-5">
 
-      <div class="form-error submit-error" i18n *ngIf="!form.valid">
+      <div class="form-error submit-error" i18n *ngIf="!form.valid && serverConfig.allowEdits">
         There are errors in the form:
 
         <ul>
@@ -77,7 +77,11 @@
         You cannot allow live replay if you don't enable transcoding.
       </span>
 
-      <input (click)="formValidated()" type="submit" i18n-value value="Update configuration" [disabled]="!form.valid || !hasConsistentOptions()">
+      <span i18n *ngIf="!serverConfig.allowEdits">
+        You cannot change the server configuration because it's managed externally.
+      </span>
+
+      <input (click)="formValidated()" type="submit" i18n-value value="Update configuration" [disabled]="!form.valid || !hasConsistentOptions() || !serverConfig.allowEdits">
     </div>
   </div>
 </form>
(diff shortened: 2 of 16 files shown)


pingiun:
@iMartyn could you take a look at this to see if it solves your issue?

iMartyn:
> @iMartyn could you take a look at this to see if it solves your issue?

To me this looks excellent, and yes, by not writing to `/config`, it solves my issue.

pingiun:
I don't think the styling is exactly right yet, some elements are greyed out but others aren't. I'm not sure which scss files need to be changed to make it look better though.

pingiun:
I have improved styling somewhat in https://github.com/Chocobozzz/PeerTube/pull/4315/commits/7726bd0b7d8f32adb17d26e584556ad196804a22

Here is a screenshot:
image

config/default.yaml
```diff
@@ -1,5 +1,7 @@
 # /!\ YOU SHOULD NOT UPDATE THIS FILE, USE production.yaml instead /!\ #
 
+allow_webadmin_config: true
```
Chocobozzz, line 3: put it in below `peertube.check_latest_version` instead and add it to `production.yaml.example`

Prefer `webadmin.config.edition.allowed: boolean` instead. So we can (in the future) enable only some settings for example

pingiun: What do you think of:

```yaml
webadmin:
  configuration:
    allow_edits: true
```

or do you see the possibility for new fields in the future?


```yaml
webadmin:
  configuration:
    # i could this happening:
    viewing:
      allowed: true
    edit: 
      allowed: true
      # but what other fields can exist here?
```

client/src/sass/include/_mixins.scss
```diff
@@ -358,12 +358,15 @@
   font-size: 15px;
 
   &.disabled {
-    background-color: #E5E5E5;
+    background-color: #f9f9f9;
```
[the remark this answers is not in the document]
pingiun: I can't remember so I will change it back

config/default.yaml
Chocobozzz: Sorry for my late replay

I'd prefer the second option, just in case :)

pingiun: Alright I changed it

Chocobozzz:
Thanks, will be merged after 3.4 because I will need time to test the new container :)

Issue #4335, opened 2021-08-18
Button Contact Admin wording does dot include several admins and is weirdly translated in French


kimsible:
The wording of button to contact the instance admins is a little bit confusing : 

- The instance can be administrated by several admins as « ADMINISTRATORS & SUSTAINABILITY » shows.
- The French translation does not include a verb as a button refers to an action :

![image](https://user-images.githubusercontent.com/1877318/129925550-9f7892a1-13a5-4128-9ca1-73082fd3e8c2.png)

I propose something larger and generic : EN  « Contact us » - FR « Nous contacter »

Issue #4333, opened 2021-08-18
Alert user on publish/upload page if the video will be blacklisted by default


kimsible:
When clicking on Publish button, it would be nice to inform the user if the video will be blacklisted by default in the same way than an account with no quota : https://github.com/Chocobozzz/PeerTube/pull/2966

PR #4336, opened 2021-08-18
Alert user for low quota and video auto-block on upload page

kimsible:
## Description

This PR introduces:

1. Display alerts on the top of the upload page when the user has for its account
- quota left lower than 10%
- video review turned on 

2. Display `Contact us` button in alerts of upload page when contact form is enabled

3. Shorter text for the contact button of the about page:

`Contact us` instead of `Contact administrator` as it's more explicit for a call-to-action button, and an instance can be maintained by more than one admin
image

**For the french translation I suggest «`Nous contacter`» instead of «~`Contact de l'administrateur`~**».
image

## Related issues

Fixes https://github.com/Chocobozzz/PeerTube/issues/4335 and https://github.com/Chocobozzz/PeerTube/issues/4333

## Has this been tested?

- [x] :+1: yes, I edited the test suite

### Video review turned on with contact form disabled

auto-block-classic

### Video review turned on with contact form enabled

auto-block-contact

### Quota left under 10% with contact form disabled

quota-left-classic

### Quota left under 10% with contact form enabled

quota-left-contact

### No quota with contact form disabled

no-quota-classic

### No quota with contact form enabled

no-quota-contact

### Video review turned on + quota left under 10%

both-alert-classic

### DIFF
--- a/client/src/app/+about/about-instance/contact-admin-modal.component.html
+++ b/client/src/app/+about/about-instance/contact-admin-modal.component.html
@@ -1,6 +1,6 @@
 <ng-template #modal>
   <div class="modal-header">
-    <h1 i18n class="modal-title">Contact instance administrators</h1>
+    <h1 i18n class="modal-title">Contact the administrator(s)<p class="modal-subtitle">{{ instanceName }}</p></h1>
     <my-global-icon iconName="cross" aria-label="Close" tabindex="0" role="button" (click)="hide()" (keydown.enter)="hide()"></my-global-icon>
   </div>
 
--- a/client/src/app/+about/about-instance/contact-admin-modal.component.html
+++ b/client/src/app/+about/about-instance/contact-admin-modal.component.html
@@ -1,6 +1,6 @@
 <ng-template #modal>
   <div class="modal-header">
-    <h1 i18n class="modal-title">Contact {{ instanceName }} administrator</h1>
+    <h1 i18n class="modal-title">Contact instance administrators</h1>
     <my-global-icon iconName="cross" aria-label="Close" tabindex="0" role="button" (click)="hide()" (keydown.enter)="hide()"></my-global-icon>
   </div>
 
--- a/client/src/app/+videos/+video-edit/video-add.component.html
+++ b/client/src/app/+videos/+video-edit/video-add.component.html
@@ -1,10 +1,18 @@
-<div *ngIf="user.isUploadDisabled()" class="no-upload">
+<div *ngIf="user.isUploadDisabled()" class="upload-messages upload-disabled">
   <div class="alert alert-warning">
     <div i18n>Sorry, the upload feature is disabled for your account. If you want to add videos, an admin must unlock your quota.</div>
     <a i18n routerLink="/about/instance" *ngIf="!isContactFormEnabled" class="about-link">Read instance rules for help</a>
     <a i18n routerLink="/about/contact" *ngIf="isContactFormEnabled" class="contact-link">Contact us</a>
   </div>
   <img src="/client/assets/images/mascot/defeated.svg" alt="defeated mascot">
+</div>
+
+<div *ngIf="!user.isUploadDisabled() && user.isAutoBlocked()" class="upload-messages auto-blocked">
+  <div class="alert alert-warning">
+    <div i18n>Uploaded videos are reviewed before publishing for your account. If you want to add videos without moderation review, an admin must turn off your videos auto-block.</div>
+    <a i18n routerLink="/about/instance" *ngIf="!isContactFormEnabled" class="about-link">Read instance rules for help</a>
+    <a i18n routerLink="/about/contact" *ngIf="isContactFormEnabled" class="contact-link">Contact us</a>
+  </div>
 </div>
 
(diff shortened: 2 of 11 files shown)


ranjithrajv:
Does this provide easy customization for admins? Like if an admin wants to ask donations if a user wants to increase quota.

kimsible:
> Does this provide easy customization for admins? Like if an admin wants to ask donations if a user wants to increase quota.

No, it's really basic and works with available settings. 

@Chocobozzz, do you see easy customization for admins more as a plugin (so a filter / hook) or a core feature ?

client/src/app/+about/about-instance/contact-admin-modal.component.html
```diff
@@ -1,6 +1,6 @@
 <ng-template #modal>
   <div class="modal-header">
-    <h1 i18n class="modal-title">Contact {{ instanceName }} administrator</h1>
+    <h1 i18n class="modal-title">Contact instance administrators</h1>
```
Chocobozzz, line 3: Why? Is the title too long with the instance name?

client/src/app/+videos/+video-edit/video-add.component.html
```diff
@@ -1,18 +1,36 @@
-<div *ngIf="user.isUploadDisabled()" class="no-upload">
-  <div class="alert alert-warning">
+<ng-template [ngIf]="user.isUploadDisabled()">
```
Chocobozzz, line 1: Prefer using `<ng-container>` instead of `<ng-template>`

client/src/app/+videos/+video-edit/video-add.component.html
```diff
@@ -1,18 +1,36 @@
...
-  <div class="alert alert-warning" *ngIf="isRootUser()" i18n>
+  <div class="upload-image">
+    <img src="/client/assets/images/mascot/defeated.svg" alt="defeated mascot">
+  </div>
+</ng-template>
+
+<ng-template [ngIf]="!user.isUploadDisabled()">
+  <div *ngIf="user.isAutoBlocked()" class="upload-message auto-blocked alert alert-warning">
+    <div i18n>Uploaded videos are reviewed before publishing for your account. If you want to add videos without moderation review, an admin must turn off your videos auto-block.</div>
+    <a i18n routerLink="/about/instance" *ngIf="!user.hasNoQuotaLeft() && !isContactFormEnabled" class="about-link">Read instance rules for help</a>
```
Chocobozzz, line 16: Use `<ng-template>` to reuse HTML code: https://stackoverflow.com/a/56510232

client/src/app/core/users/user.model.ts
```diff
@@ -133,4 +133,19 @@ export class User implements UserServerModel {
   isUploadDisabled () {
     return this.videoQuota === 0 || this.videoQuotaDaily === 0
   }
+
+  isAutoBlocked () {
+    return this.role === UserRole.USER && this.adminFlags !== UserAdminFlag.BYPASS_VIDEO_AUTO_BLACKLIST
+  }
+
+  hasNoQuotaLeft () {
```
Chocobozzz, line 141: What about daily quota?

Chocobozzz:
NB: the client test failure is a weird bug in the CI

client/src/app/+about/about-instance/contact-admin-modal.component.html
kimsible: Sorry, I forgot to explain this part. It was indeed to get a shorter title, and to keep the information that you contact the administrator(s) of the instance; the action button cannot tell precisely who you contact, I thought the modal title was intended to give more details without being too long.

Do you think we should keep the instance name and find something else like put `contact the administrator(s)` in a sub-title?

client/src/app/core/users/user.model.ts
kimsible: In a moment I thought it was useless for the daily quota because it would be reset in the next 24 hours. 
You're right, I'm gonna add an helper and alert for the the daily quota too.

kimsible:
> NB: the client test failure is a weird bug in the CI

I will explore it more today, I didn't understand either this test failure, maybe it's caused by the `adminFlags`

Chocobozzz:
> I will explore it more today, I didn't understand either this test failure, maybe it's caused by the adminFlags

No, it's a bug from develop

client/src/app/+about/about-instance/contact-admin-modal.component.html
Chocobozzz: > Do you think we should keep the instance name for example in a sub-title and name the main title contact the administrator(s)?

:+1:

Chocobozzz:
> do you see easy customization for admins more as a plugin (so a filter / hook) or a core feature ?

A filter hook I thinj

kimsible:
Screenshot 2021-08-24 at 16-06-13 Login - PeerTube
Screenshot 2021-08-24 at 16-45-50 Contact - PeerTube

@Chocobozzz I suggest for french translation of `Contact the administrator(s)` => « `Contact des administrateur·ice(s)` » (`des` cannot really be inclusive)

Issue #3953, opened 2021-04-11
Translation strings longer than source breaks action button in federation admin


kimsible:
Since PeerTube 3.1, the french translation breaks the action buttons on the left in admin federation : 

![image](https://user-images.githubusercontent.com/1877318/114323941-da187d80-9b27-11eb-96e7-bdc8f00b3685.png)

PR #4378, opened 2021-09-06
Fix broken delete buttons of admin federation lists

kimsible:
## Description
Fixes broken my-delete buttons into admin federation lists depending on language.
Use  the generic label «Remove » on following list to keep consistency with follower list.

## Related issues

Resolves https://github.com/Chocobozzz/PeerTube/issues/3953

## Has this been tested?

## Screenshots

Screenshot 2021-09-06 at 14-26-55 Suivant - PeerTube

### DIFF
--- a/client/src/app/+admin/follows/followers-list/followers-list.component.scss
+++ b/client/src/app/+admin/follows/followers-list/followers-list.component.scss
@@ -23,8 +23,5 @@ a {
 }
 
 my-delete-button {
-  ::ng-deep .button-label {
-    @include ellipsis;
-    width: 70px;
-  }
+  max-width: 130px;
 }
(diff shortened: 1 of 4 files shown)


client/src/app/+admin/follows/followers-list/followers-list.component.scss
```diff
@@ -21,3 +21,10 @@ a {
     @include margin-right(10px);
   }
 }
+
+my-delete-button {
+  ::ng-deep .button-label {
```
Chocobozzz, line 26: I think you should set a width on the button, that automatically uses ellipsis and does not take more width that defined by the parent. It would prevent `::ng-deep` use

kimsible: thanks for the tip !

Issue #4379, opened 2021-09-06
Undefined this.user.videoChannels on production build


kimsible:
**Describe the current behavior**

**Steps to reproduce:**

1. Build for production and start PeerTube
2. Go to /videos/upload or /my-library/video-channels
3. `this.user.videoChannels` appears to be undefined in the console

**Describe the expected behavior**

 `this.user.videoChannels`  is used in `hasChannelNotConfigured` before all user information loaded.
![error-videoChannels](https://user-images.githubusercontent.com/1877318/132233121-a962dd1b-b6de-4368-87f0-cc2a2fda4586.png)

**Additional information**

* PeerTube v3.4.0-rc1

PR #4380, opened 2021-09-06
Fix undefined this.user.videoChannels on production build

kimsible:
## Description

Check if `this.user.videoChannels` not undefined in the getter `hasChannelNotConfigured` to avoid errors until user information fully loaded.

## Related issues

Resolves https://github.com/Chocobozzz/PeerTube/issues/4379

## Has this been tested?

### DIFF
--- a/client/src/app/shared/shared-main/misc/channels-setup-message.component.ts
+++ b/client/src/app/shared/shared-main/misc/channels-setup-message.component.ts
@@ -16,14 +16,10 @@ export class ChannelsSetupMessageComponent implements OnInit {
     private authService: AuthService
   ) {}
 
-  get userInformationLoaded () {
-    return this.authService.userInformationLoaded
-  }
-
   get hasChannelNotConfigured () {
-    return this.user.videoChannels
-      .filter((channel: VideoChannel) => (!channel.avatar || !channel.description))
-      .length > 0
+    if (!this.user.videoChannels) return
+
+    return this.user.videoChannels.filter((channel: VideoChannel) => (!channel.avatar || !channel.description)).length > 0
   }
 
   ngOnInit () {


client/src/app/shared/shared-main/misc/channels-setup-message.component.ts
```diff
@@ -16,14 +16,11 @@ export class ChannelsSetupMessageComponent implements OnInit {
...
 
-  get userInformationLoaded () {
-    return this.authService.userInformationLoaded
-  }
-
   get hasChannelNotConfigured () {
-    return this.user.videoChannels
-      .filter((channel: VideoChannel) => (!channel.avatar || !channel.description))
-      .length > 0
+    if (this.user.videoChannels)
```
Chocobozzz, line 20: Prefer `if (!this.user.videoChannels) return` form

Chocobozzz (commented):
Please fix lint

client/src/app/shared/shared-main/misc/channels-setup-message.component.ts
kimsible: it's weird with this form the alter does not show anymore

kimsible: Thanks for the tip :+1:

client/src/app/shared/shared-main/misc/channels-setup-message.component.ts
```diff
@@ -16,15 +16,13 @@ export class ChannelsSetupMessageComponent implements OnInit {
     private authService: AuthService
   ) {}
 
-  get userInformationLoaded () {
-    return this.authService.userInformationLoaded
-  }
-
   get hasChannelNotConfigured () {
+    if (!this.user.videoChannels) return
```
Chocobozzz, line 20: Sorry, I think it should be `return false`

kimsible: I've re-tried, it seems to work, it's just very long process because I need to re-build for each changes with that bug.

Issue #4108, opened 2021-05-19
Customize default playlist image


kontrollanten:
**Describe the problem to be solved**
The current default playlist image is gray with a PeerTube logo. We'd like to have a custom image.

![image](https://user-images.githubusercontent.com/6680299/118820383-829fc700-b8b6-11eb-82c8-71a002117bdd.png)

**Describe the solution you would like:**
Possibility to change the default playlist image.

**Describe alternatives you have considered**  
Add the playlist image to client-overrides.

Alterak:
After all, you can change the playlist thumbnail yourself in the playlist settings. Also, the thumbnail will change when you add at least one video to this playlist.

PR #4392, opened 2021-09-09
extend client-overrides with default images

kontrollanten:
## Description
Add:
* default-playlist.png
* default-avatar-account.png
* default-avatar-videochannel.png

## Related issues

solves #4108


ROBERT-MCDOWELL:
is this PR allow changing the live background too?

Issue #4430, opened 2021-09-24
youtube-dl not working on Gentoo


SimplyCorbett:
Recently swapped to Gentoo because I like to tinker. Transferred the installation successfully and everything works other then youtube-dl.

I'm assuming this is because like on debian 10 you need python-is-python3 or python-is-python2. These packages don't exist in Gentoo.

Can you update the install instructions to include the required packages under the Gentoo area in any os? Thank you!

rigelk:
Hi @SimplyCorbett,

None of us works with Gentoo ; this is mainly a part maintained by the community, so unless someone steps up to correct it, it will stay as is (or be deleted).

Please also use the documentation template.

SimplyCorbett:
Hi @rigelk 

I understand. However I feel that this is a bug that should be fixed regardless. youtube-dl should work out of the box without a workaround solution like python-is-python2.

As far as gentoo goes peertube is fully functional on it minus the above issue, so removing all support from the manual/documentation seems premature.

rigelk:
You're welcome to PR a doc fix at https://framagit.org/framasoft/peertube/documentation

> youtube-dl should work out of the box without a workaround solution like python-is-python2.

Isn't it a question better fit for the youtube-dl community?

frankstrater:
Hi @SimplyCorbett 

I'll update the dependencies guide for Gentoo soon. For all the distros that don't have the `python-is-python2` or `python-is-python3` package in their respective repositories, you can 'fix' this by doing:

`sudo ln -s /usr/bin/python2 /usr/bin/python`

or

`sudo ln -s /usr/bin/python3 /usr/bin/python`

depending on the Python version you have installed.

frankstrater:
@rigelk

Which part of the dependencies guide does the PeerTube team maintain? Just the Debian/Ubuntu one?

Maybe it's a good idea that it's mentioned that the dependencies guide is maintained by the community (some note in the intro of that page). I've seen a lot of frustration in the community about the sometimes outdated dependencies guide and you are constantly blamed for it.

rigelk:
> Which part of the dependencies guide does the PeerTube team maintain? Just the Debian/Ubuntu one?

Pretty much, yes.

> Maybe it's a good idea that it's mentioned that the dependencies guide is maintained by the community

Maybe. Invoking @SansPseudoFix there.

frankstrater:
[...]

PR #4436, opened 2021-09-27
Add Python symlink step for Gentoo

frankstrater:
## Description

Extra step for Gentoo to make youtube-dl function properly after first install.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/4430

## Has this been tested?

- [x] 🙋 no, because I need help (someone needs to confirm this works on an actual install of Gentoo)

### DIFF
--- a/support/doc/dependencies.md
+++ b/support/doc/dependencies.md
@@ -554,6 +554,12 @@ rc-service redis start
 rc-service postgresql-11 start
 ```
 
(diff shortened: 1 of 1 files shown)


frankstrater:
Confirmed by https://github.com/Chocobozzz/PeerTube/issues/4430#issuecomment-928177278

PR #4477, opened 2021-10-21
Keep it snimple snupple

comradekingu:
Wanted to translate this since one of the PeerTube Live strings links to its URL (which can be localized).
Hoping this alternative to all of which I don't find agreeable can be used.

---

To keep it short, the former document is authoritarian because it changes power-structure, and removes humanity, dialogue and agency where it shouldn't.
It hinges on the idea of a "harassment-free experience", which does sounds good, but either is impossible, or incompatible with anything that resembles a (human) community. This is further emphasized in its appeal to professionalism as commercial corporatism.

In the end, nobody is going to be thwarted from being anti-social because it means also disobeying a text-document. The structure of required secrecy and escalation of conflict really attracts that sort of character though. Meanwhile people most vulnerable to attack are left believing that won't happen because of how easy it is to misinterpret what it says, and fail to understand what it does.

Anything that can and should be done can be done ultimately by maintainers, and I argue should.

### DIFF
--- a/CONTACT.md
+++ b/CONTACT.md
@@ -1,0 +1,5 @@
+Welcome to the PeerTube community
+
+Give and grant critique observing desired privacy.
+Settle disputes within these confines.
+Finding yourself unable, e-mail the maintainer at chocobozzz@framasoft.org.


comradekingu:
@rigelk I take it this was not put in place to get the green light then?
I'm just a contributor that didn't pledge allegiance to it, and don't pull any strings here, so I am just at a loss as to why.
I really want to avoid the types of scenarios the likes of it has caused in any of the projects I contribute to,
and so far I have removed a lot of these to great effect.

That doesn't mean I haven't missed something, so I am always eager to hear why it is.
Also happy to state my reasons, which along with the CONTACT.md I write up every time.
What parts of the former text do you remember or find agreeable?

@Chocobozzz What requests would you answer differently if they stemmed from the new document?

comradekingu:
1. is an assumption. It is historically unsound, and practically testable in every case it is dropped.
A leadership decision is different from a community decision.
2. I never said "unenforceable" I said "authoritarian", with a summary, and then some reasons. I'll happily elaborate further.
3. I presented an argument with my change. I can't put an argument in the actual PR, and I can't discuss with myself, so it isn't without discussion other than being closed without one.
4. "Community leader" is a term from 2.0, 1.4 has "project maintainers". It is no more accepted than its authority in being authoritative. Its position as such is a problem though.
5. How does retroactively claiming someone pledges allegiance carry weight?

If you purport "helps to protect the community", what do you mean, and why isn't it subject to my critique or being secondary to the anti-social nature of the behavior outweighing the moral impetus to also disobey a text-file, with false confidence arising from it?

My action is to put up a simple file that doesn't assume anything, one which specifically doesn't shift the power structure into something anti-meritocratic, requiring secrecy, and wielding ultimate power for what really is "I don't agree".
To the contrary, my goal is to maximize communication between rational actors to settle disputes, instead of escalating conflict (authority to an authoritarian process that removes it in full).

I still don't know if you are arguing my case and somehow think I tried to remove the new text and put in the old, but there is nothing democratic about the old text being there, or its nature.
I can't make sense of "evidenced by work on other projects". Are you presenting my case, or are you saying I am doing something nefarious?

I didn't merge this without community discussion, I invited it.
As these often are put up to have green lights, I don't mind it falling to similar effort (as is also often the case).
I am not even a developer, and I want you to object if there is anything objectionable in the change.
I really thought long and hard about making it so, and (again) "That doesn't mean I haven't missed something".

>Give and grant critique observing desired privacy.

I am not forcing a tirade of arguments on anyone, because that is at the discretion of the reader and those trying to follow along.
I know full well discussing the minutia to great detail isn't very productive, and that is another CoC problem.

>Settle disputes within these confines.

To make sure inevitable conflicts are settled amicably between involved parties, try again to keep it at critique, and be mindful of peoples privacy.

Then there is a presentation of the position of the person you are contacting (again no name, because it isn't in the profile though I would argue that is nice),

The title is a play on "keep it simple stupid", or "keep it stupid simple", which is a good concept, but sounds a bit harsh.

JigmeDatse:
@comradekingu 

> @JigmeDatse
> I think maybe you have what I intended to replace mistaken for what I actually sent?

Nice plausible deniability there.  While I attributed a "reason" to your actions, most of what I talked of was your actions.  I also referenced behaviours you have demonstrated in other repositories.  Where you have decided that you had some issue to handle...  But when the maintainers have decided that they didn't like everything in your PR, you have doubled down.

What you did, was take a purposely progressive Code of Conduct, and replaced it with an purposely regressive Code of Conduct.  Whether you did so *intentionally* to bring in the regressive attitude, you can deny.  What you can't deny is that you are aware that is what your PR is accomplishing.

PR #4481, opened 2021-10-22
Add migrate-to-object-storage script

### LINKED ISSUE #4467: Migrate videos to native object storage
If you've previously used s3fs-fuse and are migrating to PeerTubes object storage you still need to use s3fs-fuse to be able to read the previous videos. That means that newly uploaded file will be double synced, first via s3fs-fuse and then moved by PeerTube to the object storage. I first thought that we could just change the streaming playlist path in the PeerTube config, but that will break old videos. Are there any thought out plan here?

I'd propose that files uploaded with s3fs-fuse will be uploaded to the PeerTube tmp folder instead.

PS. I've read the docs about how to migrate and couldn't find any help in this case. https://docs.joinpeertube.org/admin-remote-storage?id=object-storage-migration

kontrollanten:
## Description
Add script to migrate videos to object storage. The script will iterate over all videos that are using file storage and create a `move-to-object-storage` job.

## Related issues
closes #4467

## Has this been tested?

It's tested and verified manually, but it fails on some videos since the playlist filename returned by `getHlsResolutionPlaylistFilename` (server/libs/paths.ts) is wrong in some cases (when playlist filename is `[resolution].m3u8` and videoFilename is `[uuid]-[resolution]-fragmented.mp4`.


kontrollanten:
I've added a script that renames the playlist files and updates the DB. The issue is that the master m3u8 and segments file aren't moved to the object storage, even though the jobs runs successfully.

Chocobozzz:
I'll update this PR

Chocobozzz:
In the added commit:
 * I moved the migration script to `migrations/peertube-4.0.ts`
 * The job queue is in "produce only" mode
 * I tried to improve script robustness (if `move` fail we still continue, put each independent section in a `try/catch`)
 * It should be more efficient since we don't need to `updateSha256VODSegments`, and I used another way to update the master playlist content (to not use ffprobe that could be long for big instances)
 * I renamed the move to object storage script to `create-move-video-storage-job.ts` so we can support other storage type in the future (moving the file on the filesystem for example)
 * `create-move-video-storage-job.ts` can move a specific video or all videos
 * Added tests & doc

@kontrollanten could you review, tell me if these changes are okay and test the scripts?

server/lib/video-state.ts
```diff
@@ -61,10 +57,29 @@ function moveToNextState (video: MVideoUUID, isNewVideo = true) {
   })
 }
 
+async function moveToExternalStorageState (video: MVideoFullLight, isNewVideo: boolean, transaction: Transaction) {
+  const videoJobInfo = await VideoJobInfoModel.load(video.id, transaction)
+  const pendingTranscode = videoJobInfo?.pendingTranscode || 0
+
+  // We want to wait all transcoding jobs before moving the video on an external storage
+  if (pendingTranscode !== 0) return
```
kontrollanten, line 65: Does this mean that if we start the job when there's a pending transcode no migration will be done? Should we log a message to the user about that?

server/lib/job-queue/job-queue.ts
```diff
@@ -124,6 +124,12 @@ class JobQueue {
 
     for (const handlerName of (Object.keys(handlers) as JobType[])) {
       const queue = new Bull(handlerName, queueOptions)
+
+      if (produceOnly) {
+        queue.pause(true)
```
kontrollanten, line 129: Don't we need to resume it somewhere as well?

support/doc/tools.md
```diff
@@ -303,6 +303,33 @@ $ cd /var/www/peertube-docker
 $ docker-compose exec -u peertube peertube npm run create-import-video-file-job -- -v [videoUUID] -i [videoFile]
 ```
 
+### create-move-video-storage-job.js
+
+Use this script to move a specific video file in object storage.
```
kontrollanten, line 308: Use this script to move all video files, or a specific video file, to object storage.

server/lib/job-queue/job-queue.ts
Chocobozzz: No because this instance of the job queue for the current process is just to produce new tasks in the job queue. 

(we use the `local` boolean https://github.com/OptimalBits/bull/blob/develop/lib/queue.js#L805)

(I may have misunderstood your comment)

scripts/migrations/peertube-4.0.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  if (playlist.playlistFilename === 'master.m3u8') {
+    try {
+      const newName = generateHLSMasterPlaylistFilename(video.isLive)
+
+      const dst = join(hlsDirPath, newName)
+      await move(join(hlsDirPath, playlist.playlistFilename), dst)
+      playlist.playlistFilename = newName
+    } catch (err) {
+      console.error(`Cannot rename ${video.name} master.m3u8 file to a new name`, err)
```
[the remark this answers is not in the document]
Chocobozzz: I removed your restore function to reduce script complexity because in v4.0 we consider all playlist files have the new name and thus if a playlist has `master.m3u8` it is considered as broken. Then, restoring a broken video in a broken state is not really useful. I prefer admins to manually fix these specific videos. 
Moreover, I think that most of the time if this step fails it means the source file does not exist (and so the video is already broken)

PR #4525, opened 2021-11-08
Add video transcoding_failed state

### LINKED ISSUE #4500: Video state is still "to transcode" when transcoding job failed
When a video-transcode job has failed, the video state is still in "to transcode".

**Steps to reproduce:**

1. Upload a video and wait till transcoding jobs are created.
2. Break the jobs (by deleting the video file, restaring the server, or something else)
3. Check the video and see its state.

When a transcoding fails, it should be displayed for the publisher so it doesn't wait for the transcoding. This could be solved by adding a "transcoding failed" state.

* PeerTube instance:
  * version: 3.4.1

kontrollanten:
## Description
When a job fails to transcode a video, move it to transcoding failed state.

## Related issues
closes #4500

## Has this been tested?

I couldn't find any way to stub or simulate a way to make a transcoding job to fail.


server/lib/video-state.ts
```diff
@@ -57,10 +57,17 @@ function moveToNextState (video: MVideoUUID, isNewVideo = true) {
   })
 }
 
+function moveToFailedState (video: MVideoFullLight) {
```
Chocobozzz, line 60: `moveToFailedTranscodingState`

Chocobozzz (commented):
Please add this state in https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/shared/shared-video-miniature/video-miniature.component.ts#L166 and https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/+videos/+video-watch/shared/information/video-alert.component.html

> I couldn't find any way to stub or simulate a way to make a transcoding job to fail.

I think we can skip tests here

PR #4528, opened 2021-11-09
Trigger rebuild of Gitpod's image for PeerTube

Poslovitch:
## Description

Gitpod's docker image for PeerTube hasn't been rebuilt for some time, and started to have compatibility issues between dependencies and the node engine. Triggering this rebuild should force Gitpod to rebuild the image with node v16.13.0 instead of node v12.20.0.

## Related issues

None.

## Has this been tested?

But in fact, yes : I did trigger a rebuild of the image on my fork, and it's working great!


Chocobozzz:
Thanks!

Issue #4201, opened 2021-06-21
Expose Typescript type definition


lutangar:
**Describe the problem to be solved**

In order to develop a __PeerTube__ plugin in Typescript, I should be able to import __PeerTube__ plugin API type definitions.

**Describe the solution you would like:**

Writing a plugin could look like the following :
```ts
import { RegisterServerOptions } from '@peertube/plugin';

export async function register({ registerSetting }: RegisterServerOptions) {
  registerSetting({
    // ...
  });
}

export async function unregister() {
  return;
}
```

Publishing to [https://github.com/DefinitelyTyped/DefinitelyTyped](DefinitelyTyped) is another option:
```ts
import { RegisterServerOptions } from '@types/peertube';
```
Or maybe:
```ts
import { RegisterServerOptions } from '@types/peertube-plugin';
```

I don't know if it makes much sense to publish either part of `client` or `server` as a npm package.
But if it does I would consider publishing plugin typedef with these.
For ex:
- `@peertube/client`
- `@peertube/server`

> Just so you know there is an npm package named `peertube` with no connection with this project:
> https://www.npmjs.com/package/peertube

**Describe alternatives you have considered**  

If it doesn't make sense the alternatives would be to either create another `package.json` in `./server/types/plugins` for example.
Or maybe create a dedicated directory in this repository and publish it to https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package

As for the best name of such package, I'd love to hear other opinions, because I'm still unsure as to weather it would make sense to expose typedef other than the plugin ones.

JohnXLivingston:
My plugin https://github.com/JohnXLivingston/peertube-plugin-livechat is completely written in Typescript (backend and frontend).

I had discussions with @Chocobozzz  on how we can have type files. Our conclusions was something like: I will write a npm package that brings Peertube's type by manually copying them (we found no easy way to export them automatically).
But I did not manage to do that. Especially because I have no clue on how to properly handle types depending on peertube's versions. First idea was to synchronise types version with peertube version. But in my plugin I want to be compatible with multiple versions of peertube...
So I gave up.

What I did instead: I wrote my own type files in my plugin : 
https://github.com/JohnXLivingston/peertube-plugin-livechat/blob/main/client/%40types/peertube.d.ts
https://github.com/JohnXLivingston/peertube-plugin-livechat/blob/main/server/%40types/peertube.d.ts

[...]

PR #4531, opened 2021-11-09
Generate & Publish PeerTube types package

lutangar:
## Description

Generate a `@peertube/peertube-types` package exposing __PeerTube__ Typescript  definitions files to the world. :earth_africa: 

Today, main usage is being able to write __PeerTube__ plugins in __Typescript__.

```
yarn add --dev @peertube/peertube-types
```

> Tomorow, the approach(es) described below, might help to uses -- at least -- some part of __PeerTube__ as libraries.

## Related issues

Initially discussed in #4201

## Background

I spent quite some finding the best approach to address this issue:
1. Publish a unique package `@peertube/peertube` packaging the __client__ and __server__ as __librairies__ exposing their types.
    - exposing them as librairies would be awesome, but requires deep knowledge of the domain and would be time consuming
    - it requires ahaving a single `package.json` for the whole project (see how [`nx`](https://nx.dev/) is working). I think it would be a natural fit for the __PeerTube__ codebase *plus many benefits*. The only tradeoff: you can't install 2 versions of the same library. Which  , after giving it some toughts, rather sounds like an advantage. As an example at the moment __client__ and __server__ are using a different version of __Typescript__... :crying_cat_face: 

2. Publish many packages `@peertube/client`, `@peertube/server`, etc... all as librairies which would also exposes their typings.

> This is kind of a [lerna](https://github.com/lerna/lerna) approach with a `package.json` in each projects.

Both approches had the main drawback to being tightly coupled to the current folder structure (which I didn't too mess too much).

The initial goal was to offer simple imports, for example:
```
import { Register } from '@peertube/server/plugins'
```

But without moving too far from the original organisation, it would have left us with things like this:
```
import { Register } from '@peertube/server/dist/server/types/plugins'
```

Another drawback was, at some point, I had to tweak a `package.json` before publication.
For example:
- changing the name to something `@peertube/server-types` to publish a types only package
- remove some dependencies which weren't required by the types

## Proposal

Since I was required to do some heavy `package.json` lifting anyway I ended up created a __types__ *only* package `@peertube/peertube-types` which could bring the best of both world:

It bring the best Developer eXperience without *too many* impacts on the current project organisation -- apart from the circular dependency refacto thing.

This approach consists in crafting a package from scratch using various inference mechanisms. 
> For example by guessing the required package `dependencies` by using [depcheck](https://github.com/depcheck/depcheck) programatically.
> This might look a bit crazy but it turns out to work pretty well.

> See commit https://github.com/Chocobozzz/PeerTube/pull/4531/commits/ad23baea71a8617bdc19cbc38794ee1a834ee00c to review the diff of the packaging part alone.

### Project references
This implementation propose to take advantage of the typescript [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) feature.

This feature epxlicit dependencies between each typescript *project* :
- `scripts`
- `server`
- `shared`
- `client`

But this also strictly forbid circular references.
Before, `shared` had a dependency on `server` (have a look at previous `imports`) but `server` depended on `shared`; 
hence there was a circular dependency which this PR addresses.

> You might want to review the diff of this part alone see:
> https://github.com/Chocobozzz/PeerTube/pull/4531/commits/0511e80d5dd7fd04ce115324886af8909f082d44

Main benefit is being able to only recompile the modified project and to recompile dependents project.

Unfortunately, __project references__ aren't yet available in the Angular build process, see:
https://github.com/angular/angular/issues/37276

But this is actually not a problem here since we don't need to run trough `ng build` to get the `*.d.ts` files.

### Evolutions

I noticed many little things which could be improved on the overall project organization which I would I live to discuss around a beer :beer: sometimes!

> On the top of my head, I'm thinking about the `root` function and the `register-path` thing:
> https://github.com/Chocobozzz/PeerTube/blob/develop/server/helpers/core-utils.ts#L164
> I might have some ideas to simplify this.

## Has this been tested?

I still need to figure out why some tests are failing and I might need some help to do so!

## Resources
- https://github.com/TypeStrong/ts-loader/blob/main/REFERENCES.md


server/tools/cli.ts
```diff
@@ -180,6 +182,7 @@ function getServerCredentials (program: Command) {
 }
 
 function buildServer (url: string) {
+  loadLanguages()
```
Chocobozzz, line 185: :+1:

shared/core-utils/ffprobe.ts
```diff
@@ -0,0 +1,180 @@
+import { ffprobe, FfprobeData } from 'fluent-ffmpeg'
```
Chocobozzz, line 1: This file should not be in `core-utils` since it imports a non nodejs core module

server/helpers/requests.ts
```diff
@@ -194,6 +203,39 @@ async function findLatestRedirection (url: string, options: PeerTubeRequestOptio
   return url
 }
 
+function makePOSTAPRequest (url: string, body: any, httpSignature: any, headers: any) {
```
Chocobozzz, line 206: This is only used by test files, so prefer to put it in something like `server/tests/shared`

server/helpers/requests.ts
```diff
@@ -194,6 +203,39 @@ async function findLatestRedirection (url: string, options: PeerTubeRequestOptio
...
+    method: 'POST' as 'POST',
+    json: body,
+    httpSignature,
+    headers
+  }
+
+  return doRequest(url, options)
+}
+
+async function makeFollowRequest (to: { url: string }, by: { url: string, privateKey }) {
```
Chocobozzz, line 217: This is only used by test files, so prefer to put it in something like `server/tests/shared`

support/doc/development/lib.md
```diff
@@ -8,3 +8,16 @@
 $ cd client/src/standalone/player/
 $ npm run build
 ```
+
+## @peertube/peertube-types
+
+Typescript definition files generation is controlled by the various `tsconfig.types.json` files, see :
```
Chocobozzz, line 14: Extra space for `:`

support/doc/development/lib.md
```diff
@@ -8,3 +8,16 @@
...
+```
+yarn tsc -b --verbose tsconfig.types.json
+```
+
+But the complete types package is generated via :
```
Chocobozzz, line 19: Extra space for `:`

support/doc/plugins/guide.md
```diff
@@ -858,6 +858,39 @@ Now you can register hooks or settings, write CSS and add static directories to
 **Caution:** It's up to you to check the code you write will be compatible with the PeerTube NodeJS version,
 and will be supported by web browsers.
 If you want to write modern JavaScript, please use a transpiler like [Babel](https://babeljs.io/).
+If you want to use __Typescript__ see section below.
+
+### Typescript
+
+You can add __PeerTube__ types as dev dependencies :
```
Chocobozzz, line 865: Extra space for `:`

support/doc/plugins/guide.md
```diff
@@ -858,6 +858,39 @@ Now you can register hooks or settings, write CSS and add static directories to
...
+```
+npm install --dev @peertube/peertube-types
+```
+
+This package exposes *server* definition files by default :
```
Chocobozzz, line 870: Extra space for `:`

support/doc/plugins/guide.md
```diff
@@ -858,6 +858,39 @@ Now you can register hooks or settings, write CSS and add static directories to
...
+
+export async function register ({ registerHook }: RegisterServerOptions) {
+  registerHook({
+    target: 'action:application.listening',
+    handler: () => displayHelloWorld()
+  })
+}
+```
+
+But it also exposes client types and various models used in __PeerTube__ :
```
Chocobozzz, line 882: Extra space for `:`

client/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './types'
```
Chocobozzz, line 1: I would prefer to not export types in from `/client/src/index.ts` and `/index.ts` to not mix types package and PeerTube main code. IMHO types is just an additional package, so we should not have them as the main TS entries.

Could we imagine creating a `/types/` or `/packages/types` directory containing these index files?
Then we can put compiled files in `/types/dist` or `/packages/types/dist` because the current `/types` directory could be confusing (is it a build directory or a source directory?).

scripts/build/server.sh
```diff
@@ -4,7 +4,10 @@ set -eu
 
 rm -rf ./dist
 
-npm run tsc
-cp "./tsconfig.json" "./dist"
+npm run tsc -- -b --verbose
+cp "./tsconfig.base.json" "./tsconfig.json" "./dist/"
+cp "./scripts/tsconfig.json" "./dist/shared/"
```
Chocobozzz, line 9: Typo for the destination?

scripts/generate-types-package.ts
```diff
@@ -0,0 +1,78 @@
+import {copyFile, readJson, writeFile, writeJSON } from 'fs-extra'
+import { resolve } from 'path'
+import { cwd } from 'process'
+import { execSync } from 'child_process'
+import depcheck, { PackageDependencies } from 'depcheck'
+
+run()
+    .then(() => process.exit(0))
```
Chocobozzz, line 8: 2 spaces for indentation please :)

Chocobozzz (commented):
Very interesting, thanks @lutangar!

> I noticed many little things which could be improved on the overall project organization which I would I live to discuss around a beer beer sometimes!

Sure, you have my email in my github profile :)

scripts/generate-types-package.ts
lutangar: Sure, will do. I noticed `eslint` doesn't fix this special case since the rule option `memberExpression` is set to to `off` in the `.eslintrc.json`:

> `"MemberExpression"` (default: 1) enforces indentation level for multi-line property chains. This can also be set to `"off"` to disable checking for MemberExpression indentation.
> https://eslint.org/docs/rules/indent

Shall I switch this option to `2` spaces instead?

client/src/index.ts
lutangar: From what I understand, in the client codebase, what you're putting in `client/src/types/` aren't external types but rather __PeerTube__ client own *interfaces*.
As __types__ and  __typings__  are usually used as synonymous, I would rather not put anything part of the __PeerTube__ codebase inside this directory. Ideally I would only put *missing* external typings.

For example on the `server` I find it rather confusing to find:
- `server/models` :heavy_check_mark:
- `server/typings` :heavy_check_mark:
- `server/types` which is containing __domain__ related `models` (`server/types/models` for example)
I'd rather merge `server/models` and`server/typings` directories.

...after some thinking I think I understand why this directory is named `types`, it contains everything declared with the `type` keyword, is that right? Might sounds silly but wasn't obvious to me... :sweat_smile:  
Is there another rational behind this that I'm not getting?

I still think most of these `types` could live close to their related model. 
I'm having a look at the `Actor` model right now... Maybe `MActor`, `MActorAccountChannelId`, etc. `types` could live in the same file? Or same dir? Just thinking out loud here. I wonder how other projects are doing.
By the way, what's the `M` of `MActor` is standing for?

> *Just a few interesting resources I browsed while wondering if some of the types could have been interfaces.*
> https://stackoverflow.com/questions/37233735/interfaces-vs-types-in-typescript/52682220#52682220
> https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

Let's get back to our issue, 

> IMHO types is just an additional package, so we should not have them as the main TS entries.

IMHO this `@peertube/peertube-types` package shouldn't exist at all but `@peertube/peertube` should instead (or `@peertube/*` with the different packages).
It only exist because I didn't want to mess the current codebase too much and at the moment `types` are needed for third parties developments.
I think that a `@peertube/peertube` should be set as a mid to long term target, and that doing as such, it would help drive future code organization choices.

I wonder... is this `types` directory exposing only publicly visible types?
https://www.typescriptlang.org/tsconfig#stripInternal
Could the `@peertube/peertube-types` only exposing the different `types` directories and expose everything needed?
If so, it fails as doing so, but I might getting ahead of myself here.

> Could we imagine creating a /types/ or /packages/types directory containing these index files?
> Then we can put compiled files in /types/dist or /packages/types/dist because the current /types directory could be confusing (is it a build directory or a source directory?).

Sure, I'll play with this idea a bit and see where I'm getting! :+1:

scripts/generate-types-package.ts
Chocobozzz: I think I disabled it so we can choose between these two cases:

```
run().then(...
       .catch(
```

Or 

```
longVariableName.subVariableName.doSomethingInteresting()
  .then(...
```

client/src/index.ts
Chocobozzz: > From what I understand, in the client codebase, what you're putting in client/src/types/ aren't external types but rather PeerTube client own interfaces.

Yes, this directory is dedicated to peertube client own interfaces. The `types` naming is not very good (same as `server/types`, that contains PeerTube server own interfaces).

> Is there another rational behind this that I'm not getting?

No, it's just a bad named directory :) I'm open to suggestions (but I'm not sure to understand why you'd "merge server/models andserver/typings directories")

>  I wonder how other projects are doing.

I never found other typescript projects that tries to type DB associations.

> By the way, what's the M of MActor is standing for?

`Model` I guess, it's just to clearly differentiate `ActorModel` (with all optional associations) than `MActor` (typed associations).

> Just a few interesting resources I browsed while wondering if some of the types could have been interfaces.

Since we use `Omit`, `Pick` etc I preferred to keep consistency with `type`. If you find a clearer way, with interfaces to describe model associations don't hesitate to explain it :)

> I think that a @peertube/peertube should be set as a mid to long term target, and that doing so, would help drive future code organization choices.

I don't understand understand the point of publishing a `@peertube/peertube` package. We don't plan to release PeerTube on NPM, so why would we publish it? Instead, I always thought we would publish packages in `@peertube/peertube-*` that could be used by other libraries (`types` of course, but we could also consider to release a `@peertube/peertube-server-helpers` package for example so other projects can use them). But I may miss your point?

> I wonder... is this types directory exposing only publicly visible types?

I'm not sure to understand what `types` directory you're talking about.

lutangar: > No, it's just a bad named directory :) I'm open to suggestions (but I'm not sure to understand why you'd "merge server/models and server/typings directories")

No worries :D
I meant to merge `server/models` and `server/types/*` or even move some of `server/types/models/*` inside their related model files. It seems most of them aren't used -- mostly -- anywhere else. Just planting seeds here :seedling:

> We don't plan to release PeerTube on NPM, so why would we publish it

Well if it's not planned, I understand.
But instead of a `@peertube/peertube-server-helpers` I was rather thinking about a standalone `@peertube/server` for example. 
> As I was saying in previous comments it could have provided the types, the code and a standalone __PeerTube__ server at the same time. There wouldn't be the need to create a `*-types` package.

I guess what I'm saying is, reasoning in *packages*, helps to set clearer boundaries between each __apps__.

> I wonder... is this types directory exposing only publicly visible types?
> >I'm not sure to understand what types directory you're talking about.

I think I misunderstand one of your point but... I was saying that if **all** of the `types` (server, client, etc.) directories were exposing *every* interfaces required by third parties, it might make more sense to me to exposes them and only them in a `*types` package.

scripts/generate-types-package.ts
lutangar: Just made a few tests with `"MemberExpression": 1`:
```
run()
    .then(() => process.exit(0)).catch(err => {
    console.error(err)
    process.exit(-1)
  })

depcheck.detector.requireCallExpression
    .toString()
```
`yarn eslint --fix scripts/generate-types-package.ts` ->
```
run()
  .then(() => process.exit(0)).catch(err => {
    console.error(err)
    process.exit(-1)
  })

depcheck.detector.requireCallExpression
  .toString()
```
It seems it doesn't "break" the expression but just fix the number of space used for indentation.
I'd usually recommend to automate any code style which can be automated, even if it means there would be a side effect. 

The side effect is usually deem acceptable. But up to you! I could submit this suggestion in another PR I guess.

Chocobozzz: > The side effect is usually deem acceptable. But up to you! I could submit this suggestion in another PR I guess.

yes please :+1:

client/src/index.ts
Chocobozzz: > I meant to merge server/models and server/types/* or even move some of server/types/models/* inside their related model files. It seems most of them aren't used -- mostly -- anywhere else. Just planting seeds here seedling

I think it's a good idea :+1:

>  I was rather thinking about a standalone @peertube/server for example.

And you'd put every server module in this package?

lutangar:
> > Could we imagine creating a /types/ or /packages/types directory containing these index files?
Then we can put compiled files in /types/dist or /packages/types/dist because the current /types directory could be confusing (is it a build directory or a source directory?).
>
> Sure, I'll play with this idea a bit and see where I'm getting! +1

I tried something with the following structure:
```
types/tsconfig.json
types/src/index.ts
types/src/client/index.ts
types/src/client/tsconfig.json
```
And output to `types/dist/` directory.

I did struggle to put something consistent in place :weary:  
I found there is more duplication between the different tsconfig and I fear it would be harder to maintain...  but I might be wrong. 

Have a look at the diff:
https://github.com/larriereguichet/PeerTube/pull/1/files

.github/workflows/publish.yml
```diff
@@ -0,0 +1,19 @@
+name: "Publish types package"
```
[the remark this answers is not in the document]
lutangar: See suggestion below:
https://github.com/Chocobozzz/PeerTube/pull/4531/commits/0e8536384adf82b4db95819e9719f85a077ae09a

lutangar:
> I found there is more duplication between the different tsconfig and I fear it would be harder to maintain...

Using a build tool like `webpack` (with `ts-loader`) would help to do more complex transformation and probably avoid some duplication. 

`webpack` could also bring some good to the `server` build process, but this is yet another story :sweat_smile:

lutangar:
I somewhat backtrack on the first refacto... I'm not sure if you noticed before but the `server.ts` file had previously migrated to `server/index.ts` file.  But anyway this is no longer the case.

While browsing the doc, I realized the path to this file is quite important for instances in productions and setup on the host system (service config file for example). 

I might later propose to move some of the logic back to the server folder AND keep the server.ts file as well.

kontrollanten:
This looks really promising. What's left to be done to be able to merge this?

Chocobozzz:
> This looks really promising. What's left to be done to be able to merge this?

It's on my TODO, just need to fix the conflicts and I think it's ready. (ping @lutangar?)

lutangar:
I'll fix the conflicts today but I'll probably need some help fixing the last failing test cases.

Chocobozzz:
> I'll fix the conflicts today but I'll probably need some help fixing the last failing test cases.

I can fix the conflicts and fix the tests. I pinged you just in case you think there is something else to do in this PR :)

lutangar:
See:
>  ```
> 1) Test multiple servers
>       Should upload the video and propagate on each server
>         Should upload the video on server 1 and propagate on each server:
>     AssertionError: expected 'Japanese' to equal undefined
>```
- https://github.com/Chocobozzz/PeerTube/runs/4517396032?check_suite_focus=true#step:8:859
- https://github.com/Chocobozzz/PeerTube/runs/4517396067?check_suite_focus=true#step:8:337

Don't know if this warning is an issue :
https://github.com/Chocobozzz/PeerTube/runs/4517396067?check_suite_focus=true#step:8:669

Many `eslint` error on the following, I suppose it's because the file weren't udpated since the rule was added :
```
Error:   269:43  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
```
Shall we ignore this rule in all tests?

What about this floating promise?
```
/home/runner/work/PeerTube/PeerTube/server/lib/job-queue/job-queue.ts
Error:   155:7  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
```

lutangar:
> I can fix the conflicts and fix the tests. I pinged you just in case you think there is something else to do in this PR :)

Oh! Too late for the conflicts I suppose... but you're still good to fix the tests! (see comment above)

I think there'll be much to discuss in the future on related subjects, but I'm done for this iteration.

Chocobozzz:
@lutangar Could you allow edits by maintainers for this PR? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Chocobozzz:
Thanks @lutangar!

Chocobozzz:
Hi,

I'm still working on the generated types.

I tried to clean up directory structure and generated types:
 * Only export client `types` directory
 * Created a `shared/typescript-utils`
 * Renamed `shared/extra-utils` to `shared/server-commands` (kept ffprobe in `extra-utils`)
 * Excluded `shared/server-commands` for generated types (I would like to provide a `server-commands` package on NPM in the future, because I think these tools could be really useful)
 * Moved functions used by server tests in `server/tests/shared`

kontrollanten:
> Only export client types directory

Is it possible to also export the job queue related types? I think that will be useful when working with #947

PR #4552, opened 2021-11-20
Apply import interval only when reasonable

mj-saunders:
## Description

When importing videos from another service, an interval can be applied between each download.  
It only really makes sense to apply this interval when the last attempted download actually happened, and not when it was skipped.

## Has this been tested?

- [x] 💭 no, because this PR is a draft and **may** still needs work

## Note

I originally had to edit code in `dist/` directly, as I needed the functionality immediately.  
I have tested it only manually and it appears to behave as I would expect.

I am not an experienced `type/javascript` programmer and would like to confirm that the fix is as you might expect or want it to be.  
The problem is trivial, but is the solutions satisfactory?

### DIFF
--- a/server/tools/peertube-import-videos.ts
+++ b/server/tools/peertube-import-videos.ts
@@ -95,14 +95,15 @@ async function run (url: string, username: string, password: string) {
 
   log.info('Will download and upload %d videos.\n', infoArray.length)
 
+  let skipInterval = 1;
   for (const [ index, info ] of infoArray.entries()) {
     try {
-      if (index > 0 && options.waitInterval) {
+      if (index > 0 && options.waitInterval && !skipInterval) {
         log.info("Wait for %d seconds before continuing.", options.waitInterval / 1000)
         await wait(options.waitInterval)
       }
 
-      await processVideo({
+      skipInterval = await processVideo({
         cwd: options.tmpdir,
         url,
         username,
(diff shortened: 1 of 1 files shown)


server/tools/peertube-import-videos.ts
```diff
@@ -184,6 +185,8 @@ async function processVideo (parameters: {
   } catch (err) {
     log.error(err.message)
   }
+
+  return 0;
```
Chocobozzz, line 189: Use `true`/`false` instead

server/tools/peertube-import-videos.ts
```diff
@@ -95,14 +95,15 @@ async function run (url: string, username: string, password: string) {
 
   log.info('Will download and upload %d videos.\n', infoArray.length)
 
+  let skipInterval = true;
```
Chocobozzz, line 98: extra `;`

server/tools/peertube-import-videos.ts
```diff
@@ -184,6 +185,8 @@ async function processVideo (parameters: {
   } catch (err) {
     log.error(err.message)
   }
+
+  return false;
```
Chocobozzz, line 189: extra `;`

mj-saunders: Well they were stupid mistakes.
Would it be acceptable to force push the change?
Or do a fresh commit, assuming perhaps that you would anyway squash merge later?

rigelk: Both are usually acceptable.

Chocobozzz:
Thanks!

PR #4562, opened 2021-11-24
[Plugin][Hook] Add token as parameters to video creation and update hooks

lutangar:
## Description

Add `token` to hook parameters `action:api.video.updated` and `action:api.video.uploaded`.

## Context

I'm developing a plugin which generates a subtitle (server side) if a video has a `language` set  and this is supported by the plugin.
This generation is done as part of the video creation/update process and I'd like to be able to add this subtitle to the video.

As far as I'm aware of, the API endpoint `PUT /videos/${videoUuid}/captions/${captionLanguage}` looks like the best entrypoint to do so, even from a plugin perspective. 
However, this endpoint is only accessible by an authenticated user.

## Rational

It seems it's possible to make use of the user token client side, it feel like a legitimate use-case to do the same server-side in the context of a user related request.

I evaluate other options, one of them, which I was about to implement was something like the following hack:
```typescript
// create server side route
router.get('/forward-authenticated-request', async (req: Request, res: Response) => {
  // 1. extract credential from the response
  // 2. retrieve subtitle file from the filesystem or generate it on the fly
  // 3. PUT `/videos/${videoUuid}/captions/${captionLanguage}`
});
```
But this would involve a silly *ping-pong* game between the server and the client...

## Going further

Does extending the usage of this parameter to others hooks involving the user would make sense?
These hooks parameters could also be typed (and then exposed) to enhance plugin Developer eXperience.

## Has this been tested?

### DIFF
--- a/server/lib/video-comment.ts
+++ b/server/lib/video-comment.ts
@@ -1,5 +1,6 @@
 import { cloneDeep } from 'lodash'
 import * as Sequelize from 'sequelize'
+import express from 'express'
 import { logger } from '@server/helpers/logger'
 import { sequelizeTypescript } from '@server/initializers/database'
 import { ResultList } from '../../shared/models'
@@ -10,7 +11,7 @@ import { sendCreateVideoComment, sendDeleteVideoComment } from './activitypub/se
 import { getLocalVideoCommentActivityPubUrl } from './activitypub/url'
 import { Hooks } from './plugins/hooks'
 
-async function removeComment (videoCommentInstance: MCommentOwnerVideo) {
+async function removeComment (videoCommentInstance: MCommentOwnerVideo, req: express.Request, res: express.Response) {
   const videoCommentInstanceBefore = cloneDeep(videoCommentInstance)
 
   await sequelizeTypescript.transaction(async t => {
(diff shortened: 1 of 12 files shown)


Chocobozzz:
Hello,

I think you can send `req` and `res` to all controllers hooks. Then you can add the `token` in `getAuthUser` result https://github.com/Chocobozzz/PeerTube/blob/develop/server/lib/plugins/plugin-helpers-builder.ts#L171

lutangar:
Sounds good! I'll make this happens. :sparkles:

server/lib/video-comment.ts
```diff
@@ -10,7 +11,7 @@ import { sendCreateVideoComment, sendDeleteVideoComment } from './activitypub/se
 import { getLocalVideoCommentActivityPubUrl } from './activitypub/url'
 import { Hooks } from './plugins/hooks'
 
-async function removeComment (videoCommentInstance: MCommentOwnerVideo) {
+async function removeComment (videoCommentInstance: MCommentOwnerVideo, req: express.Request, res: express.Response) {
```
lutangar, line 14: I don't like adding `req` and `res` here but I fell short on refactoring ideas.

Chocobozzz:
Thanks!

Issue #3647, opened 2021-01-27
Random language translation is selected


kontrollanten:
**Describe the current behavior**
I've selected English as default language in my browser (Brave) and Swedish as language at our PeerTube instance. Randomly when I visit our PeerTube instance I get a Finnish translation. We've also got reports from users who gets Russian translations randomly. This problem has been since version 2.4.0 and became better since 3.0.0.

**Steps to reproduce:**
Can't reproduce this since it's happening randomly.

* PeerTube instance:
  * version: 3.0.1

* Browser name, version and platforms on which you could reproduce the bug:
Brave, Linux

Chocobozzz:
What is your instance URL?

kontrollanten:
I emailed you the instance URL

Chocobozzz:
Could it be a front proxy/CDN issue? Because I could not reproduce and I've never heard of such problem

kontrollanten:
I can't really see how the CDN or proxy could cause that issue. Do you know how angular is determining the language? When the problem appears it doesn't gets solved by page reload, only by opening the site in a new tab.

Chocobozzz:
We serve the appropriate bundle from the server: https://github.com/Chocobozzz/PeerTube/blob/develop/server/controllers/client.ts#L82 

We use the request parameter to guess the language. If it does not exist, we check the cookie. If it does not exist, we send the appropriate page according to the request accept language header: https://github.com/Chocobozzz/PeerTube/blob/9aeef9aafa009762e128046bfb2f3cec3252e181/server/lib/client-html.ts#L291

kontrollanten:
I think I've figured out this. If the request accept language header is set to Russian our CDN will cache the Russian index.html version (with `<html lang="ru">` etc), right?

kontrollanten:
We're still facing this issue with the latest version. I can't reproduce it and I haven't got any problems since this was closed, but we're having users reporting this issue. One user reported the following.

1) He visited the site and everything was in Swedish.
1) He downloaded a video.
1) The site was in Russian.
1) He downloaded another video.
1) The site was in English.

Can a solution be to make it possible to disable auto lang selection (and only support one selection)? I know that won't solve the root problem, but I can't find any other way in solving this.

Chocobozzz:
Hello,

I'm sorry but we can't help. This is an issue with your current setup and not with peertube.

PR #4588, opened 2021-11-30
Add header Vary Accept-Language

kontrollanten:
## Description
Since PeerTube uses Accept-Language header to determine response language, the HTTP response should include the `Vary: Accept-Languge` header to ensure no cache will prevent language detection.

## Related issues
solves #3647

## Has this been tested?
Yes, manually.

### DIFF
--- a/server/lib/client-html.ts
+++ b/server/lib/client-html.ts
@@ -547,9 +547,12 @@ class ClientHtml {
   }
 }
 
-function sendHTML (html: string, res: express.Response) {
+function sendHTML (html: string, res: express.Response, localizedHTML: boolean = false) {
   res.set('Content-Type', 'text/html; charset=UTF-8')
-  res.set('Vary', 'Accept-Language')
+
+  if (localizedHTML) {
+    res.set('Vary', 'Accept-Language')
+  }
 
   return res.send(html)
 }
@@ -579,5 +582,5 @@ export {
 async function generateHTMLPage (req: express.Request, res: express.Response, paramLang?: string) {
   const html = await ClientHtml.getDefaultHTMLPage(req, res, paramLang)
 
-  return sendHTML(html, res)
+  return sendHTML(html, res, true)
 }
--- a/server/lib/client-html.ts
+++ b/server/lib/client-html.ts
@@ -549,6 +549,7 @@ class ClientHtml {
 
 function sendHTML (html: string, res: express.Response) {
   res.set('Content-Type', 'text/html; charset=UTF-8')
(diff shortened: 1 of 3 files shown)


Chocobozzz:
Nice catch! I did not even know the header

server/lib/client-html.ts
```diff
@@ -549,6 +549,7 @@ class ClientHtml {
 
 function sendHTML (html: string, res: express.Response) {
   res.set('Content-Type', 'text/html; charset=UTF-8')
+  res.set('Vary', 'Accept-Language')
```
Chocobozzz, line 552: Embed page does not vary so I don't think we should set the header here.

server/lib/client-html.ts
```diff
@@ -549,6 +549,7 @@ class ClientHtml {
...
```
Chocobozzz, line 553: I think you should also use `Cookie` for `Vary` since we can also serve a different HTML page depending on the user cookie.

server/lib/client-html.ts
kontrollanten: Can we add `i18n: bool` as an argument to sendHTML?

server/lib/client-html.ts
kontrollanten: In what cases are the HTML served differently based on the cookie? According to this docs it's not recommended to use Vary Cookie; https://www.fastly.com/blog/best-practices-using-vary-header

Chocobozzz: > In what cases are the HTML served differently based on the cookie?

When the user explicitly set the language in the interface: https://github.com/Chocobozzz/PeerTube/blob/develop/server/lib/client-html.ts#L335 (to remember to serve this explicit lang in future index HTML calls)

> According to this docs it's not recommended

Peertube only uses one cookie that stores client language. But since plugins like matomo can also store cookies, I agree we may not add it.

server/lib/client-html.ts
Chocobozzz: Sure, or maybe `localizedHTML: bool`

server/lib/client-html.ts
kontrollanten: Hmm. That was a hard case. For most of the time that shouldn't be a problem for the user, unless there's a CDN in the picture. But I think that most CDN's has the possibility to bypass cache based on a specific cookie. Maybe we can add something about this in the documentation?

Chocobozzz: Yep

PR #4601, opened 2021-12-02
Fix test documentation

JohnXLivingston:
## Description

The previous documentation only mentioned `yarn install --pure-lock`. But in order to develop or launch unit tests, it is required to also install devDependencies (`yarn install --pure-lock --production=false`).

If you don't do this, unit tests won't launch (because mocha and ts-node will be missing).

I also changed the command for launching only some unit tests: using «npx mocha» instead of «mocha», so that it uses the local version (no need to install mocha globally, and no risk to have the wrong mocha version).

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -85,7 +85,7 @@ to install the dependencies.
 $ git clone https://github.com/Chocobozzz/PeerTube
 $ cd PeerTube
 $ git remote add me git@github.com:YOUR_GITHUB_USERNAME/PeerTube.git
-$ yarn install --pure-lockfile
+$ yarn install --pure-lockfile --production=false
 ```
 
 Note that development is done on the `develop` branch. If you want to hack on
--- a/support/doc/development/tests.md
+++ b/support/doc/development/tests.md
@@ -2,6 +2,12 @@
 
 ## Preparation
 
+Install devDepedencies:
+
+```bash
+yarn install --pure-lockfile --production=false
+```
+
 Prepare PostgreSQL user so PeerTube can delete/create the test databases:
 
 ```bash
@@ -42,7 +48,7 @@ $ npm run test # See scripts/test.sh to run a particular suite
 Most of tests can be runned using:
 
 ```bash
-TS_NODE_TRANSPILE_ONLY=true mocha -- --timeout 30000 --exit -r ts-node/register -r tsconfig-paths/register --bail server/tests/api/videos/video-transcoder.ts
+TS_NODE_TRANSPILE_ONLY=true npx mocha -- --timeout 30000 --exit -r ts-node/register -r tsconfig-paths/register --bail server/tests/api/videos/video-transcoder.ts
 ```
 
 `server/tests/api/activitypub` tests will need different options:
--- a/support/doc/development/tests.md
+++ b/support/doc/development/tests.md
@@ -2,7 +2,7 @@
 
 ## Preparation
 
-Install devDepedencies:
+If you not already installed devDependencies:
 
 ```bash
 yarn install --pure-lockfile --production=false


.github/CONTRIBUTING.md
```diff
@@ -85,7 +85,7 @@ to install the dependencies.
 $ git clone https://github.com/Chocobozzz/PeerTube
 $ cd PeerTube
 $ git remote add me git@github.com:YOUR_GITHUB_USERNAME/PeerTube.git
-$ yarn install --pure-lockfile
+$ yarn install --pure-lockfile --production=false
```
Chocobozzz, line 88: yarn does install deve dependency by default

support/doc/development/tests.md
```diff
@@ -2,6 +2,12 @@
 
 ## Preparation
 
+If you not already installed devDependencies:
```
Chocobozzz, line 5: I think you should point to the contributing guide instead, because you also need to build peertube

support/doc/development/tests.md
```diff
@@ -42,7 +48,7 @@ $ npm run test # See scripts/test.sh to run a particular suite
 Most of tests can be runned using:
 
 ```bash
-TS_NODE_TRANSPILE_ONLY=true mocha -- --timeout 30000 --exit -r ts-node/register -r tsconfig-paths/register --bail server/tests/api/videos/video-transcoder.ts
+TS_NODE_TRANSPILE_ONLY=true npx mocha -- --timeout 30000 --exit -r ts-node/register -r tsconfig-paths/register --bail server/tests/api/videos/video-transcoder.ts
```
Chocobozzz, line 51: prefer `npm run mocha -- timeout...`

.github/CONTRIBUTING.md
JohnXLivingston: That's weird... In my case, I haven't mocha installed. I installed it manually, and then it was ts-node (if I remember correctly) that was missing. Then I made `yarn install --pure-lockfile --production=false` and it worked fine.
I have no NODE_ENV var.
I don't understand what happened.

support/doc/development/tests.md
JohnXLivingston: Building Peertube is 2 steps under :)

support/doc/development/tests.md
JohnXLivingston: Maybe I am wrong, but if you use «mocha», it will use the version installed globally. That is not necessarely compatible with the version used by Peertube.
Moreover, if I have no version installed globally, I got an error «bash: /usr/local/bin/mocha: Aucun fichier ou dossier de ce type»

support/doc/development/tests.md
Chocobozzz: Oup's,  then remove `--production=false`

support/doc/development/tests.md
Chocobozzz: `npm run mocha` will run `mocha` script defined in `package.json`

support/doc/development/tests.md
JohnXLivingston: As said in comments to the previous point, launching unit tests with `TS_NODE_TRANSPILE_ONLY=true mocha -- --timeout 30000 --exit -r ts-node/register -r tsconfig-paths/register --bail server/tests/api/videos/video-transcoder.ts` doesn't work for me without --production=false

support/doc/development/tests.md
JohnXLivingston: Oops, I misread your answer. Ok.

.github/CONTRIBUTING.md
Chocobozzz: You reproduce this bug every time you run `yarn install`?

JohnXLivingston: I don't know. You must be right. I suggest to drop the entire PR, and just keep this commit: https://github.com/Chocobozzz/PeerTube/pull/4601/commits/80281efc4dce04552b3f79d9d82abd32e1fc631d

Chocobozzz: Agreed. Done in https://github.com/Chocobozzz/PeerTube/commit/11e520b50d791a0dd48cbb2d0fc681b25eb7cd53

PR #4606, opened 2021-12-04
Update iframe allowfullscreen to current method

nicorikken:
The current PeerTube frontend generates iframe code that contains a old method
of allowing fullscreen: `allowfullscreen=""`. This method is surpassed by the
`allow="fullscreen"` property. This commit changes the rendered output to the
new code. I have tested it in Gitpod and the change worked as expected.

More documentation on this property:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-allowfullscreen

Signed-off-by: Nico Rikken <nico@nicorikken.eu>

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -553,6 +553,7 @@
  * wakutiteo
  * wonderingdane
  * zeynepeliacik
+ * Nico Rikken
 
 
 # Design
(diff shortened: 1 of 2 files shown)


nicorikken:
Just to add some context to the commit. i came here because our current PeerTube instance was rendering an embed code with `allowfullscreen` (no `=""`) which was causing XML linting errors. The current version on the `develop` branch does render XML-valid attribute `allowfullscreen=""`. But I had already prepared this commit to update it to the new standard for specifying if fullscreen is allowed (`allow="fullscreen"`). I'm not sure about backwards compatibility, but it works for my browser.

So feel free to dismiss or delay this commit. But as I already prepared it I thought why not make the pull request as well.

Chocobozzz:
Hello,

Unfortunately `allow` attribute on iframe is not supported by ios < 11.3 (peertube supports ios 11)

PR #4608, opened 2021-12-06
Give moderators access to edit channels

### LINKED ISSUE #4596: Peertube with Minio S3
Upload Content should move into Minio S3 Buckets, but nothing happend. Under the Video, in Peertube Interface, there's still the Infomration that the Video is moving on external Storage.  

S3 works, i've tested it via the Minio CLI, its possible to copy files, create buckets etc. 

**Additional information**

* PeerTube instance:
  * version: 3.4.1

### LINKED ISSUE #4598: Give admins access to edit channels
We've some users that needs help with customizing their video channels. Currently there's not possible for an admin to edit someone else's video channel.

Give the administrator access to edit all video channels.

Same solution as for user accounts; add a dropdown menu to the right of the video channel name.

kontrollanten:
## Description
Gives moderators access to edit another users video channel. A moderator won't be able to update an admins channel.

## Related issues
closes #4598

## Has this been tested?

- [x] 🙋 yes

## Screenshots

image


kontrollanten:
@Chocobozzz My second test fails due to

```
  1) Test video channels
       Should have another acccounts video channel updated:
     Error: expected 200 "OK", got 404 "Not Found"
      at buildRequest (shared/extra-utils/requests/requests.ts:175:35)
      at makeGetRequest (shared/extra-utils/requests/requests.ts:39:10)
      at ChannelsCommand.getRequest (shared/extra-utils/shared/abstract-command.ts:83:26)
      at ChannelsCommand.getRequestBody (shared/extra-utils/shared/abstract-command.ts:57:31)
      at ChannelsCommand.get (shared/extra-utils/videos/channels-command.ts:110:17)
      at Context.<anonymous> (server/tests/api/videos/video-channels.ts:242:42)
```

It seems like the update doesn't get synced to the second server. Do you know what I'm doing wrong here?

kontrollanten:
Thanks. Do you have any comments on the questions in the PR description?

Chocobozzz:
> It seems like the update doesn't get synced to the second server. Do you know what I'm doing wrong here?

You need the complete handle **and** upload a video in the channel so server 2 knows it. I added the commit that fixes tests.

> Should we display this Manage channel button, or should we add a dropdown like the account page?

I think the manage button is okay

> Should we send a notification and email to the channel owner? I think that would be a great idea.

IMHO I don't think a notification could be interesting for users. It can be confusing to receive a notification that an admin updated one of your channels. I mean, why an admin updated my channel? For what purpose?
I think that most of the time, as an admin, you'll update channels of users for a specific need that should be communicated to the user (by email for example).
But I may miss a use case, so as an admin why do you need this feature?

> In the future I think also moderators should be able to edit channels, but that was a bit tricky to solve so I skipped it for now.

I think you can just replace `REMOVE_ANY_VIDEO_CHANNEL` by `MANAGE_VIDEO_CHANNELS`. Why is this tricky?

kontrollanten:
> > Should we send a notification and email to the channel owner? I think that would be a great idea.
> 
> IMHO I don't think a notification could be interesting for users. It can be confusing to receive a notification that an admin updated one of your channels. I mean, why an admin updated my channel? For what purpose? I think that most of the time, as an admin, you'll update channels of users for a specific need that should be communicated to the user (by email for example). But I may miss a use case, so as an admin why do you need this feature?

That's my use case as well. The notification idea came from @andrew712-1 (https://github.com/Chocobozzz/PeerTube/issues/4598#issuecomment-985650413) and I thought it may be good to be transparent. But on the other hand I think as a user you don't expect to get a notification if an admin is changing anything in your channel. Since we don't send notifications for similar admin actions I think we can skip it.

> > In the future I think also moderators should be able to edit channels, but that was a bit tricky to solve so I skipped it for now.
> 
> I think you can just replace `REMOVE_ANY_VIDEO_CHANNEL` by `MANAGE_VIDEO_CHANNELS`. Why is this tricky?

I don't remember by heart, I just didn't manage to get the role permissions to work when I tried, but I'll try again.

kontrollanten:
The tricky part was how to handle when a moderator tries to update an admins channel. I found the following code and made same logic here.

https://github.com/Chocobozzz/PeerTube/blob/c4c0c31144b2cb8b6ade1cc39d676bc00c08cabf/server/middlewares/validators/users.ts#L516-L519

client/src/app/+video-channels/video-channel-edit/video-channel-update.component.ts
```diff
@@ -95,7 +95,7 @@ export class MyVideoChannelUpdateComponent extends MyVideoChannelEdit implements
 
           this.notifier.success($localize`Video channel ${videoChannelUpdate.displayName} updated.`)
 
-          this.router.navigate([ '/my-library', 'video-channels' ])
+          this.router.navigate([ '/c', this.videoChannel.name ])
```
Chocobozzz, line 98: Prefer to redirect to previous route (`redirect.service.ts`)

If there is no previous route, you could redirect to `this.router.navigate([ '/c', this.videoChannel.name ])`

client/src/app/+my-library/+my-video-channels/my-video-channels-routing.module.ts
```diff
@@ -13,24 +11,6 @@ const myVideoChannelsRoutes: Routes = [
         title: $localize`My video channels`
       }
     }
-  },
-  {
-    path: 'create',
```
Chocobozzz, line 18: Please add redirections to new routes

client/src/app/+video-channels/video-channel-edit/video-channel-create.component.ts
```diff
@@ -12,11 +12,11 @@ import {
...
-import { MyVideoChannelEdit } from './my-video-channel-edit'
+import { MyVideoChannelEdit } from './video-channel-edit'
 
 @Component({
-  templateUrl: './my-video-channel-edit.component.html',
-  styleUrls: [ './my-video-channel-edit.component.scss' ]
+  templateUrl: './video-channel-edit.component.html',
+  styleUrls: [ './video-channel-edit.component.scss' ]
 })
 export class MyVideoChannelCreateComponent extends MyVideoChannelEdit implements OnInit {
```
Chocobozzz, line 21: `VideoChannelCreateComponent`

Chocobozzz: `VideoChannelEdit`

client/src/app/+video-channels/video-channels-routing.module.ts
```diff
@@ -1,10 +1,21 @@
...
 import { RouterModule, Routes } from '@angular/router'
+import { MyVideoChannelCreateComponent } from './video-channel-edit/video-channel-create.component'
+import { VideoChannelUpdateComponent } from './video-channel-edit/video-channel-update.component'
 import { VideoChannelPlaylistsComponent } from './video-channel-playlists/video-channel-playlists.component'
 import { VideoChannelVideosComponent } from './video-channel-videos/video-channel-videos.component'
 import { VideoChannelsComponent } from './video-channels.component'
 
 const videoChannelsRoutes: Routes = [
+  {
+    path: '@create',
```
Chocobozzz, line 11: `'create'`

client/src/app/+video-channels/video-channels-routing.module.ts
```diff
@@ -37,6 +48,16 @@ const videoChannelsRoutes: Routes = [
         }
       }
     ]
+  },
+  {
+    path: ':videoChannelName/update',
```
Chocobozzz, line 53: `update/:videoChannelName` for consistency with the previous route

client/src/app/+video-channels/video-channels-routing.module.ts
```diff
@@ -37,6 +48,16 @@ const videoChannelsRoutes: Routes = [
...
+    component: VideoChannelUpdateComponent,
+
```
Chocobozzz, line 55: Extra `\n`

client/src/app/+video-channels/video-channels.component.html
```diff
@@ -6,11 +6,11 @@
   <div class="channel-info">
 
     <ng-template #buttonsTemplate>
-      <a *ngIf="isManageable()" [routerLink]="[ '/my-library/video-channels/update', videoChannel.nameWithHost ]" class="peertube-button-link orange-button" i18n>
+      <a *ngIf="isManageable()" [routerLink]="[ 'update' ]" class="peertube-button-link orange-button" i18n>
```
Chocobozzz, line 9: I think you need `videoChannel.nameWithHost` in `[routerLink]`

client/src/app/+video-channels/video-channels.component.ts
```diff
@@ -98,10 +98,14 @@ export class VideoChannelsComponent implements OnInit, OnDestroy {
     return this.authService.isLoggedIn()
   }
 
+  isOwner () {
+    return this.videoChannel?.ownerAccount.userId === this.authService.getUser().id
```
Chocobozzz, line 102: Prepend with `if (!this.userLoggedIn()) return false`

server/middlewares/user-right.ts
```diff
@@ -20,8 +20,34 @@ function ensureUserHasRight (userRight: UserRight) {
   }
 }
 
+async function ensureUserCanManageChannel (req: express.Request, res: express.Response, next: express.NextFunction) {
```
Chocobozzz, line 23: Put this function in `server/middlewares/validators/users` instead

I think you can remove `ensureAuthUserOwnsChannelValidator` because all calls can be replaced by this new function

server/middlewares/user-right.ts
```diff
@@ -20,8 +20,34 @@ function ensureUserHasRight (userRight: UserRight) {
...
+  const user = res.locals.oauth.token.user
+  const isUserOwner = res.locals.videoChannel.Account.userId !== user.id
```
Chocobozzz, line 25: `===` no?

server/middlewares/user-right.ts
```diff
@@ -20,8 +20,34 @@ function ensureUserHasRight (userRight: UserRight) {
...
+
+  if (isUserOwner && user.hasRight(UserRight.MANAGE_ANY_VIDEO_CHANNEL) === false) {
```
Chocobozzz, line 27: Then `!isUserOwner`

server/middlewares/user-right.ts
```diff
@@ -20,8 +20,34 @@ function ensureUserHasRight (userRight: UserRight) {
...
+    const message = `User ${user.username} does not have right to manage channel ${req.params.nameWithHost}.`
+    logger.info(message)
```
Chocobozzz, line 29: Please remove

server/middlewares/user-right.ts
```diff
@@ -20,8 +20,34 @@ function ensureUserHasRight (userRight: UserRight) {
...
+
+    return res.fail({
+      status: HttpStatusCode.FORBIDDEN_403,
+      message
+    })
+  }
+
+  const onUser = await res.locals.videoChannel.Account.$get('User')
+  if (user.role === UserRole.MODERATOR && onUser.role === UserRole.ADMINISTRATOR) {
```
Chocobozzz, line 38: I don't think we need this.

IMHO a moderator should be trusted, and could update admin's channels.

server/middlewares/validators/videos/video-channels.ts
```diff
@@ -151,31 +149,8 @@ export {
 
 // ---------------------------------------------------------------------------
 
-function checkUserCanDeleteVideoChannel (user: MUser, videoChannel: MChannelAccountDefault, res: express.Response) {
-  if (videoChannel.Actor.isOwned() === false) {
```
Chocobozzz, line 155: I think this `if` should be used to check all routes except `/:nameWithHost/followers`

You could put this part in a dedicated middleware, and call it in controllers

client/src/app/+video-channels/video-channels-routing.module.ts
kontrollanten: That would collide with a channel named "create"

server/middlewares/user-right.ts
kontrollanten: That's right 😄

client/src/app/+video-channels/video-channels-routing.module.ts
kontrollanten: Then it will collide if there's a channel named "update"?

client/src/app/+video-channels/video-channels-routing.module.ts
Chocobozzz: :o Nice catch

Then maybe `/manage/create` and `/manage/update/:videoChannelName`?

kontrollanten: Isn't it better with `/c/@create` and `/c/@update/:videoChannelName` then?

Chocobozzz: It's just you introduce this `@` which is a good idea but not consistent with the rest of the project

I would expect to also have `/my-library/video-playlists/@create` which is not the case and I don't want to update all these routes :grin:

server/middlewares/validators/videos/video-channels.ts
kontrollanten: Should we really use this check on regular GET routes?

kontrollanten: There's middleware called `localVideoChannelValidator` but it returns a 404 if the video channel isn't found on the local server. It's only used by activity pub routes, should I move that middleware to `validators/activitypub/video-channel.ts` and then create a new middleware in `validators/videos/video-channel.ts`?

Chocobozzz: I think I'll remove `localVideoChannelValidator`
Thanks!

Chocobozzz:
Thanks

PR #4616, opened 2021-12-08
scripts/test: make MacOS compatible

kontrollanten:
Add support for running tests with GNU xargs (which are used by MacOS).

```
man xargs:
     -r      Compatibility with GNU xargs.  The GNU version of xargs runs the utility argument at least once, even if xargs input is empty, and it supports a -r option to inhibit this
             behavior.  The FreeBSD version of xargs does not run the utility argument on empty input, but it supports the -r option for command-line compatibility with GNU xargs, but the
             -r option does nothing in the FreeBSD version of xargs.
```

### DIFF
--- a/scripts/clean/server/test.sh
+++ b/scripts/clean/server/test.sh
@@ -20,9 +20,9 @@ dropRedis () {
   port=$((9000+$1))
   host="localhost"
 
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks!

PR #4624, opened 2021-12-10
Add plugin helper function to get videos storage path

lutangar:
## Description

Add plugin helper function to get videos storage path `CONFIG.STORAGE.VIDEOS_DIR`.
At the moment, a plugin doesn't have a clue on where a video is actually stored and the video file isn't accessible deterministically.

I actually exposed only the videos storage path but maybe other storage info should be exposed.
OR maybe none of the storage informations should leak to the plugin.

About the naming you might `getVideosdirectoryPath` instead of `getVideosStoragePath`, your call.

## Has this been tested?

### DIFF
--- a/server/lib/plugins/plugin-helpers-builder.ts
+++ b/server/lib/plugins/plugin-helpers-builder.ts
@@ -148,6 +148,10 @@ function buildConfigHelpers () {
 
     getServerConfig () {
       return ServerConfigManager.Instance.getServerConfig()
+    },
+
+    getVideosStoragePath () {
+      return CONFIG.STORAGE.VIDEOS_DIR
     }
   }
 }
--- a/server/tests/fixtures/peertube-plugin-test-four/main.js
+++ b/server/tests/fixtures/peertube-plugin-test-four/main.js
@@ -76,6 +76,12 @@ async function register ({
       return res.json({ serverConfig })
     })
 
+    router.get('/videos-storage-path', async (req, res) => {
+      const videosStoragePath = await peertubeHelpers.config.getVideosStoragePath()
+
+      return res.json({ videosStoragePath })
+    })
+
     router.get('/static-route', async (req, res) => {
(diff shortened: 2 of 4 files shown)


Chocobozzz:
Hello,

If the main goal is to get video file paths, maybe you should create a dedicated function for that purpose. Because it's still difficult for you with this `getVideosStoragePath` function to understand where the file is located (it could have multiple resolutions, could be a fragmented mp4 so located in hls directory etc).

lutangar:
Alright thanks, I understand... My goal is in fact to extract the video audio. In order to do that I'm feeding the video stream to `ffmpeg`. So as long as I can get a `stream` of the video I'll be fine.

I'll probably propose to expose an helper method to the `VideosHelpers` object instead.
But on which method should I rely instead ? As of now I'm relying on the first `VideoFileModel` of the video `videoFiles[0]?.filename` which is very weak from what I understand now (but since I only care about the sound I don't really care about the different resolutions).
Or is there a service method which could get me the video stream maybe?

Chocobozzz:
I'll develop an helper that will give you the paths of a video. Something like `videos.getFilesPath(videoId: number)` with the following result:

```
videoFiles: {
  webtorrent: {
    resolution: number
    path: string
  }
  hls: {
  ...
  }
}
thumbnails: {
  type: thumbnailtype
  path: string
}
```

I could also create an helper so plugins can get the JSON output of ffprobe for a specific path if you need it

Chocobozzz:
Would the API described above be okay for you?

lutangar:
I suppose it would!
I wouldn't want to add something too specific to the plugin API though. 
But if you think it does make sense for broader use-cases then let's do this! :sparkles:

Chocobozzz:
Here you go: https://github.com/Chocobozzz/PeerTube/commit/2e9c7877eb3a3c5d64cc5c3383f0a7c0b51f5481 https://github.com/Chocobozzz/PeerTube/commit/754c52b9b998a825fd830d5ac527a67e0eefeb9a

Issue #1489, opened 2018-12-15
Add an option for language: "This video has no language"


multimob:
(enhancement)
If one does not fill the language drop-down box, the video language will be shown as "Unknown".

I suggest to add an extra "language" in the list, something that would say "No language".  To be used for instrumental music or any video with no words in the audio track.

Why?
1. Some people might want to filter videos and select only languages they understand
-- "Unknown" = possibly a foreign language I do not know → default could be not to show.
-- "No language" = universally understandable → default could be to show.
2. Instance admins could easily spot whether there are videos with missing information; today they cannot see whether it was intentionally left blank or whether the uploader forgot to fill the information.

My PeerTube version: v1.1.0

r4dh4l:
> I suggest to add an extra "language" in the list, something that would say "No language". To be used for instrumental music or any video with no words in the audio track.

As mentioned in https://github.com/Chocobozzz/PeerTube/issues/2299:

It would be great to have additional options to specify "multiple languages" selection:
- Main language: (dropdown)
- With contributions in following languages:
(1st dropdown)
+ (2nd dropdown)
+ (3rd dropdown)
...

n76:
This issue is nearly two years old but I’d like to add a vote for implementing this.

PR #4631, opened 2021-12-12
Added "zxx" (no linguistic content) in available video languages

Poslovitch:
## Description

It adds `zxx` as part of the available "language codes" in the backend. `zxx` is the ISO-639 code for the absence of linguistic content, according to Wikipedia (https://en.wikipedia.org/wiki/ISO_639): 

> No linguistic information at all (added 2006-01-11). The content (e.g. graphics, photos or audio/video records not including text in a human language, or technical metadata and most programming source code) is usable as is with any language and should not be translated (except for its description possibly associated in separate contents, or for non-essential fragments of the content).

As this "language" also appeared in the group "All languages" in the forms, I had to create a dedicated group for it. Feel free to suggest better names 😉.

## Related issues

Closes https://github.com/Chocobozzz/PeerTube/issues/1489

## Has this been tested?

## Screenshots

In the video-edit form:

image

image

When watching the video : 
image

### DIFF
--- a/client/src/app/+videos/+video-edit/shared/video-edit.component.ts
+++ b/client/src/app/+videos/+video-edit/shared/video-edit.component.ts
@@ -178,10 +178,11 @@ export class VideoEditComponent implements OnInit, OnDestroy {
       .subscribe(res => {
         this.videoLanguages = res.languages
           .map(l => {
-            if (l.id === 'zxx') return { ...l, group: $localize`Special values`, groupOrder: -1 }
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Thanks!

WonderingDane:
This is a harmful feature imo. The selected language should or could also reflect the language of the title of the video. Using "no linguistic content" would only make it even harder to filter out content in languages you do not understand.

PR #4640, opened 2021-12-15
Fix plugin storage return value when storing a Json array

lutangar:
## Description

Fix plugin storage return value when storing a Json array.
We may as well entirely rely on `JSON.parse` which is foolproof.

## Has this been tested?


Chocobozzz:
Thanks!

Just fixed the tests because `typeof [ 1, 2 ]` is `object`

lutangar:
> Just fixed the tests because typeof [ 1, 2 ] is object

Thanks a lot! :tada:

PR #4654, opened 2021-12-18
Add config option to disable video history for new users

### LINKED ISSUE #1499: Add ability to disable videos history by default
In admin configuration.

See https://github.com/Chocobozzz/PeerTube/issues/1200#issuecomment-427305084

Poslovitch:
## Description

I added a config option (`user.history.enabled_by_default`) and a checkbox in the web config editor that allows admins to enable/disable video history by default for new users.

## Related issues

Closes #1499

## Has this been tested?

What unit tests should I do, and where should I put them in?

Aside from these, I manually tested the feature and it worked. 

## Screenshots

image

### DIFF
--- a/server/controllers/api/users/index.ts
+++ b/server/controllers/api/users/index.ts
@@ -188,6 +188,7 @@ async function createUser (req: express.Request, res: express.Response) {
     role: body.role,
     videoQuota: body.videoQuota,
     videoQuotaDaily: body.videoQuotaDaily,
+    videosHistoryEnabled: CONFIG.USER.HISTORY.ENABLED_BY_DEFAULT,
     adminFlags: body.adminFlags || UserAdminFlag.NONE
   }) as MUser
 
--- a/server/controllers/api/users/index.ts
+++ b/server/controllers/api/users/index.ts
@@ -238,6 +238,7 @@ async function registerUser (req: express.Request, res: express.Response) {
     role: UserRole.USER,
     videoQuota: CONFIG.USER.VIDEO_QUOTA,
     videoQuotaDaily: CONFIG.USER.VIDEO_QUOTA_DAILY,
+    videosHistoryEnabled: CONFIG.USER.HISTORY.ENABLED_BY_DEFAULT,
     emailVerified: CONFIG.SIGNUP.REQUIRES_EMAIL_VERIFICATION ? false : null
   })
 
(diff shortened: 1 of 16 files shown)


server/controllers/api/users/index.ts
```diff
@@ -188,6 +188,7 @@ async function createUser (req: express.Request, res: express.Response) {
     role: body.role,
     videoQuota: body.videoQuota,
     videoQuotaDaily: body.videoQuotaDaily,
+    videosHistoryEnabled: CONFIG.USER.HISTORY.ENABLED_BY_DEFAULT,
```
Chocobozzz, line 191: Also add it in https://github.com/Chocobozzz/PeerTube/blob/develop/server/lib/auth/oauth-model.ts#L224

server/controllers/static.ts
```diff
@@ -275,7 +275,10 @@ async function generateNodeinfo (req: express.Request, res: express.Response) {
         },
         user: {
           videoQuota: CONFIG.USER.VIDEO_QUOTA,
-          videoQuotaDaily: CONFIG.USER.VIDEO_QUOTA_DAILY
+          videoQuotaDaily: CONFIG.USER.VIDEO_QUOTA_DAILY,
+          history: {
```
Chocobozzz, line 279: We don't need this info here

server/lib/server-config-manager.ts
```diff
@@ -204,7 +204,10 @@ class ServerConfigManager {
       },
       user: {
         videoQuota: CONFIG.USER.VIDEO_QUOTA,
-        videoQuotaDaily: CONFIG.USER.VIDEO_QUOTA_DAILY
+        videoQuotaDaily: CONFIG.USER.VIDEO_QUOTA_DAILY,
+        history: {
```
Chocobozzz, line 208: We don't need this info here

server/tests/api/server/config.ts
```diff
@@ -60,6 +60,7 @@ function checkInitialConfig (server: PeerTubeServer, data: CustomConfig) {
 
   expect(data.user.videoQuota).to.equal(5242880)
   expect(data.user.videoQuotaDaily).to.equal(-1)
+  expect(data.user.history.enabledByDefault).to.be.false
```
Chocobozzz, line 63: I'm not sure to understand why it's `false` by default :thinking:

Chocobozzz (commented):
For the tests you can add a section in https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/server/config-defaults.ts

Where you set the default history to false, then create and register users and check their history value.

Chocobozzz:

PR #4664, opened 2021-12-22
Move zxx to its own group in select-languages component

Poslovitch:
## Description

#4631 introduced the `zxx` language code and put it in a "Other" group in the language selector of the video-edit page. However, `zxx` was not in a dedicated group in the more widely-used `my-select-languages` component.

This PR adds a `groupOrder` attribute to `SelectOptionsItem` and puts `zxx` in group "Other" (order 0), and all languages remain in the "All languages" group (order 1).

## Related issues

None.

## Has this been tested?

## Screenshots

image

### DIFF
--- a/client/src/app/shared/shared-forms/select/select-languages.component.ts
+++ b/client/src/app/shared/shared-forms/select/select-languages.component.ts
@@ -39,12 +39,12 @@ export class SelectLanguagesComponent implements ControlValueAccessor, OnInit {
     this.server.getVideoLanguages()
       .subscribe(
         languages => {
-          this.availableLanguages = [{
+          this.availableLanguages = [ {
             label: $localize`Unknown language`,
             id: '_unknown',
             group: this.allLanguagesGroup,
             groupOrder: this.allLanguagesGroupOrder
-          }]
+          } ]
 
           this.availableLanguages = this.availableLanguages
             .concat(languages.map(l => {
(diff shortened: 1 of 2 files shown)


client/src/types/select-options-item.model.ts
```diff
@@ -4,6 +4,7 @@ export interface SelectOptionsItem {
   description?: string
   group?: string
   groupLabel?: string
+  groupOrder?: number
```
Chocobozzz, line 7: Don't add this to the global type, it's only supported on `select-languages` component

client/src/app/shared/shared-forms/select/select-languages.component.ts
```diff
@@ -23,6 +23,7 @@ export class SelectLanguagesComponent implements ControlValueAccessor, OnInit {
   availableLanguages: SelectOptionsItem[] = []
 
   allLanguagesGroup = $localize`All languages`
+  allLanguagesGroupOrder = 1
```
Chocobozzz, line 26: You don't need to create a variable here, prefer to put `1` line `52` so we can easily identify the difference between the two groups

client/src/types/select-options-item.model.ts
Poslovitch: Without this, I cannot compile the code due to an error in : `this.availableLanguages = [ { label: $localize`Unknown language`, id: '_unknown', group: this.allLanguagesGroup, groupOrder: this.allLanguagesGroupOrder } ]` (namely the fact there is no `groupOrder` attribute in the `SelectOptionsItem`.

Chocobozzz: Because you need to extend `availableLanguages` type: `availableLanguages: (SelectOptionsItem & { groupOrder: number })[]`

Chocobozzz:
Thanks!

Issue #4682, opened 2022-01-02
Cannot fetch remote information of this URL.


tio-trom:
I have read past issues opened about this issue but cannot find a solution. This happened after the latest upgrade to peertube 4. 

* PeerTube instance: https://videos.trom.tf
  * URL: https://videos.trom.tf
  * version: 4.0.0
  * NodeJS version:
  * Ffmpeg version:

* Browser name, version and platforms on which you could reproduce the bug: Firefox
```

info[1/2/2022, 4:59:52 PM] Cannot fetch information from import for URL https://www.youtube.com/watch?v=AsPmJdlrBiI.

{
  "err": {
    "stack": "Error: Command failed with exit code 1: python /home/yunohost.app/peertube/storage/bin/yt-dlp --dump-json -f bestvideo[vcodec^=avc1][height=2160]+bestaudio[ext=m4a]/bestvideo[vcodec!*=av01][vcodec!*=vp9.2][height=2160]+bestaudio/bestvideo[vcodec^=avc1][height<=2160]+bestaudio[ext=m4a]/bestvideo[vcodec!*=av01][vcodec!*=vp9.2]+bestaudio/best[vcodec!*=av01][vcodec!*=vp9.2]/best https://www.youtube.com/watch?v=AsPmJdlrBiI\nTraceback (most recent call last):\n  File \"/usr/lib/python2.7/runpy.py\", line 174, in _run_module_as_main\n    \"__main__\", fname, loader, pkg_name)\n  File \"/usr/lib/python2.7/runpy.py\", line 72, in _run_code\n    exec code in run_globals\n  File \"/home/yunohost.app/peertube/storage/bin/yt-dlp/__main__.py\", line 16, in <module>\n  File \"/home/yunohost.app/peertube/storage/bin/yt-dlp/yt_dlp/__init__.py\", line 4\n    f'You are using an unsupported version of Python. Only Python versions 3.6 and above are supported by yt-dlp'  # noqa: F541\n                                                                                                                ^\nSyntaxError: invalid syntax\n    at makeError (/var/www/peertube/node_modules/execa/lib/error.js:60:11)\n    at handlePromise (/var/www/peertube/node_modules/execa/index.js:118:26)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)",
[...]

PR #4706, opened 2022-01-09
Give the ability to define python path for youtube-dl / yt-dlp

fflorent:
## Description

On Debian, the `python` often points to `/usr/bin/python2`.

This PR aims at giving the ability to define the python path so the administrator can choose the path to the python with which PeerTube executes youtube-dl or yt-dlp.

## Related issues

 - https://github.com/Chocobozzz/PeerTube/issues/4682
 - https://github.com/YunoHost-Apps/peertube_ynh/issues/295

## Has this been tested?

(*not sure whether it is relevant here as long as the tests continue passing ?*)

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -446,6 +446,9 @@ import:
         # yt-dlp is also supported
         name: 'youtube-dl'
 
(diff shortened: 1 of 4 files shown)


Chocobozzz:
Thanks!

PR #4709, opened 2022-01-10
Add `me` attribute to `rel` for links for Mastodon verification

ThatOneCalculator:
It's a super simple change, but it'll allow people who put their Mastodon links to have them be verified. 
Screenshot_2022-01-10-12-42-49_2560x1440
Screenshot_2022-01-10-12-43-02_2560x1440


Chocobozzz:
Hello,

You are setting the `me` `rel` attribute to every link built using markdown

ThatOneCalculator:
> Hello,
> 
> You are setting the `me` `rel` attribute to every link built using markdown

Ah, gotcha. How would you recommend going about it for *only* the channel description?

ThatOneCalculator:
Actually @Chocobozzz I think I did it. Could you check and see if I did it right?

Chocobozzz:
I don't think it will work because you may put URL in channel's description that would not be related to you

ThatOneCalculator:
> I don't think it will work because you may put URL in channel's description that would not be related to you

The `me` attribute is only for semantic purposes regardless, I think adding it for Mastodon/Pleroma/etc users would give more benefit than not, given the peertube link would have to be on the link sent regardless.

Chocobozzz:
We prefer to not add semantic meaning on links based only on assumptions. It's just semantic but could be interpreted by remote software. For example, you could write in your description that you're a student of a [school](https://example.com), but we don't want other software to understand the school and you share the same identity.

So I'm sorry for this PR work, but I think Mastodon's system with key/value is the way to go instead.

PR #4711, opened 2022-01-10
Fix missing user.Account.name on getAuthUser return type.

JohnXLivingston:
## Description

I'm trying to use the new @peertube/peertype-types npm package for my peertube-plugin-livechat.
The function getAuthUser return type was missing the user.Account.name property.
This is an expected property, so that we can have the user displayName. See for example this test file:
https://github.com/Chocobozzz/PeerTube/blob/73e64592b9b949bf9b895e2fae8dafcfab0b5124/server/tests/fixtures/peertube-plugin-test-four/main.js#L101

## Has this been tested?


Chocobozzz:
Thanks!

PR #4736, opened 2022-01-20
Fix plugin settings manager definition

lutangar:
## Description

This mainly fix the `onSettingsChange` argument typing.
I'm not 100% sure the setting `value` can be a boolean though.
But this is how it was typed before so I just made it consistent.


Chocobozzz:
Thanks!

PR #4750, opened 2022-01-29
Fix HLS transcoding upon WebTorrent transcoding

kontrollanten:
## Description
Previously the web torrent transcoding was the on initializing HLS transcoding. This PR will separate the logic and make the HLS transcoding independent of the WebTorrent transcoding.

## Related issues
closes #4749 

## Has this been tested?
- [x] 👍 yes, I added tests to the test suite

PR #4754, opened 2022-01-30
Fix retranscoding HLS with object storage enabled

### LINKED ISSUE #4746: Retranscoding doesn't work properly with object storage
When object storage is used and create-transcoding-job is ran, the master playlist and segments file won't be moved to object storage nor updated in the database. The transcoded files will be some kind of orphans that's not connected to any video in the database.

**Steps to reproduce:**

1. Turn on object storage
2. Upload a video
3. Run npm run create-import-video-file-job
4. Run npm run create-transcoding-job

master playlist and segments file should be updated so that the transcoding job will have effect.

**Additional information**
I think it's this `continue` statement that breaks it; https://github.com/Chocobozzz/PeerTube/blob/ebe4b3df5c4e595f3fcf92b8ecf6c5174b13b3b0/server/lib/job-queue/handlers/move-to-object-storage.ts#L85-L87

What's the point with doing this check?

To reduce the complexity it may be easier to create new entries of videoFile and videoStreamingPlaylists in the database upon video-file-import. That would need to have some kind of `active` column in the database to know which videoFile/videoStreamingPlaylist is the accurate. That would also make it possible to rollback a file import, and will probably also help to create a video journal as proposed in https://github.com/Chocobozzz/PeerTube/issues/4320

* PeerTube instance:
  * version: 4.0

kontrollanten:
## Description
Upon trying to solve #4746 I found the solution became too complex in handling different cases. So instead I did some refactor that in broad terms means that when doing retranscoding a new HLS playlist is created, instead of updating the existing. Possible benefits (each would need some additional code):

* Avoid disruption or downtime during transcoding
* Rollback a broken retranscoding/video file import
* Preview a new video file import before publishing

What has changed:
* Video streaming playlists unique index has changed from video, type > video, type, playlistFilename
* playlistFilename, p2pMediaLoaderPeerVersion and segmentsSha256Filename can now be null (during transcoding)
* WebTorrent and HLS transcoding jobs less coupled
* Removed support to transcode a particular resolution just to be consistent (with the new way of doing it this would have been creating a new playlist with only one resolution)

This PR is based on #4750.

## Related issues
closes #4746 

## Has this been tested?

### DIFF
--- a/client/src/app/+admin/overview/videos/video-list.component.html
+++ b/client/src/app/+admin/overview/videos/video-list.component.html
@@ -81,8 +81,8 @@
       </td>
 
       <td>
-        <span *ngIf="isHLS(video)" class="badge badge-blue">HLS</span>
-        <span *ngIf="isWebTorrent(video)" class="badge badge-blue">WebTorrent</span>
+        <span *ngIf="isHLS(video)" class="badge badge-blue">HLS ({{ video.streamingPlaylists.length }})</span>
+        <span *ngIf="isWebTorrent(video)" class="badge badge-blue">WebTorrent ({{ video.files.length }})</span>
         <span *ngIf="video.isLive" class="badge badge-blue">Live</span>
 
         <span *ngIf="!isImport(video) && !video.isLive && video.isLocal">{{ getFilesSize(video) | bytes: 1 }}</span>
@@ -112,11 +112,16 @@
           <div *ngIf="isHLS(video)">
             HLS:
 
-            <ul>
-              <li *ngFor="let file of video.streamingPlaylists[0].files">
-                {{ file.resolution.label }}: {{ file.size | bytes: 1 }}
-              </li>
-            </ul>
+            <div *ngFor="let playlist of video.streamingPlaylists; let i = index;" class="hls-playlist-list">
+              <strong>Playlist {{ playlist.id }}</strong>
+              <span *ngIf="i === video.streamingPlaylists.length - 1" class="badge badge-yellow">Active</span>
+
+              <ul>
+                <li *ngFor="let file of playlist.files">
+                  {{ file.resolution.label }}: {{ file.size | bytes: 1 }}
+                </li>
+              </ul>
+              </div>
           </div>
 
           <my-embed class="ml-auto" [video]="video"></my-embed>
--- a/client/src/app/+admin/overview/videos/video-list.component.scss
+++ b/client/src/app/+admin/overview/videos/video-list.component.scss
@@ -14,8 +14,17 @@ my-embed {
 
 .video-info > div {
   display: flex;
+
+  > div {
+    width: 25%;
+  }
 }
 
 .loading {
   opacity: 0.5;
 }
+
+.hls-playlist-list {
+  padding-left: 15px;
+}
+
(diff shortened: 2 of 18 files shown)


Chocobozzz:
Hi and thanks for this PR!

Unfortunately it adds complexity and potential bugs since this changes the mental model of a peertube streaming playlist (only one playlist per type, without null values for filenames for example). Moreover having multiple versions of the same playlist is an headache to maintain for me and for admins (disk storage, we would have to add a management system to delete them etc).

Regarding the benefits:

> Avoid disruption or downtime during transcoding

I think we could handle this case in the client or use a cache to handle this specific case. But we would need more details on these disruption/downtime: when does this happen? After a move to object storage job? Before?

> Rollback a broken retranscoding/video file import

Maintaining multiple versions of a video is a good idea (which requires a lot of work though) but we should not use the playlist table and provide all versions to the client in the same `GET /video/:id` endpoint. Instead, we should store previous versions in a dedicated table/directory (ideally just keeping the max resolution to reduce disk storage).

> Preview a new video file import before publishing

Same as the previous point, even if I don't think this feature is really worth it.

So, we decided to fix this in an alternative way:
 * Adding an argument to explicitly specify if we want to automatically generate HLS versions (https://github.com/Chocobozzz/PeerTube/commit/0f11ec8dd32b50897c18588db948e96cf0fc2c70)
 * Uploading new HLS master playlist/segment sha file versions to object storage on re transcoding (https://github.com/Chocobozzz/PeerTube/commit/a2caee9f5162232234de2e8aae6957cc7f38c853)

We tried to keep some interesting code of this PR (like the number of webtorrent files in the UI, or additional logs in the move to object storage job handler). 

As a side note, while the "Run WebTorrent" transcoding option in web admin should only generate webtorrent version of a video, the CLI script `create-transcoding-job` creates a transcoding job according to the instance configuration. So it's expected to also run HLS transcoding when we use this script if the instance enabled HLS transcoding.

kontrollanten:
> Moreover having multiple versions of the same playlist is an headache to maintain for me and for admins (disk storage, we would have to add a management system to delete them etc).

Can't we just solve that with running a cleanup job upon the last transcoding job (pendingTranscode === 0)? If we add rollback support, then it can be configurable to cleanup old versions after X days.

> > Avoid disruption or downtime during transcoding
> 
> I think we could handle this case in the client or use a cache to handle this specific case. But we would need more details on these disruption/downtime: when does this happen? After a move to object storage job? Before?

From the first transcoding job until the last. If I remember correctly the playlist is cleaned empty upon the first, and then the master playlist is created in the last, during that time the video will be unavailable. Generally I think it's a bad idea to manipulate the streaming playlists when they're live, if something goes wrong during the way it's no way to fix a broken video (that may be widely spread and has a a lot of viewers).

> > Rollback a broken retranscoding/video file import
> 
> Maintaining multiple versions of a video is a good idea (which requires a lot of work though) but we should not use the playlist table and provide all versions to the client in the same `GET /video/:id` endpoint. Instead, we should store previous versions in a dedicated table/directory (ideally just keeping the max resolution to reduce disk storage).

I agree upon not serving all versions in the GET /video:id endpoint (my plan was to filter them out in another PR, by just showing the latest playlist with available resolutions).

Issue #3407, opened 2020-12-06
Verify video validity


kontrollanten:
**Describe the problem to be solved**
We've imported a YouTube video to PeerTube HLS. It appeared that the PeerTube video where corrupt at a specific format (720p), which meant that the video crashed at a certain minute during a specific resolution. I downloaded the file and rand the following command.

```
$ ffmpeg -v error -i [uuid]-720-fragmented.mp4 -f null
[h264 @ 0x56055d962580] Invalid NAL unit size (0 > 1569).
[h264 @ 0x56055d962580] Error splitting the input into NAL units.
[h264 @ 0x56055d938d80] error while decoding MB 43 4, bytestream -6
Error while decoding stream #0:0: Invalid data found when processing input
[+ 15 000 similar lines]
```
I've tested the 240p version of the video, and it works fine. After doing a new import it works fine.

**Describe the solution you would like:**
Since data corruption always is a risk when working with data, I guess we should find a way to detect corruptions to avoid a bad user experience. I'd propose a solution where an admin can opt in for validation. When validation is enabled the user have three choices:

A) Don't publish a video before validation for all resolutions has succeeded.
B) Publish the video and all resolutions that didn't fail. If a specific resolution fails, then let the user to replace the file.
C) Run validation after publish and just send a notification the user.

**Describe alternatives you have considered**  

I think the easiest way forward would be to create a plugin that listens for video creations. Upon every creation the file is validated by running ffmpeg. If they fails a notification is sent to the user.

Chocobozzz:
Hello,

Do you have errors in your logs related to this video/resolution?

kontrollanten:
Unfortunately I couldn't find anything. The transcoding job reported no errors and I couldn't find anything somewhere else either.

Chocobozzz:
Do you think your issue is related to https://github.com/Chocobozzz/PeerTube/issues/3596?

kontrollanten:
I don't thinks so. The issue with #3596 lead to audio only was played during the whole video. But in this issue the video just crashed at a certain minute.

Chocobozzz:
https://github.com/Chocobozzz/PeerTube/commit/aaedadd5386b580e9ebac540201399c25c7f0b0f should fix this issue

kontrollanten:
We've found three corrupt videos the last week, so I'd say this would be a really helpful feature.

Chocobozzz:
[...]

PR #4756, opened 2022-01-31
Fix media player error handling

### LINKED ISSUE #4748: HLS: Video player tries to fallback non-existent webtorrent
When a HLS file for some reasons is corrupt and the media player can't proceed to play it, it will just stop and display a black screen in the media player with a loading spinner that continues for some minutes. After some minutes an error message is displayed for the user; "The media playback was aborted due to a corruption problem or because the media used features your browser did not support.".

**Steps to reproduce:**

1. Make a HLS video corrupt (I don't know exactly how to do that manually though)
2. Play it

At least the error message should be displayed earlier, few users will wait for many minutes. It would be even better if the player tries to go some seconds forward in the video, to skip the corrupted part.

**Additional information**
* Web Torrent disabled.
* HLS enabled, with 240, 360, 480 and 720 resolutions.

Even though web torrent is disabled, it will try to fallback to web torrent and it seems to succeed in fetching the web torrent for resolution 360, and then it will fetch the fragmented mp4 for some seconds.

* PeerTube instance:
  * version: 4.0

Latest Chrome, MacOS
Immediately upon crash the following is printed to the web console:
```
VIDEOJS: ERROR: (CODE:3 MEDIA_ERR_DECODE) The media playback was aborted due to a corruption problem or because the media used features your browser did not support. 
Fallback to webtorrent.
Adding magnet:?xs=...
Adding magnet:?xs=...
ERROR: (CODE:4 MEDIA_ERR_SRC_NOT_SUPPORTED) The media could not be loaded, either because the server or network failed or because the format is not supported.
Destroying IndexDB database webtorrent-chunks-3f42c441b9b9e547e3a0c7912be07127589aa3b2.
Removed magnet:?xs=...
webtorrent-plugin.ts:355 DOMException: Failed to load because no supported source was found.
Dexie.delete('webtorrent-chunks-3f42c441b9b9e547e3a0c7912be07127589aa3b2') was blocked
```

kontrollanten:
## Description
* Don't try to fallback to webtorrent when webtorrent is missing.
* Display a notification for the user when decoding issues appear.
* Try to fast forward to get rid of the decoding issue.

## Related issues
closes #4748

## Has this been tested?

### DIFF
--- a/client/src/app/+videos/+video-watch/video-watch.component.ts
+++ b/client/src/app/+videos/+video-watch/video-watch.component.ts
@@ -83,6 +83,10 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
 
   private hotkeys: Hotkey[] = []
 
+  private anonymousOrLoggedUser: User
+
+  private videojsDecodeErrors = 0
+
   constructor (
     private elementRef: ElementRef,
     private route: ActivatedRoute,
@@ -245,23 +249,14 @@ export class VideoWatchComponent implements OnInit, OnDestroy {
       this.userService.getAnonymousOrLoggedUser()
     ]).subscribe({
       next: ([ video, captionsResult, loggedInOrAnonymousUser ]) => {
-        const queryParams = this.route.snapshot.queryParams
+        this.anonymousOrLoggedUser = loggedInOrAnonymousUser
 
         const urlOptions = {
-          resume: queryParams.resume,
-
-          startTime: queryParams.start,
-          stopTime: queryParams.stop,
-
-          muted: queryParams.muted,
-          loop: queryParams.loop,
-          subtitle: queryParams.subtitle,
-
-          playerMode: queryParams.mode,
+          ...this.buildUrlOptions(),
           peertubeLink: false
         }
 
-        this.onVideoFetched({ video, videoCaptions: captionsResult.data, loggedInOrAnonymousUser, urlOptions })
+        this.onVideoFetched({ video, videoCaptions: captionsResult.data, urlOptions })
             .catch(err => this.handleGlobalError(err))
       },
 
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Hi,

I think we should just display an error message instead of trying to recover the error:
 * This error is rare but fatal
 * This fix works for your specific error case, but we're not sure it works in most of them
 * It might be fixed in future HLS/ffmpeg versions
 * Users prefer that admins fix this error completely, by rencoding the video. An error message is the better way to be notified about a corrupted video
 * You add code in the watch page, but what about the embed one?
 * You add player error handling code in watch page, but it should be handled by the player manager (so this PR adds complexity and could conflict with player manager error handling)

Then, I suggest to keep the part where you check there is available webtorrent fallback, and if it does not correctly display an error message.

kontrollanten:
>  * This error is rare but fatal

Yes, but we've got it two times the last week.

> * This fix works for your specific error case, but we're not sure it works in most of them

It will just try 20 times, otherwise an error will be shown. Shouldn't that be safe in all cases?

> * Users prefer that admins fix this error completely, by rencoding the video. An error message is the better way to be notified about a corrupted video

In our case it doesn't work to reencode it though. Today there's no existing functionality to fix it, but remove the video and upload a new.

> * It might be fixed in future HLS/ffmpeg versions

Yes, this specific error, but shouldn't we always assume that some files will be corrupt and write the logic according to that?

I think that a long term solution should be either

A) This PR + functionality for the channel owner to update the video.
B) Video validation before publish as proposed in #3407

Chocobozzz:
I investigated a little bit and it seems the input file AAC stream is "broken", so ffmpeg **decoder** cannot correctly decodes the stream. `channelmap` audio filter may fix the file, but I did not test this solution: https://www.mail-archive.com/ffmpeg-user@ffmpeg.org/msg28011.html

Regarding your PR, I'll try to move the fast forward code to the player manager so it's also used by the embed.

kontrollanten:
> I investigated a little bit and it seems the input file AAC stream is "broken", so ffmpeg **decoder** cannot correctly decodes the stream. `channelmap` audio filter may fix the file, but I did not test this solution: https://www.mail-archive.com/ffmpeg-user@ffmpeg.org/msg28011.html

Thanks!

> Regarding your PR, I'll try to move the fast forward code to the player manager so it's also used by the embed.

I can do that if you want to?

Chocobozzz:
> I can do that if you want to?

No thanks! I'll also try to refactor this file :)

Chocobozzz:
Please test https://github.com/Chocobozzz/PeerTube/commit/c4207f978e23c77f09c4646b940dfd532281300f and tell me if it's okay for you :)

kontrollanten:
It seems that the fast forward part is missing?

Chocobozzz:
No the fast forward part should work :thinking: 

https://github.com/Chocobozzz/PeerTube/commit/c4207f978e23c77f09c4646b940dfd532281300f#diff-268545880f4b55b1d625c2b3d4eb5b07ab87092cdf6b15f30733392453482c87R98 & https://github.com/Chocobozzz/PeerTube/commit/c4207f978e23c77f09c4646b940dfd532281300f#diff-268545880f4b55b1d625c2b3d4eb5b07ab87092cdf6b15f30733392453482c87R152

kontrollanten:
I just tried it and it works well. Thanks a lot!

PR #4811, opened 2022-02-24
Possibility to set custom RTMP/RTMPS hostname

### LINKED ISSUE #4786: Possibility to configure hostname for rtmp
We're using a CDN which only accepts incoming traffic on port 80 and 443. Therefore incoming rtmp stream won't work for us.

Possible to configure which hostname is shown for the user when they get their rtmp URI. 

`live.rtmp/rtmps.hostname`, which defaults to web server hostname.

kontrollanten:
## Description
Adds config to set hostname for incoming RTMP/RTMPS connections.

## Related issues
closes #4786

## Has this been tested?

I couldn't find out how to easily change the config in a test case.


config/default.yaml
```diff
@@ -396,6 +396,7 @@ live:
   rtmp:
     enabled: true
     port: 1935
+    hostname: '' # Defaults to webserver.hostname
```
Chocobozzz, line 399: Prefer to not default to other configuration value, it's more difficult to maintain et understand for admins
Just put `'localhost'`

Chocobozzz:
Please check the config is set in https://github.com/Chocobozzz/PeerTube/blob/develop/server/initializers/checker-before-init.ts#L12

Issue #4812, opened 2022-02-24
python3 is missing in docker image for 4.1.0


LecygneNoir:
Hello !

It seems that for the recent 4.1.0 release, python3 was forgotten inside the docker image.
Thus it causes impossibility to use ytp-dl to fetch video :
```
 2022-02-24 12:55:26.786 Cannot fetch information from import for URL https://www.youtube.com/watch?v=6n-Ic9CMaL4.
 "shortMessage": "Command failed with ENOENT: /usr/bin/python3 /data/bin/youtube-dl --dump-json -f [...]
```

Indeed when looking inside the image, python is correctly installed (as in https://github.com/Chocobozzz/PeerTube/blob/develop/support/docker/production/Dockerfile.bullseye#L5) BUT it's python 2
```
# /usr/bin/python --version
Python 2.7.18
```

The binary directly called by peertube is not present:
```
# ls /usr/bin/python3
/usr/bin/python3: File not Found
```

When installing python3 directly inside the image, the import feature works:
```
apt update
apt install python3
=> import from peertube working
```

I have no idea why debian bullseyes still point to python2 when installing python, but I guess we need to specify python3 in the dockerfile :-/

Thanks!

LecygneNoir:
Okay, when trying to install the python virtual package inside the docker image, it auto-select `python-is-python2` instead of python-is-python3 😭 

```
# apt install python
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'python-is-python2' instead of 'python'
python-is-python2 is already the newest version (2.7.18-9).
```

This is the explanation why we have python2 instead of 3 in the docker image. However, I found anything from debian that explain this, and why it could be different in prior versions

Chocobozzz:
Fixed by https://github.com/Chocobozzz/PeerTube/pull/4813

PR #4813, opened 2022-02-24
Install python3 instead of global python in docker image

LecygneNoir:
Hello !

Regarding issue #4812, python virtual package uses python-is-python2 and so install python2 instead of python3
But we need to explicitly use python3 for PeerTube tools (especially ytp-dl)

May need an hotfix to propagate changes as currently import is probably broken for everyone using the official docker image :-/

## Related issues

Issue #4812 

## Has this been tested?

- [x ] 🙅 no, because this PR does not update server code

However, I test installing python3 inside the official docker image and it solves the problem with ytp-dl

### DIFF
--- a/support/docker/production/Dockerfile.bullseye
+++ b/support/docker/production/Dockerfile.bullseye
@@ -2,7 +2,7 @@ FROM node:14-bullseye-slim
 
 # Install dependencies
 RUN apt update \
- && apt install -y --no-install-recommends openssl ffmpeg python ca-certificates gnupg gosu build-essential curl \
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Oup's, thanks!

decentral1se:
Thanks for raising / fixing :rocket: Did this get hotfixed into https://hub.docker.com/layers/chocobozzz/peertube/v4.1.0-bullseye/images/sha256-37f6ea01873864a87a370920b5ad82ea3a9eb4497fed895fa8d27bc962854fdc?context=explore? We're re-deploying and not seeing the fix. `apt install python3` live does work ofc for now.

Chocobozzz:
@decentral1se should be available in 4.1.1 :)

PR #4826, opened 2022-03-01
open privacy info in user settings dialogue in new window

rriemann:
## Description

When a user clicks on *sharing system* in the user settings, then the privacy notice is opened, but still covered by the settings. Closing the settings is confusing. Hence, we open a new window with the privacy notice.

## Has this been tested?

- [x] 🙋 no, because I need help (I don't have a working dev environment here)


Chocobozzz:
Thanks!

PR #4857, opened 2022-03-14
add client.videos.upload.maxChunkSize config

kontrollanten:
## Description
Add client.videos.upload.maxChunkSize config. We're using a CDN with a max upload limit at 100 mb, therefore we need to configure this.

## Has this been tested?

### DIFF
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -639,6 +639,10 @@ client:
       # By default PeerTube client displays author username
       prefer_author_display_name: false
       display_author_avatar: false
+    upload:
+      # Max size, given in bytes, of upload chunks
+      # If not set it will be calculated based on network speed
+      max_chunk_size: 0
 
   menu:
     login:
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -639,10 +639,10 @@ client:
       # By default PeerTube client displays author username
       prefer_author_display_name: false
       display_author_avatar: false
-    upload:
-      # Max size, given in bytes, of upload chunks
-      # If not set it will be calculated based on network speed
-      max_chunk_size: 0
+    resumable_upload:
+      # Max size of upload chunks, e.g. '90MB'
+      # If null, it will be calculated based on network speed
+      max_chunk_size: null
 
   menu:
     login:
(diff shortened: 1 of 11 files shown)


config/production.yaml.example
```diff
@@ -639,6 +639,10 @@ client:
       # By default PeerTube client displays author username
       prefer_author_display_name: false
       display_author_avatar: false
+    upload:
+      # Max size, given in bytes, of upload chunks
+      # If not set it will be calculated based on network speed
+      max_chunk_size: 0
```
Chocobozzz, line 645: * Prefer using `null` value instead of `0`
 * I think we should parse bytes here
 * Add an example: (`'100MB'` for example)

Chocobozzz (changes requested):

config/production.yaml.example
kontrollanten: In the quota config we're using `-1` to "unset" the value, may it be inconsistent to use null here?

server/controllers/api/config.ts
```diff
@@ -171,6 +171,9 @@ function customConfig (): CustomConfig {
       videos: {
         miniature: {
           preferAuthorDisplayName: CONFIG.CLIENT.VIDEOS.MINIATURE.PREFER_AUTHOR_DISPLAY_NAME
+        },
+        upload: {
+          maxChunkSize: CONFIG.CLIENT.VIDEOS.UPLOAD.MAX_CHUNK_SIZE
```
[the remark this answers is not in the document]
kontrollanten: I assume you meant we should run it in the initializer?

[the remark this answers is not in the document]
Chocobozzz: Oup's yes!

config/production.yaml.example
Chocobozzz: `-1` for quota means "Unlimited", not "unset" :)

shared/models/server/custom-config.model.ts
```diff
@@ -57,6 +57,9 @@ export interface CustomConfig {
       miniature: {
         preferAuthorDisplayName: boolean
       }
+      resumableUpload: {
```
Chocobozzz, line 60: If you don't plan to expose this setting in admin config, you don't need to add it to `custom-config.model.ts`

Chocobozzz:
Thanks!

SKJoy:
An **environment** variable for this would be far more convenient.

PR #4861, opened 2022-03-15
Make object storage ACL configurable

### LINKED ISSUE #4850: Customize upload ACL when uploading in object storage
I'm using S3 object storage with a bucket that is not publicly readable. Instead, it feeds a CloudFront distribution. This works smashingly well and I'm very grateful that the ability to store videos to S3 was added to the code base.

After upgrading to 4.1.0, attempts to upload to the bucket result in an error:

```
2022-03-11 09:55:59.076 error: Cannot move video https://candidates-viz.ligo.org/videos/watch/9b938c28-9fe1-4726-8a0b-d84785d40cb3 to object storage. {
  "err": {
    "stack": "AccessDenied: Access Denied\n    at deserializeAws_restXmlPutObjectCommandError (/var/www/peertube/versions/peertube-v4.1.0/node_modules/@aws-sdk/client-s3/dist-cjs/protocols/Aws_restXml.js:8266:41)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async /var/www/peertube/versions/peertube-v4.1.0/node_modules/@aws-sdk/middleware-serde/dist-cjs/deserializerMiddleware.js:7:24\n    at async /var/www/peertube/versions/peertube-v4.1.0/node_modules/@aws-sdk/middleware-signing/dist-cjs/middleware.js:11:20\n    at async StandardRetryStrategy.retry (/var/www/peertube/versions/peertube-v4.1.0/node_modules/@aws-sdk/middleware-retry/dist-cjs/StandardRetryStrategy.js:51:46)\n    at async /var/www/peertube/versions/peertube-v4.1.0/node_modules/@aws-sdk/middleware-logger/dist-cjs/loggerMiddleware.js:6:22",
    "message": "Access Denied",
    "Code": "AccessDenied",
    ...
``` 
This is due to ACL being set to `public-read` in the `PutObjectCommand` in [server/lib/object-storage/shared/object-storage-helpers.ts](https://github.com/Chocobozzz/PeerTube/commit/0d4a3c62ab3036508ba1f38653580af80a4b631e#diff-5920e5b5f60e5689d78b3a47b4b60378d269f65589115abaeab6423689726f51). If I remove that option in code the upload works just fine.

What was the motivation for adding that to the S3 call? Would an addition to the config of an explicit ACL be the best way to mitigate this issue?

dougluce:
Override this value to allow uploads to non-public S3 buckets. Otherwise "AccessDenied: Access Denied" errors will end up in the log.

Fixes #4850


server/lib/object-storage/shared/object-storage-helpers.ts
```diff
@@ -25,6 +25,13 @@ type BucketInfo = {
   PREFIX?: string
 }
 
+type CommandInput =  {
```
Chocobozzz, line 28: You don't need to create a struct, you can use `PutObjectCommandInput` (import it from aws sdk)

dougluce: True enough. I didn't even look.

Chocobozzz:
Thanks!

PR #4867, opened 2022-03-18
Validate video files

### LINKED ISSUE #3407: Verify video validity

kontrollanten:
## Description
Runs a validation job after each transcoding. 

## Related issues
closes #3407

## Has this been tested?
No, it should be covered by the current tests. Some tests are failing though, I don't know why. When I run the tests locally all of them succeeds.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,7 +44,7 @@ jobs:
     env:
       PGUSER: peertube
       PGHOST: localhost
-      NODE_PENDING_JOB_WAIT: 250
+      NODE_PENDING_JOB_WAIT: 450
       ENABLE_OBJECT_STORAGE_TESTS: true
 
     steps:
--- a/server/helpers/ffmpeg/ffmpeg-validate.ts
+++ b/server/helpers/ffmpeg/ffmpeg-validate.ts
@@ -1,0 +1,35 @@
+import ffmpeg from 'fluent-ffmpeg'
+import { Job } from 'bull'
+import { lTags } from '@server/lib/object-storage/shared'
+import { logger } from '../logger'
+import { runCommand } from './ffmpeg-commons'
+
+async function validateVideoFile (options: {
+  job?: Job
+  path: string
+}) {
+  logger.debug('Will validate video file.', { options, ...lTags() })
+
+  let validationFailed = false
+
+  const command = ffmpeg(options.path)
+    .addOption('-loglevel', 'error')
+    .addOption('-f', 'null')
+    .output('/dev/null')
+    .on('stderr', () => {
+      if (validationFailed) {
+        return command.kill('SIGKILL')
+      }
+      validationFailed = true
+    })
+
+  await runCommand({ command, job: options.job })
+
+  if (validationFailed) {
+    throw Error(`Video validation failed for file ${options.path}`)
+  }
+}
+
+export {
+  validateVideoFile
+}
--- a/server/lib/transcoding/transcoding.ts
+++ b/server/lib/transcoding/transcoding.ts
@@ -29,6 +29,7 @@ import {
 } from '../paths'
 import { VideoPathManager } from '../video-path-manager'
 import { VideoTranscodingProfilesManager } from './default-transcoding-profiles'
+import { validateVideoFile } from '@server/helpers/ffmpeg/ffmpeg-validate'
 
 /**
  *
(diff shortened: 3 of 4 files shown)


kontrollanten:
@Chocobozzz This one is ready for review when you've time

kontrollanten:
I found out that this can lead to errors with som videos when using older versions of ffmpeg; https://stackoverflow.com/a/65501876 When I tried with ffmpeg 5.0 the error is solved.

Chocobozzz:
> This one is ready for review when you've time

But most of the tests fail

> I found out that this can lead to errors with som videos when using older versions of ffmpeg; https://stackoverflow.com/a/65501876 When I tried with ffmpeg 5.0 the error is solved.

Unfortunately we must support older ffmpeg versions

kontrollanten:
> > This one is ready for review when you've time
> 
> But most of the tests fail
Yes, but as I wrote above all of them are working when I run them locally. Can't figure out why it doesn't work in the pipeline.

> > I found out that this can lead to errors with som videos when using older versions of ffmpeg; https://stackoverflow.com/a/65501876 When I tried with ffmpeg 5.0 the error is solved.
> 
> Unfortunately we must support older ffmpeg versions

Shall we make this configurable then?

Chocobozzz:
> Yes, but as I wrote above all of them are working when I run them locally. Can't figure out why it doesn't work in the pipeline.

I'm sorry but I can't merge something that breaks all of our tests.

> Shall we make this configurable then?

Yes, or you can try to change `max_muxing_queue_size` option like we already do in https://github.com/Chocobozzz/PeerTube/blob/develop/server/helpers/ffmpeg/ffmpeg-presets.ts#L15

kontrollanten:
> Yes, or you can try to change `max_muxing_queue_size` option like we already do in https://github.com/Chocobozzz/PeerTube/blob/develop/server/helpers/ffmpeg/ffmpeg-presets.ts#L15

Great, that solved the test issues.

kontrollanten:
We've tried this on some users now and found that some validations throws the following error from ffmpeg; "Non monotonically increasing dts to muxer in stream". It seems that this error will be fixed by ffmpeg when converting the codec, so I'll see if I can find a way to fix this in the cases when the codec will be converted.

kontrollanten:
I just removed the validation on the optimizing encoding, it should be as safe as without.

Chocobozzz:
I'm really hesitant about this PR :/ It seems to add potential new errors, while not resolving the root cause of https://github.com/Chocobozzz/PeerTube/issues/3407

What's your feedback with your users? Do this PR detects broken files?

kontrollanten:
> I'm really hesitant about this PR :/ It seems to add potential new errors, while not resolving the root cause of https://github.com/Chocobozzz/PeerTube/issues/3407

The current failing test (a rate limit test) shouldn't have anything to do with this? Since we're creating and sending files, shouldn't we always assume that the files may be corrupt for some reason?

> What's your feedback with your users? Does this PR detects broken files?

Yes it has detected a lot of broken files. Since implementing #4903 I haven't seen any broken files though.

Chocobozzz:
> Since implementing https://github.com/Chocobozzz/PeerTube/pull/4903 I haven't seen any broken files though.

Then I prefer to close this issue

PR #4885, opened 2022-03-27
store uploaded video filename

### LINKED ISSUE #4731: Display video file name for video owner
When a user creates different versions of a video it good to know which of the file that's published on the PeerTube instance.

Video/channel owners can see the filename of the uploaded video file.

Either the filename is shown in the watch view or in the update view. I think YouTube has this functionality, but I don't know how they're showing it.

kontrollanten:
## Description
Stores filename of uploaded video files. Shown under Advanced settings.

## Related issues
closes #4731

## Has this been tested?

## Screenshots
image

### DIFF
--- a/client/src/app/+videos/+video-edit/shared/video-edit.component.html
+++ b/client/src/app/+videos/+video-edit/shared/video-edit.component.html
@@ -343,6 +343,17 @@
           </div>
 
           <div class="col-md-12 col-xl-4">
+            <div class="form-group">
+              <label i18n for="filename">Filename</label>
+              <my-help>
+                <ng-template ptTemplate="preHtml">
+                  <ng-container i18n>
+                    Name of the uploaded file
+                  </ng-container>
+                </ng-template>
+              </my-help>
+              <input type="text" [disabled]="true" id="filename" class="form-control" [value]="videoToUpdate.sources[0].filename" />
+            </div>
             <div class="form-group originally-published-at">
               <label i18n for="originallyPublishedAt">Original publication date</label>
               <my-help>
--- a/server/controllers/api/videos/upload.ts
+++ b/server/controllers/api/videos/upload.ts
@@ -44,6 +44,7 @@ import {
 import { ScheduleVideoUpdateModel } from '../../../models/video/schedule-video-update'
 import { VideoModel } from '../../../models/video/video'
 import { VideoFileModel } from '../../../models/video/video-file'
+import { VideoSourceModel } from '@server/models/video/video-source'
 
 const lTags = loggerTagsFactory('api', 'video')
 const auditLogger = auditLoggerFactory('videos')
@@ -151,6 +152,7 @@ async function addVideo (options: {
   video.url = getLocalVideoActivityPubUrl(video) // We use the UUID, so set the URL after building the object
 
   const videoFile = await buildNewFile(videoPhysicalFile)
+  const originalFilename = videoPhysicalFile.filename
 
   // Move physical file
   const destination = VideoPathManager.Instance.getFSVideoFileOutputPath(video, videoFile)
@@ -180,6 +182,11 @@ async function addVideo (options: {
     await videoFile.save(sequelizeOptions)
 
     video.VideoFiles = [ videoFile ]
+
+    await VideoSourceModel.create({
+      filename: originalFilename,
+      videoId: video.id
+    }, { transaction: t })
 
     await setVideoTags({ video, tags: videoInfo.tags, transaction: t })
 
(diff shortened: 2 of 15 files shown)


server/initializers/migrations/0710-video-source.ts
```diff
@@ -0,0 +1,50 @@
+import * as Sequelize from 'sequelize'
+
+async function up (utils: {
+  transaction: Sequelize.Transaction
+  queryInterface: Sequelize.QueryInterface
+  sequelize: Sequelize.Sequelize
+  db: any
+}): Promise<void> {
+  {
+    await utils.queryInterface.createTable('videoSource', {
```
Chocobozzz, line 10: Prefer using sequelize raw query (available in logs at startup)

server/models/video/formatter/video-format-utils.ts
```diff
@@ -182,7 +182,7 @@ function videoModelToFormattedDetailsJSON (video: MVideoFormattableDetails): Vid
 
   const tags = video.Tags ? video.Tags.map(t => t.name) : []
 
-  const detailsJSON = {
+  const detailsJSON: any = {
```
Chocobozzz, line 185: Please don't use `any` type

server/models/video/video.ts
```diff
@@ -597,6 +598,15 @@ export class VideoModel extends Model<Partial<AttributesOnly<VideoModel>>> {
   })
   VideoPlaylistElements: VideoPlaylistElementModel[]
 
+  @HasMany(() => VideoSourceModel, {
```
Chocobozzz, line 601: I'm not sure to understand how a video may have multiple source files

kontrollanten: If uploading a new file it would be possible to store the old version while transcoding the new. Theoretically it would also be possible to save the old version to let the user be able to rollback to the old version, I'm not sure if there's any use case for that though.

server/controllers/api/videos/upload.ts
```diff
@@ -181,6 +183,11 @@ async function addVideo (options: {
 
     video.VideoFiles = [ videoFile ]
 
+    await VideoSourceModel.create({
```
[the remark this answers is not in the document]
kontrollanten: Yes, but currently I guess it wouldn't add any value since they don't have any meaningful filename.

server/initializers/migrations/0710-video-source.ts
kontrollanten: May I ask why?

client/src/app/+videos/+video-edit/shared/video-edit.component.html
```diff
@@ -343,6 +343,17 @@
           </div>
 
           <div class="col-md-12 col-xl-4">
+            <div class="form-group">
```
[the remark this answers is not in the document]
kontrollanten: It's disabled, the form-group class is only for the styling. I couldn't come up with any other design that was consistent with the rest of the page.

server/initializers/migrations/0710-video-source.ts
Chocobozzz: To ensure we have the same table schema as a new peertube installation

server/models/video/video.ts
Chocobozzz: I prefer to not design the database according to a potential future use case. So let's keep things simple and just have 1:1 relation between these tables.

Chocobozzz (changes requested):
Please add openapi documentation

server/models/video/video.ts
```diff
@@ -597,6 +598,15 @@ export class VideoModel extends Model<Partial<AttributesOnly<VideoModel>>> {
   })
   VideoPlaylistElements: VideoPlaylistElementModel[]
 
+  @HasOne(() => VideoSourceModel, {
+    foreignKey: {
+      name: 'videoId',
+      allowNull: true
+    },
+    onDelete: 'set null'
```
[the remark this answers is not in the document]
kontrollanten: True, I defined that in the VideoSource model. I'll remove it from there.

server/middlewares/validators/videos/video-source.ts
```diff
@@ -0,0 +1,44 @@
...
+} from '../shared'
+
+const videoSourceGetValidator = [
+  isValidVideoIdParam('id'),
+
+  async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    logger.debug('Checking videoSourceGet parameters', { parameters: req.params })
+
+    if (areValidationErrors(req, res)) return
+    if (!await doesVideoExist(req.params.id, res, 'for-api')) return
```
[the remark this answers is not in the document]
Chocobozzz: Hum I think you're right. I assumed that we should use the SEE right, since we just see the video filename. But maybe the UPDATE right is more appropriate.

Then you can use `checkUserCanManageVideo` function

kontrollanten:
Ping @Chocobozzz

PR #4903, opened 2022-04-07
object-storage: @aws-sdk/lib-storage for multipart

kontrollanten:
## Description
Use [@aws-sdk/lib-storage](https://github.com/aws/aws-sdk-js-v3/tree/main/lib/lib-storage) for multipart upload

## Related issues
#3407 

## Has this been tested?

- [x] 👍 yes, current tests covers this

### DIFF
--- a/server/tests/api/object-storage/videos.ts
+++ b/server/tests/api/object-storage/videos.ts
@@ -3,7 +3,13 @@
 import 'mocha'
 import * as chai from 'chai'
 import { merge } from 'lodash'
-import { checkTmpIsEmpty, expectLogDoesNotContain, expectStartWith, generateHighBitrateVideo, MockObjectStorage } from '@server/tests/shared'
+import {
+  checkTmpIsEmpty,
+  expectLogDoesNotContain,
+  expectStartWith,
+  generateHighBitrateVideo,
+  MockObjectStorage
+} from '@server/tests/shared'
 import { areObjectStorageTestsDisabled } from '@shared/core-utils'
 import { HttpStatusCode, VideoDetails } from '@shared/models'
 import {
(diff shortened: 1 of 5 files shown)


server/tests/api/object-storage/videos.ts
```diff
@@ -107,17 +107,18 @@ async function checkFiles (options: {
 }
 
 function runTestSuite (options: {
+  fixture?: string
```
Chocobozzz, line 110: I don't understand why you removed `maxUploadPart` and instead use a big fixture

kontrollanten: Since lib-storage has a minimum limit at 5 mb for the parts, so the old test will make lib-storage to through an error.

https://github.com/aws/aws-sdk-js-v3/blob/4d9f581de6633dd0a7cde71b094fdb07fe2910dc/lib/lib-storage/src/Upload.ts#L27

Chocobozzz: Okay then could you try to generate the big video file to not increase repo side? Like https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/shared/generate.ts#L17

server/tests/api/object-storage/videos.ts
```diff
@@ -390,12 +397,17 @@ describe('Object storage for videos', function () {
     })
   })
 
-  describe('Test object storage with small upload part', function () {
+  describe('Test object storage with file bigger than upload part', function () {
+    let fixture
```
Chocobozzz, line 401: Please type `fixture`

server/tests/api/object-storage/videos.ts
```diff
@@ -390,12 +397,17 @@ describe('Object storage for videos', function () {
...
+
+    before(async function () {
+      fixture = await generateHighBitrateVideo()
```
Chocobozzz, line 404: Please ensure `fixture` size is greater than `max_upload_part`

Chocobozzz (changes requested):

Chocobozzz:
Thanks!

Issue #4934, opened 2022-04-22
Support forward proxying of object storage


noellabo:
**Describe the problem to be solved**

If you intend to configure external access from a service through a forward proxy, access to object storage is not supported. 

**Describe the solution you would like:**

Allows access to object storage via proxy when the administrator intends to use the forward proxy.

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/pull/4973

PR #4935, opened 2022-04-22
Support forward proxying of object storage

noellabo:
## Description

Allows access to object storage via proxy when the administrator intends to use the forward proxy. 

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/4934

## Has this been tested?

I dont know what unit tests to write for this.

### DIFF
--- a/server/lib/object-storage/shared/client.ts
+++ b/server/lib/object-storage/shared/client.ts
@@ -1,21 +1,18 @@
 import { NodeHttpHandler } from "@aws-sdk/node-http-handler"
-import { HttpsProxyAgent } from 'https-proxy-agent'
-import { getProxy, isProxyEnabled } from '@server/helpers/proxy'
+import { getAgent } from '@server/helpers/requests'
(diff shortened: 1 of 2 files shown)


server/lib/object-storage/shared/client.ts
```diff
@@ -1,8 +1,23 @@
...
 import { lTags } from './logger'
 
+function getProxyRequestHandler () {
+
+  if (!isProxyEnabled()) { return new NodeHttpHandler() }
+
+  const proxy = getProxy()
+
+  return new NodeHttpHandler({
+    httpAgent: new HttpsProxyAgent(proxy),
```
Chocobozzz, line 16: Could you use `hpagent` instead? https://github.com/Chocobozzz/PeerTube/blob/develop/server/helpers/requests.ts#L165

Chocobozzz:
Closing in favour of https://github.com/Chocobozzz/PeerTube/pull/4973

PR #4977, opened 2022-05-10
(#4959) Add 'Originally Published At' sort option in videos pages

### LINKED ISSUE #4959: On Channel Page Add Option Sort by Original Publication Date

A lot of channels have a huge amount of videos uploaded in one day. They all have the same published on date but often users set the Original publication date to reflect when they first created the video or made it public on other video platforms etc. So it can be difficult to have an idea of the chronological order of the videos based on sorting by Recently Added.

On the channel page, eg., https://framatube.org/c/joinpeertube/videos
Add to the sort by drop down menu the option: Sort by "Original Publication Date"
And then it would sort newest to the top and oldest to the bottom based on video Original Publication Date. If a video doesn't have an Original Publication Date filled in then it would use the published on date.

Thanks!

Jlll1:
## Description
This allows the user to sort videos by the **Original publication date**, allowing them to differentiate between the date of publication on PeerTube and for example the date of publication on other platforms.

## Related issues
Resolves #4959 

## Has this been tested?

## Screenshots
image

### DIFF
--- a/client/src/app/shared/shared-video-miniature/video-filters-header.component.html
+++ b/client/src/app/shared/shared-video-miniature/video-filters-header.component.html
@@ -44,6 +44,7 @@
       [searchable]="false"
     >
       <ng-option i18n value="-publishedAt">Sort by <strong>"Recently Added"</strong></ng-option>
+      <ng-option i18n value="-originallyPublishedAt">Sort by <strong>"Originally Published At"</strong></ng-option>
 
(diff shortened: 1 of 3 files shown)


Jlll1:
Thank you!

Mike35538:
I don't see the option (Sort by "Original Publication Date") in the latest 4.2.0 or 4.2.1 versions. Did this get added to the latest PeerTube release?

Thanks

Chocobozzz:
@Mike35538 this  feature will be available in 4.3

PR #4484, opened 2021-10-23
Added "total views" in the my channels list

Poslovitch:
## Description

I added a "total views" counter for each channel in the "my-channels" list. (1)

Also made the followersCount consistent with other counters (if =0, "No followers"). (2)

However, I couldn't find a way to get around the "pagination" in the VideoService.

## Related issues

Closes https://github.com/Chocobozzz/PeerTube/issues/4331

## Has this been tested?

- [x] 🙅 no, because this PR does not update server code

## Screenshots

image

PR #5008, opened 2022-05-27
Display "No subscribers" in my-library instead of "0 subscribers"

Poslovitch:
## Description

It displays "No subscribers" instead of "0 subscribers" in the my-library, to follow what happens with the videos count.

## Related issues

It was originally part of https://github.com/Chocobozzz/PeerTube/pull/4484, but I decided to split it into its own PR.

## Has this been tested?

## Screenshots

image


Chocobozzz:
Thanks!

PR #5020, opened 2022-06-06
Use correct URI for schema.org in `@context`

### LINKED ISSUE #5019: schema.org URI in `@context` is not correct
**Steps to reproduce:**

1. Expand or compact a JSON-LD document from Peertube

`@context` uses `http://schema.org#` which leads to incorrect namespacing, for example `https://schema.org#Boolean`

`@context` uses `http://schema.org/` which leads to correct namespacing, for example `https://schema.org/Boolean`

**Additional information**

https://github.com/mastodon/mastodon/issues/18361

trwnh:
## Description

schema.org should resolve properly, but it currently does not, due to the usage of a hash instead of a slash

## Related issues

Fix #5019

## Has this been tested?

i do not know enough about how peertube processes JSON(LD) in order to know for sure that this will not cause any other issues. if peertube uses plain json, then it should be fine. if peertube uses json-ld, then some additional compatibility code might be necessary.

### DIFF
--- a/server/lib/activitypub/context.ts
+++ b/server/lib/activitypub/context.ts
@@ -179,7 +179,7 @@ function buildContext (contextValue?: ContextValue) {
 
     {
       pt: 'https://joinpeertube.org/ns#',
-      sc: 'http://schema.org#',
+      sc: 'http://schema.org/',
 
       ...contextValue
     }


Chocobozzz:
Thanks! To include this bug fix in 4.2, I committed directly in https://github.com/Chocobozzz/PeerTube/commit/4d526901349b35bfac9811acc45854c65639cee3

Issue #4974, opened 2022-05-09
Video not played from defined parameters in embed service


raazkarkee405:
**Describe the current behavior**
When setting different parameters in the video URL, the video is played without considering the parameter setting in the embed service.
**Steps to reproduce:**

1. Click on the Share button of any public video 
2. Customize the URL adding different parameters. Example: https://example.com/w/o7fL1ZRMqQw5Z5cSxtxXbc?start=1m39s&muted=1
3. Open embed service with URL: https://example.com/services/oembed?url=https://example.com/w/o7fL1ZRMqQw5Z5cSxtxXbc?start=1m39s&muted=1

**Describe the expected behavior**
When opening the link on the browser I am able to play the video as per the parameter but with the embed service, I am not able to play the video as per the parameter.
For example: If the video URL has a parameter set to start=1m39s, the video is played from the same time on the browser but not with the embed service.

**Additional information**
![image](https://user-images.githubusercontent.com/26194365/167384067-4a8a1686-b27c-4f56-8f81-105abae07691.png)

![image](https://user-images.githubusercontent.com/26194365/167385635-1ab8997c-fde6-4635-8c99-58f85527aae2.png)
So, If we see the youtube embed URL it is clearly mentioned the start duration of the video in the HTML section which plays the video from the defined start time.

But if we see the PeerTube embed URL defining the start time and other customization parameters do not get included in the HTML section which is restricting the video to be played from the defined start time and with custom parameters.
![image](https://user-images.githubusercontent.com/26194365/167382965-ee9020f3-a1e5-4bf7-8e16-16bf457b5408.png)

* PeerTube instance:
  * version: 4.1.1
  * NodeJS version: v14.19.1
  * Ffmpeg version: 4.2.4-1ubuntu0.1

smilekison:
@Chocobozzz, Are you trying to resolve this issue?

Chocobozzz:
Hi,

I'm not sure the OEmbed service should mimic the current URL query params. Youtube, for example, doesn't add the timecode query param to oembed.

Can you detail your use case?

raazkarkee405:
Hello, 

Our organization is currently using Mattermost as a messaging channel and PeerTube as solution provider. And the members in organization share videos with specific start times to directly provide solutions which points to that specific time, making it easier and faster for other members to access relevant information.

Isn't this already working?

Chocobozzz:
Thanks!

[...]

PR #5023, opened 2022-06-07
fixes video not played from defined parameters in embed service 

raazkarkee405:
## Description

Since the oembed service of peertube ignores the parameters passed in the URL, here with the changes the parameters are added to the oembed URL by extracting the query parameters.

## Related issues

This fixes issue [#4974](https://github.com/Chocobozzz/PeerTube/issues/4974)

## Has this been tested?

## Screenshots
image
image

### DIFF
--- a/server/controllers/services.ts
+++ b/server/controllers/services.ts
@@ -1,10 +1,9 @@
 import express from 'express'
-import { EMBED_SIZE, PREVIEWS_SIZE, WEBSERVER, THUMBNAILS_SIZE } from '../initializers/constants'
-import { asyncMiddleware, oembedValidator } from '../middlewares'
-import { accountNameWithHostGetValidator } from '../middlewares/validators'
 import { MChannelSummary } from '@server/types/models'
 import { escapeHTML } from '@shared/core-utils/renderer'
-import { pick } from '@shared/core-utils'
+import { EMBED_SIZE, PREVIEWS_SIZE, THUMBNAILS_SIZE, WEBSERVER } from '../initializers/constants'
+import { asyncMiddleware, oembedValidator } from '../middlewares'
+import { accountNameWithHostGetValidator } from '../middlewares/validators'
 
 const servicesRouter = express.Router()
 
@@ -37,7 +36,7 @@ function generatePlaylistOEmbed (req: express.Request, res: express.Response) {
   const json = buildOEmbed({
     channel: playlist.VideoChannel,
     title: playlist.name,
-    embedPath: playlist.getEmbedStaticPath(),
+    embedPath: playlist.getEmbedStaticPath() + buildPlayerURLQuery(req.query.url),
     previewPath: playlist.getThumbnailStaticPath(),
     previewSize: THUMBNAILS_SIZE,
     req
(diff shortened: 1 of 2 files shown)


server/controllers/services.ts
```diff
@@ -48,10 +49,13 @@ function generatePlaylistOEmbed (req: express.Request, res: express.Response) {
 function generateVideoOEmbed (req: express.Request, res: express.Response) {
   const video = res.locals.videoAll
 
+  const regex = /\?[^?]*$/
+  const params = regex.test(req.query.url) ? req.query.url.match(regex) : ''
```
Chocobozzz, line 53: Please use https://nodejs.org/api/url.html to extract query parameters

and only reinject params that we know (create a `Set` for example)

raazkarkee405: hello @Chocobozzz,

I need your suggestions here. I have found two approaches for now.

To extract query parameters:
First Approach would be to use WHATWG URL API[#](https://nodejs.org/api/url.html#the-whatwg-url-api) as you suggested:
```
  const paramSearch = new URL(req.query.url).search.split('?').join('').split('=')
  const key = paramSearch[0]
  const { url, ...queryParams } = req.query
  const params = validateParams({ [key]: paramSearch[1], ...queryParams }) || ''
```

The other approach would be to use `decodeQueryString`:
```
  const decode = decodeQueryString(req.query.url)
  const params = validateParams(decode) || ''
```
To reinject params that we know I have created a function similar to `decorateVideoLink` which checks the params validity.

```
function validateParams (options: {
  start?: string
  stop?: string
  loop?: string
  autoplay?: string
  muted?: string
}) {

  const params = new URLSearchParams()

  if (options.start !== undefined && options.start !== null) {
    params.set('start', options.start)
  }

  if (options.stop !== undefined && options.stop !== null) {
    params.set('stop', options.stop)
  }

  if (options.loop !== undefined && options.loop !== null) {
    params.set('loop', options.loop)
  }

  if (options.autoplay !== undefined && options.autoplay !== null) {
    params.set('autoplay', options.autoplay)
  }

  if (options.muted !== undefined && options.muted !== null) {
    params.set('muted', options.muted)
  }

  return buildParams(params)
}

function buildParams (params: URLSearchParams) {
  let hasParams = false
  params.forEach(() => {
    hasParams = true
  })
  if (hasParams) return '?' + params.toString()
}

```
Both the approach works fine but the problem here is using the first approach I get the following result: 
image

And with the second approach I get the following result:
image

The first approach does not work when the URL is encoded. 
The second approach does not add the query to the response html.

Chocobozzz: > The other approach would be to use decodeQueryString:

Please use the first approach, the URL object should correctly parse the `url` query param. You can use `searchParams` properties:

```js
new URL(req.query.url).searchParams.forEach((v, k) => {
  console.log(k + '=' + v)
})
```

> To reinject params that we know I have created a function similar to decorateVideoLink which checks the params validity.

I think you can just use `pick` with the parameter names you want to support: https://github.com/Chocobozzz/PeerTube/blob/develop/shared/core-utils/common/object.ts#L1

raazkarkee405: I added a section as you suggested, please review it and let me know what you think!

Chocobozzz:
Thanks!

PR #5043, opened 2022-06-10
fix(doc): fix links to 'anchors' in install doc

dingcycle:
Broken since the production guide was *emojified* 6 months ago.


Chocobozzz:
Thanks!

PR #5097, opened 2022-07-04
Custom error view upon 401 API response

### LINKED ISSUE #4768: Show "access denied" when visiting a private/internal video
When a user is visiting a watch page of a video that's private/internal, the user will be redirected to login form if it's not logged in. This may be confusing for the user since it's not clear why they're going to the login page.

It would be better if an access denied message would be shown, with a link to the login page.

kontrollanten:
## Description
Show a message telling the user is not authorized to see a resource when 401 is responded from the API, instead of redirect to /login. I decided to not redirect to login and show a message since this is more clear for the user, and it's an uncommon case where an extra click won't harm.

## Related issues
closes #4768

## Has this been tested?

## Screenshots
image


Chocobozzz:
Thanks, it's much more better!

PR #5125, opened 2022-07-13
feat: show contained playlists under My videos

### LINKED ISSUE #4769: Show which playlists a video is added to under My videos
We use playlists as a way to categorize our own videos to make it easier for the users to navigate and find interesting content. Today there's no easy way for us to know which playlists a certain video contains to.

A list of videos showing which playlists they contain to.

Under "My videos", show each playlist the videos is added to.

kontrollanten:
## Description
Shows which playlists videos are added to under My Videos.

I couldn't figure out how the cache is working in the VideoPlaylistService, so now these calls aren't added to the cache.

## Related issues
closes #4769

## Has this been tested?

## Screenshots
image

### DIFF
--- a/client/src/app/shared/shared-video-miniature/video-miniature.component.html
+++ b/client/src/app/shared/shared-video-miniature/video-miniature.component.html
@@ -52,6 +52,12 @@
             <ng-container *ngIf="displayOptions.privacyText && displayOptions.state && getStateLabel(video)"> - </ng-container>
             <ng-container *ngIf="displayOptions.state">{{ getStateLabel(video) }}</ng-container>
           </div>
+
+          <div *ngIf="displayOptions.containedInPlaylists" class="video-contained-in-playlists">
+            <a *ngFor="let playlist of containedInPlaylists" class="chip rectangular bg-secondary text-light" [routerLink]="['/w/p/', playlist.playlistShortUUID]" i18n>
+              {{ playlist.playlistDisplayName }}
+            </a>
+          </div>
         </div>
       </div>
 
--- a/client/src/app/shared/shared-video-miniature/video-miniature.component.ts
+++ b/client/src/app/shared/shared-video-miniature/video-miniature.component.ts
@@ -11,7 +11,7 @@ import {
   Output
 } from '@angular/core'
 import { AuthService, ScreenService, ServerService, User } from '@app/core'
-import { HTMLServerConfig, VideoPlaylistType, VideoPrivacy, VideoState } from '@shared/models'
+import { HTMLServerConfig, VideoExistInPlaylist, VideoPlaylistType, VideoPrivacy, VideoState } from '@shared/models'
 import { LinkType } from '../../../types/link.type'
 import { ActorAvatarSize } from '../shared-actor-image/actor-avatar.component'
 import { Video } from '../shared-main'
@@ -19,6 +19,7 @@ import { VideoPlaylistService } from '../shared-video-playlist'
 import { VideoActionsDisplayType } from './video-actions-dropdown.component'
 
 export type MiniatureDisplayOptions = {
+  containedInPlaylists?: boolean
   date?: boolean
   views?: boolean
   by?: boolean
@@ -38,8 +39,10 @@ export type MiniatureDisplayOptions = {
 export class VideoMiniatureComponent implements OnInit {
   @Input() user: User
   @Input() video: Video
+  @Input() containedInPlaylists: VideoExistInPlaylist[]
 
   @Input() displayOptions: MiniatureDisplayOptions = {
+    containedInPlaylists: false,
     date: true,
     views: true,
     by: true,
--- a/client/src/app/shared/shared-video-playlist/video-playlist.service.ts
+++ b/client/src/app/shared/shared-video-playlist/video-playlist.service.ts
@@ -75,6 +75,10 @@ export class VideoPlaylistService {
                )
   }
 
+  listPlaylistsForVideos (videoIds: number[]) {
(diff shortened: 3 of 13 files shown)


Chocobozzz:
Hello,

 * Server files should not change since you only use existing server methods
 * Miniature should check itself playlists existence, like we do for the watch later playlist

Chocobozzz:
> Then we'll get at least a 500 ms delay which I think is a bad user experience.

Okay then :+1:

client/src/app/shared/shared-video-miniature/video-miniature.component.html
```diff
@@ -52,6 +52,12 @@
             <ng-container *ngIf="displayOptions.privacyText && displayOptions.state && getStateLabel(video)"> - </ng-container>
             <ng-container *ngIf="displayOptions.state">{{ getStateLabel(video) }}</ng-container>
           </div>
+
+          <div *ngIf="displayOptions.containedInPlaylists" class="video-contained-in-playlists">
+            <a *ngFor="let playlist of containedInPlaylists" class="chip rectangular bg-secondary text-light" [routerLink]="['/w/p/', playlist.playlistShortUUID]" i18n>
```
Chocobozzz, line 57: You don't need `i18n`

client/src/app/shared/shared-video-miniature/video-miniature.component.ts
```diff
@@ -11,14 +11,15 @@ import {
...
-import { HTMLServerConfig, VideoPlaylistType, VideoPrivacy, VideoState } from '@shared/models'
+import { HTMLServerConfig, VideoExistInPlaylist, VideoPlaylistType, VideoPrivacy, VideoState } from '@shared/models'
 import { LinkType } from '../../../types/link.type'
 import { ActorAvatarSize } from '../shared-actor-image/actor-avatar.component'
 import { Video } from '../shared-main'
 import { VideoPlaylistService } from '../shared-video-playlist'
 import { VideoActionsDisplayType } from './video-actions-dropdown.component'
 
 export type MiniatureDisplayOptions = {
+  containedInPlaylists?: boolean
```
Chocobozzz, line 22: I think we can consider that providing `containedInPlaylists` is enough to say we want playlists to be displayed

client/src/app/shared/shared-video-playlist/video-playlist.service.ts
```diff
@@ -75,6 +75,10 @@ export class VideoPlaylistService {
                )
   }
 
+  listPlaylistsForVideos (videoIds: number[]) {
+    return this.doVideosExistInPlaylist(videoIds)
```
Chocobozzz, line 79: I think you can just set `doVideosExistInPlaylist` as `public`

client/src/app/shared/shared-video-playlist/video-playlist.service.ts
```diff
@@ -188,14 +192,6 @@ export class VideoPlaylistService {
     return this.authHttp.post<{ videoPlaylistElement: { id: number } }>(url, body)
                .pipe(
                  tap(res => {
-                   const existsResult = this.videoExistsCache[body.videoId]
```
Chocobozzz, line 191: Why did you remove this code?

server/types/models/video/video-playlist.ts
```diff
@@ -26,6 +30,10 @@ export type MVideoPlaylistWithElements =
   MVideoPlaylist &
   Use<'VideoPlaylistElements', MVideoPlaylistElementLight[]>
 
+export type MVideoPlaylistSummaryWithElements =
```
Chocobozzz, line 33: Just noticed that `MVideoPlaylistWithElements` is not used anywhere
I think it's also the same for `MVideoPlaylistIdWithElements`

client/src/app/shared/shared-video-playlist/video-playlist.service.ts
Chocobozzz: This breaks the "add to playlist" dropdown (video miniature menu for example)

Chocobozzz: > I remember the caching makes the results in the "Add to playlist" dropdown inconsistent, depending on which playlists are loaded in the cache

Don't hesitate to create an issue with the reproducing steps so I can fix this behaviour.

>  By the way I just found out that the cache isn't cleared upon logout.

Thanks, I'll fix this!

> In which way does it break? When I test it seems to work.

https://user-images.githubusercontent.com/5180488/192965688-e4479f5c-77c3-4ca5-ba5a-219e7bec0541.mp4

> If we want to cache the result, isn't it better/easier to just cache the API result

The cache is more like a store, so we don't have to do API calls every time we add/remove a video from a playlist.

client/src/app/shared/shared-video-playlist/video-playlist.service.ts
```diff
@@ -190,16 +193,19 @@ export class VideoPlaylistService {
...
     return this.authHttp.post<{ videoPlaylistElement: { id: number } }>(url, body)
                .pipe(
                  tap(res => {
                    const existsResult = this.videoExistsCache[body.videoId]
+
                    existsResult.push({
-                     playlistId,
                      playlistElementId: res.videoPlaylistElement.id,
+                     playlistId: playlist.id,
+                     playlistDisplayName: playlist.displayName,
```
Chocobozzz, line 207: I don't think we need to put `playlistDisplayName` to this cache/store. It can lead to some issues when the playlist is renamed for example.

Why don't you limit the cache to IDs since we don't need more information in it?

Chocobozzz:
Thanks!

PR #5135, opened 2022-07-20
Channel sync

### LINKED ISSUE #754: Automatically import videos from other platforms
In order to simplify the PeerTube adoption from current Youtubers, it's important to simplify the importing process.

The [import script](https://github.com/Chocobozzz/PeerTube/blob/master/server/tools/import-videos.ts) is a good step in this direction, but it is possible to do better: automatically import Youtube videos, so the PeerTube import can be totally transparent for the Youtuber.

We can do this without the Youtube API, by using atom feeds: `https://www.youtube.com/feeds/videos.xml?channel_id=<channel_id>`.

I'm not used with TypeScript but I wrote a small Python script as a proof-of-concept example, available [here](https://framagit.org/snippets/1891).

Used with the import script and a 5min-cronjob, a PeerTube user could feed its instance by only specifying its Youtube channel name.

Note that we also need a way to check if the video is not already on the PeerTube instance, can we provide the youtube video id in some video metadata?

fflorent:
## Description

Implement channel synchronization, but with some lacks detailed below.

## Related issues

Fixed #754
Also it paves the way for the importation of channel videos from other platforms in the UI (like peertube-import-videos).

## Has this been tested?

- :+1: Manual tests done
- 👍 Unit tests done

## Screenshots

"My synchronizations" button

Synchronization table

Synchronization creation form

## EDIT1 : Screencast of admin configuration

https://user-images.githubusercontent.com/371705/181291121-42a3f474-4420-478d-850f-1d201014f3bc.mp4


server/lib/job-queue/handlers/video-channels-sync.ts
```diff
@@ -0,0 +1,266 @@
...
+    'filter:api.video.import-url.video-attribute.result'
+  )
+
+  const video = new VideoModel(videoData)
+  video.url = getLocalVideoActivityPubUrl(video)
+
+  return video
+}
+
+async function addYoutubeDLImport (parameters: {
```
Chocobozzz, line 202: Could you try to refactor this function? It shares a lot of code with the `addYoutubeDLImport` function form the controller

server/middlewares/validators/videos/video-channel-sync.ts
```diff
@@ -0,0 +1,79 @@
...
+  body('externalChannelUrl').custom(isUrlValid).withMessage('Should have a valid channel url'),
+  body('videoChannelId').isInt().withMessage('Should have a valid video channel id'),
+
+  async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    logger.debug('Checking videoChannelsAdd parameters', { parameters: req.body })
+
+    if (areValidationErrors(req, res)) return
+
+    const body: VideoChannelSyncCreate = req.body
+    if (!await doesVideoChannelIdExist(body.videoChannelId, res)) {
```
Chocobozzz, line 32: Just `return`, `doesVideoChannelIdExist` will automatically use `.fail` on 404

server/lib/job-queue/handlers/video-channels-sync.ts
```diff
@@ -0,0 +1,266 @@
+import { logger } from '@server/helpers/logger'
```
fflorent, line 1: Not sure whether the naming of this file is cool or not for this file.

I assumed we trigger an action (*sync*) for *video-channels*. So *video-channels* is here in plural. Are you happy with this name?

server/middlewares/validators/videos/video-channel-sync.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    return next()
+  }
+]
+
+export const ensureSyncExists = [
+  async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+
+    const syncId = parseInt(req.params.id, 10)
+    const sync = await VideoChannelSyncModel.loadById(syncId)
```
Chocobozzz, line 43: `loadById` -> `load` (we implicitly load by id)

server/middlewares/validators/videos/video-channel-sync.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+]
+
+export const ensureSyncTargetChannelExists = [
+  async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    const sync = res.locals.videoChannelSync
+    if (!await doesVideoChannelIdExist(sync.videoChannelId, res)) {
+      return
+    }
+    await sync.reload({ include: VideoChannelModel })
```
Chocobozzz, line 61: Do you need this?

server/middlewares/validators/videos/video-channel-sync.ts
```diff
@@ -0,0 +1,75 @@
...
+    return next()
+  }
+]
+
+const routeWithIdValidator = [
+  param('id').exists().isInt().withMessage('Should have an sync id'),
```
Chocobozzz, line 67: I think you can move all this logic in `ensureSyncExists` that you can rename `getVideoChannelSyncValidator` or something like that

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,163 @@
...
+type AvailableForListOptions = {
+  accountId: number
+  search?: string
+  host?: string
+  handles?: string[]
+  forCount?: boolean
+}
+
+@Table({
+  tableName: 'videoChannelSync'
```
Chocobozzz, line 36: Add an index on `videoChannelId`

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,163 @@
...
+})
+export class VideoChannelSyncModel extends Model<Partial<AttributesOnly<VideoChannelSyncModel>>> {
+
+  @AllowNull(false)
+  @Default(null)
+  @Is('VideoChannelExternalChannelUrl', value => throwIfNotValid(value, isVideoImportTargetUrlValid, 'externalChannelUrl', true))
```
Chocobozzz, line 42: `isVideoImportTargetUrlValid` -> `isUrlValid`

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,163 @@
...
+  @Column(DataType.STRING(CONSTRAINTS_FIELDS.VIDEO_CHANNEL_SYNCS.EXTERNAL_CHANNEL_URL.max))
+  externalChannelUrl: string
+
+  @CreatedAt
+  createdAt: Date
+
+  // We have no use of this field
```
Chocobozzz, line 49: Please remove this comment. Even if we don't use this field, I think it can be interesting in the future.

Also, please add a `lastSyncAt` field that you update every time you check/run the sync

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,163 @@
...
+        offset: options.start,
+        limit: options.count,
+        order: getSort(options.sort),
+        include: [
+          {
+            model: VideoChannelModel.unscoped(),
+            required: true,
+            include: [
+              {
+                model: accountModel,
```
Chocobozzz, line 91: You can remove this include and add `accountId: options.accountId` in `VideoChannelModel` where

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  toFormattedJSON (this: MChannelSyncFormattable): VideoChannelSync {
+    return {
+      id: this.id,
+      state: {
+        id: this.state,
+        label: VIDEO_CHANNEL_SYNC_STATE[this.state]
+      },
+      externalChannelUrl: this.externalChannelUrl,
+      createdAt: this.createdAt.toString(),
```
Chocobozzz, line 120: `toISOString()`

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,163 @@
...
+      channel: this.VideoChannel?.toFormattedSummaryJSON(),
```
Chocobozzz, line 121: Since we don't allow `null` in `videoChannelId` you don't need `?`. As a side note, it would have thrown an exception if `this.VideoChannel` was `null` or `undefined` (because `null` or `undefined` don't have `toFormattedSummaryJSON` method

server/middlewares/validators/videos/video-channel-sync.ts
```diff
@@ -0,0 +1,75 @@
...
+      })
+    }
+    res.locals.videoChannelSync = sync
+    return next()
+  }
+]
+
+export const ensureSyncTargetChannelExists = [
+  async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+    const sync = res.locals.videoChannelSync
```
Chocobozzz, line 57: Since we don't allow `null` for channels, I think the `load` can be renamed `loadAndPopulateChannel` that automatically fetches the entry with the channel.

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,163 @@
...
+        ? AccountModel.unscoped()
+        : AccountModel
+
+      return {
+        offset: options.start,
+        limit: options.count,
+        order: getSort(options.sort),
+        include: [
+          {
+            model: VideoChannelModel.unscoped(),
```
Chocobozzz, line 87: `VideoChannelModel.unscoped()` for counts to not fetch avatars (that breaks the `count()`), and `VideoChannelModel` if not counting to also fetch `ActorModel` that will include avatars.

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,163 @@
...
+            include: [
+              {
+                model: accountModel,
+                where: {
+                  id: options.accountId
+                },
+                required: true
+              },
+              {
+                model: ActorModel.unscoped()
```
Chocobozzz, line 98: You can remove this

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+
+  static loadById (id: number) {
+    return this.findOne({
+      where: { id }
+    })
+  }
+
+  static async listSyncs (): Promise<VideoChannelSyncModel[]> {
```
Chocobozzz, line 132: Please create a dedicated type for the result of this function

For example `VideoChannelSyncChannel` that would be the VideoChannelSync entity with the MVideoChannel association because the caller doesn't seem to need more than the channel name and channel actor id

server/models/video/video-import.ts
```diff
@@ -161,6 +161,23 @@ export class VideoImportModel extends Model<Partial<AttributesOnly<VideoImportMo
     ]).then(([ total, data ]) => ({ total, data }))
   }
 
+  static async urlAlreadyImported (userId: number, targetUrl: string): Promise<boolean> {
```
Chocobozzz, line 164: I think you need to check if the video has been imported in the channel, and not by the user because we may import the same video in multiple channels

server/types/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,12 @@
+import { VideoChannelSyncModel } from "@server/models/video/video-channel-sync"
+import { FunctionProperties, PickWith } from "@shared/typescript-utils"
+import { MChannelFormattable } from "./video-channels"
+
+type Use<K extends keyof VideoChannelSyncModel, M> = PickWith<VideoChannelSyncModel, K, M>
+
+export type MChannelSync = VideoChannelSyncModel
```
Chocobozzz, line 7: `export type MChannelSync = Omit<VideoChannelSyncModel, 'VideoChannel'>`

server/types/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type MChannelSyncFormattable =
+  FunctionProperties<MChannelSync> &
+  Use<'VideoChannel', MChannelFormattable> &
+  Pick<MChannelSync, 'id' | 'videoChannelId' | 'externalChannelUrl' | 'state' | 'createdAt'>
```
Chocobozzz, line 12: `MChannelSync` should be enough (no need to `Pick`)

shared/models/videos/channel-sync/video-channel-sync.model.ts
```diff
@@ -0,0 +1,14 @@
...
+import { VideoConstant } from '../video-constant.model'
+import { VideoChannelSyncState } from './video-channel-sync-state.enum'
+
+export interface VideoChannelSync {
+  id: number
+
+  externalChannelUrl: string
+
+  createdAt: string
+  channel?: VideoChannelSummary
```
Chocobozzz, line 11: I don't think channel can be optional

shared/models/videos/channel-sync/video-channel-sync.model.ts
```diff
@@ -0,0 +1,14 @@
...
+  videoChannelId: number
```
Chocobozzz, line 12: You don't need this, it's already in `channel.id`

support/doc/api/openapi.yaml
```diff
@@ -3263,6 +3263,62 @@ paths:
         '204':
           description: successful operation
 
+  '/video-channel-syncs/':
```
Chocobozzz, line 3266: `/video-channel-syncs`

support/doc/api/openapi.yaml
```diff
@@ -3263,6 +3263,62 @@ paths:
...
+    post:
+      summary: Create a synchronization for a video channels
```
Chocobozzz, line 3268: `for a video channel`

support/doc/api/openapi.yaml
```diff
@@ -3263,6 +3263,62 @@ paths:
...
+        - OAuth2: []
+      tags:
+        - Video Channels Synchronizations
+      parameters:
+        - $ref: '#/components/parameters/channelSyncId'
+      responses:
+        '204':
+          description: successful operation
+
+  '/video-channel-syncs/syncChannel/{channelSyncId}':
```
Chocobozzz, line 3307: needs a url update

support/doc/api/openapi.yaml
```diff
@@ -3263,6 +3263,62 @@ paths:
...
           description: successful operation
 
+  '/video-channel-syncs/':
+    post:
+      summary: Create a synchronization for a video channels
+      operationId: addVideoChannelSync
+      security:
+        - OAuth2: []
+      tags:
+        - Video Channels Synchronizations
```
Chocobozzz, line 3273: `Channels Sync` (we need to be concise because it is displayed on the left menu in https://docs.joinpeertube.org/api-rest-reference.html

You'll also have to add this tag to the list of available tags in https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/api/openapi.yaml#L154 and in https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/api/openapi.yaml#L300

server/middlewares/validators/videos/video-channel-sync.ts
fflorent: Yes, otherwise I don't get the VideoChannel

server/models/video/video-channel-sync.ts
fflorent: Ahah! I wondered why I didn't see any index along with `create table` instructions. That's because they're declared in models!

server/models/video/video-channel-sync.ts
fflorent: Nice catch!

server/models/video/video-import.ts
fflorent: I think I encountered some difficulties doing that. I'll try again and keep you informed of them if any.

server/controllers/api/server/debug.ts
```diff
@@ -44,6 +44,8 @@ async function runCommand (req: express.Request, res: express.Response) {
     'remove-dandling-resumable-uploads': () => RemoveDanglingResumableUploadsScheduler.Instance.execute(),
     'process-video-views-buffer': () => VideoViewsBufferScheduler.Instance.execute(),
     'process-video-viewers': () => VideoViewsManager.Instance.processViewerStats()
+    // TODO
+    // await permet d'attendre l'exécution du job
```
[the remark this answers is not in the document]
fflorent: This is a note for myself I left here for the tests. Probably not a good idea to keep it here. I'll remove that.

config/default.yaml
```diff
@@ -369,6 +369,7 @@ user:
 
 video_channels:
   max_per_user: 20 # Allows each user to create up to 20 video channels.
+  synchronization_interval: 15 minutes
```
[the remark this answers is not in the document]
fflorent: I pushed a change: the configuration on the UI is updated and the value is stored on the backend ( 9b2b63452b1df70d3458b5c95f70a93bf1818788 ).

Here is a screenast. Feedback welcome:
https://user-images.githubusercontent.com/371705/181291121-42a3f474-4420-478d-850f-1d201014f3bc.mp4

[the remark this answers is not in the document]
Chocobozzz: Nice!

server/lib/job-queue/handlers/video-channels-sync.ts
fflorent: I think I should implement test before this refactoring. Do you mind if I do that in this PR? (I would understand as it is a bit long enough)

server/middlewares/validators/videos/video-channel-sync.ts
fflorent: That being said, `ensureSyncTargetChannelExists` should not be useful: if a channel is updated or deleted, there is the `       ON DELETE CASCADE` and `ON UPDATE CASCADE` instructions for `videoChannelSync.videoChannelId`which should prevent inconsistency.

I may remove that check.

server/lib/job-queue/handlers/video-channels-sync.ts
Chocobozzz: I'd rather have the tests included in this PR. But I can work on them with you if you want

fflorent: I start to work on it :)

server/middlewares/validators/videos/video-channel-sync.ts
fflorent: I wonder if `getVideoChannelSyncValidator` would be a good name as it should rename an array and not a function, doesn't it?

server/middlewares/validators/videos/video-channel-sync.ts
fflorent: Implemented that way in 92c58e378

server/types/models/video/video-channel-sync.ts
fflorent: Not sure how to do that properly:
1. I would like to keep the video-name attribute for audit logging: https://github.com/fflorent/PeerTube/blob/2866df4eb4b5fa395fc0bc8e60c7007507cc3d34/server/helpers/audit-logger.ts#L263-L273
2. The `res.locals.videoChannelSync` (stored as `syncInstance` is of type `MChannelSync` and I use it to pass to the AuditLogger: thttps://github.com/fflorent/PeerTube/blob/2866df4eb4b5fa395fc0bc8e60c7007507cc3d34/server/helpers/audit-logger.ts#L263-L273
3. But I must pass an instance which has a `VideoChannel` property.

I wonder what could be a good way to bring this change and make the code compile.

Some options I see:
1. Casting `syncInstance` to a `VideoChannelSyncFormattable` (is it safe?)
2. Make `res.locals.videoChannelSync` of type `VideoChannelSyncFormattable`

Help welcome on this one. Thanks in advance

Chocobozzz: update `res.locals.videoChannelSync` to be `MChannelSyncChannel` (so channel sync + channel association).

fflorent: Done in [515dae8](https://github.com/Chocobozzz/PeerTube/pull/5135/commits/515dae816fe50209aa19e39ac683d28ce3e6b8fd)

server/initializers/constants.ts
```diff
@@ -155,10 +157,12 @@ const JOB_ATTEMPTS: { [id in JobType]: number } = {
   'video-live-ending': 1,
   'video-studio-edition': 1,
   'manage-video-torrent': 1,
-  'move-to-object-storage': 3
+  'move-to-object-storage': 3,
+  'video-channels-sync': 1,
+  'video-channel-import': 3
```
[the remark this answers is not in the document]
fflorent: About the naming:
 - I am not sure what `video-channel-do-sync` and `video-channels-check-sync` would refer? ( `video-channel-do-sync` would refer to `video-channels-sync` and `video-channels-check-sync` would refer to `video-channel-import`?)
 - I still think the naming of `video-channel-import` may be good as it could be reused for other uses (importing a whole video channel without syncing it).
 - Maybe `video-channel-sync-latest` (`channel` in singular and `latest` to refer to latest publications)

But as you are the maintainer of the application, I will follow your suggestion if you're not convinced of my explanation anyway :).

About the scheduler, I assume you refer to the job currently named `video-channels-sync`. Indeed, it will probably ease my job and the testing

server/controllers/api/accounts.ts
```diff
@@ -72,6 +74,15 @@ accountsRouter.get('/:accountName/video-channels',
   asyncMiddleware(listAccountChannels)
 )
 
+accountsRouter.get('/:accountName/video-channel-syncs',
+  asyncMiddleware(accountNameWithHostGetValidator),
+  paginationValidator,
+  videoChannelSyncsSortValidator,
+  setDefaultSort,
+  setDefaultPagination,
```
fflorent, line 82: @Chocobozzz I assumed that channel sync information is not confidential and can be exposed publicly. But I am not 100% sure of this choice.

Do you think the access should be restricted? Or is this fine?

server/initializers/constants.ts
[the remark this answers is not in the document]
Chocobozzz: > I am not sure what video-channel-do-sync and video-channels-check-sync would refer? ( video-channel-do-sync would refer to video-channels-sync and video-channels-check-sync would refer to video-channel-import?)

No, it's just another name for `video-channels-sync`. My initial goal is to avoid confusion with the `video-channel-import` task
`video-channel-sync-latest` seems perfect

server/controllers/api/accounts.ts
Chocobozzz: I think you can restrict it, just in case there are credentials in the URL (we can import other things than youtube channels)

fflorent: That point is interesting, although the user should not give URL with credentials because it can be seen by root.

But let's rather restrict and raise good arguments if we want to allow anyone to see the synchronizations.

server/helpers/youtube-dl/youtube-dl-import-utils.ts
```diff
@@ -0,0 +1,112 @@
+import { isVTTFileValid } from "../custom-validators/video-captions"
```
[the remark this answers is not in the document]
fflorent: Note to myself: TODO

server/controllers/api/server/debug.ts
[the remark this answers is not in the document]
fflorent: This is removed

server/lib/job-queue/handlers/video-channels-sync.ts
```diff
@@ -0,0 +1,265 @@
...
+      await sync.save()
+    } catch (err) {
+      logger.error(`Failed to synchronize channel ${sync.VideoChannel.name}`, { err })
+    }
+  }
+}
+
+export async function processVideoChannelImport (job: Job) {
+  const payload = job.data as VideoChannelImportPayload
+  logger.debug('Running processVideoChannelImport')
```
fflorent, line 80: If the import is triggered from a sync object, flag the latter as PROCESSING

client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.html
```diff
@@ -0,0 +1,64 @@
...
+          <ng-container i18n>Add synchronization</ng-container>
+        </a>
+      </div>
+    </div>
+  </ng-template>
+  <ng-template pTemplate="header">
+    <tr>
+      <th style="width: 10%"><my-global-icon iconName="columns"></my-global-icon></th>
+      <th style="width: 25%" i18n pSortableColumn="externalChannelUrl">External Channel <p-sortIcon field="externalChannelUrl"></p-sortIcon></th>
+      <th style="width: 25%" i18n>Channel</th> <!-- FIXME allow sort for this column -->
```
fflorent, line 34: To be implemented in another task

server/helpers/youtube-dl/youtube-dl-import-utils.ts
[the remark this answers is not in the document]
fflorent: Done in a1f86312d8fa1178b1df9b41ae2bfc7afc88a481

server/lib/job-queue/handlers/video-channels-sync.ts
fflorent: Done a1f86312d8fa1178b1df9b41ae2bfc7afc88a481

server/models/video/video-import.ts
fflorent: Done 1de2b573a33b4792623ec1616ed7df78e5295d11

server/lib/job-queue/handlers/video-channels-sync.ts
fflorent: Finally split video-channels-sync into 2 files:
- `server/lib/job-queue/handlers/video-channel-import.ts`
- `server/lib/schedulers/video-channel-sync-latest-scheduler.ts`

server/lib/job-queue/handlers/video-channels-sync.ts
fflorent: The more I think of it, the more I believe it should trigger a change to the VideoChannelSync object.

I even think the video-channel-import deserve its own API (`/video-channel-import`) so it can be reused by the CLI and its own UI. And the synchronization creation UI should call this API and we should remove `/video-channel-sync/$1/sync`.

What do you think @Chocobozzz?

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,177 @@
...
+              include: [ {
+                attributes: [],
+                model: UserModel.unscoped(),
+                where: {
+                  videoQuota: {
+                    [Op.ne]: 0
+                  },
+                  videoQuotaDaily: {
+                    [Op.ne]: 0
+                  }
```
fflorent, line 165: These tests prevent that users without quota clutter the synchronization job.

I have this questions:
 - how many syncs we may allow by user? (20 syncs? More?)

Related: I implemented a check that the user asking for a channel import have some quota left: 1ad1ac2bd20a5b610ac35ff1a5322ae5d60affc2 (so we avoid abusive calls to yt-dlp / youtube-dl)

server/lib/job-queue/handlers/video-channels-sync.ts
Chocobozzz: Agreed! You can do this in this PR

server/models/video/video-channel-sync.ts
Chocobozzz: > how many syncs we may allow by user? (20 syncs? More?)

You can add this as an admin config

fflorent: Alright. For the UI, does this sounds good to you?

image

config/default.yaml
```diff
@@ -539,6 +539,9 @@ import:
       # We recommend to only enable magnet URI/torrent import if you trust your users
       # See https://docs.joinpeertube.org/maintain-configuration?id=security for more information
       enabled: false
+  synchronization:
```
Chocobozzz, line 542: `video-channel-synchronization`

We can imagine in the future also sync videos, to automatically update their title etc :)

client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.ts
```diff
@@ -0,0 +1,107 @@
...
+import { mergeMap } from 'rxjs'
+
+@Component({
+  templateUrl: './my-video-channel-syncs.component.html',
+  styleUrls: [ './my-video-channel-syncs.component.scss' ]
+})
+export class MyVideoChannelSyncsComponent extends RestTable implements OnInit {
+
+  error: string = undefined
+  channelsSync: VideoChannelSync[]
```
Chocobozzz, line 16: Try to always assign `[]` to arrays to prevent errors when using `for...`

client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.ts
```diff
@@ -0,0 +1,107 @@
...
+          this.notifier.success($localize`Synchronization removed successfully for ${videoChannelsSync.channel.displayName}.`)
+          this.reloadData()
+        },
+        error: (err) => {
+          this.error = err.message
+        }
+      })
+  }
+
+  getVideoChannelCreateLink () {
```
Chocobozzz, line 92: `getSyncCreateLink`

client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.ts
```diff
@@ -0,0 +1,107 @@
...
+    return '/my-library/video-channel-syncs/create'
+  }
+
+  getVideoChannelSyncStateClass (stateId: number) {
+    return 'pt-badge ' + MyVideoChannelSyncsComponent.STATE_CLASS_BY_ID[stateId]
```
Chocobozzz, line 97: Prefer using an array for `ngClass` (`[ 'pt-badge', ... ]`)

client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.html
```diff
@@ -0,0 +1,63 @@
...
+        ></my-action-dropdown>
+      </td>
+      <td>
+        <a [href]="videoChannelSync.externalChannelUrl" target="_blank" rel="noopener noreferrer">{{ videoChannelSync.externalChannelUrl }}</a>
+      </td>
+      <td>
+        <a [href]="getChannelUrl(videoChannelSync.channel.name)" target="_blank" rel="noopener noreferrer">{{ videoChannelSync.channel.displayName }}</a>
+      </td>
+      <td>
+        <span class="badge" [ngClass]="getVideoChannelSyncStateClass(videoChannelSync.state.id)">
```
Chocobozzz, line 55: You do need the `badge` class with `pt-badge` from `ngClass`?

client/src/app/+my-library/my-video-channel-syncs/video-channel-sync-edit/video-channel-sync-edit.component.html
```diff
@@ -0,0 +1,63 @@
...
+              [items]="userVideoChannels"
+              [(ngModel)]="selectedVideoChannel"
+              formControlName="videoChannel">
+          </my-select-channel>
+          <div *ngIf="formErrors['videoChannel']" class="form-error">
+            {{ formErrors['videoChannel'] }}
+          </div>
+        </div>
+
+        <div class="form-group existing-video-strategy-container">
```
Chocobozzz, line 42: `existing-video-strategy-container` doesn't seem to be used anywhere

client/src/app/+my-library/my-video-channel-syncs/video-channel-sync-edit/video-channel-sync-edit.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+@Component({
+  selector: 'my-video-channel-sync-edit',
+  templateUrl: './video-channel-sync-edit.component.html',
+  styleUrls: [ './video-channel-sync-edit.component.scss' ]
+})
+export class VideoChannelSyncEditComponent extends FormReactive implements OnInit {
+  error: string
+  selectedVideoChannel: VideoChannel
+  userVideoChannels: SelectChannelItem[]
```
Chocobozzz, line 20: `= []`

server/controllers/api/video-channel-sync.ts
```diff
@@ -0,0 +1,103 @@
...
+  ensureSyncExists,
+  ensureSyncIsEnabled,
+  videoChannelSyncValidator
+} from '@server/middlewares'
+import { VideoChannelSyncModel } from '@server/models/video/video-channel-sync'
+import { HttpStatusCode, VideoChannelSyncState } from '@shared/models'
+import express from 'express'
+
+const videoChannelSyncRouter = express.Router()
+const auditLogger = auditLoggerFactory('channels')
```
Chocobozzz, line 20: Wrong domain I think for audit logger

server/controllers/api/video-channel-sync.ts
```diff
@@ -0,0 +1,103 @@
...
+
+async function createVideoChannelSync (req: express.Request, res: express.Response) {
+  const syncCreated = new VideoChannelSyncModel({
+    externalChannelUrl: req.body.externalChannelUrl,
+    videoChannelId: req.body.videoChannelId,
+    state: VideoChannelSyncState.WAITING_FIRST_RUN
+  })
+  await sequelizeTypescript.transaction(async t => {
+    await syncCreated.save({ transaction: t })
+    await syncCreated.reload({ transaction: t })
```
Chocobozzz, line 58: Why do you need this reload?

server/initializers/checker-after-init.ts
```diff
@@ -200,6 +201,12 @@ function checkTranscodingConfig () {
   }
 }
 
+function checkImportConfig () {
+  if (CONFIG.IMPORT.SYNCHRONIZATION.ENABLED && !CONFIG.IMPORT.VIDEOS.HTTP) {
+    throw new Error("You need to enable HTTP import to allow synchronization")
```
Chocobozzz, line 206: Use single quote

server/initializers/constants.ts
```diff
@@ -451,6 +461,8 @@ const VIDEO_PRIVACIES: { [ id in VideoPrivacy ]: string } = {
   [VideoPrivacy.INTERNAL]: 'Internal'
 }
 
+const VIDEO_CHANNEL_MAX_SYNC = 3
```
Chocobozzz, line 464: I'm not sure to understand the purpose of this constant?

server/lib/job-queue/handlers/video-channel-import.ts
```diff
@@ -0,0 +1,41 @@
...
+      secondsToWait: 30
+    })
+    if (errors > 0) {
+      logger.error(`Finished importing videos to "${videoChannel.name}" with failures` +
+        ` (failures: ${errors}, imported: ${successes}, ignored because already imported: ${alreadyImported}). Please check the logs.`)
+    } else {
+      logger.info(`Finished importing videos to "${videoChannel.name}" successfully` +
+        ` (imported: ${successes}, ignored because already imported: ${alreadyImported})`)
+    }
+  } catch (ex) {
```
Chocobozzz, line 38: `catch (err)` for consistency

server/lib/job-queue/handlers/video-channel-import.ts
```diff
@@ -0,0 +1,41 @@
...
+    logger.error(`Failed to import channel ${videoChannel.name}: ${ex.stack}`)
```
Chocobozzz, line 39: `logger.error('...', { err })`

server/lib/video-import-channel.ts
```diff
@@ -0,0 +1,87 @@
...
+  const result: ChannelSyncInfo = {
+    total: 0,
+    errors: 0,
+    successes: 0,
+    alreadyImported: targetUrls.length
+  }
+
+  for (const targetUrl of targetUrls) {
+    try {
+      // TODO retry pour l'import d'une chaîne ?
```
Chocobozzz, line 65: English comment :)

server/lib/video-import-channel.ts
```diff
@@ -0,0 +1,87 @@
...
+          importDataOverride: {
+            privacy: VideoPrivacy.PUBLIC
+          }
+        })
+        await job.finished()
+        result.successes += 1
+      } else {
+        result.alreadyImported += 1
+      }
+    } catch (ex) {
```
Chocobozzz, line 80: `err`

server/lib/video-import-channel.ts
```diff
@@ -0,0 +1,87 @@
...
+      result.errors += 1
+      logger.error(`An error occured while importing ${targetUrl}: ${ex.stack}`)
```
Chocobozzz, line 82: `logger.error('...', { err })`

server/middlewares/validators/videos/video-channel-sync.ts
```diff
@@ -0,0 +1,65 @@
...
+    const syncId = parseInt(req.params.id, 10)
+    const sync = await VideoChannelSyncModel.loadWithAccount(syncId)
+    if (!sync) {
+      return res.fail({
+        status: HttpStatusCode.NOT_FOUND_404,
+        message: 'Synchronization not found'
+      })
+    }
+    res.locals.videoChannelSync = sync
+    res.locals.videoChannel = sync.VideoChannel
```
Chocobozzz, line 51: If you need the channel in another middleware, prefer to use `res.locals.videoChannelSync.VideoChannel` directly.
This way you can only fetch video sync channel attributes you need, and not fetch all attributes also needed by unrelated middlewares that use `res.locals.videoChannel`

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,177 @@
...
+  UpdatedAt
+} from "sequelize-typescript"
+import { AccountModel } from "../account/account"
+import { getSort, throwIfNotValid } from "../utils"
+import { VideoChannelModel } from "./video-channel"
+import { Op } from "sequelize"
+import { UserModel } from "../user/user"
+import { isUrlValid } from "@server/helpers/custom-validators/activitypub/misc"
+
+type AvailableForListOptions = {
```
Chocobozzz, line 27: I don't think you need this

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  @AllowNull(false)
+  @Default(VideoChannelSyncState.WAITING_FIRST_RUN)
+  @Is('VideoChannelSyncState', value => throwIfNotValid(value, isVideoChannelSyncStateValid, 'state'))
+  @Column
+  state: VideoChannelSyncState
+
+  @AllowNull(true)
+  @Column(DataType.DATE)
+  lastSyncAt?: Date
```
Chocobozzz, line 85: `lastSyncAt: Date`

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,177 @@
...
+          model: VideoChannelModel,
+          required: true,
+          where: {
+            accountId: options.accountId
+          }
+        }
+      ]
+    }
+    return Promise.all([
+      VideoChannelSyncModel.count(query),
```
Chocobozzz, line 107: `count` should use `VideoChannelModel.unscoped` include or the count may be wrong because of channel multiple avatars
`VideoChannelSyncModel` should also be `unscoped()` to prevent double `VideoChannelModel` include

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,177 @@
...
+      VideoChannelSyncModel.findAll(query)
```
Chocobozzz, line 108: `VideoChannelSyncModel.unscoped()`

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,177 @@
...
+    ]).then(([ total, data ]) => ({ total, data }))
+  }
+
+  toFormattedJSON (this: MChannelSyncFormattable): VideoChannelSync {
```
Chocobozzz, line 112: Please put this below static methods

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,177 @@
...
+      },
+      externalChannelUrl: this.externalChannelUrl,
+      createdAt: this.createdAt.toISOString(),
+      channel: this.VideoChannel.toFormattedSummaryJSON(),
+      lastSyncAt: this.lastSyncAt?.toISOString()
+    }
+  }
+
+  static load (id: number) {
+    return this.findOne({
```
Chocobozzz, line 127: `findByPk`

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,177 @@
...
+      where: { id }
+    })
+  }
+
+  static loadWithAccount (id: number) {
+    return this.findOne({
```
Chocobozzz, line 133: `unscoped()`

server/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,177 @@
...
+        {
+          model: VideoChannelModel.unscoped(),
+          required: true,
+          include: [
+            {
+              model: AccountModel.unscoped(),
+              required: true,
+              include: [ {
+                attributes: [],
+                model: UserModel.unscoped(),
```
Chocobozzz, line 158: Should be `required` no?

server/models/video/video-import.ts
```diff
@@ -161,6 +161,26 @@ export class VideoImportModel extends Model<Partial<AttributesOnly<VideoImportMo
     ]).then(([ total, data ]) => ({ total, data }))
   }
 
+  static async urlAlreadyImported (channelId: number, targetUrl: string): Promise<boolean> {
+    const count = await (VideoImportModel.unscoped().count({
```
Chocobozzz, line 165: `findOne` will generate a faster SQL request because of `LIMIT 1`

server/tests/api/videos/video-channel-syncs.ts
```diff
@@ -0,0 +1,231 @@
...
+                    : 'https://api.github.com/repos/yt-dlp/yt-dlp/releases',
+
+                  name: mode
+                }
+              }
+            }
+          }
+        })
+        await setAccessTokensToServers([ server ])
+        await setDefaultVideoChannel([ server ])
```
Chocobozzz, line 60: Also use `setDefaultChannelAvatar` and `setDefaultAccountAvatar` to ensure we don't have bad `total` results from REST API

server/tests/api/check-params/video-channel-syncs.ts
```diff
@@ -0,0 +1,392 @@
...
+    await setDefaultVideoChannel([ server ])
+
+    command = server.channelSyncs
+
+    rootChannelId = server.store.channel.id
+
+    {
+      userInfo.username = 'user1'
+      const password = 'my super password'
+      await server.users.create({ username: userInfo.username, password })
```
Chocobozzz, line 53: You can also use `server.users.generateUserAndToken`

server/tests/api/check-params/video-channel-syncs.ts
```diff
@@ -0,0 +1,392 @@
...
+      const { videoChannels, ...otherUserInfo } = await server.users.getMyInfo({ token: userInfo.accessToken })
+      userInfo.id = otherUserInfo.id
+      userInfo.defaultQuota = otherUserInfo.videoQuota
+      userInfo.defaultDailyQuota = otherUserInfo.videoQuotaDaily
+      userInfo.channelId = videoChannels[0].id
+    }
+    await server.config.enableChannelSync()
+  })
+
+  after(async function () {
```
Chocobozzz, line 66: Put this after all tests for consistency

server/tests/api/check-params/video-channel-syncs.ts
```diff
@@ -0,0 +1,392 @@
...
+    channelId: -1,
+    syncId: -1,
+    defaultQuota: -1,
+    defaultDailyQuota: -1
+  }
+
+  async function withChannelSyncDisabled<T> (callback: () => T): Promise<T> {
+    try {
+      await server.config.disableChannelSync()
+      return callback()
```
Chocobozzz, line 32: prefer `await callback()` because it's sometimes difficult to understand the finally block will be executed after the `return` :)

server/tests/api/videos/video-channel-syncs.ts
```diff
@@ -0,0 +1,231 @@
...
+  function runSuite (mode: 'youtube-dl' | 'yt-dlp') {
+    describe('Sync using ' + mode, function () {
+      let server: PeerTubeServer
+      let command: ChannelSyncsCommand
+      let startTestDate: Date
+      const userInfo = {
+        accessToken: '',
+        username: 'user1',
+        get channelName () {
+          return this.username + "_channel"
```
Chocobozzz, line 27: prefer `channelName: 'user1_channel'` for simplicity

server/types/models/video/video-channel-sync.ts
```diff
@@ -0,0 +1,13 @@
+import { VideoChannelSyncModel } from "@server/models/video/video-channel-sync"
+import { FunctionProperties, PickWith } from "@shared/typescript-utils"
+import { MChannelFormattable } from "./video-channels"
+
+type Use<K extends keyof VideoChannelSyncModel, M> = PickWith<VideoChannelSyncModel, K, M>
+
+type MChannelSync = Omit<VideoChannelSyncModel, 'VideoChannel'>
+export type MChannelSyncChannel = VideoChannelSyncModel
```
Chocobozzz, line 8: This should be `MChannelSync` + the channel object you need

shared/models/videos/channel-sync/video-channel-sync.model.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export interface VideoChannelSync {
+  id: number
+
+  externalChannelUrl: string
+
+  createdAt: string
+  channel: VideoChannelSummary
+  state: VideoConstant<VideoChannelSyncState>
+  lastSyncAt?: string
```
Chocobozzz, line 13: `lastSyncAt` is `string` or `null`, but I don't think it can be missing or `undefined`

shared/server-commands/videos/channel-syncs-command.ts
```diff
@@ -0,0 +1,66 @@
...
+    count?: number
+    sort?: string
+  }) {
+    const { accountName, sort = 'createdAt' } = options
+    const path = `/api/v1/accounts/${accountName}/video-channel-syncs`
+    return this.getRequestBody<ResultList<VideoChannelSync>>({
+      ...options,
+
+      path,
+      query: { sort, ...pick(options, [ 'start', 'count', 'withStats', 'search' ]) },
```
Chocobozzz, line 20: You don't need `withStats` and `search`

Use `pick` from `@shared/core-utils` that is strictly typed instead of the one of lodash

server/tests/api/videos/video-channel-syncs.ts
```diff
@@ -0,0 +1,231 @@
+import { FIXTURE_URLS } from '@server/tests/shared'
```
Chocobozzz, line 1: Add this file to `server/tests/api/videos/index.ts`

server/controllers/api/video-channel-sync.ts
fflorent: Otherwise I don't retrieve the `syncCreated.VideoChannel` property (even with the `@DefaultScope` added in the model).

If you have any tip to avoid the reload, I would appreciate!

server/initializers/checker-after-init.ts
fflorent: Alright. If single quotes should be preferred, do you mind if I later open a PR to add this eslint rule?
```
"quotes": "off",
"@typescript-eslint/quotes": ["error", "single", { "avoidEscape": true }]
```

With all the fixes.

server/initializers/constants.ts
fflorent: It's passed to youtube-dl / yt-dlp. It is the number of the latest of videos of the remote channel to check and maybe to upload. Without that limitation, there are performance issues (it can take very long).

The naming is bad. However I miss an idea for a better one.

server/controllers/api/video-channel-sync.ts
Chocobozzz: In fact you should check video channel existence in middleware and populate `res.locals.videoChannel`. Then you can just do `syncCreated.VideoChannel = res.locals.videoChannel`

server/initializers/checker-after-init.ts
Chocobozzz: It's in my todo, it's weird it was not activated

server/initializers/constants.ts
Chocobozzz: You can promote this constant as an admin setting (not mandatory to provide it on the web interface) :)

server/lib/job-queue/handlers/video-channels-sync.ts
fflorent: Done in 485d19649fb8e224c96579962b54ca4376156f9d

Also I added an option to fully synchronize a channel (i.e. call this API even after having created a synchronization).

Screenshot:
image

client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.ts
```diff
@@ -0,0 +1,101 @@
...
+    this.serverConfig = this.serverService.getHTMLConfig()
+    this.initialize()
+    this.videoChannelSyncActions = [
+      [
+        {
+          label: $localize`Delete`,
+          iconName: 'delete',
+          handler: videoChannelSync => this.deleteSync(videoChannelSync)
+        }
+      ]
```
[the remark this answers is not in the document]
fflorent: Actually done in ca4c56be8ff8fba0ba8233f75df4df5d3d93ce7b

client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.html
fflorent: I removed it.

server/models/video/video-channel-sync.ts
fflorent: Finally I changed my mind: I made no UI, as I am not sure it is important enough. I can make one if you prefer.

Done in 37082c1d3e3040be8580fb5b53d92d32e3e9e626

server/initializers/constants.ts
fflorent: Done in 37082c1d3e3040be8580fb5b53d92d32e3e9e626

fflorent (commented):
I took into account your remarks

fflorent:
I improved a little bit the UI of the list of the synchronizations: 
image

client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.html
fflorent: Finally done in 6a4820a9d

server/controllers/api/accounts.ts
```diff
@@ -72,6 +75,17 @@ accountsRouter.get('/:accountName/video-channels',
...
+accountsRouter.get('/:accountName/video-channel-syncs',
+  authenticate,
+  asyncMiddleware(accountNameWithHostGetValidator),
+  ensureCanManageUser,
+  paginationValidator,
+  videoChannelSyncsSortValidator,
+  setDefaultSort,
+  setDefaultPagination,
+  asyncMiddleware(listAccountChannelsSync)
+)
```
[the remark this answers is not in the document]
fflorent: Issue reported here #5181

server/lib/job-queue/handlers/video-channel-import.ts
```diff
@@ -0,0 +1,41 @@
+import { logger } from '@server/helpers/logger'
+import { YoutubeDLCLI } from '@server/helpers/youtube-dl'
+import { CONFIG } from '@server/initializers/config'
+import { VideoChannelModel } from '@server/models/video/video-channel'
+import { VideoChannelImportPayload } from '@shared/models'
+import { Job } from 'bull'
+import { synchronizeChannel } from '@server/lib/video-import-channel'
+
+export async function processVideoChannelImport (job: Job) {
+  const payload = job.data as VideoChannelImportPayload
```
fflorent, line 10: I wonder if there could not be some issues with video orders when:
- the user creates a synchronization, requesting a channel import
- the channel import runs while a sync is run

I wonder if we should no passe an optional `syncId` in the payload to flag the synchronization as being processed, and skip any of sync in `PROCESSING`.

What do you think @Chocobozzz?

Chocobozzz:
Thanks, I'll rebase from develop and merge this PR

server/lib/job-queue/handlers/video-channel-import.ts
Chocobozzz: Maybe we can use the same queue for these two tasks so they are processed sequentially.
I'll check taht

Chocobozzz: I'll check how we can handle this case and update the code

Chocobozzz:
:tada: :tada: :tada:  :pray:  :)

emansom:
I have migrated a YouTube channel to my PeerTube instances a few weeks ago, when this feature wasn't available. Is there any way I can flag the existing videos as being "imported" already?

fflorent:
> I have migrated a YouTube channel to my PeerTube instances a few weeks ago, when this feature wasn't available. Is there any way I can flag the existing videos as being "imported" already?

@emansom If these videos were imported through the [Import with URL](https://docs.joinpeertube.org/use-create-upload-video?id=import-with-url) feature, you should have nothing to do, these videos should already be recognized as imported.

emansom:
> @emansom If these videos were imported through the [Import with URL](https://docs.joinpeertube.org/use-create-upload-video?id=import-with-url) feature, you should have nothing to do, these videos should already be recognized as imported.

They were imported using the `peertube-import-videos.js` CLI and don't show up in the imported tab.
https://docs.joinpeertube.org/maintain-tools?id=peertube-import-videosjs

fflorent:
> > @emansom If these videos were imported through the [Import with URL](https://docs.joinpeertube.org/use-create-upload-video?id=import-with-url) feature, you should have nothing to do, these videos should already be recognized as imported.
> 
> They were imported using the `peertube-import-videos.js` CLI and don't show up in the imported tab. https://docs.joinpeertube.org/maintain-tools?id=peertube-import-videosjs

I see.

That being said, if you create a new sync without importing the existing videos, only videos newer than the creation date are eligible for importation.

emansom:
> I see.
> 
> That being said, if you create a new sync without importing the existing videos, only videos newer than the creation date are eligible for importation.

Which database table and which column determines the "is imported" state? Or is it another table with video ID references that I need to add entries into?

fflorent:
I meant that you should have nothing to do, except creating a synchronization with the option "only watch for new publications". Only the videos uploaded on Youtube after the synchronizations will be imported.

If you are still willing to play with the database, the table is named `videoImport` and place the `state` value to `2`. Ensure to use the sequence (`videoImport_id_seq`) for the `id` column. Note that's ok for this table if you know what you are doing, but avoid manipulating PeerTube DB tables in general (otherwise that can lead to confusion on ActivityPub).

tio-trom:
I just want to say a huge THANK YOU! Amazing.

Seyferto:
> I meant that you should have nothing to do, except creating a synchronization with the option "only watch for new publications". Only the videos uploaded on Youtube after the synchronizations will be imported.

There's no option to import all prior ones?

emansom:
> > I meant that you should have nothing to do, except creating a synchronization with the option "only watch for new publications". Only the videos uploaded on Youtube after the synchronizations will be imported.
> 
> There's no option to import all prior ones?

There is, manually. The way I did it was:

1. Change the state as described [above by @fflorent](https://github.com/Chocobozzz/PeerTube/pull/5135#issuecomment-1210388067)
2. Setting up [pgAdmin](https://github.com/postgres/pgadmin4) and connecting to the PeerTube database
2. Export the videos table as CSV, opening it in LibreOffice Calc and transforming it into a structure compatible with the schema of videoImports

Seyferto:
This could also be dealt better with just an interface click that automates it... but perhaps I should get used to how mirroring and syncing will always be complex tasks, if not impossible elsewhere...

PR #5170, opened 2022-08-04
Fix lint on server/lib/hls.ts

fflorent:
npm run lint report an error on server/lib/hls.ts

And indeed, the `async` is useless in this function as we just need to return the result of `playlistFilesQueue.add` which is a Promise.


Chocobozzz:
Thanks!

PR #5171, opened 2022-08-04
Disallow sending analytics data to Angular team

fflorent:
## Description

Each time we run `npm run build`, a prompt asks us whether we want to send analytics data to angular.

I think we should disallow that by default.


Chocobozzz:
Thanks!

PR #5190, opened 2022-08-11
Tests: Add response error along with unexpected status in report

fflorent:
## Description

Show the error message the server returned in the response when the status is not the one expected.

Note: I proposed to help the user so they can see more details in log. The last line displays a URL. I am not very sure of the relevancy, feedback welcome.

## Has this been tested?

### DIFF
--- a/shared/server-commands/requests/requests.ts
+++ b/shared/server-commands/requests/requests.ts
@@ -172,7 +172,6 @@ function buildRequest (req: request.Test, options: CommonRequestParams) {
   if (options.accept) req.set('Accept', options.accept)
   if (options.host) req.set('Host', options.host)
   if (options.redirects) req.redirects(options.redirects)
-  if (options.expectedStatus) req.expect(options.expectedStatus)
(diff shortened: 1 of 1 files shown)


shared/models/http/http-error-codes.ts
```diff
@@ -4,7 +4,7 @@
  *
  * WebDAV and other codes useless with regards to PeerTube are not listed.
  */
-export const enum HttpStatusCode {
+export enum HttpStatusCode {
```
Chocobozzz, line 7: Please keep the `const` here that reduce client bundle size

I'm not sure the HTTP status label is very interesting. If you think so, you'll have to build the object yourself

fflorent: Yeah, that was the price for this feature. If you don't think it is worth it to mention the label, I revert that.

fflorent: Done

Chocobozzz:
Thanks!

PR #5200, opened 2022-08-16
Allow continusously printing logs through tail and parse-log

fflorent:
## Description

While testing APIs, one may want to only print latest logs and continuously follow them in order to ease the logs analysis.

This PR aims at allowing parse-log to read stdin so it can be piped with the output of commands like `tail` and `tail -f`.

## Has this been tested?


Chocobozzz:
Thanks!

PR #5214, opened 2022-08-24
server: serve files from storage/well-known

### LINKED ISSUE #5206: Add support for Apple app site association

> When you support universal links, iOS users can tap a link to your website and get seamlessly redirected to your installed app without going through Safari. If your app isn’t installed, tapping a link to your website opens your website in Safari.

https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html

To be enable this feature you have to put a file in either the web root or /.well-known directory. Since PeerTube is managing all the routing this has to be solved in either the nginx config or in the web app.

Add a location directive in the nginx config that points `/.well-known/apple-app-site-association` to a apple-app-site-association file in the storage folder. I'm not sure if it should be put directly in the storage folder or in some sub folder.

kontrollanten:
## Description
Give the server admin the possibility to add files in storage/well-known to be served by `/.well-known/`. Makes it possible to add Apple app site association file.

## Related issues
closes #5206

## Has this been tested?

Since all logic is handled by express this may not need tests?


Chocobozzz:
Hello,

Why don't you prefer to add your custom directive in your nginx configuration? It seems to be very specific to your peertube instance.

config/default.yaml
```diff
@@ -122,6 +122,7 @@ storage:
   captions: 'storage/captions/'
   cache: 'storage/cache/'
   plugins: 'storage/plugins/'
+  well_known: 'storage/well-known/'
```
Chocobozzz, line 125: Please fix the path

Chocobozzz (changes requested):
Please add server tests.

> Since all logic is handled by express this may not need tests?

Yes you need to add tests to ensure non existing well known file returns `404`, and if you add a file it answers it. We may wrongly configure the express middlewares.

config/default.yaml
Chocobozzz: Sorry, the comment was on the wrong line :)

config/production.yaml.example
```diff
@@ -120,6 +120,7 @@ storage:
   captions: '/var/www/peertube/storage/captions/'
   cache: '/var/www/peertube/storage/cache/'
   plugins: '/var/www/peertube/storage/plugins/'
+  well_known: 'storage/well-known/'
```
Chocobozzz, line 123: Please fix the path :)

Chocobozzz (changes requested):

PR #5215, opened 2022-08-24
Add additional certbot configuration information

nixuno:
## Description
As described [here](https://community.letsencrypt.org/t/auto-renewal-started-failing-with-error-missing-command-line-flag-or-config-entry-for-this-setting/111211/4) there must be a `[[webroot_map]]` entry for each domain in the certbot .conf file. This should be included in the documentation.

## Related issues
N/A

## Has this been tested?


Chocobozzz:
Hi,

Are you sure this entry is needed? From the documentation it seems it will just override the default `webroot_path`. Do you have an error when you run certbot with the current config?

Issue #3282, opened 2020-11-10
Add "a live video started" notifications for admins


Chocobozzz:
To help moderating lives for example.

Niquarl:
This could help greatly indeed on top of an allowlist of users. Thank you for creating this issue.

PR #5239, opened 2022-09-07
Expose PeerTube socket to plugins

kontrollanten:
## Description
Expose socket to plugins, so that they can send notifications and update live video state.

## Related issues
#3282

## Has this been tested?
Doesn't make sense to test this.


AnvarKhatik:
Can I Contribute This Project & Can You Help Me To Guide

Chocobozzz:
> I can't see that my changes affect this test, may it be flaky?

I'll relax them!

server/lib/plugins/plugin-helpers-builder.ts
```diff
@@ -218,6 +222,17 @@ function buildPluginRelatedHelpers (plugin: MPlugin, npmName: string) {
   }
 }
 
+function buildSocketHelpers () {
+  return {
+    sendNotification: (userId, notification) => {
```
Chocobozzz, line 227: Please type parameters

server/tests/fixtures/peertube-plugin-test-four/main.js
```diff
@@ -128,6 +128,22 @@ async function register ({
 
       return res.json(result)
     })
+
+    router.post('/send-notification', async (req, res) => {
+      peertubeHelpers.socket.sendNotification(1, {
```
Chocobozzz, line 133: Most of the type `root` is id `1`, but it may change. It would be better for the tests to send the user id in body

Chocobozzz (changes requested):
> To really ensure it works I guess we'd need to setup a WS client and check the message received. I couldn't find more ways in testing it. Any input?

Yes, you can take inspiration of https://github.com/Chocobozzz/PeerTube/blob/develop/server/tests/api/live/live-socket-messages.ts#L72

PR #5240, opened 2022-09-07
p2p-media-loader: dont validate segments from http

kontrollanten:
## Description
Validating segments shouldn't be needed when they're fetched from http.

## Has this been tested?


Chocobozzz:
Hello,

They need to be validated because they can come from other PeerTube instances :)

Chocobozzz:
I think you can already alter

PR #5241, opened 2022-09-07
server: treat WAITING_FOR_LIVE as available

### LINKED ISSUE #5235: Add admin setting to choose to display scheduled lives in browse videos page

A live video isn't visible until the stream starts.

### Steps to reproduce

1. Create a normal live
2. Set privacy to Public
3. Go to Recent videos, the video isn't published

The user should be possible to publish the live before the stream starts, in order to announce that a live is upcoming. Eventually upcoming lives should be placed in the top of the list to highlight it, this assumes that users will create the live relatively shortly before stream starts in order to not spam the video list page.

### Additional information

* PeerTube instance:
  * Version: 4.2.2

kontrollanten:
## Description
Treat public lives as available, even if they don't have a stream since it can be a way to announce that a live is upcoming.

## Related issues
closes #5235

## Has this been tested?


Chocobozzz:
See https://github.com/Chocobozzz/PeerTube/issues/5235#issuecomment-1246796814

PR #5267, opened 2022-09-15
Escape % signs in crontab example (CLI tools docs)

thomaskuntzz:
% is a special character for crontab. From man 5 crontab: A "%" character in the command, unless escaped with a backslash (\), will be changed into newline characters, and all data after the first % will be sent to the command as standard input.


Chocobozzz:
Thanks!

PR #5338, opened 2022-10-10
Sync channel: move the list imports button #5337

### LINKED ISSUE #5337: Sync: List imports button cut

The `List imports` button of the synchronization screen is cut when we resize the browser window. Depending on the language, it can make the sense hardly comprehensible.

### Steps to reproduce

1. Go to http://[instance.tld]/my-library/video-channel-syncs
2. Reduce the width of the browser window (or use responsive mode in browser devtools)

You should see the button cut as below:
list imports cut button

Especially in French, we can hardly understand what the button is meant for:
list imports cut button - french

Options I see:
1. Add a tooltip can describe what the button is meant for (but the tooltip is something hardly discoverable and cannot help on mobile / tablets)
2. Move this button in the actions (but it is semantically wrong, as this button does not imply changes on the synchronization)
3. Change the layout of the table when the button begins to be cropped

Personal opinion: I think the option 2 is the most convenient and answer to the problem. Some action button in the PeerTube interface don't imply a change on the item, like viewing the stats on a video (if this example is not meant to be an exception about what is allowed to be an action):

video stats button

### Additional information

* PeerTube instance:
  * URL: localhost
  * Version: 4.3.0

* Browser name, version and platforms on which you could reproduce the bug: Firefox developer edition (106.0b6 (64 bits)) or Nightly

fflorent:
## Description

On channel sync screen, move the list imports button form the left of the table into the action buttons list.

## Related issues

Fixes #5337

## Has this been tested?

## Screenshots

Previously:

list imports cut button

With the patch:

list imports in actions list

### DIFF
--- a/client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.ts
+++ b/client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.ts
@@ -46,6 +46,13 @@ export class MyVideoChannelSyncsComponent extends RestTable implements OnInit {
     this.videoChannelSyncActions = [
       [
         {
+          label: $localize`List imports`,
+          linkBuilder: (videoChannelSync) => [ `/my-library/video-imports?search=videoChannelSyncId:${videoChannelSync.id}` ],
+          iconName: 'cloud-download'
+        }
+      ],
+      [
+        {
           label: $localize`Delete`,
           iconName: 'delete',
           handler: videoChannelSync => this.deleteSync(videoChannelSync)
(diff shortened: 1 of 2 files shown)


client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.ts
```diff
@@ -44,6 +44,13 @@ export class MyVideoChannelSyncsComponent extends RestTable implements OnInit {
     this.initialize()
 
     this.videoChannelSyncActions = [
+      [
+        {
+          label: $localize`List imports`,
```
fflorent, line 49: Note that the button is not localized anymore.

Is there any way to get the translation back here?

Chocobozzz: It will when I'll update translations I think

client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.ts
```diff
@@ -44,6 +44,13 @@ export class MyVideoChannelSyncsComponent extends RestTable implements OnInit {
...
+          linkBuilder: (videoChannelSync) => [ `/my-labrary/video-imports?search=videoChannelSyncId:${videoChannelSync.id}` ],
```
Chocobozzz, line 50: `/my-labrary/` -> typo :)

Chocobozzz (changes requested):

client/src/app/+my-library/my-video-channel-syncs/my-video-channel-syncs.component.ts
fflorent: Wow, impressive eyes :eyes:

Issue #5134, opened 2022-07-18
To add an alphabetical classification of videos


klorydryk:
Hi,

When I open a video list, I can order them by "recent", "hot", "likes"... but not by name. Is it possible to add this way to classify the videos?

Thanks.

pyoung-fr:
In the same time, it would be interesting to have ability to save default order for each channel.
Thanks

PR #5351, opened 2022-10-13
Added filter to sort videos by name (alphabetical order)

Poslovitch:
## Description

As requested in #5134, I have added an option in the "sort" menu in order to use the already-existing "name" filter.

## Related issues
Implements https://github.com/Chocobozzz/PeerTube/issues/5134

## Has this been tested?

## Screenshots

image

(you can see it's sorting alphabetically, instead of any other sorts that would put the SkylandStyle video first (date, view count, etc.))


Poslovitch:
The test failure seems unrelated to the change brought by this PR

Chocobozzz:
Thanks!

PR #5441, opened 2022-12-02
FIX: search on private videos

koying:
## Description

Currently, it is not possible for an user to search on his private videos.
This fixes it (I think)

## Related issues

How to reproduce:
- Upload a private video, adding a unique tag, e.g. "foobar"
- Use the search box to find back your video, using "foobar" as search term
- See that the video is not found

## Has this been tested?

Tested manually

- [X] 🙋 no, because I need help 

No clue on your test framework, tbh.

### DIFF
--- a/server/models/video/sql/video/videos-id-list-query-builder.ts
+++ b/server/models/video/sql/video/videos-id-list-query-builder.ts
@@ -323,7 +323,9 @@ export class VideosIdListQueryBuilder extends AbstractRunQuery {
   private wherePrivacyAvailable (user?: MUserAccountId) {
     if (user) {
       this.and.push(
-        `("video"."privacy" = ${VideoPrivacy.PUBLIC} OR "video"."privacy" = ${VideoPrivacy.INTERNAL})`
+        `("video"."privacy" = ${VideoPrivacy.PUBLIC} OR ` +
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Hello,

Thanks for the issue and the PR but these endpoints are only for public videos. The endpoint to list your videos is https://docs.joinpeertube.org/api-rest-reference.html#tag/My-User/paths/~1api~1v1~1users~1me~1videos/get (https://peertube.example.com/my-library/videos) where there is also a search bar.

koying:
Are you talking about the "filter" bar in "My Library-Videos"?
That one, well, filters. It doesn't return the same set of result.
E.g. it doesn't search on tags...

Chocobozzz:
You can create a dedicated issue to add ability for users to search their videos with tags

binerf:
Hello,

Is there any issue created for this feature (full filtering on Private videos with tag for example) ? :thinking: 
I cannot find any.

Thank you.

Chocobozzz:
I don't think so :)

koying:
@binerf TBH, my patch here does exactly what I need, and is trivial to maintain (so far).
If you want what @Chocobozzz proposes, you'll have to open the feature request yourself ;)

PR #5484, opened 2022-12-28
Log HLS fatal error with error label

kontrollanten:
## Description
HLS.js fatal errors are currently logged as warnings, and non-fatal as errors. I assume this is a typo. Additionally I added the current time in the error logging, so it's gets easier to reproduce these errors.

## Has this been tested?


Chocobozzz:
Thanks!

PR #5486, opened 2022-12-28
feat(player/playbackRate): set rate with url param

kontrollanten:
## Description
I'm using a puppeter script that runs to test PeerTubes video streams. To increase the speed it would be great to set the playback rate via an URL parameter.

## Has this been tested?


Chocobozzz:
Thanks!

Issue #4734, opened 2022-01-20
Apple Podcast support


kontrollanten:
**Describe the problem to be solved**
A lot of our users ask for listening to our videos through their podcast platform. Personally I don't have any big experience of creating or consuming podcasts, but as I understand Apple Podcast is one of the biggest consuming platforms, therefore I think adding support for Apple Podcast is a great start. 

**Describe the solution you would like:**
Create RSS feeds for channels that are [compatible with Apple Podcast](https://podcasters.apple.com/support/823-podcast-requirements). What needs to be done:

- [x] Add RSS feeds for channels.
- [x] language tag for each channel.
- [x] itunes:image tag including channel image.
- [x] itunes:category tag for each channel.
- [x] itunes:explicit tag for each channel.
- [x] link tag containing link to the channel page.

For each video:
- [x] itunes:duration
- [x] itunes:image
- [x] mp4 file in the enclosure tag. (#4052)

**Related isssues

* https://github.com/Chocobozzz/PeerTube/issues/4052
* https://github.com/Chocobozzz/PeerTube/issues/3761
* https://github.com/Chocobozzz/PeerTube/issues/2814

emansom:
A better path for this would be PeerTube publishing a [RSS 2.0 compatible](3761) containing audio-only streams, that can then be imported by Castopod.

[Castopod](https://castopod.org/) can then take care of republishing to Apple Podcast, which it was designed for.

dhk2:
Podcast2 plugin should be able to function as a work around until this is addressed. It checks the required boxes and is passing all the validators I can find.

psaunders:
Some issues have been encountered with the Podcast2 plugin.
Is this feature being brought into the actual PeerTube codebase?

Chocobozzz:
@dhk2 Can you detail all the features you developed in the podcast2 plugin?

I improved our current implementation in https://github.com/Chocobozzz/PeerTube/commit/c88cb2166336e2fd43becbd869b2f5d380005f09

I'll check the apple validator when peertube2.cpy.re (podcast feed example: https://peertube2.cpy.re/feeds/podcast/videos.xml?videoChannelId=32421) will be updated.

Chocobozzz:
Seems to work correctly. We can close this issue!

wakest:
> I improved our current implementation in [c88cb21](https://github.com/Chocobozzz/PeerTube/commit/c88cb2166336e2fd43becbd869b2f5d380005f09)

[...]

Issue #3786, opened 2021-02-26
Publish livestreams in RSS feed


sgiraudot:
Currently, the livestreams do not seem to appear in the RSS feed. It would nice if they would, as it would for example allow users to announce their future livestreams via RSS.

__Example:__ I plan to do a livestream on Saturday, I create the livestream the day before, viewers who subscribed to the RSS of the channel get the link for the livestream and are thus aware that there's going to be a stream soon (announcing the date/time when it will actually be streamed could remain the responsibility of the user).

I imagine that you could object that links for livestream would be irrelevant in RSS feed once the stream is over, so maybe it should only be done when the replay is switched on (as, as far as I know, the link remains the same for stream and replay).

rigelk:
How do you plan to inform the user of a specific time at which the livestream is going to start? How do you deal with streams that already ended?

sgiraudot:
> How do you plan to inform the user of a specific time at which the livestream is going to start?

I don't think that would be done automatically, but users could announce it themselves. What I did when I tried (and realized it did not appear in the RSS) was to simply put in in the title ("_rendez-vous samedi à 11h_"). Of course that's not ideal but the point is that the link would pop in the RSS feed as soon as the page is created.

> How do you deal with streams that already ended?

That's probably a good argument to only publish livestreams _that are programmed to have a replay_ on RSS (that way the link should remain valid not matter when).

rigelk:
> > How do you plan to inform the user of a specific time at which the livestream is going to start?
> 
> I don't think that would be done automatically, but users could announce it themselves. What I did when I tried (and realized it did not appear in the RSS) was to simply put in in the title ("_rendez-vous samedi à 11h_"). Of course that's not ideal but the point is that the link would pop in the RSS feed as soon as the page is created.

Unfortunately, this is less than not ideal: we cannot automate it, since it relies on a specific language (here english) and not a semantic marker.

Invoking @agates here: since you have some experience with `<podcast:media>`, do you agree with my analysis that bare RSS2 doesn't provide any semantics for this? At least with Media RSS, they provide an example of using Dublin Core:

```xml
<dcterms:valid>
    start=2002-10-13T09:00+01:00;
    end=2002-10-17T17:00+01:00;
    scheme=W3C-DTF
</dcterms:valid>
```

Maybe setting a start date as soon as it has started, and an end date when the livestream ends would be enough. Non-saved livestreams would be deleted from the feed directly after the stream ends.

[...]

Issue #3761, opened 2021-02-21
Podcasting 2.0 RSS support


agates:
Hello,

I am interested in adding RSS support for [The "podcast" Namespace](https://github.com/Podcastindex-org/podcast-namespace) and possibly optional integration with the [Podcast Index](https://podcastindex.org/).

One of the goals of the project is to eventually add support for multiple "enclosures", via `podcast:alternateEnclosure` or with [my `podcast:media` proposal](https://github.com/Podcastindex-org/podcast-namespace/issues/174) (which was heavily influenced by Media RSS from PeerTube), in addition to all sorts of other things to help content creators, application developers, and hosts.  You can find out more on the [Podcasting 2.0 podcast](https://podcastindex.org/podcast/920666).

I think it's a great fit for PeerTube specifically as a decentralized video ecosystem.

With that said:
1. Would the project be interested in mainline support for this feature?
2. If not, is it currently possible to implement some of these features as a plugin?  Some initial things I'm thinking about:
   - Addition of a new RSS option for subscription
   - Addition of fields to a user's account and/or channel (primarily to include `podcast:value` support at first, but I'm sure this would expand)

agates:
Also, I think this a better solution for #2922 primarily because of the enclosure plans.

rigelk:
> Also, I think this a better solution for #2922 primarily because of the enclosure plans.

Could you elaborate?

agates:
> > Also, I think this a better solution for #2922 primarily because of the enclosure plans.
> 
> Could you elaborate?

The Podcast Namespace project is addressing the issues and lack of necessary features in standard, stagnated RSS with an official, community-supported XML namespace.

This includes, but is not limited to, a proposal to support multiple media formats and transport types.  Development here, regardless of the route taken, would be great for feedback particularly as a video platform with optional audio and multiple transport types.

In addition, some of the other proposals could provide a framework to work on other additional features.  One example is the `podcast:value` tag (still in flux) which provides a way for a client to stream bitcoin to the host/content creator via the lightning network.

I envision allowing PeerTube as a way for people to easily host their own video/audio "podcasts," but that's merely a vision in my head and I understand it's not the goal of the PeerTube project.

rosano:
Would be great to subscribe to any channel in a podcast player. Is there a way to accomplish this through plugins? If so, are there any plugins that one could use as a reference?

agates:
[...]

PR #5487, opened 2022-12-28
Add Podcast RSS feeds

agates:
## Description

The primary purpose of this PR is to add a new `podcast` Feed format that generates "Podcast" RSS feeds using [The Podcast Namespace](https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md) and to make the feeds easily consumable by standard podcast players and podcast indexes.

The secondary purpose is to add support for live streams in RSS as well as additional infrastructure to make live streams more usable in podcast applications.  It does so by clearing the cache of the Podcast RSS feed whenever the live stream goes live or ends, which ensures any client will have up-to-date information for the live stream.

It also adds the following action hook:
* `action:live.video.state.updated` - Gets fired when a live stream starts or ends.  Makes it possible for plugins to handle live stream notifications, which I will be using for a [podping.cloud](https://github.com/Podcastindex-org/podping.cloud) plugin.

And the following three filter hooks:
* `filter:feed.podcast.channel.create-custom-tags.result` - Returns a list of CustomTag objects for plugins to add custom elements to the Podcast RSS channel.
* `filter:feed.podcast.video.create-custom-tags.result` - Returns a list of CustomTag objects for plugins to add custom elements to Podcast RSS items or liveItems.
* `filter:feed.podcast.rss.create-custom-xmlns.result` - Returns a list of CustomXMLNS objects for plugins to add custom XMLNS definitions to the Podcast RSS feed.

Lastly, an `isEmailPublic` boolean field is added to the `UserModel` to allow the user to opt-in to showing their email address in the `<podcast:locked>` tag.  This allows the user to "claim" their feed in various podcast services to prove that it is theirs.  I put the toggle button for this above the "Change your email" section of the user profile.  It very clearly needs work (see screenshot).

## Related issues

[Podcasting 2.0 RSS support](https://github.com/Chocobozzz/PeerTube/issues/3761)
[Publish livestreams in RSS feed](https://github.com/Chocobozzz/PeerTube/issues/3786)

## Has this been tested?

## Screenshots

screenshot of draft is email public toggle

## Other notes

Support for [Apple Podcast](https://github.com/Chocobozzz/PeerTube/issues/4734) is not a purpose of this PR as one would need to do additional testing on an Apple device, but adding the extra itunes tags to the feed generator wouldn't be too difficult.  However, Apple requires very large images (3000x3000, I think).

Additionally, I made the decision not to include fragmented MP4s in the feeds because they offer an extremely poor user experience where some players wont play them correctly or only play the first segment.  If a webtorrent version of a video exists, it will default to the webtorrent MP4.  If only HLS exists, it defaults to the m3u8 file.

Lastly, this PR supports only channel level podcast feeds as that is the generally accepted standard use case for podcasting.  I don't think it would be difficult to extend to account- or subscription-level but as that is not a primary goal I haven't included it in testing.

## Feed dependency

I'm making this PR with the dependency on my fork of `@peertube/feed`, `pfeed-podcast`, because my PR there is still open.  This can and should be changed once that is updated.


server/controllers/feeds.ts
```diff
@@ -127,40 +133,81 @@ async function generateVideoFeed (req: express.Request, res: express.Response) {
...
+    // include: req.query.include | VideoInclude.FILES | VideoInclude.TAGS,
     hasFiles: true,
     countVideos: false,
     ...options
   })
 
-  addVideosToFeed(feed, data)
+  let feed: Feed
+
+  if (format === 'podcast') {
```
Chocobozzz, line 170: I think you should create a dedicated endpoint for podcast feed (like `/feeds/podcast/videos`). This way you would not accept `isLocal` or `sort` parameters.

server/controllers/feeds.ts
```diff
@@ -211,126 +260,357 @@ function initFeed (parameters: {
...
+      })
 
-    const categories: { value: number, label: string }[] = []
-    if (video.category) {
-      categories.push({
-        value: video.category,
-        label: getCategoryLabel(video.category)
+      // If both webtorrent and HLS are enabled, prefer webtorrent files
+      // standard files for webtorrent are regular videos
+      const groupedVideos = groupBy(videos, video => video.title)
```
Chocobozzz, line 340: Prefer using https://github.com/Chocobozzz/PeerTube/blob/develop/server/models/video/video.ts#L1618

server/controllers/feeds.ts
```diff
@@ -211,126 +260,357 @@ function initFeed (parameters: {
...
-      thumbnails: [
-        {
-          url: WEBSERVER.URL + video.getPreviewStaticPath(),
-          height: PREVIEWS_SIZE.height,
-          width: PREVIEWS_SIZE.width
+      }).filter(c => c)
+
+      const customTags: CustomTag[] = await Hooks.wrapObject(
+        [],
+        'filter:api.feed.podcast.item.custom-tags.result',
```
Chocobozzz, line 387: Prefer `filter:feed.podcast.video.create-custom-tags.result`

server/controllers/feeds.ts
```diff
@@ -211,126 +260,357 @@ function initFeed (parameters: {
...
+        { video }
+      )
+
+      const author = {
+        name: video.VideoChannel.Account.getDisplayName(),
+        href: video.VideoChannel.Account.Actor.url
+      }
+      const item = {
```
Chocobozzz, line 395: Please refactor this big object creation inside a function and reuse it to generate live JSON too

server/lib/live/live-manager.ts
```diff
@@ -435,9 +441,14 @@ class LiveManager {
...
       await fullVideo.save()
 
+      // Clear cache for Podcast RSS feed when live stream ends
+      await clearCacheRoute(`/feeds/videos.xml?videoChannelId=${fullVideo.channelId}&format=podcast`)
+
       PeerTubeSocket.Instance.sendVideoLiveNewState(fullVideo)
 
       await federateVideoIfNeeded(fullVideo, false)
+
+      Hooks.runAction('action:api.live-video.state.updated', { video: fullVideo })
```
Chocobozzz, line 451: Prefer `action:live.video.state.updated`

server/controllers/feeds.ts
```diff
@@ -127,40 +133,81 @@ async function generateVideoFeed (req: express.Request, res: express.Response) {
...
+
+  if (format === 'podcast') {
+    // If the first video in the channel is a film, that will be the only video in the feed
+    // Yes, this is a hack :)
+    const isFilm: boolean = data.length > 0 && data[data.length - 1].category === 2
+    const videos = isFilm ? [ data[data.length - 1] ] : data
+
+    const customTags: CustomTag[] = await Hooks.wrapObject(
+      [],
+      'filter:api.feed.podcast.channel.custom-tags.result',
```
Chocobozzz, line 178: filter:feed.podcast.channel.create-custom-tags.result

agates:
> What's the status of this? Looking forward to get this implemented.

I was away for a while but am working on requested changes now.

server/lib/live/live-manager.ts
```diff
@@ -369,13 +370,18 @@ class LiveManager {
 
       await wait(getLiveSegmentTime(live.latencyMode) * 1000 * VIDEO_LIVE.EDGE_LIVE_DELAY_SEGMENTS_NOTIFICATION)
 
+      // Clear cache for Podcast RSS feed when live stream starts
+      await clearCacheRoute(`/feeds/videos.xml?videoChannelId=${video.channelId}&format=podcast`)
```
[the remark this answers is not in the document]
agates: The issue with cache at all is it messes with live notifications -- it's been standard practice up until now to invalidate cache for podcast live streams and I believe it's worth the effort vs regenerating the feed every minute.

The reason for this is we have seen live notifications for podcasts occur only for clients to retrieve a cached version without the live stream information, causing a lot of headache on the client side for the user/podcast application.

Additionally, if we decide to keep the cache and lower it, we would need to run a job to run the notification when cache is set to expire.

Agorise:
We'll implement every bit of this into hive-tube, this is so exciting! Great work Alecks!! :)

agates:
The PR [Add support for feed generation using the Podcast Namespace](https://github.com/Chocobozzz/feed/pull/1) is also a dependency.

server/controllers/feeds.ts
```diff
@@ -207,43 +228,304 @@ async function generateVideoFeedForSubscriptions (req: express.Request, res: exp
...
+    start,
+    count: CONFIG.FEEDS.VIDEOS.COUNT,
+    sort: '-publishedAt',
+    displayOnlyForFollower: {
+      actorId: server.id,
+      orLocalVideos: true
+    },
+    nsfw,
+    // Prevent podcast feeds from listing videos in other instances
+    // helps prevent duplicates when they are indexed -- only the author should control them
```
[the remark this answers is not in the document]
agates: Essentially this boils down to the concept of ownership, duplicate feeds, and the user experience in the end user application.

Feeds get user-submitted and often automatically crawled for inclusion into podcast indexes such as https://podcastindex.org.  When multiple feeds of one podcast get submitted there comes into a question of who owns it and which one is "correct".

For the same reasons I don't allow additional URL parameters because all of the different parameters get included as duplicate.

That said, I'm actually not sure how relevant this is on the channel level.  Querying by the video channel ID will only include the local videos anyway, right?

server/controllers/feeds.ts
```diff
@@ -207,43 +228,304 @@ async function generateVideoFeedForSubscriptions (req: express.Request, res: exp
...
+    isLocal: true,
+    include: req.query.include | VideoInclude.FILES,
+    // TODO: include tags for future inclusion into the RSS feed
+    // include: VideoInclude.FILES | VideoInclude.TAGS,
```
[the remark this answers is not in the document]
agates: The draft specification for the RSS feed on what I wanted to do with this is still in flux.  I will remove it.

server/controllers/feeds.ts
```diff
@@ -207,43 +228,304 @@ async function generateVideoFeedForSubscriptions (req: express.Request, res: exp
...
+    hasFiles: true,
+    countVideos: false,
+    ...options
+  })
+
+  // If the first video in the channel is a film, that will be the only video in the feed
+  // Yes, this is a hack :)
+  const isFilm: boolean = data.length > 0 && data[data.length - 1].category === 2
```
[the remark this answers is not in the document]
agates: Understood.  I'll remove it until we can come up with a better solution in the future.

server/models/video/video.ts
```diff
@@ -1706,15 +1706,23 @@ export class VideoModel extends Model<Partial<AttributesOnly<VideoModel>>> {
...
 
   getFormattedVideoFilesJSON (includeMagnet = true): VideoFile[] {
+    return videoFilesModelToFormattedJSON(this, this.VideoFiles, { includeMagnet })
+  }
+
+  getFormattedVideoStreamingPlaylistsJSON (includeMagnet = true): VideoFile[] {
+    return flatten(this.VideoStreamingPlaylists.map(p => videoFilesModelToFormattedJSON(this, p.VideoFiles, { includeMagnet })))
+  }
+
+  getFormattedVideoFilesWithStreamingPlaylistsJSON (includeMagnet = true): VideoFile[] {
```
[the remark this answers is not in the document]
agates: I had to separate out `getFormattedVideoFilesJSON` to avoid deduplicating videos after the query.  We need to get all video files and this seemed like the best approach.  Otherwise I'd just be duplicating the logic of [videoFilesModelToFormattedJSON](https://github.com/Chocobozzz/PeerTube/blob/67599b83c544e05d62112e1faed808645a8e4303/server/models/video/formatter/video-format-utils.ts#L248C10-L289).

I believe you had suggested to use [getMaxQualityFile](https://github.com/Chocobozzz/PeerTube/blob/0c9668f77901e7540e2c7045eb0f2974a4842a69/server/models/video/video.ts#L1625-L1627) but that only gets the highest quality and not all available files.

Chocobozzz:
Thanks, I'll update this PR to integrate your changes in develop. Then you'll be able to create other PRs if there things to fix/improve.

Chocobozzz:
Do we consider the channel mandatory? Or can we generate a podcast feed of local videos of the instance?

Agorise:
Down to just 1 failed check now, that's awesome! I so hope we can see all this work in v5.2 :)

Issue #5497, opened 2022-12-31
Object Storage permissions using Bucket Policies


Kinuseka:
### Describe the problem to be solved

Object storage has been certainly a big part on some instances, and now on Peertube V5.0.0 allow to control our bucket's privacy so that it can no longer be access was certainly a big thing to prevent requests from someone that still knows the direct link. This was possible with the use of ACLs, it is an old feature by Amazon S3 which by them no longer recommends using and they actually encourage to use IAM and or policies instead. 

These are the object storage I have used in the past that I have come across

## Object Store With official support to ACLs
* Microsoft Azure Blob 
* Amazon Web Services, Simple Storage Services 
* Wasabi

## Object Store with limited support to ACLs
* Backblaze B2 (Cannot do ACLs object level)
* Cloudflare R2 (Cannot do ACL private)

## Object Store that denies support to ACLs
* Minio Object Storage

----
You can see that there are many inconsistencies with the API Support of ACLs  between various buckets, 
and even the cheapest one Backblaze B2 has incomplete support of it. While bigger companies with much expensive cost have a much complete API support, these definetly is a hurdle to the proper development of a proper video share especially those that use **Minio** with no support to ACLs whatsoever

### Describe the solution you would like

ACLs are definetly easier to work with but service providers often discourage them due to how "Expensive" they could be and could tangle the functionalities of a bucket.

## My Proposal
As a work around between these inconsistencies, we can do;

## **1. We can use bucket policies, as they are widely supported by any of those Object Storage providers I have mentioned above.**

These requires `PutBucketPolicy`  permissions

**Heres a working bucket policy that allows access to all files except `file.mp4`**
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "*"
                ]
            },
            "Action": [
                "s3:GetObject"
            ],
            "Resource": [
                "arn:aws:s3:::test/*"
            ]
        },
        {
            "Effect": "Deny",
            "Principal": {
                "AWS": [
                    "*"
                ]
            },
            "Action": [
                "s3:GetObject"
            ],
            "Resource": [
                "arn:aws:s3:::test/file.mp4"
            ]
        }
    ]
}
```
Anything appended on the denied resources will be denied in anonymous access (aka public access without authentication)
while keeping everything else public. **I have tested this and it works as expected** 

Heres a flowchart of how I expect things to flow 
[...]

PR #5502, opened 2023-01-02
Bucket policy as an alternative for ACLs

Kinuseka:
## Description

Currently peertube uses ACLs to restrict access to direct video links to the bucket. 
However since ACL is a legacy feature, it is often not added by the some object storage provider (even the best ones e.g. backblaze). 
This PR aims to use Bucket policy instead to restrict access to videos on direct links due to wide compatibility.

## Related issues
solves #5497

## Has this been tested?

tested on staging app [testvid.kinuseka.us](https://testvid.kinuseka.us)

# Goals and tests 🏁
- ✔ Toggable feature
- ✔ Changes policy 
- ✔ Adds video to policy and removes it when public
- ✔ Automatically adds policy when bucket policy is empty
- ❌ Identifies deny and allow (for now the program assumes `Policy.Statement[1]` is the Deny part of the policy)

# Efficiency 🔋
- ✔ HLS Per prefix policy
- ❌ Live Per prefix policy (For now it will keep appending the .ts files if live is private mode)
- ❌ Single folder private (Good for MinIO)

Template policy where this PR works on:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "*"
                ]
            },
            "Action": [
                "s3:GetObject"
            ],
            "Resource": [
                "arn:aws:s3:::vidtest/*"
            ]
        },
        {
            "Effect": "Deny",
            "Principal": {
                "AWS": [
                    "*"
                ]
            },
            "Action": [
                "s3:GetObject"
            ],
            "Resource": [
                "arn:aws:s3:::vidtest/favicon.ico" 
            ]
        }
    ]
}
```
since Bucket policy requires to have atleast 1 Resource in there, I just added favicon.ico as a temporary file


emansom:
Is this compatible with Ceph and MinIO?

Kinuseka:
> Is this compatible with Ceph and MinIO?

Yep! I made this PR because I was fraustrated with lack of compatibility of ACLs from MinIO

emansom:
This PR is adding some S3 protocol level functionality, could this be deferred to a third party library instead?
Would lower the maintenance burden on Framasoft somewhat, potentially.

Kinuseka:
I am trying to make this as flexible as possible by just adjusting how the ACL works and translating it into a bucket policy.
Both of these are configurable on CONFIGS, so if their bucket works with ACL, keep it enabled and the if the latter enable bucket policy. 

It is hard to rely on other obscure s3 compatible providers since they often 
lack the functionality crucial for the program or they have the functionality but does not work correctly.

If you mean we could merge this Bucket policy + ACL + (any other privacy methods) into a one giant library, which handles it. 

```
Step 1: If ACL fail try step 2
Step 2: If Bucket policy fail try step 3
Step 3: If all else fails then nothing works and keep direct link public
```

then that would be fine but for now, ill stick with this solution till further notice.

Kinuseka:
Pretty happy with the results so far, migration script works and I deployed it on my production environment
image

still on bughunting Ill report if I find issues along the way

Chocobozzz:
Closing this PR as we have not found a "perfect" replacement for ACL: https://github.com/Chocobozzz/PeerTube/issues/5497#issuecomment-1464834051

PR #5548, opened 2023-01-23
additional trackers for webtorrent applicable from config

desphilboy:
## Description
Currently the only tracker used in creation of webtorrents is the server itself.
there are additional public tracker that you can use as a fail safe tracker to make presence and accessability of your torrents more robust and redundant. this PR is to fix this issue

## Related issues
- power of global public trackers could not be leveraged
- The only tracker was server itself, so if  servers tracker was down or filtered, the video could not be tracked

## Has this been tested?

- [X] 🙋 yes, torrent being created as per existing tests, also tested on local and it contains additional trackers

### DIFF
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -293,6 +293,30 @@ tracker:
   private: true
   # Reject peers that do a lot of announces (could improve privacy of TCP/UDP peers)
   reject_too_many_announces: false
+  
+customize_torrents:
+  # If additional trackers is more than 0, 
+  # then as many as additional trackers entries from the trakcers list will be 
+  # added to the torrent file
+  # entries are randomly chosen if number is less than whole list otherwise all
+  # 0 , empty or absense of this property means disabled   
+  # additional_trackers: 6
+  # trackers_list:
+  #   - 'udp://tracker.leechers-paradise.org:6969'
+  #   - 'udp://www.torrent.eu.org:451/announce'
+  #   - 'udp://tracker.opentrackr.org:1337'
+  #   - 'udp://tracker.moeking.me:6969/announce'
+  #   - 'udp://explodie.org:6969'
+  #   - 'wss://tracker.btorrent.xyz'
+  #   - 'wss://tracker.openwebtorrent.com'
+  #   - 'udp://open.stealth.si:80/announce'
+  #   - 'udp://tracker.coppersurfer.tk:6969'
+  #   - 'udp://exodus.desync.com:6969/announce'
+  #   - 'udp://retracker.lanta-net.ru:2710/announce'
+  #   - 'http://open.acgnxtracker.com:80/announce'
+  #   - 'udp://tracker.torrent.eu.org:451/announce'
+  #   - 'udp://tracker.empire-js.us:1337'
+  #   - 'udp://open.stealth.si:80/announce'
 
 history:
   videos:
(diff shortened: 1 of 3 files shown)


config/default.yaml
```diff
@@ -295,6 +295,11 @@ tracker:
   private: true
   # Reject peers that do a lot of announces (could improve privacy of TCP/UDP peers)
   reject_too_many_announces: false
+  # This is a comma-separated list of additional trackers
+  # will be used in creation of webtorrents if specified
+  # example: 
+  #webtorrent_trackers_list_csv: 'udp://tracker.leechers-paradise.org:6969,udp://tracker.coppersurfer.tk:6969,wss://tracker.btorrent.xyz'
+  webtorrent_trackers_list_csv: ''
```
Chocobozzz, line 302: Use an array and a root key instead:

```
tracker:
  enabled: true
  ...

additional-trackers:
  urls:
    - 'wss://tracker1'
    -  ...
```

Chocobozzz (changes requested):
Please add server tests to ensure your code is working as expected

config/default.yaml
desphilboy: @Chocobozzz 
I need help, I could not make this work with my server.
this happens: 
Selection_181

Looks like config is being moderated from somewhere and I am not allowed to add things to config?
or is it just a yaml space problem?

desphilboy:
@Chocobozzz 
moved list of additonal trackers to a yaml array and a new root key as suggested.
the whole functionality tested on local server and works fine,
the existing testsuits are all success so I  assume I have not broken anything and webtorrents testing the functionality
Stil dont know how to directly write a test for it as there is no related tests

desphilboy:
working now on my instance:
peertube.freestorm.online

config/production.yaml.example
```diff
@@ -293,6 +293,30 @@ tracker:
...
   reject_too_many_announces: false
+  
+customize_torrents:
+  # If additional trackers is more than 0, 
+  # then as many as additional trackers entries from the trakcers list will be 
+  # added to the torrent file
+  # entries are randomly chosen if number is less than whole list otherwise all
+  # 0 , empty or absense of this property means disabled   
+  # additional_trackers: 6
+  # trackers_list:
```
Chocobozzz, line 304: Prefer to add `additional_trackers` array in https://github.com/Chocobozzz/PeerTube/blob/develop/config/production.yaml.example#L299 object

server/helpers/webtorrent.ts
```diff
@@ -47,8 +47,9 @@ async function downloadWebTorrentVideo (target: { uri: string, torrentName?: str
           deleteDownloadedFile({ directoryPath, filepath: file.path })
         }
 
-        return safeWebtorrentDestroy(webtorrent, torrentId, undefined, target.torrentName)
-          .then(() => rej(new Error('Cannot import torrent ' + torrentId + ': there are multiple files in it')))
+        return safeWebtorrentDestroy(webtorrent, torrentId, undefined, target.torrentName).then(() =>
```
Chocobozzz, line 50: Please revert styling changes of this file

config/production.yaml.example
```diff
@@ -293,6 +293,30 @@ tracker:
...
   # Reject peers that do a lot of announces (could improve privacy of TCP/UDP peers)
   reject_too_many_announces: false
+  
+customize_torrents:
+  # If additional trackers is more than 0, 
+  # then as many as additional trackers entries from the trakcers list will be 
+  # added to the torrent file
+  # entries are randomly chosen if number is less than whole list otherwise all
+  # 0 , empty or absense of this property means disabled   
+  # additional_trackers: 6
```
Chocobozzz, line 303: I find this too complex. I prefer to remove the max count config and just put inside the torrent all additional trackers defined in the config.

Chocobozzz (changes requested):
Thanks and sorry for the late review!

PR #5551, opened 2023-01-24
fix(client): hook filter:internal.player.videojs.options.result

kontrollanten:
## Description
PT 5 breaked the hook by providing the options as a Promise.

## Has this been tested?


Chocobozzz:
Sorry about that.
Can you test this patch instead? https://github.com/Chocobozzz/PeerTube/compare/develop...release/5.0.0
So we don't do the same mistake in the future

Chocobozzz:
Sorry, wrong link.

PR #5593, opened 2023-02-10
feat(server): add redis sentinel support

### LINKED ISSUE #5141: Redis Sentinel support

Currently the only way to ensure high availability for the redis server is to use master/slave setup with manual handling when making system upgrades.

> Redis Sentinel provides high availability for Redis when not using [Redis Cluster](https://redis.io/docs/manual/scaling).
> 
> * Monitoring. Sentinel constantly checks if your master and replica instances are working as expected.
> * Notification. Sentinel can notify the system administrator, or other computer programs, via an API, that something is wrong with one of the monitored Redis instances.
> * Automatic failover. If a master is not working as expected, Sentinel can start a failover process where a replica is promoted to master, the other additional replicas are reconfigured to use the new master, and the applications using the Redis server are informed about the new address to use when connecting.
> * Configuration provider. Sentinel acts as a source of authority for clients service discovery: clients connect to Sentinels in order to ask for the address of the current Redis master responsible for a given service. If a failover occurs, Sentinels will report the new address.

Read more at https://redis.io/docs/manual/sentinel/

The currently used redis client doesn't [support sentinels](https://github.com/redis/node-redis/issues/302) though, so a switch to ioredis would be needed. https://github.com/luin/ioredis#sentinel

kontrollanten:
## Description
Adds [redis sentinel](https://redis.io/docs/management/sentinel/) support.

## Related issues
closes #5141

## Has this been tested?
Yes, manually. By setting up the following sentinel and connecting:

```
version: '2'

services:
  redis-sentinel:
    image: docker.io/bitnami/redis-sentinel:7.0
    environment:
      - REDIS_MASTER_HOST=127.0.0.1
    networks:
      - webproxy
    ports:
      - 26379:26379
    volumes:
      - redis-sentinel_data:/bitnami
  redis:
    image: docker.io/bitnami/redis:7.0
    environment:
      # ALLOW_EMPTY_PASSWORD is recommended only for development.
      - ALLOW_EMPTY_PASSWORD=yes
    networks:
      - webproxy
    ports:
      - 6379:6379
    volumes:
      - redis_data:/bitnami
volumes:
  redis-sentinel_data:
    driver: local
  redis_data:
    driver: local

networks:
  webproxy:
    driver: bridge
```

Not sure if it's worth to setup a test for this feature?

### DIFF
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -63,8 +63,15 @@ database:
 redis:
   hostname: 'localhost'
   port: 6379
-  auth: null
+  auth: null # Used by both standalone and sentinel
   db: 0
+  sentinel:
+    enabled: false
+    enable_tls: false
+    master_name: ''
+    sentinels:
+      - hostname: ''
+        port: 26379
 
 # SMTP server to send emails
 smtp:
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -65,8 +65,15 @@ database:
 redis:
   hostname: 'localhost'
   port: 6379
-  auth: null
+  auth: null # Used by both standalone and sentinel
   db: 0
+  sentinel:
+    enabled: false
+    enable_tls: false
+    master_name: ''
+    sentinels:
+      - hostname: ''
+        port: 26379
 
 # SMTP server to send emails
 smtp:
--- a/server/initializers/checker-before-init.ts
+++ b/server/initializers/checker-before-init.ts
@@ -84,7 +84,8 @@ function checkMissedConfig () {
   const requiredAlternatives = [
     [ // set
       [ 'redis.hostname', 'redis.port' ], // alternative
-      [ 'redis.socket' ]
+      [ 'redis.socket' ],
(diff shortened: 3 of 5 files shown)


config/production.yaml.example
```diff
@@ -66,6 +66,13 @@ redis:
   auth: null
   db: 0
 
+# redis_sentinel:
```
Chocobozzz, line 69: Can we put this key as `sentinel` inside `redis` key? With a `enabled` key to ensure we want to enable sentinel or not. We could also reuse `redis.auth` instead of redefining `redis_sentinel.auth`

Chocobozzz: But they are mutually exclusive no?

Chocobozzz: Then what do you think of a `mode: 'sentinel' | 'redis'` inside the `redis` object? So it's clearer for the admin that it needs to provide sentinel auth/configuration while still keeping the exclusive thing inside the peertube code? We can imagine in the future having a 'cluster' mode too?

Chocobozzz:
Thanks!

PR #5596, opened 2023-02-12
Update docker.md

ruvilonix:
1. I don't know if you're supposed to do something other than enter `$EDITOR ./docker-compose.yml` into terminal, but when I did that it gave a permission error, so I just used nano. Same with `.env`.

2. Newer versions of Docker Compose use the command `docker compose`, not `docker-compose`. 

3. Grepping the password from logs was not working. I looked at the full logs, and I didn't see anything about a password. I added how to set a custom password. Maybe the grepping part should be removed or changed to make it work.

## Has this been tested?

### DIFF
--- a/support/doc/docker.md
+++ b/support/doc/docker.md
@@ -35,13 +35,13 @@ View the source of the file you're about to download: [.env](https://github.com/
 #### Tweak the `docker-compose.yml` file there according to your needs
 
 ```shell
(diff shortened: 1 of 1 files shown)


emansom:
`$EDITOR` is not always set on all distros.

Chocobozzz:
Thanks!

PR #5598, opened 2023-02-12
Install plugins upon init

### LINKED ISSUE #5595: Install plugins upon init

Currently PeerTube assumes that the plugins “installed” (i.e. added in the database) also exists on the disk. In our case we’re creating a new server upon each PT update, which means we have to sync all the npm plugins to a S3 bucket to be able to get them to our new servers.

Upon init, PT should generate a package.json file in the plugins folder and run yarn install. If the package.json file already exists nothing should be done, so this should not affect restart time.

kontrollanten:
## Description
Upon init PeerTube will check for storage/plugins/package.json, if it doesn't exist it will be created based on DB records and `yarn install` is ran.

## Related issues
closes #5595

## Has this been tested?

### DIFF
--- a/server/initializers/installer.ts
+++ b/server/initializers/installer.ts
@@ -1,4 +1,4 @@
-import { ensureDir, readdir, remove } from 'fs-extra'
+import { ensureDir, readdir, remove, stat, writeFile } from 'fs-extra'
 import passwordGenerator from 'password-generator'
 import { join } from 'path'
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Hello,

I'm sorry but we don't have enough resource to maintain this use case for now :/ See https://github.com/Chocobozzz/PeerTube/issues/5595#issuecomment-1427539909

Issue #1845, opened 2019-05-25
Allow uploading a new version of a video


12people:
I've been wanting to create videos summarizing certain topics that are continually updated as more news on the topic comes in. This is unusual for videos, as Youtube has never allowed such a thing, but common for blog posts.

I would really appreciate an "Upload new version" feature, which would replace the video in question with another video, maintaining the same link and the same comments. Older versions could be archived and available from a screen listing previous versions.

Krafting:
good idea!

dedesite:
+1. I've got this problem with Youtube also. For example, I uploaded a video, then share it with a link and then realize that I forgot something or that something was wrong in the "cut" etc. On youtube, you need to upload a new video and share the new link, which is a shame for "marketing".

Chocobozzz:
Video file updates are generally not allowed because there can be abuses: for example a viral cat video can be replaced with some kind of propaganda etc. We could however add the ability to specify a link to the new version of the video, so users are redirected with a confirm modal: *A new version of this video is available, would you like to watch it?* or something like that.

dedesite:
> Video file updates are generally not allowed because there can be abuses: for example a viral cat video can be replaced with some kind of propaganda etc.

Oh, I never though about that ! That's kind of a problem indeed.

> We could however add the ability to specify a link to the new version of the video, so users are redirected with a confirm modal: _A new version of this video is available, would you like to watch it?_ or something like that.

It seems to be a great idea :D

12people:
@Chocobozzz 

The downside I see with a redirect dialog is that, by submitting a new version, you suddenly lose all conversations under a video as well as all ratings. The uploader then has to make a choice whether correcting a simple fact in a video or adding new content as an update on a situation is worth losing all of that.

I agree that there is potential for abuse. The way that social networks have dealt with this is by adding an "Edited" date to the post. The way blog posts often deal with this is by prefacing content with details on what was changed and how.

I think that, if it was clearly stated under the video when the latest version was uploaded, and older versions of the video were still accessible from an "Older versions" page, that would go a long way toward curbing abuse

Along with that, instance admins could have a rule against uploading new versions that vastly differ from the original, and abuse of that rule could be treated the same way as rules against spam, porn, and offensive content. As older versions would be publicly accessible, anyone could report abuse.

therahedwig:
[...]

PR #5601, opened 2023-02-12
Upload new version of video

kontrollanten:
## Description

TODO:

- [x] PoC working video replacement in both client and server.
- [ ] ~Cleanup client code and share code between VideoReplaceComponent and VideoAddComponent.~
- [ ] Move client code to Studio.
- [ ] Pretty GUI.
- [x] Make import-video-file job create transcoding jobs for web torrents.
- [x] Make import-video-file job create transcoding jobs for HLS.
- [x] Fix broken tests.
- [x] Write new tests. 
- [ ] Don't remove files until new files are in place.

Questions:
- Should we handle "regular" uploads the same way (i.e. creating a video-file-import job instead of doing everything in the controller)?

## Related issues
#1845

## Has this been tested?

## Screenshots

image


vid-bin:
Is this feature going to be admin only? If not, is it a feature that can be enabled/disabled through through the admin panel?

I would not want this enabled for my end-users. An admin only feature would be fine, though. But yeah I have serious concerns with this feature being abused by regular users or hacked accounts.

FiskFan1999:
> Is this feature going to be admin only? If not, is it a feature that can be enabled/disabled through the admin panel?
> 
> I would not want this enabled for my end-users. An admin only feature would be fine, though. But yeah I have serious concerns with this feature being abused by regular users or hacked accounts.

Sending the video back to the auto-block list if the video is replaced is another possibility to help moderation, depending on if the user does not have the auto-block disabled.

EricGuic:
Hi everyone, I'm not a developper but in a user perspective, I've listed some (I hope usefull) informations on how this feature could work.
It's published in this post on Framacolibri : **https://framacolibri.org/t/working-on-a-plugin-to-update-a-published-video/16782/4**
I would be happy to bring you some help (no dev, but on general thoughts, UI, etc.).
Best regards.

Chocobozzz:
Thanks! We can close this PR as the feature has been developed in official PeerTube branch. See https://github.com/Chocobozzz/PeerTube/issues/1845#issuecomment-1759709781

Chocobozzz:
And @EricGuic, thank you for your feedback on this feature. We did not have time to implement what you wanted but don't hesitate to create a feature request so we can add them in the future :)

PR #5616, opened 2023-02-17
Hotfix/refactoring style

wickloww:
## Description

Refactoring class helpers, mixins, and style for the about, accounts, error page, home, login and manage components.

## Has this been tested?


client/src/app/+login/login.component.scss
```diff
@@ -15,39 +13,16 @@ input[type=email] {
...
-      width: 100%;
-    }
-  }
-
-  .modal-footer .grey-button {
-    display: none;
-  }
-}
-
+//garder ça ou deux flex-grow-1 flex-shrink-1 lignes 39 et 104
```
Chocobozzz, line 16: To remove

client/src/app/+login/login.component.scss
```diff
@@ -15,39 +13,16 @@ input[type=email] {
...
 .wrapper {
   display: flex;
   justify-content: space-between;
   flex-wrap: wrap;
   margin: auto;
 
-  > div {
+  >div {
```
Chocobozzz, line 23: We still want a space after `>`

client/src/app/+login/login.component.scss
```diff
@@ -56,46 +31,24 @@ input[type=email] {
 }
 
 .alert {
-  margin: 0 auto 30px;
+  @include rfs(0 auto 2rem, margin);
```
Chocobozzz, line 34: You don't use the `margin` peertube mixin?

client/src/app/+about/about-peertube/about-peertube.component.scss
```diff
@@ -6,63 +6,14 @@
...
-  margin: 50px 0;
-  justify-content: center;
-
-  .card {
-    margin: 30px;
-    flex-basis: 300px;
-  }
-}
+.card {
+  @include rfs(2rem, margin);
```
wickloww, line 10: Is it more appropriate to write `@include margin(2rem, 2rem, 2rem, 2rem)` so that we use the peertube mixin even if the code is repeated infamously ?

client/src/app/+login/login.component.scss
wickloww: My mistake, corrected.

client/src/app/+about/about-peertube/about-peertube.component.scss
Chocobozzz: No your version is better!

You can also update the mixin so it just use `@include rfs($block-start, margin)` if `$inline-end` and other arguments are not set (something like https://stackoverflow.com/a/25402481)

client/src/app/+login/login.component.scss
```diff
@@ -1,12 +1,10 @@
 @use '_variables' as *;
 @use '_mixins' as *;
 
-@import 'bootstrap/scss/functions';
-@import 'bootstrap/scss/variables';
-
 label {
+  @include font-size(1.125rem);
```
Chocobozzz, line 5: Please use a variable instead of `1.125rem` (like you did with `font-size-rem-14px`)

client/src/sass/class-helpers/_custom-bootstrap-helpers.scss
```diff
@@ -0,0 +1,8 @@
+@use '_variables' as *;
+@use '_mixins' as *;
+
+// Font sizes
+
+.fs-5-5 {
+  @include font-size(1.125rem);
```
Chocobozzz, line 7: Same here :)

PR #5665, opened 2023-03-02
Hotfix/filter subscription videos

wickloww:
## Description

Fixes filter on subscription videos and adds tests

## Has this been tested?

### DIFF
--- a/server/tests/api/videos/videos-common-filters.ts
+++ b/server/tests/api/videos/videos-common-filters.ts
@@ -76,10 +76,18 @@ describe('Test videos filter', function () {
 
   describe('Check deprecated videos filter', function () {
 
-    async function getVideosNames (server: PeerTubeServer, token: string, filter: string, expectedStatus = HttpStatusCode.OK_200) {
+    async function getVideosNames (
+      server: PeerTubeServer,
+      token: string,
+      filter: string,
+      skipSubscription = false,
+      expectedStatus = HttpStatusCode.OK_200
+    ) {
       const videosResults: Video[][] = []
 
       for (const path of paths) {
+        if (skipSubscription && path === subscriptionVideosPath) continue
+
         const res = await makeGetRequest({
           url: server.url,
           path,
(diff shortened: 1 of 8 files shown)


server/tests/api/videos/videos-common-filters.ts
```diff
@@ -162,17 +176,21 @@ describe('Test videos filter', function () {
...
+    async function getVideosNames (
+      options: {
+        server: PeerTubeServer
+        isLocal?: boolean
+        include?: VideoInclude
+        privacyOneOf?: VideoPrivacy[]
+        token?: string
+        expectedStatus?: HttpStatusCode
+      },
+      skipSubscription = false
```
Chocobozzz, line 188: Why not put `skipSubscription` in `options`?

server/tests/api/videos/videos-common-filters.ts
```diff
@@ -162,17 +176,21 @@ describe('Test videos filter', function () {
...
+    ) {
       const videosResults: string[][] = []
 
       for (const path of paths) {
+        if (skipSubscription && path === subscriptionVideosPath) continue
```
Chocobozzz, line 193: Please add a `\n` at the end of the line, to separate cases where we skip the loop with the loop content

server/tests/api/videos/videos-common-filters.ts
```diff
@@ -70,10 +76,18 @@ describe('Test videos filter', function () {
 
   describe('Check deprecated videos filter', function () {
 
-    async function getVideosNames (server: PeerTubeServer, token: string, filter: string, expectedStatus = HttpStatusCode.OK_200) {
+    async function getVideosNames (
```
Chocobozzz, line 79: Please convert all args as an object

When you reach 3 or 4 arguments it means it's time to switch to the object arg style

server/tests/api/videos/videos-common-filters.ts
wickloww: Correct in https://github.com/Chocobozzz/PeerTube/pull/5665/commits/cee2a3fc25413173a465eb2392681605261cff78.

Issue #5682, opened 2023-03-07
Using hotkeys regardless of keyboard layout language


kirill2400:
### Describe the problem to be solved

Such combinations as B or F do not work when I have the Russian language, they work in English, in YouTube the full screen works on F on any layout

### Describe the solution you would like

Make processing any keyboard layout

wickloww:
Hello, 
Peertube's hotkeys use `mousetrap`, which does not support non-Latin keyboard layouts.
As for the player hotkeys, it should now support all keyboard layouts with this pull request https://github.com/Chocobozzz/PeerTube/pull/5684.

Chocobozzz:
@kirill2400 Can you try with latest PeerTube version? We moved away from `mousetrap` in favour of [tinykeys](https://github.com/jamiebuilds/tinykeys)

kirill2400:
> @kirill2400 Can you try with latest PeerTube version? We moved away from `mousetrap` in favour of [tinykeys](https://github.com/jamiebuilds/tinykeys)

In Player now work! But in instance no 
![image](https://github.com/Chocobozzz/PeerTube/assets/30482699/42324441-4578-4606-9567-9e2955ffdd53)

ubf-hunter:
Looks like this was already fixed and merged in #5684 back in March 2023. Should this issue be closed?

PR #5684, opened 2023-03-08
Non latin keyboard layout support player shortcut

wickloww:
## Description

Add player hotkey support for non-Latin keyboard layouts

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/5682

## Has this been tested?

### DIFF
--- a/client/src/assets/player/shared/hotkeys/peertube-hotkeys-plugin.ts
+++ b/client/src/assets/player/shared/hotkeys/peertube-hotkeys-plugin.ts
@@ -79,7 +79,7 @@ class PeerTubeHotkeysPlugin extends Plugin {
       // Fullscreen
       {
         // f key or Ctrl + Enter
-        accept: e => this.isNaked(e, 'F') || (!e.altKey && e.ctrlKey && e.key === 'Enter'),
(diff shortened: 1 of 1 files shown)


client/src/assets/player/shared/hotkeys/peertube-hotkeys-plugin.ts
```diff
@@ -79,7 +79,7 @@ class PeerTubeHotkeysPlugin extends Plugin {
       // Fullscreen
       {
         // f key or Ctrl + Enter
-        accept: e => this.isNaked(e, 'f') || (!e.altKey && e.ctrlKey && e.key === 'Enter'),
+        accept: e => this.isNaked(e, 'F') || (!e.altKey && e.ctrlKey && e.key === 'Enter'),
```
Chocobozzz, line 82: I think `isNaked` should be in charge to do a `.toUpperCase()` of `f` so the caller doesn't have to think about casing

wickloww: Done in https://github.com/Chocobozzz/PeerTube/pull/5684/commits/a8a6504c3c85507b68e9aa91cf5b756e66084128

Issue #5113, opened 2022-07-08
Option to set initial privacy of a new live replay


Mike35538:
### Describe the problem to be solved

Some live streaming users like to save a replay and publish it later. Currently the live replay takes the privacy status of the original live video. So if the live video is public, the replay will also default to public. However users don't always want the live video replays to immediately become public after they are processed by the instance. Sometimes they will trim it in Studio first and then make it public. Currently users live stream and then wait for the replay to become available in their library then they can change the privacy to private. But if they leave for awhile because the instance is still processing the replay then when the processing is done the replay is automatically public for everyone until the user gets back later and makes it private.

### Describe the solution you would like

Add an option to set the initial privacy of the replay. You could add the option under the replay settings under live settings like in the mock up below.

![replay-privacy](https://user-images.githubusercontent.com/104943871/178069290-deaa1985-97e2-47c5-b900-c50876840823.jpg)

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/pull/5692 by @wickloww :sparkles:

PR #5692, opened 2023-03-14
Feature/Add replay privacy

wickloww:
## Description

Add the ability to set a different privacy level for live playback.

## Related issues

Fixes https://github.com/Chocobozzz/PeerTube/issues/5113

## Has this been tested?


server/lib/live/live-manager.ts
```diff
@@ -470,15 +472,30 @@ class LiveManager {
...
+      })
+      : null
 
-    return liveSession.save()
+    return sequelizeTypescript.transaction(async t => {
+      if (videoLive.saveReplay) {
+        await replaySettings.save({ transaction: t })
+      }
+
+      const liveSession = new VideoLiveSessionModel({
```
Chocobozzz, line 487: I think you can directly write `return VideoLiveSessionModel.create({ ...`)

server/models/video/video-live-replay-setting.ts
```diff
@@ -0,0 +1,42 @@
...
+    })
+  }
+
+  static removeSettings (id: number) {
+    return VideoLiveReplaySettingModel.destroy({
+      where: { id }
+    })
+  }
+
+  toFormattedJSON (): { privacy: VideoPrivacy } {
```
Chocobozzz, line 37: You don't need to type the result as it will be automatically inferred from your `return`

server/models/video/video-live.ts
```diff
@@ -66,6 +83,27 @@ export class VideoLiveModel extends Model<Partial<AttributesOnly<VideoLiveModel>
...
+  @BelongsTo(() => VideoLiveReplaySettingModel, {
+    foreignKey: {
+      allowNull: true
+    },
+    onDelete: 'set null'
+  })
+  ReplaySetting: VideoLiveReplaySettingModel
+
+  @BeforeDestroy
+  static deleteReplaySetting (instance: VideoLiveModel) {
```
Chocobozzz, line 99: Also add this hook in live session model, even if (for now) we never remove them

shared/server-commands/videos/live-command.ts
```diff
@@ -118,10 +118,11 @@ export class LiveCommand extends AbstractCommand {
 
   async quickCreate (options: OverrideCommandOptions & {
     saveReplay: boolean
+    replaySettings?: { privacy: VideoPrivacy }
```
Chocobozzz, line 121: I think you can remove this argument, and automatically set it to private since it's a "quick" create

server/types/models/video/video-live.ts
```diff
@@ -1,15 +1,22 @@
...
 
 // ############################################################################
 
 export type MVideoLiveVideo =
   MVideoLive &
   Use<'Video', MVideo>
+
+// ############################################################################
+
+export type MVideoLiveVideoFormattable =
```
Chocobozzz, line 20: Prefer `MVideoLiveVideoWithSetting` name

Chocobozzz (changes requested):
:sparkles:

PR #5703, opened 2023-03-22
object-storage-proxy: set content-range

kontrollanten:
## Description
When S3 returns a HTTP 206 response the content-range should be proxied as well.


Chocobozzz:
Thanks!

PR #5728, opened 2023-04-03
feature/ability to disable video history by default

wickloww:
## Description

Add the ability to disable video history by default

## Related issues

Fixes https://github.com/Chocobozzz/PeerTube/issues/1499

## Has this been tested?

## Screenshot

image


config/default.yaml
```diff
@@ -408,6 +408,9 @@ signup:
       blacklist: []
 
 user:
+  history:
+    videos:
+      enabled: true
```
Chocobozzz, line 413: Please add a comment explaining it's only the default behaviour for new users

Issue #5710, opened 2023-03-24
Images in Markdown or in HTML


Alterak:
### Describe the problem to be solved

In the Support section (for a channel or for a video) it is possible to add an image, for example like this:
`![альт](https://peervideo.ru/static/st-files/donate.png "Donate")`
Also in the NGINX configuration file I added a directory of static files "st-files"
```
  # Bypass PeerTube for performance reasons. Optional.
  location ~ ^/static/(thumbnails|avatars|st-files)/ {
```
This works fine, but only for the "Support" section

Now I want in the description of my server, in the Financing section of Pertube I want to also add an image (barcode). I may want to add an image in other sections of my server description, but I can't do that because there is no such image support there.

### Describe the solution you would like

A big request is to add support for images in all sections of the server where Markdown or HTML support is used. If this violates security, then you can add this item to the Peertube configuration so that the administrator decides what to choose (either Yes or No).

Jolg42:
I created my first PR on PeerTube for this here https://github.com/Chocobozzz/PeerTube/pull/5732

Alterak:
> I created my first PR on PeerTube for this here #5732

Thank you for this functionality!
Can you tell me how to control the image size?

Chocobozzz:
You can't, please create another issue. https://github.com/markdown-it/markdown-it/issues/80#issuecomment-1146587035 might be the way to handle this case

PR #5732, opened 2023-04-04
feat(about): render images in markdown

Jolg42:
Closes https://github.com/Chocobozzz/PeerTube/issues/5710

## Description

It looks to me that why it works in the "Support" section is because it's using `this.markdownService.enhancedMarkdownToHTML()` 
https://github.com/Chocobozzz/PeerTube/blob/00ee545c2444c35ed0d37df9f91984246aa36124/client/src/app/shared/shared-support-modal/support-modal.component.ts

which has different rules set in https://github.com/Chocobozzz/PeerTube/blob/00ee545c2444c35ed0d37df9f91984246aa36124/shared/core-utils/renderer/markdown.ts#L17 which allows `image`

So I changed the relevant line for the "About" page fields and now it works 🎊 

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/5710

## Has this been tested?

Do we need to run or have a test for this?

## Screenshots

Before
Screenshot 2023-04-04 at 15 39 15

After
Screenshot 2023-04-04 at 15 40 08

### DIFF
--- a/client/src/app/+admin/config/edit-custom-config/edit-instance-information.component.html
+++ b/client/src/app/+admin/config/edit-custom-config/edit-instance-information.component.html
@@ -126,6 +126,7 @@
 
           <my-markdown-textarea
             name="instanceTerms" formControlName="terms"
+            [customMarkdownRenderer]="getCustomMarkdownRenderer()" [debounceTime]="500"
             [formError]="formErrors['instance.terms']"
           ></my-markdown-textarea>
         </div>
@@ -135,6 +136,7 @@
 
           <my-markdown-textarea
             name="instanceCodeOfConduct" formControlName="codeOfConduct"
+            [customMarkdownRenderer]="getCustomMarkdownRenderer()" [debounceTime]="500"
             [formError]="formErrors['instance.codeOfConduct']"
           ></my-markdown-textarea>
         </div>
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Thanks! Can you also update the admin to use these enhanced rules? https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/%2Badmin/config/edit-custom-config/edit-instance-information.component.html#L127

Jolg42:
@Chocobozzz thanks for the tip, I just did it and now on the admin the preview works too 🙌🏼 
Screenshot 2023-04-05 at 10 54 48

client/src/app/+admin/config/edit-custom-config/edit-instance-information.component.html
```diff
@@ -126,6 +126,7 @@
 
           <my-markdown-textarea
             name="instanceTerms" formControlName="terms"
+            [customMarkdownRenderer]="getCustomMarkdownRenderer()" [debounceTime]="500"
```
Chocobozzz, line 129: Just use `[markdownType]="enhanced"` instead

Jolg42: ✅ Thank you for the tip again!

Chocobozzz:
Thanks @Jolg42!

Issue #5708, opened 2023-03-24
Database Name Environment


berksmbl:
### Describe the problem

By default the database name is "peertube" and it is not possible to change it with env variables, I can use the docker image only if I create a database as "peertube". However, I host my database as a service elsewhere, and the service provider does not give complete freedom to specify the database name.

### Additional information

* PeerTube version: production-bullseye (Docker)
* Redis version: 7
* PostgreSQL version: 14

Jolg42:
It looks to me that you could set the database name with `POSTGRES_DB=mydbname` from:
https://github.com/Chocobozzz/PeerTube/blob/develop/support/docker/production/.env#L5

If this is confirmed to work we could mention it in:
- https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/docker.md
- https://github.com/Chocobozzz/PeerTube/blob/master/support/docker/production/config/custom-environment-variables.yaml

berksmbl:
> It looks to me that you could set the database name with `POSTGRES_DB=mydbname` from: https://github.com/Chocobozzz/PeerTube/blob/develop/support/docker/production/.env#L5

Hello, This variable only interacts with PostgreSQL. It has no interaction with PeerTube and continues to use its default value.

Jolg42:
@berksmbl Indeed!
I opened a PR to make the appropriate change here: https://github.com/Chocobozzz/PeerTube/pull/5734

berksmbl:
@Chocobozzz The update still hasn't been released on Docker for production.

Chocobozzz:
It's released on docker in `develop-bullseye` image

berksmbl:
> It's released on docker in `develop-bullseye` image

Hello, I am preparing an installation scheme through the one-click install logic on EasyPanel for people to use. We only create templates for production and stable versions. Unfortunately, I cannot use latest or develop versions here. Can you give me an estimated time of completion?

berksmbl:
> It's released on docker in `develop-bullseye` image

Hello @Chocobozzz, I am preparing an installation scheme through the one-click install logic on EasyPanel for people to use. We only create templates for production and stable versions. Unfortunately, I cannot use latest or develop versions here. Can you give me an estimated time of completion?

Chocobozzz:
It will be in docker production image for the next release, ~2 months I think.

berksmbl:
> It will be in docker production image for the next release, ~2 months I think.

Hello @Chocobozzz,
[...]

PR #5734, opened 2023-04-05
feat(docker): make database name configurable

Jolg42:
## Description

I found that it should be configurable via `name` from the config file here: https://github.com/Chocobozzz/PeerTube/blob/9258e9a4a37911fc64b5faada2a7e604bd1ede1f/server/initializers/database.ts#LL59C36-L59C36

## Related issues

Closes https://github.com/Chocobozzz/PeerTube/issues/5708

## Has this been tested?

I tested following the instructions from https://docs.joinpeertube.org/install/docker

Before
```
peertube-docker-peertube-1   |       "message": "database \"custom\" does not exist",
peertube-docker-peertube-1   |       "length": 93,
peertube-docker-peertube-1   |       "name": "error",
peertube-docker-peertube-1   |       "severity": "FATAL",
peertube-docker-peertube-1   |       "code": "3D000",
peertube-docker-peertube-1   |       "file": "postinit.c",
peertube-docker-peertube-1   |       "line": "875",
peertube-docker-peertube-1   |       "routine": "InitPostgres"
peertube-docker-peertube-1   |     }
peertube-docker-peertube-1   |   }
peertube-docker-peertube-1   | }
peertube-docker-peertube-1 exited with code 255
```

After the change I also added the modified `custom-environment-variables.yaml` file to this path locally `docker-volume/config/custom-environment-variables.yaml`
The error message went away.

Not sure if there are tests for this?

### DIFF
--- a/support/docker/production/config/custom-environment-variables.yaml
+++ b/support/docker/production/config/custom-environment-variables.yaml
@@ -1,3 +1,8 @@
+# 
+# This file will be read by node-config
+# See https://github.com/node-config/node-config/wiki/Environment-Variables#custom-environment-variables
+# 
+
 webserver:
   hostname: "PEERTUBE_WEBSERVER_HOSTNAME"
   port:
@@ -19,7 +24,7 @@ database:
   port:
     __name: "PEERTUBE_DB_PORT"
     __format: "json"
-  dbname: "PEERTUBE_DB_NAME"
+  name: "PEERTUBE_DB_NAME"
   suffix: "PEERTUBE_DB_SUFFIX"
   username: "PEERTUBE_DB_USERNAME"
   password: "PEERTUBE_DB_PASSWORD"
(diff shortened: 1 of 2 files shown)


support/docker/production/config/custom-environment-variables.yaml
```diff
@@ -1,3 +1,8 @@
+# 
+# This file will be read by node-config
+# See https://github.com/node-config/node-config/wiki/Environment-Variables#custom-environment-variables
+# 
+
```
Jolg42, line 5: I discovered that this file was read by `node-config` and found that useful to add as a comment because it was not obvious to me.

Chocobozzz:
Thanks!
If `PEERTUBE_DB_NAME` is empty, peertube still uses `PEERTUBE_DB_SUFFIX`?

Jolg42:
@Chocobozzz It's my understanding from this line
https://github.com/Chocobozzz/PeerTube/blob/9258e9a4a37911fc64b5faada2a7e604bd1ede1f/server/initializers/config.ts#L27
The database name will be by default `peertube` + a suffix if defined, and if `name` is defined it takes over.

Chocobozzz:
Thanks!

berksmbl:
Is the update released in the docker image?
Because the problem still persists.

Issue #1985, opened 2019-07-24
"Other Videos" sidebar suggests you videos you already watched


ghost:
As of right now on PeerTube v1.3.1 the "Other Videos" sidebar suggests you videos you already watched - so you end up in a loop and you stop using the sidebar recommendations at some point.
What we can do to make sure people see great, non-duplicated content in the sidebar?

frankstrater:
Maybe we could extend the functionality from the [video history](https://docs.joinpeertube.org/#/use-video-history) to filter the list of sidebar recommendations, but then you might end up with zero recommendations after watching them all.

realityfabric:
If you do that (filter based on view history) you may want to add a user setting to toggle that filter on and off.

frankstrater:
I would prefer just randomized 5 recommendations picked from a metedata matched (or otherwise) set of at least 20, but that's because I don't like 'personalisation' / user tracking built in by design.

LoveIsGrief:
It looks like the recommendations are built by the client in a query

https://github.com/Chocobozzz/PeerTube/blob/db5814c0472d2b9e8f0ca6bdd1175ff1cd99ed7c/client/src/app/%2Bvideos/%2Bvideo-watch/recommendations/recent-videos-recommendation.service.ts#L56

The autoplay then uses the first video from that list (originally it was a random video, not sure why that was changed :thinking: )

https://github.com/Chocobozzz/PeerTube/blob/9e2e51dc0bee5dc065a61c927e787ae021e8aa73/client/src/app/%2Bvideos/%2Bvideo-watch/video-watch.component.ts#L325

In any case, either a new API endpoint is required to return better recommendations or the client will have to make multiple queries depending on the settings we choose to introduce. I believe an API endpoint would be better, then the user-settings can be used to forge the query and only a list of videos will be returned.

My preference would be : 
 - either have the option to filter out videos I've already watched
 - or use the current API [`searchVideos`](https://docs.joinpeertube.org/api-rest-reference.html#operation/searchVideos) to make multiple queries that are then merged, filtered (to remove the current video) and finally shuffled
   1. Make a bigger query (current size is 5 so maybe multiply the result size). This would maybe even allow filtering out watched videos.
   2. Query using the category of the current video
   3. Query using title of video while removing short words (e.g <3 or <4)
   4. (Optional) If nothing really comes up then make simple query for the newest videos

[...]

PR #5739, opened 2023-04-07
Feature/filter already watched videos

wickloww:
## Description

Add a filter to not list already watched videos for a user.

## Related issues

Fixes https://github.com/Chocobozzz/PeerTube/issues/1985 for logged users

## Has this been tested?

### DIFF
--- a/server/middlewares/validators/videos/videos.ts
+++ b/server/middlewares/validators/videos/videos.ts
@@ -489,6 +489,9 @@ const commonVideosFiltersValidator = [
   query('search')
     .optional()
     .custom(exists),
+  query('excludeAlreadyWatched')
+    .optional()
+    .isBoolean(),
 
   (req: express.Request, res: express.Response, next: express.NextFunction) => {
     if (areValidationErrors(req, res)) return
@@ -520,6 +523,15 @@ const commonVideosFiltersValidator = [
       }
     }
 
+    logger.error(!user)
+    logger.error(req.query.excludeAlreadyWatched)
+    if (!user && exists(req.query.excludeAlreadyWatched)) {
+      res.fail({
+        status: HttpStatusCode.BAD_REQUEST_400,
+        message: 'Cannot use excludeAlreadyWatched parameter for unlogged users'
+      })
+      return false
+    }
     return next()
   }
 ]
--- a/server/models/video/sql/video/videos-id-list-query-builder.ts
+++ b/server/models/video/sql/video/videos-id-list-query-builder.ts
@@ -78,6 +78,8 @@ export type BuildVideosListQueryOptions = {
 
   transaction?: Transaction
   logging?: boolean
+
+  excludeAlreadyWatched?: boolean
 }
 
 export class VideosIdListQueryBuilder extends AbstractRunQuery {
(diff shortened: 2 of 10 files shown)


server/middlewares/validators/videos/videos.ts
```diff
@@ -489,6 +489,9 @@ const commonVideosFiltersValidator = [
   query('search')
     .optional()
     .custom(exists),
+  query('excludeAlreadyWatched')
```
Chocobozzz, line 492: Add a sanitizer: `.customSanitizer(toBooleanOrNull)`

server/middlewares/validators/videos/videos.ts
```diff
@@ -520,6 +523,15 @@ const commonVideosFiltersValidator = [
       }
     }
 
+    logger.error(!user)
+    logger.error(req.query.excludeAlreadyWatched)
+    if (!user && exists(req.query.excludeAlreadyWatched)) {
+      res.fail({
+        status: HttpStatusCode.BAD_REQUEST_400,
+        message: 'Cannot use excludeAlreadyWatched parameter for unlogged users'
```
Chocobozzz, line 531: `for unlogged users` -> `when auth token is not provided`

server/models/video/sql/video/videos-id-list-query-builder.ts
```diff
@@ -260,6 +262,10 @@ export class VideosIdListQueryBuilder extends AbstractRunQuery {
       this.whereDurationMax(options.durationMax)
     }
 
+    if (options.excludeAlreadyWatched) {
```
Chocobozzz, line 265: Please throw an error if `options.user.id` is not defined (you can use `exists` function)

server/models/video/sql/video/videos-id-list-query-builder.ts
```diff
@@ -598,6 +604,18 @@ export class VideosIdListQueryBuilder extends AbstractRunQuery {
     this.replacements.durationMax = durationMax
   }
 
+  private whereExcludeAlreadyWatched (userId: number) {
+    this.and.push(
+      'NOT EXISTS (' +
+      '  SELECT *' +
+      '  FROM "userVideoHistory"' +
+      '  WHERE "video"."id" = "userVideoHistory"."videoId"' +
+      '  AND "userVideoHistory"."userId" = :excludeAlreadyWatched' +
```
Chocobozzz, line 613: `excludeAlreadyWatchedUserId`

server/models/video/sql/video/videos-id-list-query-builder.ts
```diff
@@ -598,6 +604,18 @@ export class VideosIdListQueryBuilder extends AbstractRunQuery {
     this.replacements.durationMax = durationMax
   }
 
+  private whereExcludeAlreadyWatched (userId: number) {
+    this.and.push(
+      'NOT EXISTS (' +
+      '  SELECT *' +
```
Chocobozzz, line 610: `SELECT 1` for consistency (not sure if it's more efficient in this context, but we always prefer to not select columns when we don't need to so PostgreSQL doesn't have to fetch more data from the disk)

PR #5746, opened 2023-04-10
Fix #5744: Missing plugins-global.css + missing async

### LINKED ISSUE #5744: Missing plugins-global.css file during Peertube startup

When Peertube starts, some requests are failing because of a missing plugins-global.css file.
Depending on the number of plugins, and the time they take to start, Peertube can be unavailable during several seconds/minutes.

Here is the error when trying to display Peertube home page:

```
peertube_1  | [localhost:9000] 2023-04-10 12:24:42.566 error: Cannot generate HTML page. {
peertube_1  |   "err": {
peertube_1  |     "stack": "Error: ENOENT: no such file or directory, open '/data/tmp/plugins-global.css'",
peertube_1  |     "message": "ENOENT: no such file or directory, open '/data/tmp/plugins-global.css'",
peertube_1  |     "errno": -2,
peertube_1  |     "code": "ENOENT",
peertube_1  |     "syscall": "open",
peertube_1  |     "path": "/data/tmp/plugins-global.css"
peertube_1  |   }
peertube_1  | }

```

I found the code that is responsible of this, and I will make a PR in a minute.

### Steps to reproduce

1. Install some plugins
2. Restart Peertube
3. Try to open peertube in a browser

_No response_

### Additional information

* PeerTube instance:
  * URL:
  * Version: 5.1.0
  * NodeJS version:
  * Ffmpeg version:

JohnXLivingston:
## Description
On startup, people trying to display the Peertube index page could face an error, because of a missing plugins-global.css file.
There was also a missing `async` attribute on a related function.

## Related issues

See https://github.com/Chocobozzz/PeerTube/issues/5744

## Has this been tested?

I tested it by building Peertube, and running it again.

### DIFF
--- a/server/lib/client-html.ts
+++ b/server/lib/client-html.ts
@@ -1,5 +1,5 @@
 import express from 'express'
-import { readFile, pathExists } from 'fs-extra'
+import { pathExists, readFile } from 'fs-extra'
 import { join } from 'path'
 import validator from 'validator'
 import { isTestOrDevInstance } from '@server/helpers/core-utils'
@@ -412,13 +412,16 @@ class ClientHtml {
       logger.info('Plugin Global CSS file is not available (generation may still be in progress), ignoring it.')
       return htmlStringPage
     }
-    let globalCSSContent
+
+    let globalCSSContent: Buffer
+
     try {
       globalCSSContent = await readFile(PLUGIN_GLOBAL_CSS_PATH)
     } catch (err) {
       logger.error('Error retrieving the Plugin Global CSS file, ignoring it.', { err })
       return htmlStringPage
     }
+
     if (globalCSSContent.byteLength === 0) return htmlStringPage
 
     const fileHash = sha256(globalCSSContent)
--- a/server/lib/client-html.ts
+++ b/server/lib/client-html.ts
@@ -1,5 +1,5 @@
 import express from 'express'
-import { readFile } from 'fs-extra'
+import { readFile, pathExists } from 'fs-extra'
 import { join } from 'path'
 import validator from 'validator'
 import { isTestOrDevInstance } from '@server/helpers/core-utils'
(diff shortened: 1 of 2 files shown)


server/lib/plugins/plugin-manager.ts
```diff
@@ -516,7 +516,7 @@ export class PluginManager implements ServerHook {
 
   // ###################### CSS ######################
 
-  private resetCSSGlobalFile () {
+  private async resetCSSGlobalFile () {
```
Chocobozzz, line 519: This method is not async

server/lib/client-html.ts
```diff
@@ -408,7 +408,13 @@ class ClientHtml {
   }
 
   private static async addAsyncPluginCSS (htmlStringPage: string) {
-    const globalCSSContent = await readFile(PLUGIN_GLOBAL_CSS_PATH)
+    let globalCSSContent
```
Chocobozzz, line 411: Prefer using `pathExists` from `fs-extra` to check if the file exists, and if not you can use your log message in `.info` mode

JohnXLivingston: Using try{}catch() can also detect permissions issues. I can add a `pathExists`, but I suggest to keep the try{}catch(). Would it be ok for you?

JohnXLivingston: From the node documentation:
https://nodejs.org/api/fs.html
«Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile(), or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.»

JohnXLivingston:
@Chocobozzz : just make a new commit taking into account your comments.
I let the try/catch (see my responses to your comments), but I also added the pathExists you requested, with a log.info message.

Chocobozzz:
Thanks!

PR #5765, opened 2023-04-20
Add sync logic for live stream `.m3u8` files in S3

mblomdahl:
## Description
- Add `m3u8-parser` to project dependencies
- Keep track of all live stream video segments that are currently synced to S3 within a live _`MuxingSession`_
- Block publishing of updated `.m3u8` video metadata for up to ~~3 seconds (100 ms x 30)~~ 10 seconds (250 ms x 40), pending video segment uploads to S3 object storage

## Related issues

Resolves #5745

## Has this been tested?

- [x] 👍 yes, tested via [the live stream](https://pt-dev.mabl.online/w/hN6c8crzhsTciZK9yDJhvb) mentioned in the commend below
- [ ] 💭 no, because this PR is a draft and still needs work
- [ ] 🙅 no, because this PR does not update server code
- [ ] 🙋 no, because I need help; ~~I don't have a real S3-enabled PeerTube server to run on at the moment~~

Issue #5745, opened 2023-04-10
Live Playback Broken on iOS Safari and Chrome


ar9708:
### Describe the current behavior

When opening (any) PeerTube live stream in Safari or Chrome on latest iOS, the playback breaks down within a minute or two.

I have been watching these 2 live streams non-stop on latest macOS in Safari and Chrome, on Ubuntu 22 in Firefox, and on Android in Samsung Internet and Chrome, and it works really great:
* https://swebbtube.se/w/aS2nB4FQqZGkCE92iGWth2
* https://swebbtube.se/w/4sAQGBsdkEC257RBi3zLQ2

It goes on and on, as long as playback is not interrupted for too long (see #5733). However, on latest iPhone iOS, the same live video playback consistently breaks down in a minute or two. This happens on Mobile Safari, Chrome, and Edge (WebKit browsers?). Video always start up normally, but comes to a halt after 10-15 seconds or longer, sometimes it can play even up to a few minutes.

Illustration:
<img width="1691" alt="Screenshot 2023-04-10 at 14 07 41" src="https://user-images.githubusercontent.com/123388211/230902439-7a0e58dd-fba0-41ba-8e64-c9c3441a2496.png">

### Steps to reproduce

1. Start a live video stream backed by S3/object storage (permanent/recurring live stream, no replay, no transcoding, and small latency mode)
2. View live video in any browser on Mac or Android (things look fine, can play 30 minutes and more without interruption)
3. Open same live video in Safari or Chrome on latest iPhone/iOS and watch until it breaks down within minutes

### Describe the expected behavior

I would expect live video playback to work the same on the major browsers and device types. 🙃 

**Update at 2023-04-10T22 CET with respect to [findings mentioned in comment 1502250885 below](https://github.com/Chocobozzz/PeerTube/issues/5745#issuecomment-1502250885):** I would want Mobile Safari/Chrome to apply the moderate retry logic suggested in #5733, i.e. some exponential back-off between retries, a reasonable ceiling on 403/404 retries, and a fallback strategy that checks the `m3u8` metadata so we're not grabbing for long gone video segments.

### Additional information

* PeerTube instance:
  * URL: https://swebbtube.se
  * Version: 5.1.x
  * NodeJS version: 16
  * Ffmpeg version: 4.2.x

* Browser name, version and platforms on which you could reproduce the bug: Mobile Safari and Google Chrome 112.x on iOS 16.4.1 (iPhone 14)
[...]

Issue #5518, opened 2023-01-09
Separate the S3 / block-storage configuration for Live & VOD


vincib:
### Describe the problem to be solved

We are using S3 to store the video of a bunch of peertube.
As of now, we have a low-bandwidth high-volume S3 cluster, since it's storing a lot of videos 

We'd like to have a **different** s3 cluster used for the lives, which will allow us to use a distinct high-bandwidth low-volume s3 cluster

having the ability to do that in production.conf would be great

### Describe the solution you would like

in the conf we have 

object_storage > streaming_playlists
object_storage > videos

you could add a 

object_storage > live 

for lives 

and add the possibility to set access_key_id, secret_access_key, endpoint and region into those groups
if those are not set, using the global info into "credentials" 

(also, you could migrate "endpoint" and "region" into credentials, since they are deeply linked ;) )

Chocobozzz:
You can disable live stream storage in object storage with https://github.com/Chocobozzz/PeerTube/commit/68d006b6fc16e3943eae6d46bf8183ec99fb64c5

I would like to re-architecture object storage config in V7 (as it will break the current configuration schema) to support multi object storage providers and assigning the provider the admin want for each object type.

PR #5779, opened 2023-04-29
Add new `live.use_object_storage` config to opt-out of S3 for streaming

ar9708:
## Description

I've been trying out the S3 sync improvements from #5765 on bug #5745, but at production scale I find myself running into issues with timing and S3 sync. It works a bit better than the baseline (which did not work at all on iPhone, consistently broke down after 10-15 seconds), but I can't make it play smoothly for long periods of time as stably as I would like.

These struggles got me thinking, "why are we doing the S3 thing really?" I don't understand the reason for publishing assets to S3 and then removing them again after 30 seconds. In my naïve perspective, it seem more simple and robust to server short-lived assets like this directly from the server and not add the extra complexity that #5765 seek to address.

This PR adds a `live.use_object_storage` flag that can be set to false to disable object storage for live streams. Here you can see it in action, up and streaming directly from disk: https://pt-dev.mabl.online/w/hN6c8crzhsTciZK9yDJhvb

What do you think about this, @Chocobozzz?

## Related issues

See #5518 and #5745

## Has this been tested?

- [x] 🙋 no, i have only tested on a local dev server

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -475,6 +475,9 @@ transcoding:
 live:
   enabled: false
 
+  # Publish live streams via S3 object storage
+  use_object_storage: true
+
   # Limit lives duration
   # -1 == unlimited
   max_duration: -1 # For example: '5 hours'
--- a/config/dev.yaml
+++ b/config/dev.yaml
@@ -74,7 +74,7 @@ signup:
 
 live:
   enabled: true
-
+  use_object_storage: false
   allow_replay: true
 
   transcoding:
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -485,6 +485,9 @@ transcoding:
 live:
   enabled: false
 
+  # Publish live streams via object storage/S3
+  use_object_storage: false
+
   # Limit lives duration
   # -1 == unlimited
   max_duration: -1 # For example: '5 hours'
--- a/server/controllers/api/config.ts
+++ b/server/controllers/api/config.ts
@@ -243,6 +243,7 @@ function customConfig (): CustomConfig {
     },
     live: {
       enabled: CONFIG.LIVE.ENABLED,
+      useObjectStorage: CONFIG.LIVE.USE_OBJECT_STORAGE,
       allowReplay: CONFIG.LIVE.ALLOW_REPLAY,
       latencySetting: {
         enabled: CONFIG.LIVE.LATENCY_SETTING.ENABLED
--- a/server/initializers/checker-after-init.ts
+++ b/server/initializers/checker-after-init.ts
@@ -241,6 +241,11 @@ function checkSearchConfig () {
 
 function checkLiveConfig () {
   if (CONFIG.LIVE.ENABLED === true) {
+
+    if (CONFIG.LIVE.USE_OBJECT_STORAGE === true && !CONFIG.OBJECT_STORAGE.ENABLED) {
+      throw new Error('Live streaming cannot use object storage if object storage is not enabled.')
+    }
+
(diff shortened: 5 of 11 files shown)


Chocobozzz:
> but I can't make it play smoothly for long periods of time as stably as I would like

Do you know what is the cause?

stevespaw:
We desperately need a fix to the s3 live issue. We cannot produce stable livestreams at all once we connect to s3. This described fix seems like it may work, but we would also have to connect the live streams to the CDN also. How would that work? Would that also have to be properly routed separately? This may also be a better path for treating the saved livestream as a new upload and doing all of the selected transcodes as expected.

Chocobozzz:
> We cannot produce stable livestreams at all once we connect to s3

Please create an issue with logs and permanent live stream so we can fix this

stevespaw:
> > We cannot produce stable livestreams at all once we connect to s3
> 
> Please create an issue with logs and permanent live stream so we can fix this

We are experiencing the issues in https://github.com/Chocobozzz/PeerTube/pull/5765

stevespaw:
@Chocobozzz I was hoping that 5.2 might address the "live issue".
My test instance is working great for normal VOD. _**S3 object storage + CDN**_. Even a "Post Live" Video record works perfect after the final transcode it moves to s3 just fine. 
The issue is during the live stream there is no stream to the player. 
If I set object storage to "false" and restart - The live works great, but does not get to s3 of course and is not connected to the CDN. How do we fix this? There does not seem to be any errors in the logs.

normen:
I had some issues with (Strato) S3 and lives at some point but they weren't consistent. Sometimes the live stream would play on the client end and sometimes it wouldn't. It was always consistent for one session though. Meaning when the stream wouldn't work it wouldn't work on any device with no amount of reloading, it would have to be restarted.

Otherwise the whole S3 thing is great - I can serve to thousands of people with pretty much zero strain on my server. Though the up- and downloading to/from S3 seems a bit CPU intense on the peertube process..

stevespaw:
> I had some issues with (Strato) S3 and lives at some point but they weren't consistent. Sometimes the live stream would play on the client end and sometimes it wouldn't. It was always consistent for one session though. Meaning when the stream wouldn't work it wouldn't work on any device with no amount of reloading, it would have to be restarted.
> 
> Otherwise the whole S3 thing is great - I can serve to thousands of people with pretty much zero strain on my server. Though the up- and downloading to/from S3 seems a bit CPU intense on the peertube process..

@normen Are you using a CDN also for Live? We are having issues with S3 and CDN...

normen:
> @normen Are you using a CDN also for Live? We are having issues with S3 and CDN...

Nope, I serve directly from S3 and in peertube 5.2 I didn't seem to have any issues so far.

stevespaw:
> 

Oh, OK we have major issues with "Live" and use of the S3/CDN. Are you saying that **_during_** live the M3U8 file is in s3 for you? We only seem to get it there after live for the replay.

normen:
Yes, with the latest versions of Peertube (since 5.1?) thats the case - lives are served directly from S3 to the viewers, which is a huge plus.

stevespaw:
Hmmm that does not seem to be the case for us. The live M3U8 only seems to live on the local until it is over and transcodes. I wonder what is different?

stevespaw:
> Hmmm that does not seem to be the case for us. The live M3U8 only seems to live on the local until it is over and transcodes. I wonder what is different?

Is there a particular YAML setting I may have missed for this?

Chocobozzz:

Issue #5812, opened 2023-05-29
Video removed after upload 


kontrollanten:
### Describe the current behavior

Sometimes when we upload a video and click on "Publish" we're getting the following error.

Sorry, but something went wrong
Video not found

The call that's failing is `PUT /api/v1/videos/{id}`. Then the video has disappeared.

### Steps to reproduce

With object storage enabled:

1. Upload a video with privacy Private
2. Unclick "Publish when transcoding is done"
3. Click on "Publish"

### Additional information

* PeerTube instance:
  * Version: 5.1.0

* Browser name, version and platforms on which you could reproduce the bug: Latest Brave

Not sure whether this is related:

```
{
	"err": "TypeError: Expected a string but received a null
    at assertString (/var/www/peertube/versions/peertube-v5.1.0/node_modules/validator/lib/util/assertString.js:17:11)
    at Object.isInt (/var/www/peertube/versions/peertube-v5.1.0/node_modules/validator/lib/isInt.js:16:29)
    at Function.<anonymous> (/var/www/peertube/versions/peertube-v5.1.0/dist/server/lib/client-html.js:44:38)
    at Generator.next (<anonymous>)
    at /var/www/peertube/versions/peertube-v5.1.0/node_modules/tslib/tslib.js:167:75
    at new Promise (<anonymous>)
    at Object.__awaiter (/var/www/peertube/versions/peertube-v5.1.0/node_modules/tslib/tslib.js:163:16)
    at Function.getWatchHTMLPage (/var/www/peertube/versions/peertube-v5.1.0/dist/server/lib/client-html.js:42:24)
    at /var/www/peertube/versions/peertube-v5.1.0/dist/server/controllers/client.js:101:53
    at Generator.next (<anonymous>)
    at /var/www/peertube/versions/peertube-v5.1.0/node_modules/tslib/tslib.js:167:75
    at new Promise (<anonymous>)
    at Object.__awaiter (/var/www/peertube/versions/peertube-v5.1.0/node_modules/tslib/tslib.js:163:16)
    at generateWatchHtmlPage (/var/www/peertube/versions/peertube-v5.1.0/dist/server/controllers/client.js:96:20)
    at /var/www/peertube/versions/peertube-v5.1.0/dist/server/middlewares/async.js:21:32
    at Layer.handle [as handle_request] (/var/www/peertube/versions/peertube-v5.1.0/node_modules/express/lib/router/layer.js:95:5)
[...]

PR #5840, opened 2023-06-18
Fix the cleanup after a failed upload

kukhariev:
## Description

Along with the failed uploaded file, the metadata file will also be removed from the disk and cache.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/5812#issuecomment-1596060166

## Has this been tested?


Chocobozzz:
Thanks!

Issue #4226, opened 2021-07-02
Support 'authenticated fetch' for compatibility with 'limited federation' Mastodon instances. 


rscmbbng:
Dear devs,

**Describe the problem to be solved**

Mastodon supports a 'limited federation' mode which allows one to set up instances which federate only with those on allow lists. To enforce that model it requires federating servers to do an authenticated fetch. See [the Mastodon documentation](https://docs.joinmastodon.org/admin/config/) for more context. 

Currently it possible to add a peertube instance to the allow list of such Mastodon server running in 'secure mode'. Accounts on peertube can be followed and announcements of new videos reach the Mastodon instance. However, any comments made on those videos do not reach the Peertube instance.

While in that model it makes sense that these comments, which normally appear publicly, would not show up. They could and should show up to logged in peertube users on that instance.    

**Describe the solution you would like:**

Peertube supporting authenticated fetch to display interactions from Mastodon instances running in 'secure mode' which have allowlisted said Peertube instance. These interactions would only be visible to users on the peertube instance which are currently logged in.

johanfleury:
I guess it’s related, logs on my instance are full of messages like this one:

```
[peertube.br0.fr:443] 2021-07-06 15:07:23.599 warn: Cannot fetch remote actor count https://[SNIPPED]/users/[SNIPPED]/followers. {
  "err": {
    "stack": "HTTPError: Response code 403 (Forbidden)\n    at Request.<anonymous> (/app/node_modules/got/dist/source/as-promise/index.js:117:42)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejecti
ons (internal/process/task_queues.js:97:5)",
    "message": "Response code 403 (Forbidden)",
    "name": "HTTPError",
    "responseBody": {
      "error": "Forbidden"
    },
    "statusCode": 403
  }
}
```

tribela:
Got same error. Is there any plan to support signed fetch?

PR #5898, opened 2023-07-21
Add an option to sign federated fetches for mastodon compatibility

irafire:
## Description

Add a minimal set of changes to have remote comments, follows, and likes working with a mastodon instance with AUTHORIZED_FETCH enabled, guarded by a config flag.

Considering this a draft due to missing test coverage and only light manual testing, but uploading in case somebody is in a hurry to try to apply this to their setup. It is also possible that some federated functionality I have not considered is still failing.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/4226

## Has this been tested?

- [x] partially, by the existing test suite and a manual deployment 

The existing suite passed for me with flag enabled in test config (and caught an initial mistake, so some of the affected paths are exercised). I've also deployed it in a docker container built on top of v5.2.0 and it seems to pass initial manual testing i.e. comments, likes, and follow requests work with a secure fetch mastodon instance. I am not sure what tests to update/create (also I don't have Node experience), so the functionality is put behind a config flag. Marking this as need help in case contributing required testing coverage turns out to be easy for me to follow up on.

### DIFF
--- a/client/src/assets/player/shared/bezels/bezels-plugin.ts
+++ b/client/src/assets/player/shared/bezels/bezels-plugin.ts
@@ -12,7 +12,10 @@ class BezelsPlugin extends Plugin {
       player.addClass('vjs-bezels')
     })
 
-    player.addChild(new PauseBezel(player, options))
+    const component = new PauseBezel(player, options)
+    player.addChild(component)
+
+    this.on('dispose', () => player.removeChild(component))
   }
 }
 
--- a/client/src/assets/player/shared/peertube/peertube-plugin.ts
+++ b/client/src/assets/player/shared/peertube/peertube-plugin.ts
@@ -36,6 +36,8 @@ class PeerTubePlugin extends Plugin {
   private mouseInControlBar = false
   private mouseInSettings = false
 
+  private errorModal: videojs.ModalDialog
+
   private videoViewOnPlayHandler: (...args: any[]) => void
   private videoViewOnSeekedHandler: (...args: any[]) => void
   private videoViewOnEndedHandler: (...args: any[]) => void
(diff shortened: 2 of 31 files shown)


dannekrose:
Thank you for the PR!  I really hope that it can get integrated since Secure/Authorized Fetch is becoming the standard more and more.

Chocobozzz:
Do you think it's fine that we display publicly information we got via a signed fetch?

dannekrose:
> Do you think it's fine that we display publicly information we got via a signed fetch?

I would say for anonymous users, no.  But for logged in users yes because those users would share the same domain name/instance as the content that the remote account has allowed to retrieve the content.  That content also would not be federated out to following instances and only stay local to the PeerTube instance (how it works on the Microblogging platforms).

That adds a lot more complexity for sure, but I think matches the intent of Signed Fetch's idea of security and privacy. 


In addition to this, I understand the model is a bit different being a video/media sharing platform, but adding the ability to also require signed fetch for PeerTube content as an option for admins would be very useful.  It does potentially make the content less discoverable since it requires authentication to view, but for some admins, that allows them to better control their content, especially content that is NSFW or requires much stricter access controls without necessarily needing to have all viewers of the content create accounts on that same instance to view.  (i.e. restrict to Signed Fetch, and then require approval to follow accounts, or channels on the instance).

Chocobozzz:
Thanks! I refactored a little bit this PR.

I think it's okay to display publicly information we fetched using signed fetch if they're public. And PeerTube respects the remote object privacy anyway.

I decided to enable this feature by default, as I see many 403 from Mastodon. We'll see if it leads to performance issues.

PR #5910, opened 2023-07-28
Switch from debian bullseye to bookworm

uda:
## Description

Switch docker base from bullseye to bookworm

## Related issues

My motivation was to get ffmpeg >= 5.1 which is available in bookworm while bullseye still has 4.3.x

## Has this been tested?


Chocobozzz:
Thanks!

Issue #5915, opened 2023-08-01
Wide screen content (21:9) has black bars on both horizontal and vertical axis


SethFalco:
### Describe the current behavior

PeerTube supports content as a variety of aspect ratios, and works flawlessly for aspect ratios of 1.78 (16:9) or below. However, content that has a higher aspect such as 2.33 (21:9) gets black bars on both axis.

Note that the issue is with the player styles, and not the video itself.

![](https://github.com/Chocobozzz/PeerTube/assets/22801583/af0d976b-8810-4923-a5af-7e84df6cb2b9)
> A 21:9 video uploaded on a local PeerTube instance, and being viewed on a 21:9 display. Observe that there are both horizontal and veritcal black bars.

### Steps to reproduce

1. Publish a video at an aspect ratio higher than 1.78, for example, a 21:9 video. (Sample file provided.)
2. View the published video on PeerTube.

https://github.com/Chocobozzz/PeerTube/assets/22801583/b1870340-5d61-4aa4-87e1-1d1b75558bac

### Describe the expected behavior

There should always only be one set of black bars, either horizontal or vertical, to best make use of the space available.

### Additional information

* PeerTube instance:
  * URL: N/A
  * Version: `04d1da5` (`develop` at time of posting this issue)
  * NodeJS version: v18.16.0
  * Ffmpeg version: 5.1.3-1

* Browser name, version and platforms on which you could reproduce the bug: Firefox 115, Debian 12
* Link to browser console log if relevant: N/A
* Link to server log if relevant (`journalctl` or `/var/www/peertube/storage/logs/`): N/A

vid-bin:
Can we also have a third option for a 1:1 of the video being played?

It's size dependent but when on a 4k monitor I'd like to watch 1080p without stretching the video. This is especially true for 480p/720p.

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/commit/8e4fba97b26090e0c77ee9591058cd34ef9d2f55

Feedback are welcomed :)

SethFalco:
LGTM! Thanks for resolving this. :+1:

SethFalco:
@Chocobozzz 

Hmm… actually something is wrong here.
I tested this earlier and I thought it was fine, but it could be I wasn't running the right commit. :thinking: 

I've tested this again with cb909ab38a6a176edce53043bcfcec8ca2f156fb and this is _not_ resolved. 

[...]

PR #5916, opened 2023-08-01
fix: use full width of page for player

SethFalco:
## Description

Could we consider dropping the `max-width` on the player? This impedes on videos that have an aspect ratio higher than 1.78 (16:9). For example, 21:9 content gets black bars on both the top and bottom.

PeerTube performs very well with a variety of content as different aspect ratios. The primary side effect of this change is that it allows the player's controls to span the full width, but this is more inline with what other sites do anyway.

Observe 21:9 content on other sites:

### YouTube

> A 21:9 video on YouTube in theatre mode. There is only a single set of black bars, and the player controls span the full width of the player.

### Invidious

> The same 21:9 video on Invidious. Similar results to above.

## Related issues

* Closes https://github.com/Chocobozzz/PeerTube/issues/5915

## Has this been tested?

### After

> Observe that the content is wider, and the player controls fill the player width. This is inline with what other sites do, like YouTube and Invidious. (Demos included in the original issue.)

I can only confirm that I and one other PeerTube user prefer the change. However, it may warrant broader discussion or the input of a designer. If you dislike this, please let me know and I'd be happy to ask one to take a look.

### DIFF
--- a/client/src/app/+videos/+video-watch/video-watch.component.scss
+++ b/client/src/app/+videos/+video-watch/video-watch.component.scss
@@ -67,7 +67,6 @@ $video-height: 66vh;
 
   ::ng-deep .video-js {
     width: 100%;
-    max-width: getPlayerWidth($video-height);
     height: $video-height;
 
     // VideoJS create an inner video player


Chocobozzz:
Hi,

Thanks but the fix is a little bit more complex as we want to keep the size of the control bar the same as the video. 

I implemented that behaviour in https://github.com/Chocobozzz/PeerTube/commit/8e4fba97b26090e0c77ee9591058cd34ef9d2f55 where the javascript player automatically adapt the player size depending on the video ratio.

Issue #5911, opened 2023-07-29
Cannot check ffmpeg version


vid-bin:
### Describe the current behavior

This issue has existed for a long time.

Please add the ability to suppress this error.

### Steps to reproduce

1. Run peertube with a git build of ffmpeg
2. Get error:
error[7/29/2023, 3:44:26 PM] Cannot check ffmpeg version 
stack": "Error: Could not find ffmpeg version in ffmpeg version N-111626-g0ba719f726

### Describe the expected behavior

Works without error

### Additional information

* PeerTube instance:
  * Version: 5.2.0...89aa33311
  * Ffmpeg version: git

PR #5917, opened 2023-08-02
fix: handle git installation of ffmpeg

SethFalco:
## Description

Resolves the error that occurred when using an unconventional distribution of FFmpeg with a non-standard version string. For example, compiled from source, or a third-party fork.

* `#getFFmpegVersion` returns the full version string of the FFmpeg now, regardless of format.
* `#parseSemVersion` returns `null` if the string is not a semantic version string.
* `#parseSemVersion` handles normalization, such as if the patch segment is missing.
* If the FFmpeg version does not follow semver (so `#parseSemVersion` returns `null`), then we log a warning that we can't check compatibility. 

## Related issues

* Fixes https://github.com/Chocobozzz/PeerTube/issues/5911

## Has this been tested?

### DIFF
--- a/packages/ffmpeg/src/ffmpeg-version.ts
+++ b/packages/ffmpeg/src/ffmpeg-version.ts
@@ -1,6 +1,9 @@
 import { exec } from 'child_process'
 import ffmpeg from 'fluent-ffmpeg'
 
+/**
+ * @returns FFmpeg version string. Usually a semver string, but may vary when depending on installation method.
+ */
 export function getFFmpegVersion () {
   return new Promise<string>((res, rej) => {
     (ffmpeg() as any)._getFfmpegPath((err, ffmpegPath) => {
@@ -10,14 +13,10 @@ export function getFFmpegVersion () {
       return exec(`${ffmpegPath} -version`, (err, stdout) => {
         if (err) return rej(err)
 
-        const parsed = stdout.match(/ffmpeg version .?(\d+\.\d+(\.\d+)?)/)
-        if (!parsed?.[1]) return rej(new Error(`Could not find ffmpeg version in ${stdout}`))
+        const parsed = stdout.match(/(?<=ffmpeg version )[a-zA-Z\d.-]+/)
+        if (!parsed) return rej(new Error(`Could not find ffmpeg version in ${stdout}`))
 
-        // Fix ffmpeg version that does not include patch version (4.4 for example)
-        let version = parsed[1]
-        if (version.match(/^\d+\.\d+$/)) {
-          version += '.0'
-        }
+        res(parsed[0])
       })
     })
   })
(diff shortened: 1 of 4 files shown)


shared/ffmpeg/ffmpeg-version.ts
```diff
@@ -10,14 +13,10 @@ export function getFFmpegVersion () {
       return exec(`${ffmpegPath} -version`, (err, stdout) => {
         if (err) return rej(err)
 
-        const parsed = stdout.match(/ffmpeg version .?(\d+\.\d+(\.\d+)?)/)
-        if (!parsed?.[1]) return rej(new Error(`Could not find ffmpeg version in ${stdout}`))
+        const parsed = stdout.match(/(?<=ffmpeg version )[a-zA-Z\d.-]+/)
+        if (!parsed) return rej(new Error(`Could not find ffmpeg version in ${stdout}`))
 
-        // Fix ffmpeg version that does not include patch version (4.4 for example)
-        let version = parsed[1]
```
SethFalco, line 17: Just a heads-up, I don't think this check was ever working properly.
The promise here wasn't being resolved, so the caller was probably just `await`ing forever.

server/tests/helpers/core-utils.ts
```diff
@@ -148,3 +148,54 @@ describe('Bitrate', function () {
...
+  it('Should parse FFmpeg version from GitHub release', function () {
+    const actual = parseSemVersion('5.1.3')
+
+    expect(actual.major).to.equal(5)
+    expect(actual.minor).to.equal(1)
+    expect(actual.patch).to.equal(3)
+  })
+
+  it('Should parse FFmpeg version from GitHub dev release', function () {
+    const actual = parseSemVersion('5.1.git')
```
SethFalco, line 187: Just a small detail, I actually compiled and installed [n6.1-dev](https://github.com/FFmpeg/FFmpeg/releases/tag/n6.1-dev), which showed this version string.
I'm guessing it was just a typo. :thinking:

Goes to show version strings aren't too reliable either. ^-^'

SethFalco:
Hmm, seems to me the failing test isn't related to this PR.
Even locally, this test fails on a build from a clean `develop` branch.

```
    Videos search
      ✔ Should make a simple search and not have results (832ms)
      ✔ Should make a simple search and have results (581ms)
      1) Should make a simple search
Saving logs peertube-4840.log.


  6 passing (20s)
  1 failing

  1) Test index search
       Videos search
         Should make a simple search:

      AssertionError: expected 2 to equal 1
      + expected - actual

      -2
      +1
      
      at /home/runner/work/PeerTube/PeerTube/dist/server/tests/api/search/search-index.js:82:55
      at fulfilled (node_modules/tslib/tslib.js:164:62)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

Chocobozzz:
Thanks!

PR #5949, opened 2023-08-24
Corrected storage tmp location

criadoperez:
## Description

Reference to `Add storage.tmp_persistent directory in configuration file` has the wrong line.

Currently it's pointing to line 128 which is empty.
Instead it should be 148 which is:

```
  tmp_persistent: '/var/www/peertube/storage/tmp-persistent/' # As tmp but the directory is not cleaned up between PeerTube restarts
```


Chocobozzz:
Thanks!

PR #5960, opened 2023-08-31
Fix path to Peertube installation in upgrade.sh

taziden:
## Description

During latest upgrade to 5.2.1, I had a message about /var/www/peertube not existing which is normal, because I'm using a different directory for Peertube.

## Has this been tested?


Chocobozzz:
Thanks!

Issue #5962, opened 2023-09-03
Allow for both Portuguese Portugal and Portuguese Brazil for captions


tio-trom:
### Describe the problem to be solved

The two can be very different but right now we can only add a single "Portuguese" caption. 

### Describe the solution you would like

Ability to select Portuguese Portugal and Portuguese Brazil.

PR #5971, opened 2023-09-14
Add brazilian option for video/captions languages

wickloww:
## Description

Add brazilian option for video/captions languages

## Related issues

Fixes https://github.com/Chocobozzz/PeerTube/issues/5962

## Has this been tested?

### DIFF
--- a/server/server/initializers/constants.ts
+++ b/server/server/initializers/constants.ts
@@ -1398,8 +1398,8 @@ async function buildLanguages () {
   languages['tok'] = 'Toki Pona'
 
   // Override Portuguese label
-  languages['pt'] = 'Portuguese (Portugal)'
-  languages['pt-BR'] = 'Portuguese (Brazilian)'
+  languages['pt'] = 'Portuguese (Brazilian)'
+  languages['pt-PT'] = 'Portuguese (Portugal)'
 
   // Chinese languages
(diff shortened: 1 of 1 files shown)


server/server/initializers/constants.ts
```diff
@@ -1397,6 +1397,10 @@ async function buildLanguages () {
   languages['el'] = 'Greek'
   languages['tok'] = 'Toki Pona'
 
+  // Override Portuguese label
+  languages['pt'] = 'Portuguese (Portugal)'
+  languages['pt-BR'] = 'Portuguese (Brazilian)'
```
Chocobozzz, line 1402: `pt-BR` is in fact the default locale region for `pt`

> Portuguese (Brazil) [pt_BR] is the default content locale for Portuguese [pt]

https://cldr.unicode.org/translation/translation-guide-general/default-content

petitlapin: So I guess we should do something like:
```
  // Override Portuguese label
  // Portuguese (Brazil) [pt_BR] is the default content locale for Portuguese [pt]
  // https://cldr.unicode.org/translation/translation-guide-general/default-content
  languages['pt'] = 'Portuguese (Brazilian)'
  languages['pt-PT'] = 'Portuguese (Portugal)'
```

PR #5975, opened 2023-09-17
Add Valencian option for video/captions languages

petitlapin:
## Description

Add Valencian option for video/captions languages

In KDE community, we started to create subtitles for our videos and upload them automatically to our instance (https://invent.kde.org/websites/video-subtitles). In our community, we have both Brazilian Portuguese translators and Valencian translators.
Brazilian translation is being dealt with in https://github.com/Chocobozzz/PeerTube/pull/5971, but it would be great to also add Valencian.

## Has this been tested?

- [x ] 🙅 no, because this PR does not update server code

### DIFF
--- a/server/server/initializers/constants.ts
+++ b/server/server/initializers/constants.ts
@@ -1401,6 +1401,9 @@ async function buildLanguages () {
   languages['zh-Hans'] = 'Simplified Chinese'
   languages['zh-Hant'] = 'Traditional Chinese'
 
+  // Catalan languages
+  languages['ca-valencia'] = 'Valencian'
+
   return languages
 }
 


petitlapin:
if the test issues are due to the `@` in the locale, Valencian do not have a ISO 639-3 code, we can switch to `ca-valencia`

server/server/initializers/constants.ts
```diff
@@ -1401,6 +1401,10 @@ async function buildLanguages () {
   languages['zh-Hans'] = 'Simplified Chinese'
   languages['zh-Hant'] = 'Traditional Chinese'
 
+  // Catalan languages
+  languages['ca'] = 'Catalan'
```
Chocobozzz, line 1405: I don't think you need this since the `languages` object already has this value :)

Chocobozzz:
Thanks!

Issue #5941, opened 2023-08-18
Change button label from Support to Donate


kontrollanten:
### Describe the problem to be solved

Currently there's a Support button below each video. It isn't really clear what the button means, even though I'm pretty sure each instance use it as a "donate" button.

### Describe the solution you would like

Change the label to "Donate" which is more commonly used and established term.

Isoumyajit:
I can fix this issue, Please assign me this issue @kontrollanten, I have the PR ready.
Thank you

PR #5992, opened 2023-10-04
Issue with ID #5941 is fixed

Isoumyajit:
## Description
As per the open Issue [#5941 ](https://github.com/Chocobozzz/PeerTube/issues/5941) I am giving this PR I changed the Label for the mentioned button.

## Has this been tested?

## Screenshots
It's a very minute change so I didn't add any reference now but, I can add the screenshot later if there is any requirement.


Chocobozzz:
Hello,

We don't want to change the `Support` term to another one that may not be adapted (some uploaders don't accept donations for example. As explained in https://github.com/Chocobozzz/PeerTube/issues/5941#issuecomment-1695830015 we prefer an uploader choice/admin choice.

Issue #2686, opened 2020-04-27
make audit log configurabe


whohoho:
Hi,

I am trying to disable the audit log, but i can't find the option in the config file, it would be nice to be able to turn that off.

constraintAutomaton:
I would like to tackle it if it's still of actuality

Ayan-web:
Is anyone working on this Issue? this will be my first PR
please tell me where to start.
thank you

MSK1582:
Is anyone working on this? If not, I can do this and submit a PR.

Jolg42:
Note: It looks like the related file would be https://github.com/Chocobozzz/PeerTube/blob/develop/server/helpers/audit-logger.ts

Archita-kale:
Hey, I would like to work on this issue

PR #6004, opened 2023-10-12
Add ability to disable audit logs

wickloww:
## Description

Add ability to disable audit logs in admin panel

## Related issues

Fixes https://github.com/Chocobozzz/PeerTube/issues/2686

## Has this been tested?

## Screenshots

image

---
image

### DIFF
--- a/client/src/app/+admin/config/edit-custom-config/edit-advanced-configuration.component.html
+++ b/client/src/app/+admin/config/edit-custom-config/edit-advanced-configuration.component.html
@@ -136,4 +136,27 @@
     </div>
   </div>
 
+  <div class="row mt-4"> <!-- cache grid -->
+    <div class="col-12 col-lg-4 col-xl-3">
+      <div class="anchor" id="customizations"></div> <!-- customizations anchor -->
+      <h2 i18n class="inner-form-title">LOGS</h2>
+      <div i18n class="inner-form-description">
+        Slight modifications to your PeerTube instance for when creating a plugin or theme is overkill.
+      </div>
+    </div>
+
+    <div class="col-12 col-lg-8 col-xl-9">
+      <ng-container formGroupName="instance">
+        <ng-container formGroupName="logs">
+          <ng-container formGroupName="auditLogs">
+            <div class="form-group">
+              <my-peertube-checkbox inputName="auditLogsEnabled" formControlName="enabled" i18n-labelText
+                labelText="Enable audit logs"></my-peertube-checkbox>
+            </div>
+          </ng-container>
+        </ng-container>
+      </ng-container>
+    </div>
+  </div>
+
 </ng-container>
--- a/client/src/app/+admin/system/logs/logs.component.html
+++ b/client/src/app/+admin/system/logs/logs.component.html
@@ -37,7 +37,9 @@
   <div *ngIf="loading" i18n>Loading...</div>
 
   <div #logsElement>
-    <div *ngIf="!loading && logs.length === 0" i18n>No log.</div>
+    <div *ngIf="isAuditLog && !isAuditLogsEnabled" i18n>Audit logs disabled.</div>
+
+    <div *ngIf="!loading && logs.length === 0 && isAuditLogsEnabled" i18n>No log.</div>
 
     <div *ngFor="let log of logs" class="log-row" [ngClass]="{ error: log.level === 'error', warn: log.level === 'warn' }">
       <span class="log-level">{{ log.level }}</span>
--- a/client/src/app/+admin/system/logs/logs.component.html
+++ b/client/src/app/+admin/system/logs/logs.component.html
@@ -6,6 +6,7 @@
   </div>
 
   <ng-select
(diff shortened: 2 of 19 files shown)


config/default.yaml
```diff
@@ -763,6 +763,9 @@ instance:
   securitytxt: |
     Contact: https://github.com/Chocobozzz/PeerTube/blob/develop/SECURITY.md
     Expires: 2025-12-31T11:00:00.000Z'
+  logs:
```
Chocobozzz, line 766: You already have a log section: https://github.com/Chocobozzz/PeerTube/blob/develop/config/default.yaml#L230

Please also add this block to `production.yaml.example`

client/src/app/+admin/config/edit-custom-config/edit-advanced-configuration.component.html
```diff
@@ -136,4 +136,27 @@ <h2 i18n class="inner-form-title">CUSTOMIZATIONS</h2>
     </div>
   </div>
 
+  <div class="row mt-4"> <!-- cache grid -->
```
Chocobozzz, line 139: You can remove HTML comments

client/src/app/+admin/config/edit-custom-config/edit-advanced-configuration.component.html
```diff
@@ -136,4 +136,27 @@ <h2 i18n class="inner-form-title">CUSTOMIZATIONS</h2>
...
+    <div class="col-12 col-lg-4 col-xl-3">
+      <div class="anchor" id="customizations"></div> <!-- customizations anchor -->
+      <h2 i18n class="inner-form-title">LOGS</h2>
+      <div i18n class="inner-form-description">
```
Chocobozzz, line 143: The description of this block is invalid

client/src/app/+admin/system/logs/logs.component.html
```diff
@@ -30,14 +31,16 @@
...
-  <my-button i18n-label label="Refresh" icon="refresh" (click)="refresh()"></my-button>
+  <my-button *ngIf="!isAuditLog() || isAuditLogsEnabled" i18n-label label="Refresh" icon="refresh" (click)="refresh()"></my-button>
 </div>
 
 <div class="logs">
   <div *ngIf="loading" i18n>Loading...</div>
 
   <div #logsElement>
-    <div *ngIf="!loading && logs.length === 0" i18n>No log.</div>
+    <div *ngIf="isAuditLog && !isAuditLogsEnabled" i18n>Audit logs disabled.</div>
```
Chocobozzz, line 41: `isAuditLog` is a function so `isAuditLog()`

Also transform `isAuditLogsEnabled` to a function. `is..` is for functions, if you still want to have a variable then prefer `auditLogsEnabled`. But it's better to have a function to not mix variables/functions if we can avoid

client/src/app/+admin/system/logs/logs.component.html
```diff
@@ -30,14 +31,16 @@
...
+
+    <div *ngIf="!loading && logs.length === 0 && isAuditLogsEnabled" i18n>No log.</div>
```
Chocobozzz, line 43: If audit logs is disabled but you're on the normal logs, you'll never see this message

I suggest to simplify lines 41/43 using an Angular else for example (not easy to use, but will be simpler in the future with the new Angular markup language: https://blog.angular.io/meet-angulars-new-control-flow-a02c6eee7843).

packages/tests/src/api/server/logs.ts
```diff
@@ -193,6 +193,41 @@ describe('Test logs', function () {
...
+              auditLogs:{
+                enabled: false
+              }
+            }
+          }
+        }
+      })
+
+      await server.videos.upload({ attributes: { name: 'video 12' } })
+      await waitJobs([ server ])
```
Chocobozzz, line 213: Just upload one video to test audit logs. Test 403 if disabled, but also the audit log was not filled when it was disabled (so re-enable the feature and check the audit log doesn't contain your upload)

Chocobozzz (changes requested):

ThisIsMissEm:
You may not want to actually have this as a feature, as it may violate some of the requirements for the Digital Services Act (e.g., to be able to provide evidence as to what an admin/moderator did and when.

The only case that I could see this being argued for is on a single-user instance, but even then audit logs are useful in case you forget doing something.

Chocobozzz:
Audit logs would be enabled by default. But the instance is owned by the admin, so we give them the choice as this feature has been requested by some of them :)

ThisIsMissEm:
> Audit logs would be enabled by default. But the instance is owned by the admin, so we give them the choice as this feature has been requested by some of them :)

Maybe so, but they're still an important feature for trust & safety (and as noted in other issues can be used to implement features for things like actions taken on reports)

It'd be good to know *why* they want to disable the audit logs? What problem are the audit logs creating that could necessitate them needing to be disabled?

Chocobozzz:

PR #6048, opened 2023-11-25
Update video-watch.component.html

FediVideos:
Tweaked error message so that users realise that refreshing the page may help the video to load properly.

## Description

If an instance doesn't immediately respond, refreshing the page often allows the video to play properly. This tweak to the error message lets the user know about this option, so that they can watch the video straight away instead of coming back later.

## Has this been tested?


Chocobozzz:
Thanks! This is perfect. I'll just wait a few days before merging because we're releasing PeerTube v6 et we don't want to break translations for this specific message :)

FediVideos:
Brilliant! Thanks for the help!

Chocobozzz:
Thanks again :)

PR #6064, opened 2023-11-29
Fix upgrade.sh when Peertube is installed outside the standard path

### LINKED ISSUE #6063: Backward compatibility broken for upgrade.sh when Peertube is not installed in the standard path
### Describe the problem

When Peertube is not installed in the standard path, the update command was:

```bash
cd /srv/peertube/peertube-latest/scripts && sudo -H -u peertube ./upgrade.sh /srv/peertube/
```

With Peertube v6.0.0, the upgrade script was moved, and replace by this for backward compatibility:

https://github.com/Chocobozzz/PeerTube/blob/d897a04565ea5ee4b9d9d2d93440edc1b52c7015/scripts/upgrade.sh#L5-L7

But parameters are not passed to the new script. So we get:

```bash
 cd /srv/peertube/peertube-latest/scripts && sudo -H -u peertube ./upgrade.sh /srv/peertube/
Error - path "/var/www/peertube" wasn't found

If peertube was installed in another path, you can specify it with
    ./upgrade.sh <PATH>
```

As a workaround, waiting for the correct fix, we can do:

```bash
cd /srv/peertube/peertube-latest/scripts && sudo -H -u peertube ../dist/scripts/upgrade.sh  /srv/peertube/
```

### Additional information

* PeerTube version: v6.0.1

taziden:
fixes #6063

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -4,4 +4,4 @@ set -eu
 
 # Backward path compatibility now upgrade.sh is in dist/scripts since v6
 
-/bin/sh ../dist/scripts/upgrade.sh
+/bin/sh ../dist/scripts/upgrade.sh $1


Chocobozzz:
Thanks!

ROBERT-MCDOWELL:
also it would be nice to see the upgrade script without /var/www/peertube hardcoded. for instance everytime I upgrade I have to modify all path where /var/www/peertube is to my own path (/home/PEERTUBE)

Issue #5460, opened 2022-12-14
Support assets, video CDN


RinCat:
### Describe the problem to be solved

Most common servers/VPS have only 1Gb network port, which usually only support up to 100-200 users watching 1080p at the same time. Higher resolutions and multiple simultaneous users live significantly reduce the number of simultaneous plays available further. Assuming they can use the full 1Gb port, most VPS are shared. The newly supported object storage alleviates this problem in some ways, but not for users who do not use object storage.

### Describe the solution you would like

Support something like 
`assets_base_url` that replace `https://example.com/client/assets` to `https://assets.example.com/client/assets`
`video_base_url` that replace `https://example.com/static/streaming-playlists/` to `https://cdn.example.com/static/streaming-playlists/`

Just like how the `base_url` works in the `object_storage`. The users could use round-robin DNS and/or GeoDNS to do load balancing on multiple servers and reduce the load on the main server.
Unlike the object storage, the files should accessible use current `https://example.com/client/assets` and `https://example.com/static/streaming-playlists/`, so the other CDN/cache servers can get data from the main server to distribute.

Further, some resources (like `/client/en-US/runtime.c6a074343c198fd0.js`) are not under assets, maybe move them to assets or put the whole `/clients` under `assets_base_url`.

This should also works for the live streaming. The CDN / cache server should only cache the `.m3u8` for 1s.

emansom:
Great feature request!

It should also allow multiple CDNs to be defined and PeerTube handing one to each viewer based on occupancy buckets with max slots per CDN configurable.

The viewer internally then sends more details during it's viewer ping on which CDN is accessed, and the occupancy bucket data is updated based on it.

RinCat:
> It should also allow multiple CDNs to be defined and PeerTube handing one to each viewer based on occupancy buckets with max slots per CDN configurable.

If multiple CDNs exists, the client could try to pull data from the top 3 fastest instead of being bound to one.

Chocobozzz:
Hello,

PeerTube already supports CDN where your domain name points to the CDN and the CDN serves PeerTube files from your server :)

RinCat:
Hi @Chocobozzz , that's not the case. Route all traffic thought CDN / edge server is different than route only static video files. The latter can be deployed on servers that are not highly trusted, not even under the same domain name. These servers are also invulnerable to attack because they do not participate in any logical processing and only cache and forward static files.

PR #6069, opened 2023-11-30
Support alternative web videos CDN base url

RinCat:
## Description
This patch introduces the ability to specify an alternate base URL for web video storage in "local", similar to the existing functionality for object storage base URLs. This allows high traffic videos to be distributed across multiple servers or CDNs, reducing the load/cost on the main site/CDN. This is also useful if your video storage and Peertube services are hosted on separate servers.

However, it's important to note one significant limitation: **private and password-protected videos will not have secure protection**. This is because the external server acts purely as a file host without any authentication mechanisms. Thus, anyone with knowledge of the URL can access these videos.

While I'm aware of these drawbacks and doubt the likelihood of this patch being merged, I'm sharing it here for those who may find it useful for their specific needs.

I used this in my peertube and [here](https://v.rincat.ch/w/8tNUvQATwvcSmstPRWdzQg) is a demo if anyone want to check.

## Related issues
https://github.com/Chocobozzz/PeerTube/issues/5460

## Has this been tested?

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -227,6 +227,9 @@ object_storage:
     prefix: ''
     base_url: ''
 
+cdn:
+  web_videos_base_url: ''
+
 log:
   level: 'info' # 'debug' | 'info' | 'warn' | 'error'
 
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -225,6 +225,9 @@ object_storage:
     prefix: ''
     base_url: ''
 
+cdn:
+  web_videos_base_url: '' # Example: 'https://mirror.example.com'
+
 log:
   level: 'info' # 'debug' | 'info' | 'warn' | 'error'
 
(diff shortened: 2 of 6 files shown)


Chocobozzz:
> While I'm aware of these drawbacks and doubt the likelihood of this patch being merged, I'm sharing it here for those who may find it useful for their specific needs.

You can add a configuration key so it's the admin choice if PeerTube also replace URLs of private/internal/password protected videos.

Any news for this PR @RinCat?

RinCat:
> > While I'm aware of these drawbacks and doubt the likelihood of this patch being merged, I'm sharing it here for those who may find it useful for their specific needs.
> 
> You can add a configuration key so it's the admin choice if PeerTube also replace URLs of private/internal/password protected videos.
> 
> Any news for this PR @RinCat?

That's a good idea, I will add that.

PR #6086, opened 2023-12-05
transcoding.web_videos reference and documentation

criadoperez:
## Description

Reference to `transcoding.web_videos` in `productinonal.yaml` was wrong for update to v.6.0. Corrected it.
 - Was refering to `1080p: false` instead of `web_videos:`

Also corrected some minor typos in the docs and api.


Chocobozzz:
Thanks!

Issue #6130, opened 2023-12-20
Remote runner upload success file using external URL


polyhb:
### Describe the problem to be solved

This issue is related to the issue #6125, We are currently using remote runners independently from the Peertube application and we struggle with a traffic problem. When a job is finished, the runner is using the `success` endpoint and return the generated files. The server has to download theses files and then upload files to a S3. It means that it uses the server resources that can create traffic issues.

### Describe the solution you would like

Making the runner uploads its generated files could resolve this issue. 
In order to do that, one solution could be to add an optional `url` property to the `runnerJob`'s payload. If this property is set, then the runner upload the result to this url, and does not return the file in the `success` callback. It offers the possibility for the application that uses the runner to either take the responsibility to store files using its resources, or make the runner directly upload to an url (s3  storage for example).

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/commit/5b4c7fc20d90b44515bcc97d33de1d0a95b79d7b

PR #6135, opened 2023-12-21
Remote runner upload success file using external url

polyhb:
## Description

The goal is to make the remote runner upload success files using directly a given url instead of sending them to the server. It avoids the server to have to download the file and upload it again, thus reducing resources usage.

This is an implementation proposition for the `HLSTranscoding` jobs. After your feedback and if you are ok with this, I'll do the same for the other jobs.

## Related issues

#6130

## Has this been tested?


packages/models/src/runners/runner-job-payload.model.ts
```diff
@@ -23,12 +23,19 @@ export interface RunnerJobVODWebVideoTranscodingPayload {
   }
 }
 
+export interface UploadUrlTransodingResultPayload {
```
lunika, line 26:
```suggestion
export interface UploadUrlTranscodingResultPayload {
```

apps/peertube-runner/src/server/process/shared/process-vod.ts
```diff
@@ -119,11 +119,37 @@ export async function processHLSTranscoding (options: ProcessOptions<RunnerJobVO
       videoFile: videoPath
     }
 
+    if (payload.output.videoFileUrl) {
```
Chocobozzz, line 122: Please wrap these `uploadTranscodeResult` calls in `success` method, I think it will be simpler and will require less code

Chocobozzz:

thino-dev:
Hi, any update for this issue.

Chocobozzz:
We'll work on that feature in the following months

Issue #733, opened 2018-06-27
Provide option to store the source video file


rigelk:
Be it for archival, or just to provide a way to transcode again without a potentially degraded source (the highest transcoded resolution is still a lossy version of the source video file), we might want to store the source video file along, and allow to download it (but not stream it).

Not everyone might be interested by this, of course, but it could certainly interest data preservation projects in general.

rzr:
Yes this is helpful, specially when original format is free (ogv, webm).

rigelk:
Should we allow users to save the source video file by default? I can see admins wishing to prevent users from keeping the source video file since it's additional storage, but that will unevitably restrict its usage a lot. Another option is to allow it by default so long as the bitrate stays under a known limit: it prevents users just pushing rushes from their camera with no regards to the file size, and allows unoptimized albeit better quality/other format videos to stay.

Chocobozzz:
I think we should disable it by default, and not use the bitrate trick: it will be confusing for users and potentially admins to understand why original files of some videos are stored, and other are not.

flaki:
Is this issue open for contributions? I'm using a self-hosted PeerTube instance both for archival of my own content and for integrating with the rest of the fediverse, but PeerTube not keeping around the originals both defeats the first purpose (causing inconvenience, having to keep the original files around) and also often making high-bandwidth videos unusable on default transcode levels (data loss, where PeerTube was the original source and the file is no longer available in its source form).

I am currently looking into patching my own instance to keep the old files around and would be happy to contribute a patch, but it's unclear to me from the current conversation what's the stance around this issue.

chagai95:
> I am currently looking into patching my own instance to keep the old files around and would be happy to contribute a patch, but it's unclear to me from the current conversation what's the stance around this issue.

We started working on a patch, would you be interested in helping us? I think the stance is pretty clear now.

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/pull/6157

The option has to be enabled by the admin. Owners/moderators/admins can download the original file and we include it in the user export archive.

PR #6157, opened 2024-01-04
Add config option to keep original video file (basic first version)

chagai95:
## Description
We added a config option to allow keeping the originally uploaded video instead of it being deleted after the transcoding process.

We added the UUID file name to the `videoSource` table, as suggested [here](https://github.com/Chocobozzz/PeerTube/issues/4731#issuecomment-1027923458).

We found two places in the code that delete the original file, both of which need to be skipped in order to keep the original file. The original file is always kept in the `web-videos` directory, in some cases in the `private` directory therein. This seems to be unrelated to whether the video is publicly published or not. Is there a specific reason for that, which we missed?

Also, are there any other places you think we might need to take into account? Maybe we should somehow make use of the columns `pendingMove`, `pendingTranscode` in the `videoJobInfo` table?

We added a new migration file that adds the two columns: `keptOriginalFile` (boolean), `keptOriginalFileName` (UUID original file name)

Things that we would like to add in the future: 

- UI support for admin
- per video based support where the user decides (including UI)
- download option (including UI)
- adding support for an S3 bucket.
- support replacement of video file.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/733

## Has this been tested?

We would need some assurance regarding our approach, is it the correct approach, and what would be the best way to test it?


Chocobozzz:
> We would need some assurance regarding our approach, is it the correct approach, and what would be the best way to test it?

Yes I think it is. I would just use only `keptOriginalFileName` (`keptOriginalFilename` for consistency with `videoFile` table for example) as if the column is `null` then we don't have this file.

Another question is where do we store such file.

You can follow https://docs.joinpeertube.org/support/doc/development/tests for a guide to write a PeerTube test (server tests will be enough)

chagai95:
Thx for answering, so quick!  

> Another question is where do we store such file.

Until it's decided where is best to save the file we can just keep it there in the meantime or is that a problem?

We will remove `keptOriginalFile` and only use `keptOriginalFileName`.

Is there anything else we would need to do to get the PR merged besides getting the tests to run?

Chocobozzz:
> Until it's decided where is best to save the file we can just keep it there in the meantime or is that a problem?

I think we should store them in a dedicated directory:
 * Easier to backup
 * We may not want to serve original videos like we do for classic videos (and so not serve them directly from nginx)
 * They are not web videos (they may be `.mkv` etc) and they are not HLS files so they should not belong to these directories

> Is there anything else we would need to do to get the PR merged besides getting the tests to run?

Yes, I think we need to support object storage for such files too as it would be complicated to add it in the future (how do we consider a video that has the original file stored on disk and the transcoded files in object storage?)

chagai95:
Thx for clarifying! We added a new dedicated directory, changed to using only keptOriginalFileName, changed keptOriginalFileName to keptOrginalFilename (for consistency with videoFile table - lowercase) and began implementing object storage.

We might need some help with implementing `object storage`.

Chocobozzz:
Hi @chagai95 

What's the status of this PR? :)

chagai95:
I'm still hoping to make some progress with the object storage, but it will take us some time, if anyone else wants to pick it up, I don't want to be in the way, but we are still planning on getting this finished soon!

mzero:
From the discussion here, it sounds like the original filename from the upload will be lost. It would be nice (and match the similar experience on sites like SoundCloud & BandCamp that allow original download) - if the original filename were retained, and what the user got when downloaded. It doesn't matter that it is UUID'd while in the storage system.

Chocobozzz:
> From the discussion here, it sounds like the original filename from the upload will be lost.

No we keep it in the database and so use it as filename when the user downloads it :)

I'll work on this PR and add S3 support and tests

Chocobozzz:
Thanks for the PR.

I added:
 * tests
 * object storage support
 * possibility for uploaders to download the original file
 * inclusion of the original file in the user export archive if available (instead of the max quality file)

Issue #6205, opened 2024-02-05
Quotations in HTML Attributes are not properly escaped, resulting in broken embeds


letydoesstuff:
### Describe the current behavior

HTML attributes used for rendering embeds break if they contain double quote characters. The commit edc6952 attempts to address this issue using backslash escapes. However, this does not work because inserting backslash characters does not actually escape the problematic double quote character from the HTML value.

The current, resultant behavior is that the truncated description gets cut short early after the first backslash character, terminating where the double quote character would be.

![improperescape](https://github.com/Chocobozzz/PeerTube/assets/48048071/a5502398-0a07-43fe-96f2-a294c12af8f4)

### Steps to reproduce

1. Publish a PeerTube video that contains quotations in the first 250 characters of the description. 250 characters is the length of the truncated description passed on to HTML attributes.
2. Post the video somewhere that generates an embed which includes the truncated description. As an example, I posted such a link on Discord.

### Describe the expected behavior

The expected behavior is for the truncated description to render with the embed in its entirety. This would require for double quote characters to be replaced with the  `&quot;` HTML entity instead of being backslash-escaped.

![properescape](https://github.com/Chocobozzz/PeerTube/assets/48048071/3066d9b5-d80f-41fe-a05e-5ac3c873755b).

### Additional information

* PeerTube instance: [Does Stuff Social's Peertube](https://peertube.doesstuff.social), although I've since patched this issue with a linked PR forthcoming
  * URL: https://peertube.doesstuff.social
  * Version: 6.0.3
  * NodeJS version: Not relevant
  * Ffmpeg version: Not relevant

* Browser name, version and platforms on which you could reproduce the bug: All Chromium and Gecko-based browsers
* Link to browser console log if relevant: Not relevant
* Link to server log if relevant (`journalctl` or `/var/www/peertube/storage/logs/`): Not relevant

Chocobozzz:
Fixed by https://github.com/Chocobozzz/PeerTube/pull/6206, thanks again @letydoesstuff

PR #6206, opened 2024-02-05
Fix the escapeAttribute function using HTML entities instead of backslash escapes

letydoesstuff:
## Description

HTML attributes used for rendering embeds break if they contain double quote characters. The commit edc6952 attempts to address this issue using backslash escapes. However, this does not work because inserting backslash characters does not actually escape the problematic double quote character from the HTML value.

The current, resultant behavior is that the truncated description gets cut short early after the first backslash character, terminating where the double quote character would be.

improperescape

The expected behavior is for the truncated description to render with the embed in its entirety.

properescape.

This would require for double quote characters to be replaced with the  `&quot;` HTML entity instead of being backslash-escaped, which is what this PR does.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/6205

## Has this been tested?

- [X] 🙋 no, because I need help 

Sorry, I'm not sure which box I should check here. I'm really new to GitHub and any kind of code dev in general if I'm being quite honest, and parts of the testing guide still go over my head.

I have tested this fix on my own personal PeerTube server. My fix can be confirmed by checking the [HTML meta tags for this video](https://peertube.doesstuff.social/w/xsUNWzuuDcTPy3hQBL5crK) and then posting that link somewhere that would generate an embed that contains the truncated description. The example I used in my screenshots is Discord.

A video on any other PeerTube server with a similar description to mine would have its embed's description terminate early at the first double quote character.

### DIFF
--- a/packages/core-utils/src/renderer/html.ts
+++ b/packages/core-utils/src/renderer/html.ts
@@ -73,5 +73,5 @@ export function escapeHTML (stringParam: string) {
 export function escapeAttribute (value: string) {
   if (!value) return ''
 
-  return String(value).replace(/"/g, '\\"')
+  return String(value).replace(/"/g, '&quot;')
 }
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Thanks! I adapted the tests with your fix :)

Issue #6216, opened 2024-02-12
Video will be overwritten if upload same video file


r6fans:
### Describe the current behavior

When I upload the same video file with different titles, the upload result will be overwritten.

### Steps to reproduce

1. Upload a video file with title 'Test_video_00'
![tube_image_00](https://github.com/Chocobozzz/PeerTube/assets/22339010/010403bc-fb58-4beb-a925-67e329db1ca7)
Result:
![Test_video_02](https://github.com/Chocobozzz/PeerTube/assets/22339010/ee8fdeca-e966-4e75-a2e6-f055d402f925)
2. Upload the same video with title 'Test_video_001'
![TEST_VIDEO_03](https://github.com/Chocobozzz/PeerTube/assets/22339010/4e353650-e0bd-4159-83b0-63011336a382)
3. Video has been overwritten
![Test_video_05](https://github.com/Chocobozzz/PeerTube/assets/22339010/c78acf62-2939-4239-bd4f-3549c4354cc5)

### Describe the expected behavior

When uploading the same video file with different titles, there will display two videos

### Additional information

* PeerTube instance:
  * URL:
  * Version: v6.0.3
  * NodeJS version:
  * Ffmpeg version:

* Browser name, version and platforms on which you could reproduce the bug:
* Link to browser console log if relevant:
* Link to server log if relevant (`journalctl` or `/var/www/peertube/storage/logs/`):

Chocobozzz:
Thanks! Should be fixed by https://github.com/Chocobozzz/PeerTube/commit/7b86b9b458e38493e1be9f26996476f56d68d73f

PR #6231, opened 2024-02-19
Fix upload of the same video file with different titles

kukhariev:
## Description

Safe removal of a resumable upload session on closing

## Related issues

#6216

## Has this been tested?


Chocobozzz:
Hi,

Thanks! But I preferred to remove the cache directly from the server in https://github.com/Chocobozzz/PeerTube/commit/7b86b9b458e38493e1be9f26996476f56d68d73f so we keep the "resume upload" feature of your wonderful lib' when we leave the page and come back with the same video file again

PR #6252, opened 2024-02-27
Short video

ankit-git00:
## Description

Added a new column `shortVideo` in the video Table so that users can specify in the upload API parameter whether the uploaded video is a short video or not. This will provide users with more control over their video uploads and enable them to categorize their content accordingly.

## Has this been tested?


Chocobozzz:
Hi,

Thanks for the PR but I don't think we need an additional column if PeerTube doesn't use it. If it's only used by users, they should use the PeerTube tags system.

PR #6259, opened 2024-03-08
fix(server): cleanup db connections upon exit

kontrollanten:
## Description

Cleanup database connection upon server exit to avoid locking, idle connections in the database.


Chocobozzz:
Thanks!

Issue #6285, opened 2024-03-22
POST /api/v1/videos/{id}/views return 400


kontrollanten:
### Describe the current behavior

In our nginx access log we see a lot of the following errors.

```
[21/Mar/2024:16:58:04 +0000] "POST /api/v1/videos/100dea84-c1d4-49f9-826e-3bb50512b7e4/views HTTP/1.1" 400 207 "/w/2YYZTvhDe4hXyxUWZhgwoY" "Mozilla/5.0 (Linux; Android 12; T431D Build/SP1A.210812.016) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/122.0.6261.119 Mobile Safari/537.36" "82.128.208.12" rt=0.005 ua="127.0.0.1:9000" us="400" ut="0.004" ul="207" cs=-
[21/Mar/2024:16:58:09 +0000] "POST /api/v1/videos/100dea84-c1d4-49f9-826e-3bb50512b7e4/views HTTP/1.1" 400 207 "/w/2YYZTvhDe4hXyxUWZhgwoY" "Mozilla/5.0 (Linux; Android 12; T431D Build/SP1A.210812.016) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/122.0.6261.119 Mobile Safari/537.36" "82.128.208.12" rt=0.001 ua="127.0.0.1:9000" us="400" ut="0.000" ul="207" cs=-
 [21/Mar/2024:16:58:10 +0000] "POST /api/v1/videos/100dea84-c1d4-49f9-826e-3bb50512b7e4/views HTTP/1.1" 400 207 "/w/2YYZTvhDe4hXyxUWZhgwoY" "Mozilla/5.0 (Linux; Android 12; T431D Build/SP1A.210812.016) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/122.0.6261.119 Mobile Safari/537.36" "82.128.208.12" rt=0.003 ua="127.0.0.1:9000" us="400" ut="0.004" ul="207" cs=-
```

We can't see any warnings or errors from Peertube related to these requests.

### Steps to reproduce

1.
2.
3.

### Describe the expected behavior

I assume that the reason for 400 returns is this line https://github.com/Chocobozzz/PeerTube/blob/v6.0.3/server/core/middlewares/validators/videos/video-view.ts#L45

It would be great to log a warning to give more information about why the time is invalid.

### Additional information

* PeerTube instance:
  * Version: 6.0.3
[...]

PR #6288, opened 2024-03-25
fix(server/video-view): log invalid currentTime req

kontrollanten:
## Description
Log a warning when a POST request is done to /api/v1/videos/{id}/views with an invalid currentTime.

## Related issues
#6285

## Has this been tested?

### DIFF
--- a/server/core/middlewares/error.ts
+++ b/server/core/middlewares/error.ts
@@ -1,11 +1,11 @@
+import { HttpStatusCode } from '@peertube/peertube-models'
+import { logger } from '@server/helpers/logger.js'
 import express from 'express'
 import { ProblemDocument, ProblemDocumentExtension } from 'http-problem-details'
-import { logger } from '@server/helpers/logger.js'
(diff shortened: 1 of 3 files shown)


Chocobozzz:
HTTP bad status are already logged in debug mode by https://github.com/Chocobozzz/PeerTube/blob/develop/server/core/middlewares/error.ts#L32

I suggest instead to improve the error message sent to the client to include current time and video duration to see what happens

kontrollanten:
> HTTP bad status are already logged in debug mode by https://github.com/Chocobozzz/PeerTube/blob/develop/server/core/middlewares/error.ts#L32

Bad status can be expected in many cases, such as wrong links, etc. I think it's better to have warnings here since it's unexpected that it should happen at all.

> I suggest instead to improve the error message sent to the client to include current time and video duration to see what happens

That's a good idea to add it to the response. In our case we can't reproduce it ourselves so therefore we would like to have it logged as warnings as well.

Chocobozzz:
Agreed! I updated a little bit the PR.

Issue #6219, opened 2024-02-12
Possibility to group users


kontrollanten:
### Describe the problem to be solved

We'd like to group users that's donating money in order to be able to provide them with some "donators only" videos.

## Scalable solution
* Possibility to add user groups via the web interface.
* Possibility to add/remove users from groups via web interface.
* Possibility to set which groups can see specific videos.

This would be an easy solution for web admins and can be well integrated to the user management in Peertube. But it will add a lot of new features to Peertube, which has to be maintained.

## Easy to implement solution
* Adding a ` filter:api.user.me.get.result`  API from where plugin developers can add custom attributes/data to users in order to group them

In this way plugin developers can create custom API endpoints to add/remove users from groups and store the result with storageManager. The upside with this solution is the small burden laid on Peertube core.

heapdavid:
Is this currently possible with either of the two methods above? We'd also find this very handy

Chocobozzz:
Merging this issue into https://github.com/Chocobozzz/PeerTube/issues/6559

PR #6292, opened 2024-03-26
feat(plugins): add filter:api.user.me.get.result

kontrollanten:
## Description
Adds `filter:api.user.me.get.result` to let plugins modify the response of GET /api/v1/users/me

## Related issues
relates to #6219

## Has this been tested?

### DIFF
--- a/packages/models/src/plugins/server/server-hook.model.ts
+++ b/packages/models/src/plugins/server/server-hook.model.ts
@@ -137,7 +137,9 @@ export const serverFilterHookObject = {
   // Peertube >= 5.2
   'filter:feed.podcast.channel.create-custom-tags.result': true,
   // Peertube >= 5.2
-  'filter:feed.podcast.video.create-custom-tags.result': true
+  'filter:feed.podcast.video.create-custom-tags.result': true,
+  // Peertube >= 6.1
+  'filter:api.user.me.get.result': true
 }
 
 export type ServerFilterHookName = keyof typeof serverFilterHookObject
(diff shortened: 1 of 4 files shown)


server/core/controllers/api/users/me.ts
```diff
@@ -160,7 +160,13 @@ async function getUserInformation (req: express.Request, res: express.Response)
   // We did not load channels in res.locals.user
   const user = await UserModel.loadForMeAPI(res.locals.oauth.token.user.id)
 
-  return res.json(user.toMeFormattedJSON())
+  const result = await Hooks.wrapObject(
+    user.toMeFormattedJSON(),
+    'filter:api.user.me.get.result',
+    user
```
Chocobozzz, line 166: Prefer `{ user }` so we can add other properties in the future

packages/models/src/plugins/server/server-hook.model.ts
```diff
@@ -137,7 +137,8 @@ export const serverFilterHookObject = {
   // Peertube >= 5.2
   'filter:feed.podcast.channel.create-custom-tags.result': true,
   // Peertube >= 5.2
-  'filter:feed.podcast.video.create-custom-tags.result': true
+  'filter:feed.podcast.video.create-custom-tags.result': true,
+  'filter:api.user.me.get.result': true
```
Chocobozzz, line 141: Please add `PeerTube >= 6.1` comment

Chocobozzz:
Thanks!

PR #6300, opened 2024-03-29
feat(plugins): add req.rawBody for routes

kontrollanten:
## Description
Stripe webhooks endpoints requires to read the raw request body.
https://docs.stripe.com/webhooks#verify-webhook-signatures-with-official-libraries

## Has this been tested?

### DIFF
--- a/server/server.ts
+++ b/server/server.ts
@@ -211,6 +211,10 @@ app.use(express.json({
         message: 'Invalid digest'
       })
     }
+
+    if (req.originalUrl.startsWith('/plugins/')) {
+      req.rawBody = buf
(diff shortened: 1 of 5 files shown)


server/server.ts
```diff
@@ -211,6 +211,8 @@ app.use(express.json({
         message: 'Invalid digest'
       })
     }
+
+    req.rawBody = buf
```
Chocobozzz, line 215: I suggest to add this only if `req.originalUrl` starts with `/plugins/` because we don't need it in peertube

Chocobozzz:
Thanks:

PR #6302, opened 2024-03-29
feat(plugins): add peertubeHelpers.loadWithFiles

kontrollanten:
## Description
Let plugin authors to load a video with all its files. This can be useful when a plugin want to switch video files based on the users capabilities.

## Has this been tested?

### DIFF
--- a/server/core/types/plugins/register-server-option.model.ts
+++ b/server/core/types/plugins/register-server-option.model.ts
@@ -17,7 +17,7 @@ import {
   VideoBlacklistCreate
 } from '@peertube/peertube-models'
 import { ActorModel } from '@server/models/actor/actor.js'
-import { MUserDefault, MVideo, MVideoThumbnail, UserNotificationModelForApi } from '../models/index.js'
+import { MUserDefault, MVideo, MVideoThumbnail, MVideoWithAllFiles, UserNotificationModelForApi } from '../models/index.js'
 import {
   RegisterServerAuthExternalOptions,
   RegisterServerAuthExternalResult,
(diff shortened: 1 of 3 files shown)


server/core/types/plugins/register-server-option.model.ts
```diff
@@ -34,6 +34,7 @@ export type PeerTubeHelpers = {
 
   videos: {
     loadByUrl: (url: string) => Promise<MVideoThumbnail>
+    loadWithFiles: (id: number | string) => Promise<MVideoWithAllFiles>
```
Chocobozzz, line 37: `loadByIdOrUUIDWithFiles`

Chocobozzz:
`getAllFiles` helper don't do the job for your use case?

kontrollanten:
Yes, I need to call `getHLSPlaylist()` to get URL to master playlist and segments file.

Chocobozzz:
> Yes, I need to call getHLSPlaylist() to get URL to master playlist and segments file.

Okay then, but `getHLSPlaylist` is not considered as a plugin API and can break at any moment.

Don't you prefer to create a dedicated plugin helper, like we did for video files, to fetch HLS playlist information?

Chocobozzz:
> That would be nice, but I think this use case is too niche to deserve its own plugin API. We use it like this:

Thanks! `filter:api.video.get.result` filter hook cannot be used to change HLS playlist URLs?

kontrollanten:
> > That would be nice, but I think this use case is too niche to deserve its own plugin API. We use it like this:
> 
> Thanks! `filter:api.video.get.result` filter hook cannot be used to change HLS playlist URLs?

It can, that's where we're running the code I pasted above. We return different playlists depending whether the user is a "premium user" or not.

PR #6346, opened 2024-04-23
Feature 1322-v2

kntndrsn:
## Description

Second iteration of #1322 using the embedded player and creating the image from the player.

## Related issues

#1322

## Has this been tested?

## Screenshots

Advanced Settings section has been updated with a "Select from video" option.

image

When the user selects "Select from frame", the embedded player is loaded.

image

The user can play through the video to select an image.

image

When the user selects "Use frame" the preview is updated.

image

When the user clicks the normal "Update" button, the update is applied.

image

### DIFF
--- a/client/src/app/modal/account-setup-warning-modal.component.ts
+++ b/client/src/app/modal/account-setup-warning-modal.component.ts
@@ -1,6 +1,7 @@
 import { CommonModule } from '@angular/common'
 import { Component, ElementRef, ViewChild } from '@angular/core'
 import { FormsModule } from '@angular/forms'
+import { RouterLink } from '@angular/router'
 import { Notifier, ServerService, User, UserService } from '@app/core'
 import { PeertubeCheckboxComponent } from '@app/shared/shared-forms/peertube-checkbox.component'
 import { GlobalIconComponent } from '@app/shared/shared-icons/global-icon.component'
@@ -13,7 +14,7 @@ import { peertubeLocalStorage } from '@root-helpers/peertube-web-storage'
   templateUrl: './account-setup-warning-modal.component.html',
   styleUrls: [ './account-setup-warning-modal.component.scss' ],
   standalone: true,
-  imports: [ CommonModule, GlobalIconComponent, PeertubeCheckboxComponent, FormsModule ]
+  imports: [ CommonModule, GlobalIconComponent, PeertubeCheckboxComponent, FormsModule, RouterLink ]
 })
 export class AccountSetupWarningModalComponent {
   @ViewChild('modal', { static: true }) modal: ElementRef
--- a/client/src/sass/bootstrap.scss
+++ b/client/src/sass/bootstrap.scss
@@ -390,7 +390,7 @@ body {
 // RTL compatibility
 // ---------------------------------------------------------------------------
 
-.modal .modal-header .modal-title {
+:root[dir=rtl] .modal .modal-header .modal-title {
   margin-inline-end: auto;
   margin-right: unset;
 }
--- a/packages/tests/src/api/server/stats.ts
+++ b/packages/tests/src/api/server/stats.ts
@@ -34,11 +34,14 @@ describe('Test stats (excluding redundancy)', function () {
     const { account } = await servers[0].users.create({ username: user.username, password: user.password })
     userAccountId = account.id
 
-    const { uuid } = await servers[0].videos.upload({ attributes: { fixture: 'video_short.webm' } })
+    {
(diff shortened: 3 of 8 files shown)


kntndrsn:
@Chocobozzz Hi. I tried using the my-embed component instead of making my own iframe but I'm having some challenges getting the embed api to work with it. I added a parameter allowing the API to be enabled, but it seems that the dynamically inserted iframe isn't accessible. I've tried what I can find online including ngAfterViewInit to setup the embed-api, but no luck. Any suggestions?

kntndrsn:
@Chocobozzz Hello. Please lmk if there is anything else I can do on this one.

client/src/app/+videos/+video-edit/shared/thumbnail-manager/thumbnail-manager.component.html
```diff
@@ -0,0 +1,29 @@
+<div class="root">
+    <div *ngIf="!selectingFromVideo" class="preview-container">
+  
+      <img *ngIf="imageSrc" [ngStyle]="{ width: previewWidth, height: previewHeight }" [src]="imageSrc" class="preview"
+        alt="Preview" i18n-alt />
+      <div *ngIf="!imageSrc" [ngStyle]="{ width: previewWidth, height: previewHeight }" class="preview no-image"></div>
+  
+    </div>
+  
+    <div id="embedContainer" class="" style="position: relative;"></div>
```
Chocobozzz, line 10: You can use and improve our embed component instead of injecting your embed code? https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/app/shared/shared-main/video/embed.component.ts

client/src/app/+videos/+video-edit/shared/thumbnail-manager/thumbnail-manager.component.html
```diff
@@ -0,0 +1,29 @@
...
+  
+    <div class="inputs">
+      <my-reactive-file *ngIf="!selectingFromVideo" inputName="uploadNewThumbnail" inputLabel="Upload image"
```
Chocobozzz, line 13: Can you try to use `@if`/`@else` to it's clearer to read? https://blog.angular-university.io/angular-if/

client/src/standalone/videos/embed-api.ts
```diff
@@ -195,4 +197,18 @@ export class PeerTubeEmbedApi {
...
+    const video = this.element
+
+    const canvas = document.createElement('canvas')
+
+    canvas.width = video.videoWidth
+    canvas.height = video.videoHeight
+
+    canvas.getContext('2d').drawImage(video, 0, 0, canvas.width, canvas.height)
+
+    return canvas.toDataURL('image/jpeg')
```
Chocobozzz, line 212: Do you think we can generate a blob directly so we don't have to change server code? https://stackoverflow.com/questions/6850276/how-to-convert-dataurl-to-file-object-in-javascript

Chocobozzz (commented):
Thanks a lot and sorry for the late review!

Chocobozzz:
@kntndrsn Do you need help to update the PR?

kntndrsn:
> @kntndrsn Do you need help to update the PR?

I actually didn't mean to close. I had to rebase my repo to get back to the current position and it must have decided to close it. I did initially try the two points you mentioned 

1. Use my-embed component (see comment above)
2. Create as blob to attach so we can avoid changing server code.

Neither worked for me, but was going to give it another go when I found some time.

client/src/standalone/videos/embed-api.ts
kntndrsn: I can create as a blob, but I'm unable to figure out how to attach it such that the server code will detect it as an attached file.

client/src/app/+videos/+video-edit/shared/thumbnail-manager/thumbnail-manager.component.html
kntndrsn: I made a comment above about this. I made an honest effort to make the my-embed component work, however, I couldn't get the Embed API to work with it. I did search the code for other places it was used but couldn't find anything. If there is something I have missed or if you have some thoughts on how to approach, I can try again.

client/src/standalone/videos/embed-api.ts
kntndrsn: I was able to get this one figured out. Just had to create a File object with the proper MIME type to set as the form value.

Chocobozzz:
Please create another PR I can try so I can update your code to include my suggestions :)

kntndrsn:
> Please create another PR I can try so I can update your code to include my suggestions :)

#6424

PR #6357, opened 2024-04-26
fix(server/plugins): avoid duplicate settings

### LINKED ISSUE #6327: Possibility for plugins to update settings options value

We've a plugin setting that populates its options based on a REST API response:

```
    try {
      stripePlans = await getStripeSubscriptionPlans(apiKey)
    } catch (err: any) {
      logger.info('Couldn\'t fetch Stripe plans', { err })
    }

    registerSetting({
      name: SETTING_STRIPE_SUBSCRIPTION_PLAN_ID,
      label: 'Stripe plan used for subscription',
      type: 'select',
      options: stripePlans.map((plan) => ({
        value: plan.id,
        label: (plan.product as Stripe.Product)?.name ?? plan.id
      })),
      private: true
    })
```

Therefore we need to be able to update the available options. When calling `registerSettings` multiple times it will create a new setting for each call, instead we would like to just update the options.

Some way of updating the options. Possible solutions:

a) When calling `registerSetting` with a name that already exists, the new options will override the existing.
b) Add `unregisterSetting` where you pass the setting name as argument.
c) Add `deleteSetting` to `settingsManager` API.

### LINKED ISSUE #6356: registerSetting adds multiple settings when called multiple times with same setting name

When calling registerSetting with the same name property, the setting will be added multiple times.

```
registerSetting({
  name: SETTING_STRIPE_SUBSCRIPTION_PLAN_ID,
  label: 'Stripe plan used for subscription',
  type: 'select',
  options: stripePlans.map((plan) => ({
    value: plan.id,
    label: (plan.product as Stripe.Product)?.name ?? plan.id
  })),
  private: true
})

...

whenSomethingHasChanged.then(() => {
  // Update options values
  registerSetting({
    name: SETTING_STRIPE_SUBSCRIPTION_PLAN_ID,
    label: 'Stripe plan used for subscription',
    type: 'select',
    options: stripePlans.map((plan) => ({
      value: plan.id,
      label: (plan.product as Stripe.Product)?.name ?? plan.id
    })),
    private: true
  })
})
```

### Steps to reproduce

1. Run the code above.

Since the name property is unique it shouldn't be possible to have multiple settings with the same name, instead it should be updated.

### Additional information

* PeerTube instance:
  * URL:
  * Version: 6.0.3
  * NodeJS version:
  * Ffmpeg version:

kontrollanten:
## Description
Filter settings so that the name property is unique.

## Related issues
closes #6327 

## Has this been tested?


Chocobozzz:
Thanks!

PR #6358, opened 2024-04-26
feat(client/plugins): add getUser to helpers

### LINKED ISSUE #6355: Give plugins access to user object on the client

We'd like to show menu items based on properties we've added to the user object on the server side.

Add a peertube helper function that returns the user object, including custom added properties from the `filter:api.user.me.get.result` filter.

kontrollanten:
## Description
Let plugins access the user object.

## Questions
* I wasn't sure how to handle the plugin helpers for the standalone version, so I set it to unimplemented.
* Should getUser return `ServerMyUserModel` instead of `AuthUser`? By doing that the methods in `AuthUser` doesn't need to be a part of the public API.

## Related issues
closes #6355

## Has this been tested?

There's no test suites for the client.

### DIFF
--- a/client/src/app/core/plugins/plugin.service.ts
+++ b/client/src/app/core/plugins/plugin.service.ts
@@ -250,6 +250,10 @@ export class PluginService implements ClientHook {
         return firstValueFrom(obs)
       },
 
+      getUser: () => {
+        return this.authService.getUser()
+      },
+
       getServerConfig: () => {
         const obs = this.server.getConfig()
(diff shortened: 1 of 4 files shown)


kontrollanten:
I don't know why all checks are failing, it seems unrelated to my changes?

kontrollanten:
Solved it with 8e2d1e5090b1b848eba05bd3eb80163c172b68a0

Chocobozzz:
Thanks!

PR #6379, opened 2024-05-06
Fix typo at log.rotation.enabled

skid9000:
## Description

In default.yaml and production.yaml.example there is what seems to be a typo at log.rotation.enabled

## Related issues

I'm not sure this typo caused any problem as i couldn't see any config parsing error on my instance but for coherence sake, i though i would still PR that.

## Has this been tested?


Chocobozzz:
Thanks!

PR #6394, opened 2024-05-12
feat(server): dont client cache HTML files

### LINKED ISSUE #6393: Explicitly don't cache HTML files

Currently there's no Cache-Control header in HTML responses. This is causing a heuristic caching. As I understand it most browsers won't cache it since there's no Last-Modified header. But proxies and CDN's may behave different.

### Steps to reproduce

1.
2.
3.

Return `Cache-Control: max-age=0, no-cache, must-revalidate` for HTML pages.

### Additional information

* PeerTube instance:
  * URL: 6.1.0
  * Version:
  * NodeJS version:
  * Ffmpeg version:

kontrollanten:
## Description
Tell the clients to not cache the HTML response.

## Related issues
closes #6393


Chocobozzz:
Thanks!

PR #6402, opened 2024-05-17
fix(server/geo-ip): throttle db download

### LINKED ISSUE #6381: Instance is getting slow upon failed geoip db update
Our instances fails to download the geoip db, which seems to slow down the whole server. Upon start the response time is OK, but then it seems to add on response time and then gets unresponsive. When disabling geoip in the config the server works well.

### Steps to reproduce

1. Fail to download geoip city db.
2. Have a lot of concurrent watchers.
3. Watch the response time get longer and longer.

Peertube should handle failed geoip db downloads gently and continue to work just as when the geoip config is disabled.

### Additional information

* PeerTube instance:
  * URL:
  * Version: 6.1.0
  * NodeJS version:
  * Ffmpeg version:

kontrollanten:
## Description
Throttle geo ip db download so that it's not ran more often than once each 10 minutes. This is to handle when the geo ip db can't be downloaded, for instance because of the remote server is unreachable.

## Related issues
Closes #6381

## Has this been tested?

### DIFF
--- a/server/core/helpers/geo-ip.ts
+++ b/server/core/helpers/geo-ip.ts
@@ -6,6 +6,7 @@ import { CONFIG } from '@server/initializers/config.js'
 import { logger, loggerTagsFactory } from './logger.js'
 import { isBinaryResponse, peertubeGot } from './requests.js'
 import { isArray } from './custom-validators/misc.js'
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks!

PR #6418, opened 2024-05-27
feat: config option object_storage.max_request_attempts

### LINKED ISSUE #6415: Possibility to increase maxAttempts for s3 uploads

[Backblaze architecture is designed to return 500/503 in certain situations](https://www.backblaze.com/blog/b2-503-500-server-error/). The instruction from Backblaze is to continue to retry until 2xx is returned. Currently the default value of 3 is used for maxAttempts, which may cause the move-to-object-storage jobs fail. I've talked to Backblaze support and they've verified that three failures is something you've to count with.

Make it configurable for admins to increase the max attempts.
https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html

kontrollanten:
## Description
Backblaze recommends to have a high amount of attempts since they've designed their architecture so that it will return 5xx errors to indicate that the client should do a new attempt.

https://www.backblaze.com/blog/b2-503-500-server-error/

## Related issues

closes #6415

## Has this been tested?

Only manual test with a local Minio server.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -213,6 +213,10 @@ object_storage:
   # Maximum amount to upload in one request to object storage
   max_upload_part: 100MB
 
+  # Maximum attempts to make on a request to object storage. Some object storage providers, for instance Backblaze,
+  # expects the client to retry upload upon 5xx errors. If you're using such a provider then you can increase this value.
+  max_request_attempts: 3
+
   streaming_playlists:
     bucket_name: 'streaming-playlists'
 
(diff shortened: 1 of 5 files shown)


config/default.yaml
```diff
@@ -213,6 +213,10 @@ object_storage:
   # Maximum amount to upload in one request to object storage
   max_upload_part: 100MB
 
+  # Maximum attempts to make on a request to object storage. Some object storage providers, for instance Backblaze,
+  # expects the client to retry upload upon 5xx errors. If you're using such a provider then you can increase this value.
+  max_attempts: 3
```
Chocobozzz, line 218: `max_request_attempts`

Chocobozzz:
Thanks!

Issue #5618, opened 2023-02-18
Increase thumbnail resolution (or make configurable)


aeharding:
### Describe the problem to be solved

A large and growing segment of computer displays are running at 2x and larger pixel densities.

Currently, PeerTube thumbnail images are the same size as the actual container (1x):

<img width="595" alt="Screenshot 2023-02-17 at 9 50 50 PM" src="https://user-images.githubusercontent.com/2166114/219829988-a594229e-fa34-4638-a8d3-186b23a0cedb.png">

However, this leads to thumbnails that look noticeably blurry/low quality.

### Describe the solution you would like

Double the size of thumbnails, or greater (to allow for more flexibility in customizing the UI, for example zoom animations).

This would also align with what other video streaming platforms are doing, such as YouTube:

<img width="950" alt="Screenshot 2023-02-17 at 9 54 53 PM" src="https://user-images.githubusercontent.com/2166114/219830157-888d19c1-dde1-4cca-8f2e-15090cc64c62.png">

Lastly, mobile devices already use the larger `PREVIEWS_SIZE` as thumbnails, so increasing `THUMBNAILS_SIZE` would have no impact on mobile devices.

Similar issue, but for `PREVIEWS_SIZE`: #5464

eddyacthergal:
Hello,
We suffer from the same prob, and we would love to have better quality resolution for thumbnail.
Maybe a way to configure it on server side (if not already available ?)

If already exists, I would be interesting to know how to do it .

Thx in advance

gergob:
I am planning to do a PR for making the `PREVIEWS_SIZE` and `THUMBNAILS_SIZE` configurable from some of the configs. The default values would stay the same, so we maintain backwards compatibility:

```js
// Videos thumbnail size
const THUMBNAILS_SIZE = {
  width: 280,
  height: 157,
  minWidth: 150
}
const PREVIEWS_SIZE = {
  width: 850,
  height: 480,
  minWidth: 400
}
```

What do you think ?

Chocobozzz:
Please do :)

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/commit/3fbe84918392141e16976b952932e7eac10e99be & https://github.com/Chocobozzz/PeerTube/commit/1e3a5b25c367b47e49808b443c504769db21ebe6

PR #6423, opened 2024-05-29
#5618 - Configuration fields for Thumbnails and Previews

gergob:
## Description

Feature Change:
1. Updated the configurations, so `Thumbnail` and `Preview` image sizes can be configured in `yaml` files. In case the configuration is not there, there are some defaults left there.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/5618

## Has this been tested?

## Test resuts

```text
> peertube@6.1.0 mocha
> mocha --exit --bail packages/tests/src/api/server/config.ts



  Test static config
    ✔ Should tell the client that edits are not allowed (48ms)
    ✔ Should error when client tries to update

  Test config
    Config keys
      ✔ Should have the correct default config
      ✔ Should have a correct config on a server with registration enabled
      ✔ Should have a correct config on a server with registration enabled and a users limit (168ms)
      ✔ Should have the correct video allowed extensions
      ✔ Should get the customized configuration
      ✔ Should update the customized configuration
      ✔ Should have the correct updated video allowed extensions (1274ms)
      ✔ Should have the configuration updated after a restart (933ms)
      ✔ Should fetch the about information
      ✔ Should remove the custom configuration
      ✔ Should enable/disable security headers (995ms)
    Image files
      Banner
        ✔ Should update instance banner (434ms)
        ✔ Should re-update an existing instance banner (114ms)
        ✔ Should remove instance banner
      Avatar
        ✔ Should update instance avatar (930ms)
        ✔ Should have the avatars in the AP representation of the instance
        ✔ Should remove instance avatar
        ✔ Should not have the avatars anymore in the AP representation of the instance


  20 passing (10s)
```

## Screenshots

peertube-image-size-changes

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -468,6 +468,16 @@ thumbnails:
     # Increasing this value will increase CPU and memory usage when generating the thumbnail, especially for high video resolution
     # Minimum value is 2
     frames_to_analyze: 50
+  size:
+    width: 280
+    height: 157
+    min_width: 150
+
+previews:
+  size:
+    width: 850
+    height: 480
+    min_width: 400
 
 stats:
   # Display registration requests stats (average response time, total requests...)
--- a/server/core/initializers/config.ts
+++ b/server/core/initializers/config.ts
@@ -370,6 +370,18 @@ const CONFIG = {
   THUMBNAILS: {
     GENERATION_FROM_VIDEO: {
       FRAMES_TO_ANALYZE: config.get<number>('thumbnails.generation_from_video.frames_to_analyze')
+    },
+    SIZE: {
+      WIDTH: config.get<number>('thumbnails.size.width'),
+      HEIGHT: config.get<number>('thumbnails.size.height'),
+      MIN_WIDTH: config.get<number>('thumbnails.size.min_width'),
+    }
+  },
+  PREVIEWS: {
+    SIZE: {
+      WIDTH: config.get<number>('previews.size.width'),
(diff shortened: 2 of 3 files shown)


Chocobozzz:
Thanks! Rebased manually in https://github.com/Chocobozzz/PeerTube/commit/3fbe84918392141e16976b952932e7eac10e99be & https://github.com/Chocobozzz/PeerTube/commit/1e3a5b25c367b47e49808b443c504769db21ebe6

gergob:
@Chocobozzz thank you!

Sakrecoer:
Apologies if i am doing this wrong.. This issue is mentioned in the release notes of 6.2.0. However, these settings does not seem to be present in the latest version of [default.yaml](https://github.com/Chocobozzz/PeerTube/blob/18f13f43ec6f17df457d6cf61e5030179670949c/config/default.yaml#L470) After upgrade from `6.1.0` to `6.2.0`, it was also not present in `production.yaml.new`

gergob:
@Sakrecoer FYI, here is a sample config, which you can use:

```yaml
thumbnails:
  generation_from_video:
    # How many frames to analyze at the middle of the video to select the most appropriate one
    # Increasing this value will increase CPU and memory usage when generating the thumbnail, especially for high video resolution
    # Minimum value is 2
    frames_to_analyze: 80
  sizes:
    - width: 1280
      height: 720

    - width: 1280
      height: 720
```

PR #6424, opened 2024-05-29
Feat-1322 Select Thumbnail from Video Frame

kntndrsn:
## Description

Version 3 of the select from the select thumbnail from video frame documented in #1322 . 

This version does all the work in the client without requiring an update to the server side code.

## Related issues

#1322 

## Has this been tested?

## Screenshots

Screenshots included in pull #6346.


Chocobozzz:
Thanks a lot! I applied the suggestions in https://github.com/Chocobozzz/PeerTube/pull/6424/commits/2f478c73b287ea1dfb317b8ce308af266ebb9ac1

PR #6443, opened 2024-06-17
server(plugins): upate without downtime

### LINKED ISSUE #4828: Update plugins without downtime
When updating a plugin the plugin will be unavailable during the installation. We're storing all our plugins at s3fs, which means it takes 15-60 minutes to update. 

Let the plugin be available during the whole installation.

Install in the new version in tmp folder and just do a quick switch when it's done.

kontrollanten:
## Description

Create a new folder each time a new plugin/theme is installed or updated. The folder name is created based on the package.json content hash.

## Related issues
closes #4828

## Has this been tested?


kontrollanten:
I don't know why the test is failing, it seems unrelated to the changes and it works when I test locally.

@Chocobozzz , currently this PR doesn't handle multiple parallell installs. The package.json is copied when install process begins, so if there's multiple parallell installations only the last one will be stored in package.json. I propose to move the plugin install/update/uninstall processes to be handled by a job instead. Cons:

a) More stable server. If the server restarts just after an admin has started a plugin installation it won't be aborted.
b) Better UX. Today the user get an error message saying the installation failed when there's a HTTP timeout.

kontrollanten:
@Chocobozzz Let me know what you think of this draft. If you're positive to the solution I'll fix the broken tests.

kontrollanten:
@Chocobozzz Ping

Chocobozzz:
> @Chocobozzz Ping

I'm sorry but I unfortunately don't have time at the moment to review your work

kontrollanten:
Hi,

I don't remember the details how this is implemented, but I can't remember it should increase the process memory. Why do you fear it?

Chocobozzz:
> I don't remember the details how this is implemented, but I can't remember it should increase the process memory. Why do you fear it?

Because every time we update a plugin we re-load all of them, and since we can't control the ESM cache I wonder if the RAM would not increase indefinitely

Chocobozzz:
>  A more robust solution could be to run plugins as spawned microservices that communicates with PT over HTTP.

I think it's called an "integration", a choice [ghost blog platform](https://ghost.org/integrations/) has made. It's an interesting choice for better security and performance, but it's much more work to implement on PeerTube side and less flexible.

Do you think that our recent move to PNPM can help to upgrade plugins without downtime in a simpler way than this PR? I'm thinking about aliases for example: https://pnpm.io/aliases

kontrollanten:
> Do you think that our recent move to PNPM can help to upgrade plugins without downtime in a simpler way than this PR? I'm thinking about aliases for example: https://pnpm.io/aliases

It sounds that it can solve the issue. At least worth a try since it seems easy to implement. The solution would then be to run `pnpm add peertube-plugin-my-custom@npm:peertube-plugin-my-custom@[version-number]` and then run the decache?

Chocobozzz:
> It sounds that it can solve the issue. At least worth a try since it seems easy to implement. The solution would then be to run pnpm add peertube-plugin-my-custom@npm:peertube-plugin-my-custom@[version-number] and then run the decache?

Yes, I think so! It's worth testing! As a side note we still `hoisted` `nodeLinker` for plugins, to not break existing plugins

PR #6454, opened 2024-06-25
Add Express req, res to filter:api.search.videos.local.list.params

grahhnt:
## Description

Add Express request & response to `filter:api.search.videos.local.list.params`'s hook context.

I believe this was needed because the query given to the hook has already been filtered for unwanted parameters before the hook executes, this adds the request & response to allow access to those query parameters that might've been removed.

## Related issues

N/A

## Has this been tested?

I do not believe this change needs tests written for it


Chocobozzz:
Thanks!

PR #6527, opened 2024-08-01
Fix documentation:

JohnXLivingston:
## Description

The documentation says "reload" instead of "start" or "restart".

### DIFF
--- a/support/doc/production.md
+++ b/support/doc/production.md
@@ -163,7 +163,7 @@ To generate the certificate for your domain as required to make https work you c
 ```bash
 sudo systemctl stop nginx
 sudo certbot certonly --standalone --post-hook "systemctl restart nginx"
-sudo systemctl reload nginx
+sudo systemctl restart nginx
 ```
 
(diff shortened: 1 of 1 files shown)


Chocobozzz:
The post certbot hook has not already restarted nginx?

JohnXLivingston:
> The post certbot hook has not already restarted nginx?

Hum...
I did not check. It just seemed wrong to me. Maybe you are right.

PaliPalo:
I've read that `--post-hook` works only with the `renew` action. So this command `--pre-hook ..` should be removed and the `start` should be done instead of `reload` (just `start` no need to stop nginx twice since the `stop` is done before `certbot`)

Chocobozzz:
Let's change this to `restart` then, just in case ;)

JohnXLivingston:
Here you go :)

Chocobozzz:
Thanks @JohnXLivingston :pray:

JohnXLivingston:
> Thanks @JohnXLivingston 🙏

That was a hard work, now i need to rest ;)

PR #6548, opened 2024-08-16
Allow installing plugins declaratively

### LINKED ISSUE #6428: Declarative plugin management

For the [NGI-funded Summer of Nix project](https://github.com/ngi-nix/summer-of-nix), we (@infinidoge and I) are looking into how [peertube plugins can be managed declaratively](https://github.com/ngi-nix/ngipkgs/issues/8) within a NixOS module.

From our understanding, plugins can only be installed imperatively: Start the CLI & tell it to install a plugin, and peertube will trigger a state change in the database to register the plugin and look for it in its plugin directory on startup. This extra requirement on a certain database state is a major pain point for our endeavour.

A previous effort around this found a workaround by

1. Collecting a list of every plugin that we want to have available
2. Getting a hash of peertube's `storage.plugins` package.json
3. Launching peertube
4. Running the plugin install script for every desired plugin
5. Comparing the list of previously-desired plugins against now-desired plugins, and running the uninstall script for every plugin that isn't desired anymore
6. Saving the list from step 1 under `storage.plugins`
7. Comparing the hash from step 2 against the current hash of that file, triggering a restart of peertube if they differ

But we think this extra plugin management around your own plugin management is very unsatisfying & hacky solution.

We would like to find solution that doesn't rely on us doing plugin management around your plugin management.

Maybe there could be a way to tell peertube to read a list of additional plugins from somewhere, or have it automagically find additional plugins if they're in a certain directory.

Ideal for us (I think) would be a way to completely lock down the imperative plugin management and have it be fully managed by the external mechanism, so we can ensure that only the plugins the user desires are available to peertube.

Infinidoge:
## Description

Implements the ability for PeerTube to self-install plugins that are declared upfront in a file, checking the file on startup to install/uninstall/update accordingly.

I am not a seasoned TypeScript developer by any stretch of the imagination, so please let me know wherever the code can be improved.

TODO:
- [ ] Add usage to documentation
- [ ] Add unit test

## Related issues

Resolves #6428

## Has this been tested?

### DIFF
--- a/server/core/lib/plugins/plugin-manager.ts
+++ b/server/core/lib/plugins/plugin-manager.ts
@@ -20,6 +20,7 @@ import { decachePlugin } from '@server/helpers/decache.js'
 import { ApplicationModel } from '@server/models/application/application.js'
 import { MOAuthTokenUser, MUser } from '@server/types/models/index.js'
 import { isLibraryCodeValid, isPackageJSONValid } from '../../helpers/custom-validators/plugins.js'
+import { execShell } from '../../helpers/core-utils.js'
 import { logger } from '../../helpers/logger.js'
 import { CONFIG } from '../../initializers/config.js'
 import { PLUGIN_GLOBAL_CSS_PATH } from '../../initializers/constants.js'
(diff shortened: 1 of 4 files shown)


synq:
I have been pushing to get Peertube updated in the NixOS packages repository. 

Unfortunately it does not build because the program itself references to a plugin (which is outside of the program itself).
https://github.com/NixOS/nixpkgs/pull/358194

If this PR would be implemented it would fix that problem and we could get Peertube upgraded from 6.0.4 to 6.3.3

Chocobozzz:
Hi @synq

I'm not sure to understand "it does not build because the program itself references to a plugin" because we don't install plugins on PeerTube installation

Infinidoge:
Having burnt myself out over yarn while working on this, I will fully admit that it's kinda dead in the water. Because of yarnpkg/yarn#7870 (5 YEARS. COME ON YARN.), making this work with how PeerTube manages plugins internally is... extremely difficult. (Or well, more specifically, making it behave the way I want it to for the greater goal of making it work with Nix is extremely difficult.) 

To the best of my memory the fundamentals of the PR work. However, due to not hosting a PeerTube instance myself and having many other things higher in priority, I don't have much drive to keep working on it.

Chocobozzz:
Would `npm` or `pnpm` solve your problem?

Infinidoge:
Yes. (And if i remember correctly, the current NixOS module patches PeerTube to use `npm` for this reason.)

Chocobozzz:
Thanks, I created an issue: https://github.com/Chocobozzz/PeerTube/issues/6836

Chocobozzz:
Hi,

Does the [move to PNPM](https://github.com/Chocobozzz/PeerTube/commit/906b5f7f2cf10b588c74a8d9939dcc19b28bec6c) unblock this PR?

Chocobozzz:

PR #6562, opened 2024-08-30
oEmbed API should use CORS

singingwolfboy:
## Description

It's handy to be able to access the [oEmbed API](https://oembed.com) from a browser, in order to construct `<iframe>` tags to embed videos. However, web browsers have security policies that prevent cross-domain communication -- unless that security policy is specifically bypassed using a [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) header. In this case, the oEmbed API returns no personal data about the logged-in user, so there is no reason to prevent cross-domain communication.

By adding this header, it allows third-party websites to construct `<iframe>` embeds directly in client-side JavaScript, without needing a backend server to act as a proxy.

## Has this been tested?

### DIFF
--- a/packages/tests/src/api/check-params/services.ts
+++ b/packages/tests/src/api/check-params/services.ts
@@ -201,6 +201,9 @@ function checkParamEmbed (
   return makeGetRequest({
     url: server.url,
     path,
+    headers: {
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Thanks!

Issue #3166, opened 2020-09-20
Show number of comments on a video on thumbnail


TILvids:
I'm trying to track comments on my instance, and it's really hard to do. One thing that might be nice is to expose the number of comments a given video has at a higher level, such as on the thumbnail for different views (Recent, Trending, etc.) I made a mockup to show what that might look like (you can see this on the "The History of Mozilla Firefox" thumbnail).

![peertube-comments-example](https://user-images.githubusercontent.com/69435791/93714197-14d8ab00-fb2f-11ea-9e6c-eb80b7af8321.png)

Thanks for considering!

Chocobozzz:
Hello,

https://github.com/Chocobozzz/PeerTube/issues/492 and https://github.com/Chocobozzz/PeerTube/issues/2736 should address your issue.

PR #6635, opened 2024-09-27
Add Comment Count to Video Preview Components

### LINKED ISSUE #492: Show number of comments in "my videos"
* **PeerTube version or URL:** 1.0.0.beta-3

Only the age of the video, the number of views and the author are visible in the video lists (trending, recently added or local and others videos on the watching view).
It would be interesting to also have the number of comments.

RF9A5V:
## Description

For backend changes, I added a migration that adds a commentCount column to the video table.

The migration:

1. Creates a column for commentCount on the "video" table that is at first nullable
2. Batches all videos in groups of 1000 to populate the actual values of comment counts for existing data
3. Sets default value of commentCount to 0 for future videos after population
4. Enforces presence of commentCount after population

From there, I've also updated the video comment model to add triggers for events that impact visibility of comments on a video, with those being:

- Creating a comment
- Deleting a comment
- When a comment becomes held for review

At this time, I think I've missed the case when we ban/unban or delete a user, will try to find remaining cases if the approach seems sound so far.

From there, also added the commentCount field to the video model and made the field visible by adding it to relevant formatter definitions.

Front end changes are less significant, just some setup to allow the commentCount to come through from the API, then the UI changes to display the value.

Also added commentCount to the displayOptions, but tbh I still need to dig further into that. Still not sure exactly how this bit works yet, but it seems reasonable to have the commentCount as an optional display, would just need to trace the behavior and implement.

## Related issues

Resolves #492 

Implements UI based on screenshot provided from #3166 

## Has this been tested?

Some manual testing at this time does show the comment count coming through, but still need to implement unit tests.

## Screenshots

Screenshot from 2024-09-27 12-33-54


Chocobozzz:
Hi,

What's the status of this PR? :)

RF9A5V:
@Chocobozzz 

Sorry about the delay, updated with latest and added some basic tests.

Chocobozzz:
Is it ready to review? :)

RF9A5V:
@Chocobozzz 

Ah, it's ready for review, let me know if anything looks weird

Chocobozzz:
Thank you @RF9A5V!

I added changes in the PR to merge develop that includes a new panel to list user's videos. I preferred to inject comments count in this table (hidden by default). I also added the possibility to sort by comments count, simplified the migration script and used a more robust way to recalculate total comments of a video when a comment is created/updated/deleted.

PR #6648, opened 2024-10-01
feat(API): permissive email check in login, reset & verification

### LINKED ISSUE #6570: Use case insensitive email for reset password

We get a lot of emails from users that says that they've requested a password reset, but they don't receive any email. When I search for their email, user id and username in the logs and audit log I can't find anything.

Relevant config:

```
log:
  level: 'info' # 'debug' | 'info' | 'warn' | 'error'

smtp:
  # smtp or sendmail
  transport: smtp
  # Path to sendmail command. Required if you use sendmail transport
  sendmail: null
  hostname: 'smtp.eu.mailgun.org'
  port: 587 # If you use StartTLS: 587
```

Whilst I know it works for other users. Possible causes:
* Handling errors by the users / bad UX
* Peertube fails to create the email job
* Mailgun issue

To investigate this further we may need to add more logging to determine what's going wrong. As far as I can see there's no audit logs for password resets, can we add that? Also we could add some info log here to tell that an email is being prepared. https://github.com/Chocobozzz/PeerTube/blob/a06c2ea020a353cdf773708832d3fa7c20f47999/server/core/lib/job-queue/job-queue.ts#L355

### Steps to reproduce

1.
2.
3.

_No response_

### Additional information

* PeerTube instance:
  * URL:
  * Version: 6.2.1
  * NodeJS version:
  * Ffmpeg version:

kontrollanten:
## Description
In order to not force users to be case sensitive when asking for password reset or resend email verification. When there's multiple emails where the only difference in the local is the capitalized letters, in those cases the users has to be case sensitive.

- [x] When a user tries to login: fetch all emails in the database case insensitive. If there's only one match, then treat it as the correct email even if they differ in capitalized letters.
- [x] When a user tries to reset password: fetch all emails in the database case insensitive. If there's only one match, then treat it as the correct email even if they differ in capitalized letters.
- [x] When a user tries to get a new verification email: fetch all emails in the database case insensitive. If there's only one match, then treat it as the correct email even if they differ in capitalized letters.

## Related issues
closes #6570

## Has this been tested?

### DIFF
--- a/packages/tests/src/api/users/oauth.ts
+++ b/packages/tests/src/api/users/oauth.ts
@@ -28,6 +28,8 @@ describe('Test oauth', function () {
     })
 
     await setAccessTokensToServers([ server ])
+    await server.users.create({ username: 'user1', email: 'user@example.com' })
+    await server.users.create({ username: 'user2', email: 'User@example.com', password: 'AdvancedPassword' })
 
     sqlCommand = new SQLCommand(server)
   })
@@ -79,7 +81,7 @@ describe('Test oauth', function () {
     })
 
     it('Should not login with an invalid password', async function () {
-      const user = { username: server.store.user.username, password: 'mew_three' }
+      const user = { username: 'User@example.com', password: 'password' }
       const body = await server.login.login({ user, expectedStatus: HttpStatusCode.BAD_REQUEST_400 })
 
       expectInvalidCredentials(body)
@@ -87,6 +89,9 @@ describe('Test oauth', function () {
 
     it('Should be able to login', async function () {
       await server.login.login({ expectedStatus: HttpStatusCode.OK_200 })
+
+      const user = { username: 'User@example.com', password: 'AdvancedPassword' }
+      await server.login.login({ user, expectedStatus: HttpStatusCode.OK_200 })
     })
 
     it('Should be able to login with an insensitive username', async function () {
@@ -98,6 +103,14 @@ describe('Test oauth', function () {
 
       const user3 = { username: 'ROOt', password: server.store.user.password }
       await server.login.login({ user: user3, expectedStatus: HttpStatusCode.OK_200 })
+    })
+
+    it('Should be able to login with an insensitive email when no similar emails exist', async function () {
+      const user = { username: 'ADMIN' + server.internalServerNumber + '@example.com', password: server.store.user.password }
+      await server.login.login({ user, expectedStatus: HttpStatusCode.OK_200 })
+
+      const user2 = { username: 'admin' + server.internalServerNumber + '@example.com', password: server.store.user.password }
+      await server.login.login({ user: user2, expectedStatus: HttpStatusCode.OK_200 })
     })
   })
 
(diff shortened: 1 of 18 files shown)


packages/tests/src/api/users/oauth.ts
```diff
@@ -79,14 +81,17 @@ describe('Test oauth', function () {
     })
 
     it('Should not login with an invalid password', async function () {
-      const user = { username: server.store.user.username, password: 'mew_three' }
+      const user = { username: 'User@example.com', password: 'password' }
```
Chocobozzz, line 84: Please duplicate the test instead of replacing it (you can for example use multiple login in this test)

packages/tests/src/api/users/oauth.ts
```diff
@@ -99,6 +104,14 @@ describe('Test oauth', function () {
       const user3 = { username: 'ROOt', password: server.store.user.password }
       await server.login.login({ user: user3, expectedStatus: HttpStatusCode.OK_200 })
     })
+
+    it('Should be able to login with an insensitive email when no similar emails exist', async function () {
```
Chocobozzz, line 108: Can you also do an offensive test where you test you cannot login if there are multiple email (insensitive cases)

server/core/lib/auth/oauth-model.ts
```diff
@@ -119,7 +119,14 @@ async function getUser (usernameOrEmail?: string, password?: string, bypassLogin
 
   logger.debug('Getting User (username/email: ' + usernameOrEmail + ', password: ******).')
 
-  const user = await UserModel.loadByUsernameOrEmail(usernameOrEmail)
+  const users = await UserModel.loadByUsernameOrEmailCaseInsensitive(usernameOrEmail)
+  let user: MUserDefault
+
+  if (usernameOrEmail.includes('@')) {
+    user = getUserByEmailPermissive(users, usernameOrEmail)
+  } else {
```
Chocobozzz, line 127: `else if (users.length === 1) {` is cleaner (`users[0]` may not exist)

server/core/models/user/user.ts
```diff
@@ -673,6 +673,18 @@ export class UserModel extends SequelizeModel<UserModel> {
     return UserModel.findOne(query)
   }
 
+  static loadByEmailCaseInsensitive (email: string): Promise<MUserDefault[]> {
```
Chocobozzz, line 676: I think you can remove `loadByEmail` method

server/core/models/user/user.ts
```diff
@@ -673,6 +673,18 @@ export class UserModel extends SequelizeModel<UserModel> {
...
+        fn('LOWER', col('email')),
+        '=',
+        email.toLowerCase()
+      )
+    }
+
+    return UserModel.findAll(query)
+  }
+
   static loadByUsernameOrEmail (username: string, email?: string): Promise<MUserDefault> {
```
Chocobozzz, line 688: Remove this method and replace the existing call by `loadByUsernameOrEmailCaseInsensitive`. Starting from now we want to prevent username/email duplicates in the database

Chocobozzz (changes requested):
Thanks for the PR! Please add verification email test for user registration too

kontrollanten:
@Chocobozzz Your comments are resolved. But now some of the new tests fails since the email exists check is case insensitive.

Chocobozzz:
Thanks!

PR #6667, opened 2024-10-09
Removed old categories, added new, set server config with enumerated …

CaptTofu:
…values for the new

## Has this been tested?


Chocobozzz:
Seems like a PR sent by mistake, we had no discussion about these new categories and we can't update existing category ids like that

PR #6692, opened 2024-10-25
Plugin support for sophisticated spam handling

### LINKED ISSUE #6264: Plugin support for adding action links to users and comments

Spam protection plugins, like the akismet plugin, should have the possibility to mark existing users and comments as spam.

Possibility to add actions in the following places:
* user list
* comments list
* abuse list
* video comments
* account page

kontrollanten:
## Description
Add support for plugins to filter action menus on:
* Admin user list, bulk and single
* Admin comment list, bulk and single
* Admin abuse list, single
* Admin user moderation (available in multiple views)

Add support for `doAction` plugin API. Supported actions:
* Increment/decrement loader
* Admin user list, load data
* Admin video comment list, load data
* Admin user list, load data
* Admin abuse list, load data
* Video watch comment list, load data

Proof of concept with a plugin available at https://framagit.org/kontrollanten/official-plugins/-/merge_requests/new/diffs?merge_request%5Bsource_branch%5D=feat-akismet-mark-as-spam

## Related issues
closes #6264 

## Has this been tested?

## Screenshots
image
image
image
image
image
image
image


client/src/app/core/plugins/plugin.service.ts
```diff
@@ -184,6 +198,14 @@ export class PluginService implements ClientHook {
     return firstValueFrom(obs)
   }
 
+  private doAction (actionName: ClientDoActionName) {
+    try {
+      return this.actions[actionName]()
+    } catch (err: any) {
+      logger.warn(`Plugin tried to do unknown action: ${actionName}`)
```
Chocobozzz, line 205: Action may also crash :)

```
logger.warn(`Cannot run action ${actionName}`, err)
```

packages/models/src/plugins/client/client-hook.model.ts
```diff
@@ -97,7 +99,22 @@ export const clientFilterHookObject = {
   'filter:internal.player.videojs.options.result': true,
 
   // Filter p2p media loader options built for PeerTube player
-  'filter:internal.player.p2p-media-loader.options.result': true
+  'filter:internal.player.p2p-media-loader.options.result': true,
+
+  // Filter bulk actions in user list
+  'filter:admin-user-list.bulk-actions.create.result': true,
```
Chocobozzz, line 105: `filter:admin-user-list.bulk-actions.create.result` for consistency (`s` to `user`). Please also do the same with `video-comment` -> `video-comments`

packages/models/src/plugins/client/client-hook.model.ts
```diff
@@ -193,3 +210,17 @@ export type ClientHookName = keyof typeof clientHookObject
...
+  'application:increment-loader': true,
+  'application:decrement-loader': true,
+  'admin-user-list:load-data': true,
+  'admin-video-comment-list:load-data': true,
+  'video-watch-comment-list:load-data': true
+}
+
+export type ClientDoActionName = keyof typeof clientDoActionObject
+
+export type ClientDoActionCallback = () => Observable<any>
```
Chocobozzz, line 225: Replace `Observable` by a `Promise` here. We don't want to force to use `Observable`, and we can easily convert a `Promise` to an `Observable` in the client

client/src/app/+admin/overview/users/user-list/user-list.component.ts
```diff
@@ -133,10 +145,12 @@ export class UserListComponent extends RestTable <User> implements OnInit {
     this.saveSelectedColumns()
   }
 
-  ngOnInit () {
+  async ngOnInit () {
     this.initialize()
 
-    this.bulkActions = [
+    this.pluginService.addAction('admin-user-list:load-data', this.reloadDataInternal.bind(this))
```
Chocobozzz, line 151: So we don't accidentally break plugin API when we update `reloadDataInternal` parameters below, prefer to create a dedicated callback. For example

```ts
this.pluginService.addAction('admin-user-list:load-data', (options: { ... }) => this.reloadDataInternal({ ... }))
```

client/src/app/+admin/overview/users/user-list/user-list.component.ts
```diff
@@ -326,11 +346,14 @@ export class UserListComponent extends RestTable <User> implements OnInit {
...
 
   protected reloadDataInternal () {
-    this.userAdminService.getUsers({
+    const obs = this.userAdminService.getUsers({
       pagination: this.pagination,
       sort: this.sort,
       search: this.search
-    }).subscribe({
+    })
+    .pipe(shortCacheObservable())
```
Chocobozzz, line 354: Is the purpose of `shortCacheObservable` to cache too many calls of `admin-user-list:load-data` in a short period of time)? If yes, prefer to move this to the `addAction` because we don't want the root component to have this side effect, especially since you don't use a cache based on the `getUsers` options.

client/src/app/+videos/+video-watch/shared/comment/video-comments.component.ts
```diff
@@ -145,11 +150,14 @@ export class VideoCommentsComponent implements OnInit, OnChanges, OnDestroy {
...
 
-  loadMoreThreads () {
+  loadMoreThreads (reset = false) {
     const params = {
       videoId: this.video.uuid,
       videoPassword: this.videoPassword,
-      componentPagination: this.componentPagination,
+      componentPagination: {
+        ...this.componentPagination,
+        currentPage: reset === true ? 1 : this.componentPagination.currentPage
```
Chocobozzz, line 159: I think that if `reset === true` you must do `this.componentPagination.currentPage = 1` or you'll loose pagination in future calls

client/src/app/shared/shared-moderation/user-moderation-dropdown.component.ts
```diff
@@ -271,20 +273,20 @@ export class UserModerationDropdownComponent implements OnInit, OnChanges {
...
 
     const myAccountModerationActions = this.buildMyAccountModerationActions()
     const instanceModerationActions = this.buildInstanceModerationActions()
 
-    if (myAccountModerationActions.length !== 0) this.userActions.push(myAccountModerationActions)
-    if (instanceModerationActions.length !== 0) this.userActions.push(instanceModerationActions)
+    if (myAccountModerationActions.length !== 0) userActions.push(myAccountModerationActions)
+    if (instanceModerationActions.length !== 0) userActions.push(instanceModerationActions)
+
+    this.userActions = await this.hooks.wrapObject(userActions, 'admin-users', 'filter:admin-user-moderation.actions.create.result')
```
Chocobozzz, line 289: The user moderation dropdown can be called in many pages. So I think you can create a `moderation` scope and use `filter:user-moderation.actions.create.result` as hook name

client/src/app/shared/shared-video-comment/video-comment-list-admin-owner.component.ts
```diff
@@ -113,6 +122,11 @@ export class VideoCommentListAdminOwnerComponent extends RestTable <VideoComment
         iconName: 'tick'
       }
     ]
+    this.bulkActions = await this.hooks.wrapObject(
+      bulkActions,
+      'admin-comments',
```
Chocobozzz, line 127: If you only want admins to use this hook, then check `this.mode === 'admin'`

Chocobozzz (changes requested):
Thank you very much for the PR and sorry for the late review.

Can you also update the plugin documentation? https://github.com/Chocobozzz/PeerTube/blob/develop/support/doc/plugins/guide.md and https://framagit.org/framasoft/peertube/documentation/-/blob/master/docs/api/plugins.md?ref_type=heads

client/src/app/+admin/overview/users/user-list/user-list.component.ts
Chocobozzz: Oh sorry I didn't see you could not pass params. But is is a `load-data` or a `reload-data` action then?

client/src/app/+admin/overview/users/user-list/user-list.component.ts
Chocobozzz: I suggest to remove this cache to avoid side effects, and if it's necessary throttle or debounce the method call

client/src/app/+admin/overview/users/user-list/user-list.component.ts
kontrollanten: When a plugin calls the action it will reload the data (reset). In one way `load-data` may be better to keep consistency if we'd like to add parameters to give plugins ability to pass query params to filter the data. On the other hand `reload-data` is more clearly for the plugin developer.

Chocobozzz: I'll let you decide the implementation you want to develop. Both solutions are okay for me :)

kontrollanten: Lets keep load-data to be able to have it flexible.

client/src/app/+admin/overview/users/user-list/user-list.component.ts
kontrollanten: The cache was due to multiple API calls made since we had multiple subscribes. I fixed this in the latest commit.

kontrollanten:
@Chocobozzz All comments fixed. The failing tests seem to be the same as in develop branch.

kontrollanten:
I did a last (?) update with changing action `admin-user-list:load-data` to `admin-users-list:load-data` and also updated the docs. I tried to update the dedicated API docs but got this error when I referenced to the new file `client-action.model.ts`:

image

Chocobozzz:
Thanks! I added some commits for styling and removed the cache observable because I don't think it worked (we create a new observable everytime the method is called so I'm not sure piping it to a replay factory would work).

If you notice issues with the main branch please don't hesitate to create an issue!

Issue #6167, opened 2024-01-13
Allow instance admin to set autoplay off by default


FediVideos:
### Describe the problem to be solved

I would rather have autoplay off on my instance, I find it annoying on other sites when videos play automatically.

At the moment if someone isn't logged in and they visit my instance on a laptop, videos play automatically. As far as I can tell there is no setting that allows the instance admin to switch off autoplay for people who aren't logged in?

### Describe the solution you would like

There should be a setting on the admin's GUI controls which allows admins to optionally set autoplay off by default.

EricGuic:
Good idea, and if I'm not wrong, this settings is also required for complying with good practices for accessibility for people with disabilities.

PARVNEMA:
i think can switch the autoplay  through javascript and videotag (if used)

Khyvodul:
Hi,

I'm interested in working on this new feature request.

I'm wondering where is the best place to put this new settings in the admin's GUI:
Should it be under _Configuration > Basic > Videos_ or would you prefer it somewhere else ?

Also, should it be a
- positive setting: "Enable automatic video playback start by default", which would be set to true by default (when no user preference) to keep the behavior consistent with the current situation without this new setting
- negative setting: "Disable automatic video playback start by default", which would then be set to false by default to keep the behavior consistent with the current situation without this new setting ?

Thanks

Chocobozzz:
Prefer to put the setting in "New user" section, because this is just the "default" setting when the user has not changed it.

Also prefer the "Enable..." form when using checkboxes

Khyvodul:
Ok noted. I'll make the changes accordingly.

Thanks for the feedback

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/pull/6788

PR #6788, opened 2024-12-23
feat: 6167 Add autoplay setting in the admin settings section

Khyvodul:
## Description

Add section in the admin's GUI to enable or disable the video autoplay for new or anonymous users

The autoplay feature is kept enabled by default to ensure consistency when upgrading from previous versions.

As already existing, user can still set autoplay on or off in its own preferences, which will override the instance's value.

## Related issues

Allow instance admin to set autoplay off by default
https://github.com/Chocobozzz/PeerTube/issues/6167

## Has this been tested?

## Screenshots

screen_capture


Chocobozzz:
HI and thanks! Please revert your changes in `clietn/src/locale`, they are automatically generated in `develop` branch :)

Khyvodul:
Hi,
I was wondering if I should include these changes or not (hence the separate commit just for them), now I have my answer :-).
I've reverted the commit and removed all changes to _client/src/locale/_.

Please let me know if more changes are needed.
Thanks

Chocobozzz:
Thanks!

I just moved the setting to the `defaults` section in the config. Since this section cannot be updated in the admin config (yet), I removed this part from the PR.

A nice addition in the future would to add all `defaults` config in the admin client :)

Issue #2424, opened 2020-01-23
Docker should respect user provided PUID and PGID


ghost:
The Docker image does not respect user provided PUID and  PGID environment values (preferring to create as 999) which causes PeerTube to endlessly reboot when it can't access its own config folder. Please see any LinuxServer image for how to do this right.

rigelk:
@Spass1966 have you tried running the container with the `--user` option? Our entrypoint explicitly supports it.

PR #6809, opened 2025-01-04
Add the ability to specify peertube UID and GID via environment variables

zbirkenbuel:
## Description

This adds support for specifying the peertube user and group IDs using the standard environment variable pattern similar to linuxserver.io images.  This allows for better external permission management, specifically this makes it easier to support mapping volumes to NFS shares using nfs4 id mapping.

This changes elects *not* to use the linuxserver.io variable names of PUID and PGID since the peertube container ends up having both a "node" and "peertube" user created and it would not be obvious which of those PUID/PGID should apply to.

New variables are `PEERTUBE_UID` and `PEERTUBE_GID` respectively.

## Related issues

[Existing feature request that was closed due to inactivity
](https://github.com/Chocobozzz/PeerTube/issues/2424)

## Has this been tested?

### DIFF
--- a/support/docker/production/Dockerfile.bookworm
+++ b/support/docker/production/Dockerfile.bookworm
@@ -8,9 +8,18 @@ RUN apt update \
  && gosu nobody true \
  && rm /var/lib/apt/lists/* -fR
 
+# Node images hardcode the node uid to 1000 so that number is not available.
+# The "peertube" user is created as a system account which selects a UID from 
+# the range of SYS_UID_MIN to SYS_UID_MAX (-1 to 1000] and consistently
+# selects 999 given the current image build steps. The same is true for the
+# system group range SYS_GID_MIN and SYS_GID_MAX. It is fine to manually assign
+# them an ID outside of that range.
+DEFAULT_PEERTUBE_UID=999
+DEFAULT_PEERTUBE_GID=999
+
 # Add peertube user
-RUN groupadd -r peertube \
-    && useradd -r -g peertube -m peertube
+RUN groupadd -r -g ${PEERTUBE_GID:-${DEFAULT_PEERTUBE_GID}} peertube \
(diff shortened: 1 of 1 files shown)


zbirkenbuel:
Just commenting that the suggestion to use the `--user` argument mentioned in https://github.com/Chocobozzz/PeerTube/issues/2424 doesn't solve the NFS permission issue since it only changes the user that the container "root" user runs as.  Since the container image manually creates a peertube user account that argument is effectively ignored.

Chocobozzz:
Thanks @zbirkenbuel :pray: 

Where do you think we should document this feature? `docker-compose.yml`? Or https://docs.joinpeertube.org/install/docker?

normen:
This only changes the GID when the docker container is built, right? So theres no point in documenting this on the user side, is there?

I still have to do this in the docker hub container when I want peertube to run as a different user:

```
usermod -u 10000 peertube
groupmod -g 1004 peertube
```

zbirkenbuel:
It should be evaluated at runtime.  Looking at the image on docker hub (https://hub.docker.com/layers/chocobozzz/peertube/production-bookworm/images/sha256-414fbff0adcc1ab369fb1b0630f5943834cad2e1ac1031e38c189cc94deb37e6) step 14 includes reading the environment variable when executing the useradd commands.  basically you should be able to define those values in your environment (or you docker compose file) and not have to execute the usermod and groupmod commands anymore.

normen:
> It should be evaluated at runtime. Looking at the image on docker hub (https://hub.docker.com/layers/chocobozzz/peertube/production-bookworm/images/sha256-414fbff0adcc1ab369fb1b0630f5943834cad2e1ac1031e38c189cc94deb37e6) step 14 includes reading the environment variable when executing the useradd commands. basically you should be able to define those values in your environment (or you docker compose file) and not have to execute the usermod and groupmod commands anymore.

Hm, okay, doesn't seem to work that way for me. They stay at 999/999

Issue #6858, opened 2025-01-27
Possibility to login via one-time authorization code (OTAC)


kontrollanten:
### Describe the problem to be solved

Currently users has to login via email and static password. Static passwords are hard to remember and is vulnerable for different kind of attacks.

### Describe the solution you would like

Provide API and UX for plugins to implement login via OTAC. Example design:

<img src="https://github.com/user-attachments/assets/226b7ae6-3b89-4ea4-9614-00b6eb6d467f" width="300"/><img src="https://github.com/user-attachments/assets/ad0f3a45-8669-49c5-96ec-6278fb443d6d" width="300"/>

This flow can be worked with regular login as well; in the first step only the email field is shown and in the next step the password field is shown.

What needs to be done in order to get plugins ability to implement this proposal:
- [ ] Divide the login flow into two steps.
- [ ] API for plugins to disable/hide the static password login and its settings (i.e. hide password from account settings).
- [ ] API for plugins to define a client login route.
- [ ] API for plugins to send emails. Not required, but would be nice to have.

kontrollanten:
@Chocobozzz Would you prefer this functionality as a plugin or into PeerTube core?

PR #6870, opened 2025-02-01
WIP: Divide login flow into two steps

kontrollanten:
## Description
Divide the login flow into two steps in order to:

* Let plugins jack into the second step #6858 
* Make it possible to make this as a part of the register flow, like https://www.disneyplus.com/identity/login/enter-email When OTAC is used an email can be sent to the user asking to create an account, if an account not already exists.

### Todo
- [ ] Add back button.
- [ ] Show entered username/email in the second step.

## Related issues
related to #6858

## Has this been tested?

## Screenshots

image
image


Chocobozzz:
Hi and thank you for this PR. But I prefer to implement 2 login steps when we'll implement OTAC because dividing the form in 2 steps introduces UX friction. So let's add it when we'll have the associated feature ready. A config to disable OTAC and just use a classic form would be also interesting for admins.

PR #6871, opened 2025-02-03
add user agent video stats

### LINKED ISSUE #6832: Add user-agent / device to video stats

Currently there's good stats about videos, but it would be interesting to group the viewers also by client type; mobile, tablet, browser and mobile app.

Store the user-agent for each viewer in order to be able to group them in the stats view. It would also be good to add a plugin filter to let plugins add custom grouping.

kontrollanten:
## Description
Add video stats dividing viewers into:

* Device family
* OS family
* Browser family

Todo:
- [x] Add tests
- [x] Add hook to parse user agent strings 

Since the current used [useragent](https://github.com/3rd-Eden/useragent) lib couldn't distinguish between Edge and Chrome I used [ua-parser-js](https://uaparser.dev/) instead.

## Related issues
closes #6832

## Has this been tested?

## Screenshots
image
image
image

### DIFF
--- a/client/src/app/+videos-publish-manage/shared-manage/stats/video-stats.component.scss
+++ b/client/src/app/+videos-publish-manage/shared-manage/stats/video-stats.component.scss
@@ -94,6 +94,7 @@ my-embed {
 
 .chart-container {
   margin-bottom: 10px;
+  min-height: 300px;
 }
 
 .zoom-container {
--- a/client/src/app/+videos-publish-manage/shared-manage/stats/video-stats.component.ts
+++ b/client/src/app/+videos-publish-manage/shared-manage/stats/video-stats.component.ts
@@ -31,7 +31,7 @@ import { VideoEdit } from '../common/video-edit.model'
 import { VideoManageController } from '../video-manage-controller.service'
 import { VideoStatsService } from './video-stats.service'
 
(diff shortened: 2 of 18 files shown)


Chocobozzz:
@kontrollanten sorry about the conflicts, but we had a redesign of the "manage video" pages planned. Can you rebase from develop? (stats pages were moved in `client/src/app/+videos-publish-manage/shared-manage/stats/video-stats.component.ts`)

kontrollanten:
@Chocobozzz Now it's rebased.

Chocobozzz:
Thanks, I added a commit:
 * We don't need to add index in the migration (they are automatically added by sequelize at startup)
 * I removed the indexes since the database can be big, and I think PG will only use `videoId`
 * I added start date/end date support to user agent endpoint

May I ask why you did not choose to add user agent information in overall stats endpoint?

And what would you think if the browser sent this information, instead of parsing the user agent on server side? (to reduce server loads, especially on this endpoint that can be under pressure).

kontrollanten:
> May I ask why you did not choose to add user agent information in overall stats endpoint?

I don't remember, but probably because either it seemed like a used pattern or to reduce unnecessary database queries.

> And what would you think if the browser sent this information, instead of parsing the user agent on server side? (to reduce server loads, especially on this endpoint that can be under pressure).

It's a good idea. I've added a commit with client side parsing (test isn't updated though).

Issue #6881, opened 2025-02-11
Plugins with a scope in the name can't be installed


pipoprods:
### Describe the current behavior

Installing a plugin that is scoped to an organization `@org-name/peertube-plugin-...` is not possible.

### Steps to reproduce

1. `peertube-cli plugins install -n "@worteks_com/peertube-plugin-cleanup-unviewed-videos"`

### Describe the expected behavior

The plugin should install.

PR #6882, opened 2025-02-11
fix: plugin/theme names with scope are not allowed

pipoprods:
## Description

This PR make validating plugin/theme names with scopes (`@org-name/peertube-plugin-...`) succeed.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/6881

## Has this been tested?

- [X] 🙋 no, because I need help as the name gets rejected by https://packages.joinpeertube.org afterwards

### DIFF
--- a/server/core/helpers/custom-validators/plugins.ts
+++ b/server/core/helpers/custom-validators/plugins.ts
@@ -6,6 +6,10 @@ import { exists, isArray, isSafePath } from './misc.js'
 
 const PLUGINS_CONSTRAINTS_FIELDS = CONSTRAINTS_FIELDS.PLUGINS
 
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thanks, do you have an example of such plugin so I can add it in the server tests?

pipoprods:
Sure: `@worteks_com/peertube-plugin-cleanup-unviewed-videos`

Chocobozzz:
Unfortunately I had to revert the commit because it doesn't work, PeerTube plugin manager expect a `peertube-plugin-` or `peertube-theme-` NPM name in many places. Supporting scoped package requires more work.

PR #6891, opened 2025-02-17
server: hide meta duration when duration is 0

kontrollanten:
## Description
Currently Google Search Console reports issue with `Value for property duration must be positive` for permanent lives where the duration value is `PT0S`. This PR will hide the duration property when the duration is falsy.


Chocobozzz:
Thanks

Issue #6836, opened 2025-01-16
Migrate from yarn to npm or pnpm


Chocobozzz:
### Describe the problem to be solved

yarn is not "useful" anymore as npm and pnpm have a lockfile now
We're stuck with yarn v1 that admins have to install as a global dependency. NixOS also has issues with yarn (see https://github.com/Chocobozzz/PeerTube/pull/6548#issuecomment-2593806097)

### Describe the solution you would like

We still need to investigate if we should move to npm or pnpm

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/commit/906b5f7f2cf10b588c74a8d9939dcc19b28bec6c

Very happy with the change so far, development UX is much easier and production install of node modules should be faster.
Dependencies documentation has been updated to install both `yarn` and `pnpm` (we'll remove `yarn` installation doc in a few months): https://docs.joinpeertube.org/support/doc/dependencies

PR #6898, opened 2025-02-18
replace yarn with pnpm

egasimus:
### what's done

* [x] all instances of `yarn.lock` replaced with top-level `pnpm-lock.yaml`
  * [x] `postinstall` in top-level package no longer needed
* [x] `workspaces` key in `package.json` replaced with `pnpm-workspace.yml`
* [x] changed how `ng` command is invoked
  * preserving max heap size flag
  * because `pnpm` installs shell wrappers in `node_modules/.bin` in place of js entrypoints
* [x] `dependencies` of each module specified using `workspace:*` instead of depending on top-level hoisting
  * [x] some deps were only referenced in `yarn.lock` (scary!) brought them out of there and into the appropriate `package.json`s
  * ⚠️ `workspace:*` version identifiers are pnpm-only
    * since packages are `"private": true` this doesnt matter in practice
    * publishing them to NPM (seamlessly and without the risk to bork local `node_modules`) may take some more specific version config but i assume for now it's a non-goal?

### what's left

* [ ] not sure how to provide `process` polyfill to client?
* [ ] some top-level deps may now be redundant and need to be removed from top-level `package.json`
* [ ] some workspace packages that i didn't see on my first pass may need `dependencies`/`devDependencies` provided
* [ ] some peer dependency warnings that i might try to resolve while i'm at it
* [ ] speculatively speaking, i'm curious to see whether the server can run on deno (feels like deno's built-in ts may tighten the feedback loop somewhat, and after 2.0 the node/npm compatibility is top notch)

## Related issues

* https://github.com/Chocobozzz/PeerTube/issues/6836
* https://github.com/Chocobozzz/PeerTube/pull/6548

## Has this been tested?

- [x] ⏳ **testing in progress.** no new tests need to be written.
  acceptance criterion: `pnpm i --frozen-lockfile && pnpm test` pass


Chocobozzz:
Hi and thanks for the issue,

Unfortunately we don't plan to move away from `yarn` before v8 (planned at the end of 2025) because it requires a system dependency change + an update of our https://github.com/Chocobozzz/PeerTube/blob/develop/server/scripts/upgrade.sh#L82 script. We also have to migrate plugin management system to use `pnpm` instead of `yarn`.

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/commit/906b5f7f2cf10b588c74a8d9939dcc19b28bec6c

Thank you for your PR @egasimus that helped me a lot to do the changes.

Issue #5699, opened 2023-03-19
Increase Object storage reliance


Kinuseka:
### Describe the problem to be solved

Peertube stores thumbnails, images, and redundancy files on the local file system, which can pose challenges for instances with limited storage resources due to the amount of space these files consume. Additionally, when managing a large instance, there can be hundreds to thousands of thumbnail files, making the migration process more time-consuming when transferring many small files (especially over ssh).

### Describe the solution you would like

To make it easier to migrate between servers, for example Mastodon stores all media files in a separate storage system, such as Amazon S3. As a result, when migrating an instance, only the configuration and database are moved, and the media storage can remain untouched. This simplifies the migration process, making it easier to move an instance to a new server without having to transfer large amounts of media files.

Chocobozzz:
@stevespaw Yes please!

MikeHuntington:
Hello! Just curious if this is still a priority for PeerTube? @Chocobozzz

Kinuseka:
Unfortunately due to lack of any attention for this issue, has anyone tried using s3fs-fuse as an alternative way for this issue?

If you did, what issues did you encounter?

fabiopicchi:
I haven't, but I'll try and see what happens.

PR #6905, opened 2025-02-27
WIP: store thumbnails on object storage

kontrollanten:
## Description
* Store thumbnails on object storage
* Switch from Jimp to [sharp](https://sharp.pixelplumbing.com/). The reason was primary because of typescript issues when handling the output during generation of thumbnails in different sizes (biggestImage). Sharp also seems easier and [should also be faster](https://sharp.pixelplumbing.com/performance).

@Chocobozzz Will you have time to review this anytime soon? Otherwise I won't go further with ending this PR.

### Known issues
* Fix and add tests
* Handle regenerate thumbnails CLI command

## Related issues
#5699 

## Has this been tested?


Chocobozzz:
> @Chocobozzz Will you have time to review this anytime soon? Otherwise I won't go further with ending this PR.

Unfortunately no in the following weeks, but I also wanted to refactor the way we manage thumbnails to mimic the avatar behaviours: having a `thumbnails` array of images with multiple width/height, that can be lazy loaded by remote servers. This particular work must be done before adding the possibility to move thumbnails to object storage.

kontrollanten:
Okay. Have you thought about moving the image processing to another service such as [imgproxy](https://imgproxy.net/) or [imagor](https://github.com/cshum/imagor)/[imagorvideo](https://github.com/cshum/imagorvideo)? Original/compressed images could be handled by Peertube, while the rest is handled by another service proxied by nginx.

Chocobozzz:
> Original/compressed images could be handled by Peertube, while the rest is handled by another service proxied by nginx.

What's the main purpose? Have the perfect image size on client size or reduce the compression/resize processing time for PeerTube?

kontrollanten:
> > Original/compressed images could be handled by Peertube, while the rest is handled by another service proxied by nginx.
> 
> What's the main purpose? Have the perfect image size on client side or reduce the compression/resize processing time for PeerTube?

Main purpose is to remove complexity from Peertube. Probably perfect image size and reduced processing time will come as a bonus.

Chocobozzz:
We don't want PeerTube to require an external service, so it won't reduce code complexity.

kontrollanten:
By external service, do you include services installed and hosted by the Peertube admin? The only difference from ffmpeg is the communication protocol.

Chocobozzz:
> By external service, do you include services installed and hosted by the Peertube admin? The only difference from ffmpeg is the communication protocol.

Unfortunately yes because every "hard" dependency is another step for admins to install PeerTube, and can increase PeerTube's maintenance burden. It's the reason why open-source & easy self installable projects have different specifications than private/internal projects. For example we chose Redis as key/value and pub/sub engine because it's well known and easy to install on any system. On a private project I might have chosen a different tool etc.

Chocobozzz:
Hello,

I re-architectured media files (including thumbnails) management in https://github.com/Chocobozzz/PeerTube/commit/526f1c6df529fa7620970c216f8915707b9f755e

This commit also moves from jimp to sharp. Unfortunately this PR is not compatible with the "new architecture" (too many conflicts), so I'm closing it. But the new consistent system to store thumbnails, avatars, captions, etc. paves the road to easily move these files in object storage in the future.

PR #6906, opened 2025-02-27
client: add nofollow to markdown links

kontrollanten:
## Description
Avoid negative SEO impact upon user input. I also removed duplicate rel values, don't know if it has any impact but it's wrong. :-)

## Has this been tested?


tcitworld:
I think `rel="ugc"` might be more appropriated: https://developers.google.com/search/docs/crawling-indexing/qualify-outbound-links?hl=en

Chocobozzz:
Thanks @tcit & @kontrollanten! I added the rel in https://github.com/Chocobozzz/PeerTube/commit/dd012bc22d204687d130e551c7b7e703e943e445

(also added to `HtmlRendererService` because we also support raw `<a>` tags in the markdown)

PR #6914, opened 2025-03-06
Fix header hit box

lil5:
## Description

Hitbox of logo + title `<a>` link is too large

This is how it is currently:
Screenshot 2025-03-06 at 10 07 03

This fix changes it to so:
Screenshot 2025-03-06 at 10 06 42

## Has this been tested?

- [x] 🙋 no, because I need help: I edited a live instance and saved the changes (for such a small fix I thought it wouldn't be necessary)


Chocobozzz:
Thanks!

PR #6945, opened 2025-03-24
feat: add user password constraints

### LINKED ISSUE #6777: Admin config to add user password constraints

_No response_

Minimum length for example.

serendipty01:
## Description

Added config to set the minimum and maximum password lengths for a new user signup, which is used through the below code
https://github.com/Chocobozzz/PeerTube/blob/56a28cb95eb0462435438bb131bf43fbe2d90de2/server/core/helpers/custom-validators/users.ts#L9-L11

However, that validation is only for server side checks, as for the client side, how do i pick from config ?
https://github.com/Chocobozzz/PeerTube/blob/56a28cb95eb0462435438bb131bf43fbe2d90de2/client/src/app/shared/form-validators/user-validators.ts#L73-L95

## Related issues

closes #6777 

## Has this been tested?

- [ ] 🙋 no, because I need help in client angular code


Chocobozzz:
Hi,

You need to inject your new config values in the config endpoint: https://github.com/Chocobozzz/PeerTube/blob/develop/server/core/controllers/api/config.ts#L36

Then use these values to customize the Angular validators

MadMan247:
Why set a max password length?

the-sinner:
@MadMan247 255 is more than enough for all use cases. Is there a need for more than that ?

According to [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#implement-proper-password-strength-controls), very long passwords might cause an DDOS attack on the server. 

> Maximum password length should be at least 64 characters to allow passphrases ([NIST SP800-63B](https://pages.nist.gov/800-63-3/sp800-63b.html)). Note that certain implementations of hashing algorithms may cause [long password denial of service](https://www.acunetix.com/vulnerabilities/web/long-password-denial-of-service/).

Relevant [Reddit](https://www.reddit.com/r/cybersecurity/comments/17fzq2h/is_there_any_good_reason_to_limit_maximum_length/) Discussion

Bitwarden password generator has a maximum length of 128 characters.

MadMan247:
I would (possibly incorrectly) assume that peertube already has a maximum password either through implementation or through limitation. My fear is that providing a config enables ignorant instance operators to limit the max password length to 16 characters.
Thank you for the explanation though. 255 is 100% more than enough.

the-sinner:
You are correct that, previously, PeerTube had a maximum limit of 255 which was hardcoded. Now, after the change, it is configurable. Though, the default is 255, until someone manually reduces it.
I’ll defer to @Chocobozzz  on the final decision, and if needed, I can revert the changes to the max password length and let it remain 255.

Chocobozzz:
Keep 255 for now :+1:

Chocobozzz:
Thank you!

PR #6963, opened 2025-04-03
feat (docker): Expose NGINX logs folder.

RiQuY:
This commit adds the folder /var/log/ngnix of the webserver container to the volume mount.

## Description

This commit adds the folder /var/log/ngnix of the webserver container to the volume mount. The motivation to add this change is to make it easier for the users to extract logs when asked.

## Has this been tested?


Chocobozzz:
I suggest to create a separate directory in root `nginx-logs` to not mix nginx site file with logs

RiQuY:
> I suggest to create a separate directory in `docker-volume`, `nginx-logs`, to not mix nginx site file with logs

Suggestion applied and tested on my deployment.

Chocobozzz:
Thanks!

Issue #6094, opened 2023-12-07
Add hint for drag and drop for replace video


kirill2400:
### Describe the problem to be solved

There is no intuitive hint for dnd in replace video form

### Describe the solution you would like

Add dotted border for hovered drag and drop for replace video stage.
Like this
![image](https://github.com/Chocobozzz/PeerTube/assets/30482699/1bd18434-4e08-48a6-944c-75377000f72e)
But there
![image](https://github.com/Chocobozzz/PeerTube/assets/30482699/f6873376-cdda-4a09-bb13-7efe2e4ad6f3)

kirill2400:
> Hi,
> 
> I'm sorry but I don't understand: you can drag and drop to upload a video but the "replace video file" doesn't support drag and drop 🤔

But I can use DnD for replace video. See video:
https://spacecomfort.miracleinsense.ru/w/sWo98ujB1QvNxLdEruAYev

allmiha2:
Hi, I am a new contributor to this project, can I be assigned this issue?

Chocobozzz:
Sure!

kirill2400:
@allmiha2 Hello. Regarding the request. When uploading a video to the site using drag and drop, I have intuitive dot border prompts. So I understand that js scripts of the site are not broken and DnD will work correctly. On the video change page I can upload a replacement video using DnD, but there is no hint if I hit the right zone. And if I am not in the right zone, instead of selecting the necessary video I will get a new tab with viewing my video from local file system. I hope I've explained the problem correctly. (I did this because I found that my video demo has already been deleted)

Translated with DeepL.com (free version)

allmiha2:
> [@allmiha2](https://github.com/allmiha2) Hello. Regarding the request. When uploading a video to the site using drag and drop, I have intuitive dot border prompts. So I understand that js scripts of the site are not broken and DnD will work correctly. On the video change page I can upload a replacement video using DnD, but there is no hint if I hit the right zone. And if I am not in the right zone, instead of selecting the necessary video I will get a new tab with viewing my video from local file system. I hope I've explained the problem correctly. (I did this because I found that my video demo has already been deleted)
> 
> Translated with DeepL.com (free version)

Hi! I was wondering about some details and your explanation covered them, thanks a lot!

Chocobozzz:
Implemented by https://github.com/Chocobozzz/PeerTube/pull/6970

PR #6970, opened 2025-04-06
Add Drag-and-Drop to Replace File

allmiha2:
Added the drag and drop functionality (that was already used for publishing videos) for replacing the video files

## Description

Now there is the possibility to drag a file from the file exporer to the "Replace file" area to replace a video (add a new version of the video), as well as upload that file using the button.

## Related issues

Solves issue "Add hint for drag and drop for replace video #6094".
Note: When dragging the file over other elements in the drop area, the DnD hint flickers, and if you happen to drop the file in between the flickers (when the hint does not appear), the file will not be selected. As far as I looked into that, that has to do with hovering over nested components within the drop area which messes with the onEnter and onLeave methods of the Drag and Drop directive. The exact same thing happens for the "Publish video" section, where the DnD directive is also used, therefore I think a separate issue should be opened to fix that problem with the DnD component.

## Has this been tested?

## Screenshots

image

### DIFF
--- a/client/src/app/+videos-publish-manage/shared-manage/common/video-manage-page-common.scss
+++ b/client/src/app/+videos-publish-manage/shared-manage/common/video-manage-page-common.scss
@@ -75,6 +75,10 @@ p-calendar {
   }
 }
 
+.dragover {
+    border: 3px dashed pvar(--primary);
+}
+
 .plugin-fields {
   display: flex;
   flex-wrap: wrap;
--- a/client/src/app/+videos-publish-manage/shared-manage/replace-file/video-replace-file.component.html
+++ b/client/src/app/+videos-publish-manage/shared-manage/replace-file/video-replace-file.component.html
@@ -1,39 +1,42 @@
-<h2>
-  <my-global-icon iconName="upload"></my-global-icon>
+<div myDragDrop (fileDropped)="onFileDropped($event)">
+	<h2>
+		<my-global-icon iconName="upload"></my-global-icon>
 
-  <ng-container i18n>Replace file</ng-container>
-</h2>
+		<ng-container i18n>Replace file</ng-container>
+	</h2>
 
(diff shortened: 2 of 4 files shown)


Chocobozzz:
May I suggest to create the same component as the thumbnail block?
screen_2025-04-11-06_58_10

So a rectangle with the name of the current file in bold, a small sentence "Current uploaded file", "Drag and drop your new video file here or click on the button below to replace your video. ⚠️ Uploading a new version of your video will completely erase the current version", an upload button.

allmiha2:
@Chocobozzz Sure, I can give it a go.

allmiha2:
@Chocobozzz quick question, when the user uploads a new and different version of the video but does not yet save, what should be in the "current uploaded file" section, the filename of the video that actually is uploaded or the name of the video to be the new version?

Chocobozzz:
> the filename of the video that actually is uploaded or the name of the video to be the new version?

You can keep the current one. But if we have the info in the drag and drop component (with the "Current file: ..." sentence) I think we can remove the text input

allmiha2:
> > the filename of the video that actually is uploaded or the name of the video to be the new version?
> 
> You can keep the current one. But if we have the info in the drag and drop component (with the "Current file: ..." sentence) I think we can remove the text input

so if i understood correctly, I should remove the text in bold that says what filename the video version that is currently uploaded (and not the potential new version) and just leave the drag and drop component part which says what filename the video that was just drag and dropped to be the new version has. is that accurate?

allmiha2:
I have worked a bit on it and this is how it currently looks:
image
image
image
@Chocobozzz let me know if it needs any corrections

Chocobozzz:
Perfect! Is the PR ready for a code review?

allmiha2:
@Chocobozzz Yes, it should be

Chocobozzz:
Thank you!

Issue #4678, opened 2021-12-31
Improve OpenRC script using start-stop-daemon


avdb13:
**Describe the current behavior**
I don't know whose idea it was to create an init script that doesn't just use `start-stop-daemon` since that would have made things a million times easier instead of reinventing the wheel.
Currently peertube just silently crashes on startup after starting it automatically on boot with `rc-update` and the biggest problem is that it doesn't report a crash so OpenRC can do something about it (which is the purpose of an init system after all).
I tried editing the existing script but it's hopeless.

**Steps to reproduce:**

1. Install Peertube on an operating system with OpenRC
2. Add it to the default runlevel
3. Find out it's actually not running on startup

**Describe the expected behavior**
`/etc/init.d/peertube` should use `start-stop-daemon` or just make sure it works on startup with the current init script.

**Additional information**
Nothing.

* PeerTube instance:
All not relevant.

* Browser name, version and platforms on which you could reproduce the bug: No.
* Link to browser console log if relevant: No.
* Link to server log if relevant (`journalctl` or `/var/www/peertube/storage/logs/`):
Nothing gets logged on startup.

Chocobozzz:
You can rewrite the script. Unfortunately I don't use/know openrc.

avdb13:
Good to hear. I will open a pull request once I did.

PR #6983, opened 2025-04-15
Modernize the OpenRC service with openrc-run and improve production guide

q0ntinuum:
## Description

This change uses openrc-run instead of a simple shell script which allows to mainly use a declarative configuration, manages dependencies, pid file, privileges drop and more. It also uses supervise-daemon as supervisor so that the service can be restarted when crashing.

The production guide now also contains instructions when updating on an OpenRC system.

## Related issues

#4678 (This PR doesn't use start-stop-daemon but supervise-daemon that brings actual supervision)

## Has this been tested?

### DIFF
--- a/support/conf.d/peertube
+++ b/support/conf.d/peertube
@@ -1,0 +1,1 @@
+directory="/var/www/peertube/peertube-latest"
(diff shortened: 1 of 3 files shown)


Chocobozzz:
Thanks!

PR #7009, opened 2025-05-08
client: dont use rel=ugc for plugins markdown

### LINKED ISSUE #6959: Add possibility to skip rel=ugc when plugin is rendering markdown

Since https://github.com/Chocobozzz/PeerTube/commit/dd012bc22d204687d130e551c7b7e703e943e445 `rel="ugc"` is added to user generated content, which is good. But currently this means that markdown provided from admins in plugin settings will get rel=ugc. Maybe other admin generated content is affected as well (about instance?).

Let plugins decide whether they want to skip the rel=ugc attribute when calling `peertubeHelpers.markdownRenderer.enhancedMarkdownToHTML`.

kontrollanten:
## Description
This PR should only affect markdown generated with custom home page markdown and plugins. Since those fields are only editable by admins they shouldn't be marked as user generated content. 

## Related issues
closes #6959

## Has this been tested?

### DIFF
--- a/client/src/app/core/plugins/plugin.service.ts
+++ b/client/src/app/core/plugins/plugin.service.ts
@@ -316,11 +316,11 @@ export class PluginService implements ClientHook {
 
       markdownRenderer: {
         textMarkdownToHTML: (textMarkdown: string) => {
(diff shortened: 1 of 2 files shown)


Chocobozzz:
Thanks!

PR #7034, opened 2025-05-17
add user 2fa status in admin page

### LINKED ISSUE #6997: Add 2FA column to user list

We'd like to easily see which users who has two factor authentication enabled in the admin user list.

Add a column showing the value of `user.twoFactorEnabled`.

the-sinner:
## Description

Add 2fa status

## Related issues

fixes #6997 

## Has this been tested?


the-sinner:
@kontrollanten should the 2FA status be shown by default ?

kontrollanten:
I personally think it's good behavior to show it per default, since it'll help the admin to keep the instance secure.

Chocobozzz:
Hi & thank you! This has been implemented in https://github.com/Chocobozzz/PeerTube/commit/37e13bbcd27730ebead63216060b2050bd8de96c :)

Issue #7104, opened 2025-06-18
Improve miniature thumbnails


letydoesstuff:
### Describe the current behavior

Video thumbnails are currently blurred with `filter: blur(20px);` when the sensitive content policy is set to “Blur”. This results in an unpleasant appearance with white borders around the blurred video edges. 

<img src="https://github.com/user-attachments/assets/05f68ab8-6706-44a2-9f87-d539e3bc18d6" height="270" />

As 20px is an absolute value, this also creates a varying blur effect depending on the exact size of the displayed thumbnail. On mobile view, where videos are displayed as a list instead of the desktop grid pattern, the thumbnail becomes so small that blurring it by 20 pixels creates a blob of so few colors that it no longer resembles the original thumbnail in any capacity.

<img src="https://github.com/user-attachments/assets/2290ade0-8d9f-4970-85b6-ba95e59af72e" height="270" />

Separately, the property `background-color: #ececec;` occasionally peaks through at the corner radii of video thumbnails at some resolutions, resulting in a strange visual glitch in the corner of some videos.

<img src="https://github.com/user-attachments/assets/c8ce409f-a6c0-48ef-9bb2-4aeb987d7303" height="270" />

### Steps to reproduce

1. These issues are most easily reproduced by [visiting this channel page at video.infosec.exchange](https://video.infosec.exchange/c/letydoesstuff_channel@peertube.doesstuff.social/videos), which is running stock PeerTube v7.2.1, and setting the sensitive content policy to “Blur”

### Describe the expected behavior

Video thumbnails should be blurred edge-to-edge with no white borders.

<img src="https://github.com/user-attachments/assets/c40c9d7d-fd90-4b78-9c8e-e031ab137b0f" height="270" />

Video thumbnails should be consistently blurred regardless of rendered size.

<img src="https://github.com/user-attachments/assets/2061233b-ab36-4dc5-a419-cf03846352af" height="270" />

Video thumbnails should never have gray-colored glitching at their rounded corners.

<img src="https://github.com/user-attachments/assets/43756f38-8ff7-48a5-b428-b274fb143c41" height="270" />

### Additional information

* PeerTube instance: video.infosec.exchange
  * URL: https://video.infosec.exchange
  * Version: 7.2.1
  * NodeJS version: Irrelevant
  * Ffmpeg version: Irrelevant

[...]

PR #7105, opened 2025-06-18
Use SVG filter for thumbnail blurs and fix thumbnail corner coloring

letydoesstuff:
## Description

Use SVG filter for thumbnail blurs and fix thumbnail corner coloring

The values for the SVG filter were derived from the current `filter: blur(20px);` property. I took the video height from the video miniatures on channel pages and rounded up to 160.

`0.125` comes from `20 ÷ 160`, and `0.07` comes from `20 ÷ ( 160 ÷ 9 × 16)`.

## Related issues

Chocobozzz/PeerTube#7104

## Has this been tested?

- [x] ❔yes, but only locally on my own server

### DIFF
--- a/client/src/sass/include/_miniature.scss
+++ b/client/src/sass/include/_miniature.scss
@@ -27,7 +27,7 @@
   width: 100%;
   height: 100%;
   overflow: hidden;
-  background-color: #ececec;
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thank you!

PR #7117, opened 2025-06-22
Fix settings menu overlapping with Picture-in-Picture button

### LINKED ISSUE #7114: Picture-in-Picture icon always takes the mouse input, even when a menu is shown on top of it

When you click an option in a menu that happens to be displayed on top of the PiP icon, the Picture-in-Picture mode activates and the menu option is not selected.

### Steps to reproduce

1. Click on the cog icon to open parameters
2. Select the speed option, as it's the biggest one : the menu should now be displayed over the PiP icon on the right of the video.
3. Click on the speed option that is right on top of the PiP icon : the PiP mode activates.

I think the menu option should be selected, and the PiP mode not activated. I'm not 100% sure, but i think that's what makes the most sense in UX.

### Additional information

I tested both in basic mode, and on integrated mode on other websites. I only used Firefox browser. I hope this is not a duplicate, sorry if it's the case !

Vikx001:
## Description
Resolves issue where clicking on settings menu options (especially speed settings) would trigger Picture-in-Picture mode instead of selecting the menu option.

## Changes Made
- Increase z-index of settings dialog to `2147483647` to ensure it appears above native browser controls
- Fix TypeScript compilation errors in settings menu button component following PeerTube patterns

## Testing
- Settings menu now properly appears above PiP button
- Menu options can be clicked without triggering PiP mode
- TypeScript compilation passes without errors
- Maintains existing code structure and patterns

## Files Changed
- `client/src/standalone/player/src/sass/shared/settings-menu.scss` - CSS z-index fix
- `client/src/standalone/player/src/shared/settings/settings-menu-button.ts` - TypeScript fixes

Fixes #7114

### DIFF
--- a/client/src/standalone/player/src/sass/shared/settings-menu.scss
+++ b/client/src/standalone/player/src/sass/shared/settings-menu.scss
@@ -23,7 +23,7 @@ $setting-transition-easing: ease-out;
     opacity: $primary-foreground-opacity;
     margin: 0 auto;
     font-size: $font-size !important;
-    z-index: 100;
+    z-index: 2147483647;
     border-radius: 10px;
 
     width: auto;
--- a/client/src/standalone/player/src/shared/settings/settings-dialog.ts
+++ b/client/src/standalone/player/src/shared/settings/settings-dialog.ts
@@ -57,7 +57,7 @@ class SettingsDialog extends Component {
     document.addEventListener('keydown', this.keydownHandler)
 
     setTimeout(() => {
-      this.el().focus()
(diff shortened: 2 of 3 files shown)


Chocobozzz:
Hello,

This PR doesn't resolve the bug. See https://bugzilla.mozilla.org/show_bug.cgi?id=1742585, https://bugzilla.mozilla.org/show_bug.cgi?id=1742585 and https://bugzilla.mozilla.org/show_bug.cgi?id=1754604

PR #7143, opened 2025-07-07
Fix scroll position not being preserved when pages load slowly

### LINKED ISSUE #7097: Back History bug

In My videos, when searching for a video (when 100 videos are set per page), when opening the video and returning back using the “back” button in the browser, the page scroll position is lost and you have to scroll down again and find where you left off.

### Steps to reproduce

1. Open My videos
2. Scroll down
3. Open video
4. Click "Back" button
5. You will be on top of the page

When you hit the back button, the scroll should go back to where it was when you switched pages, or at least scroll right in front of my eyes. One of these two options was used in the endless list before the redesign.

### Additional information

* PeerTube instance:
  * URL: https://spacecomfort.miracleinsense.ru
  * Version: 7.2.1

bojidar-bg:
## Description

Before this change, ScrollService attempted to scroll down immediately after navigating, before before the page had fully loaded.
I considered making the Table component (and other data-loading componets) responsible for telling the ScrollService that data has been loaded—or to use Angular's Route.resolve property, but realized that using ResizeObserver to get the page size is a lot more reliable (and a small code change overall, so less painful :grin:).

I'm pretty sure there is some way to use RxJS observables for the `unhandledScrollEvent` variable I added, but I couldn't figure out a good way to do so. (Perhaps `combineLatestAll` with something to mark an event as "handled", but that "something" has been a bit elusive.) Would appreciate any pointers in that regard.

Also, with the unhandled event approach, there is a chance that if only part of a page loads (or perhaps, if the page loads with a smaller size than the one it had when it was stored in history), the user might start interacting with the page, and trigger the unhandled event much later, in case something resizes the page to be tall enough.

## Related issues

Fixes #7097.

To reproduce the issue, I would:
1. Scroll down on the My Videos page.
2. Open a video's "Manage" button
3. Switch the connection in Firefox's Dev Tools / Network to Throttled (otherwise, reproducing on localhost is near impossible)
4. Press the back button on the browser

## Has this been tested?

### DIFF
--- a/client/src/app/core/routing/scroll.service.ts
+++ b/client/src/app/core/routing/scroll.service.ts
@@ -2,11 +2,13 @@ import { ViewportScroller } from '@angular/common'
 import { ApplicationRef, Injectable, inject } from '@angular/core'
 import { logger } from '@root-helpers/logger'
 import debug from 'debug'
-import { pairwise } from 'rxjs'
+import { pairwise, Observable } from 'rxjs'
 import { RouterSetting } from '../'
+import { Scroll } from '@angular/router'
 import { PeerTubeRouterService } from './peertube-router.service'
 
 const debugLogger = debug('peertube:main:ScrollService')
+
 
 @Injectable()
 export class ScrollService {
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Nice! Thank you!

bojidar-bg:
BTW, this PR was made during a [live stream](https://watch.bojidar-bg.dev/w/uHW7Rgkc9MsM3nRL8dbB73)—you can watch see the final reproduction and demonstration at [2:41:24](https://watch.bojidar-bg.dev/w/uHW7Rgkc9MsM3nRL8dbB73?start=2h41m24s) (:
~~Apologies for the delay with posting that link; uploading and transcoding the recording took forever :joy:~~

Issue #6604, opened 2024-09-21
Schedule live and schedule reminder posts 


christophehenry:
### Describe the problem to be solved

Peertube allows to create a permanent video for performing live. However, there's no ability to advertise scheduled live unless posting a comment under the video.

It would be nice to add features to live streams to schedule live and advertise them. 

### Describe the solution you would like

On top of my mind, a first solution could be to modify the stream settings to add a "schedule next live" input in the settings form amongst with a "schedule reminder" section to schedule posts to be posted at precise date and hour.

Chocobozzz:
Implemented by https://github.com/Chocobozzz/PeerTube/pull/7144

christophehenry:
Does it also implement some form of communication with followers like posting to their feed? If no, this only implements parts of the issue. Should I open another issue for that?

Maybe this issue would require the work of a UX so that scheduling a live allows to efficiently reach your community without being invasive.

For instance, if I schedule a live in a week and it produces a post in the follower's feed right now, they may have forgotten the event the day it happens so I may want the schedule to produce other posts later. But it shouldn't flood either.

Chocobozzz:
> Does it also implement some form of communication with followers like posting to their feed? 

No it doesn't

> Should I open another issue for that?

Yes please :)

PR #7144, opened 2025-07-07
Add Scheduled Lives functionality

bojidar-bg:
## Description

Implements scheduled live videos by reusing the `originallyPublishedAt` field of `isLive` videos to mark "waiting for live" videos as scheduled at a set time.
In order to differentiate between scheduled live videos that have already started and scheduled live videos that are yet to start, I made `VideosIdListQueryBuilder` return all `WAITING_FOR_LIVE` videos that have `originallyPublishedAt` set to a non-null value, and then changed `video-api-format`'s `guessAdditionalAttributesFromQuery` to always return the state of videos for `isLive: true` queries.

I believe this could also be implemented as a plugin; let me know if I should do that instead.

BTW, this PR was made during a [live stream](https://watch.bojidar-bg.dev/w/uHW7Rgkc9MsM3nRL8dbB73): you can watch the final demonstration of the feature at [6:11:45](https://watch.bojidar-bg.dev/w/uHW7Rgkc9MsM3nRL8dbB73?start=6h11m45s) :blush:

## Related issues

Implements #6604 

## Has this been tested?

## Screenshots

Multiple lives on a channel: (Current live is in the middle, because the future scheduled live was created while it was still going (and thus is counted "more recently added"))
image

Hovering over the "Scheduled live" label to see the scheduled-for date:
image

Customization page:
image


Chocobozzz:
Thank you!

I think we should store the scheduled live date in a dedicated feed, that you can put in the `videoLive` table. We can also add a REST API param to include scheduled lives so it's the client choice if it want to display them or not.

bojidar-bg:
I added the field; haven't added the REST api yet.

Would appreciate if you could look over the latest commit and confirm that I did not miss any place in which I should be adding the new field to.

Chocobozzz:
Thank you! I added commits to use a dedicated table to store live schedules so we can imagine add multiple dates with titles in the future.

bojidar-bg:
Oh. I just imagined one would make multiple non-recurrent lives with different scheduled dates for that usecase (:

bojidar-bg:
Oh wow, thanks for polishing up the PR! :green_heart: (I'll have to make sure to study the commits you did there! (: )

Chocobozzz:
> Oh. I just imagined one would make multiple non-recurrent lives with different scheduled dates for that usecase (:

I think streamers would prefer to use the same live stream URL for multiple live sessions. :) (I'm thinking about a FOSDEM room for example that can use a permanent live stream)

Issue #6468, opened 2024-07-05
Automatic transcription: a way for video publishers to choose to enable it or not, video per video


JohnXLivingston:
### Describe the problem to be solved

Peertube v6.2.0-RC1 comes with the new automatic transcription feature, to generate subtitles.

But some videos don't need subtitles (music show, video without people speaking, ...).
Or maybe, the publisher has subtitles that they will publish later (maybe just after the upload, maybe 2 or 3 days later).

In v6.2.0-RC1, transcription will be automatic for all videos, even if not needed/wanted.

### Describe the solution you would like

There should be an option in the video upload form, to choose if we want to generate automatic subtitles.
This option could be checked by default.

JohnXLivingston:
Related: https://ideas.joinpeertube.org/posts/272/make-it-possible-to-disable-a-videos-subtitles

PR #7161, opened 2025-07-17
[fix] remove useless transcription file

romtou:
## Description

The automatic transcription creation feature generates hallucinations: videos without subtitles end up with useless subtitles, and a notification indicates the creation of subtitles in a random language. To address this, we simply added the option to keep transcription files only if they exceed a certain size, so the majority of hallucinations are not stored.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/6468 

## Has this been tested?


romtou:
I am closing this merge request because I want to verify it internally first.

Issue #7181, opened 2025-08-10
Let external frontend get an API token for users using external auth


jakmeier:
### Describe the problem to be solved

Right now, a user from an external auth plugin, such as OIDC, cannot directly use the /token endpoint to receive an API token because that would require a PeerTube password.

With https://github.com/Chocobozzz/PeerTube/issues/6258 resolved, an API token for such users can be obtained. But it requires that the user's browser goes to the PeerTube frontend and initiates the external login from there.

For an external frontend as I am building it, I would like a way to get an API token through the API alone, without the need to go to PeerTube frontend.

At first I though I can solve this inside the auth plugin without PeerTube core changes. The plugin can set up redirectUrl with the bypass token in the query parameters and thus have PeerTube core answer with a "302 Found" response. This works perfectly well to make manual requests from CLI.

However, when a JS client in the browser performs this request, the browser will not let JS code inspect the full details of the redirection. All I get to see is a, `opaqueredirect` without query parameters or location headers. The automated redirect also does not work because a 302 redirection will be followed with a GET and not a POST by browsers.

### Describe the solution you would like

To get around these limitations, I propose that external auth plugins can decide to send the bypass token back as a JSON response, rather than redirecting. This way, a JS fetch can read the bypass token and use it to make the second request to the /token endpoint and get an access + refresh token pair for further API requests.

I'm open to other solutions, this is just the first simplest thing that I could come up with.

jakmeier:
I have a draft solution of this that allows my frontend to upload videos purely using the API.

The PeerTube core changes are here: https://github.com/Chocobozzz/PeerTube/pull/7182

The plugin changes are here: https://framagit.org/framasoft/peertube/official-plugins/-/merge_requests/33

And for completeness, a PoC of how I'm able to use it in my frontend, using a JWT from Keycloak and exchanging it for a PeerTube OAuth 2 token: https://github.com/jakmeier/bouncy-feet/blob/bdf1970f56ee3d4dc0708a346cd11f48b697c40c/bouncy_frontend/src/lib/peertube.js#L80-L166

I'm looking for feedback if this is a valid approach and if the PeerTube core changes can be upstreamed like this.

Also if you are interested to have the plugin changes to OIDC plugin changes upstreamed, or if that goes beyond what you want to be officially supported. (I can also use my own unofficial fork of the plugin for it, no problem.)

Chocobozzz:
[...]

PR #7182, opened 2025-08-10
API-access for external auth plugins

jakmeier:
This allows auth plugins to return the bypass token as a JSON response, rather than in a redirection.

Using a redirect and putting the bypass token in the query parameter works when the browser goes through the login flow using the PeerTube frontend and redirect back to an external frontend afterwards. But for a purely API based access from an external frontend, the redirect does not work.

See https://framagit.org/framasoft/peertube/official-plugins/-/merge_requests/33 for more details on why it does not work and how the proposed change can be used in a plugin.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/7181

## Has this been tested?

### DIFF
--- a/server/core/lib/auth/external-auth.ts
+++ b/server/core/lib/auth/external-auth.ts
@@ -45,7 +45,7 @@ async function onExternalUserAuthenticated (options: {
     return
   }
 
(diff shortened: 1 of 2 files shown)


Chocobozzz:

Issue #7212, opened 2025-09-11
Browse Videos config check does not report errors during server initialization


gianantoniopini:
### Describe the current behavior

If a custom config contains an invalid `DEFAULT_SORT` or `DEFAULT_SCOPE` in the `CLIENT.BROWSE_VIDEOS` section, the config check performed during server initialization does not report errors.

### Steps to reproduce

1. Enter an invalid value (e.g., "test") for the `default_sort` (or `default_scope`) key under the `client.browse_videos` section of the `config/local-dev.json` file and save the file
2. In the terminal, run `npm run dev:server` to start the server
3. The server starts and no errors are shown in the terminal

### Describe the expected behavior

The expected behavior is that the configuration consistency check (`checkConfig` function in the `checker-after-init.ts` file) performed during server initialization should report an error, and the process should be terminated.

After some initial investigation, the issue appears to be related to the use of the `i18n` `t` function in `server/core/helpers/custom-validators/browse-videos.ts`. Because `i18n` is still not initialized when the `checkConfig` function is called, calls to the `t` function result in `undefined` and then no error is thrown.
This is not an issue when the `server/core/helpers/custom-validators/browse-videos.ts` functions are called during a custom config update via the api, as `i18n` is initialized then.

As a simple fix, the `server/core/helpers/custom-validators/browse-videos.ts` functions could be modified so that, when called from `checker-after-init.ts`, the `i18n` `t` function is not used. Instead the plain error text in english is returned.
I would be happy to submit a PR with this simple fix for review.

### Additional information

* PeerTube instance:
  * URL:
  * Version: 7.3.0
  * NodeJS version:
  * Ffmpeg version:

* Browser name, version and platforms on which you could reproduce the bug:
* Link to browser console log if relevant:
* Link to server log if relevant (`journalctl` or `/var/www/peertube/storage/logs/`):

Chocobozzz:
Yes please!

Chocobozzz:
Fixed by https://github.com/Chocobozzz/PeerTube/pull/7222

PR #7222, opened 2025-09-15
fix(config): init config check does not report Browse Videos errors

gianantoniopini:
## Description

This PR modifies the `server/core/helpers/custom-validators/browse-videos.ts` functions so that, when called from `checker-after-init.ts`, the `i18n` `t` function is not used. Instead the plain error text in english is returned.

This change was required because `i18n` is still not initialized when the `checkConfig` function in `checker-after-init.ts` is called, therefore calls to the `t` function would result in `undefined`.

## Related issues

#7212

## Has this been tested?

I did not add tests because these changes are related to the server initialization, which is currently not covered by the existing test suite. Please correct me if I am wrong.


gianantoniopini:
@Chocobozzz This is ready for review.

Chocobozzz:
Thank you! I just updated the PR so we can correctly extract strings to translate :)

gianantoniopini:
Thank you!

Issue #7300, opened 2025-12-01
Browse Videos custom config validation errors have missing details


gianantoniopini:
### Describe the current behavior

If a custom config contains an invalid `default_sort` or `default_scope` in the `client.browse_videos` section, the validation errors returned by the config update validator have some missing details.

### Steps to reproduce

1. Add the following line at the end of the `buildSortItems` function in the `client/src/app/shared/shared-forms/select/select-videos-sort.component.ts` file:
    `this.sortItems.push({ id: '-invalid', label: 'Invalid Option' })`
2. From your terminal run: `npm run dev:client`
3. Once the frontend is running, go to http://localhost:3000/
4. Login as administrator
5. Go to the Settings -> Configuration -> General page
6. Scroll down to the BROWSE VIDEOS section and change the value of the 'Default sort' drop-down list to 'Invalid Option'
7. Click Save
8. The following Error is shown: "Browse videos default sort should be {options}, instead of {value}"

A similar issue also exists for 'Default scope'.

### Describe the expected behavior

The Error shown in Step 8 should be: "Browse videos default sort should be '-publishedAt' or '-originallyPublishedAt' or 'name' or '-trending' or '-hot' or '-likes' or '-views', instead of '-invalid'"

The issue appears to be related to the translations (`i18n` `t` function calls) used in the `browse-videos.ts` custom validator. It appears as if the integration of dynamic values (interpolation) is not working as expected in the translation function.

I would be happy to submit a PR with a fix for review.

### Additional information

* PeerTube instance:
  * URL:
  * Version: 8.0.0-rc.1
  * NodeJS version:
  * Ffmpeg version:

* Browser name, version and platforms on which you could reproduce the bug:
* Link to browser console log if relevant:
* Link to server log if relevant (`journalctl` or `/var/www/peertube/storage/logs/`):

Chocobozzz:
Should be fixed in develop branch, thank you for the bug report!

PR #7302, opened 2025-12-02
fix(config): Browse Videos custom config validation errors have missing details

gianantoniopini:
## Description

This PR modifies the `server/core/helpers/custom-validators/browse-videos.ts` functions so that the integration of dynamic values (interpolation) works as expected when calling the translation (`t`) function.
Quotes (') around the dynamic values have been removed in both TypeScript code and .json translation files, as that pattern was breaking the `t` function interpolation.

Additionally, tests in `packages/tests/src/api/check-params/config.ts` have been updated to also check for the expected validation error.

## Related issues

#7300 

## Has this been tested?


gianantoniopini:
@Chocobozzz PR is ready for review.

Chocobozzz:
Thank you for the PR, but the bug should be already fixed in develop branch. Don't hesitate to create another PR so we keep your integration tests though?

gianantoniopini:
I've checked the develop branch. The interpolation works as expected when calling the `t` function.

However, the first validation error related to 'Default sort' is now: "Browse videos default sort should be -publishedAt' or '-originallyPublishedAt' or 'name' or '-trending' or '-hot' or '-likes' or '-views, instead of hello". So some single quotes (') are still present in the error string, which is confusing in my opinion. With the changes in this PR, the following validation error would be produced instead: "Browse videos default sort should be -publishedAt or -originallyPublishedAt or name or -trending or -hot or -likes or -views, instead of hello".
The validation error related to 'Default scope' is now: "Browse videos default scope should be local' or 'federated, instead of hello". So same issue as above. With the changes in this PR, the following validation error would be produced instead: "Browse videos default scope should be local or federated, instead of hello".

For consistency and code clarity, the changes in this PR also remove the use of quotes (') in the validation errors produced during server init: these are the validation errors that do not use the `t` function.

Chocobozzz:
Can you re-apply your fix after a rebase of the develop branch, and not update the translation files?

gianantoniopini:
Done.

Chocobozzz:
Thank you!

gianantoniopini:
Thank you!

Issue #7328, opened 2025-12-11
Live stream returns 416 response for ts files


kontrollanten:
### Describe the current behavior

During our live streams some requests for the `.ts` files returns HTTP status 416. The error seems to become more often when we've more viewers (the response is cached for a short time in our CDN).

<img width="870" height="315" alt="Image" src="https://github.com/user-attachments/assets/d8134feb-be6d-481a-becc-16d3059f9570" />

In our nginx error log we also see a lot of `No such file or directory` errors.

```
2025/12/11 07:52:51 [error] 281583#281583: *1223 open() "/storage/streaming-playlists/hls/b50f8738-27b3-45b6-8993-fe0f662dbbb6/0.m3u8" failed (2: No such file or directory), request: "GET /static/streaming-playlists/hls/b50f8738-27b3-45b6-8993-fe0f662dbbb6/0.m3u8 HTTP/1.1"

... 

2025/12/11 08:02:09 [error] 281583#281583: *60195 open() "/storage/streaming-playlists/hls/b50f8738-27b3-45b6-8993-fe0f662dbbb6/2-000098.ts" failed (2: No such file or directory), , request: "GET /static/streaming-playlists/hls/b50f8738-27b3-45b6-8993-fe0f662dbbb6/2-000098.ts HTTP/1.1"
```

It's both `m3u8` and `ts` files that are recorded as not found, but only `ts` files that returns 416. Requests for `m3u8` files are often cancelled after 10 seconds.

### Steps to reproduce

1. Create a live behind a CDN with 120 seconds cache for `ts` files.
2. Add +300 viewers.
3. Watch the console.

### Describe the expected behavior

According to the code it seems that neither `tsWatcher` nor `m3u8Watcher` in `apps/peertube-runner/src/server/process/shared/process-live.ts` is using the `awaitWriteFinish` option, which will make it possible to request those files before they're written to disk.

Even with the `awaitWriteFinish` option I guess it's possible that some clients are requesting the files during the write operation. Maybe it's possible to get nginx to retry in those cases, or if they should be created as new files every time.

### Additional information

* PeerTube instance:
  * Version: 7.3.0
  * NodeJS version: 20.19.5
  * Ffmpeg version: 4.4.2

Chocobozzz:
[...]

PR #7331, opened 2025-12-12
fix(runner): wait for `.ts` segments to be fully written before upload

ar9708:
## Description

In my analysis of 416 errors mentioned in #7328, I've seen this related to the use of Remote Runners. I have [a runner solution](https://github.com/OwnTube-tv/peertube-runner) that I'm using, and I suspect that it could produce incomplete segments as input for PeerTube. See illustrations below on how I understand the race conditions.

**The Race Condition**

```plain
┌─────────────────────────────────────────────────────────────────────┐
│                    [[ Remote Runner ]]                              │
│                                                                     │
│   FFmpeg ──writes──► segment-001.ts (incomplete)                    │
│                           │                                         │
│                           ▼                                         │
│  chokidar ────'add'────► triggers IMMEDIATELY (no awaitWriteFinish) │
│                           │                                         │
│                           ▼                                         │
│  sendPendingChunks() ──► reads incomplete file                      │
│                           │                                         │
└───────────────────────────┼─────────────────────────────────────────┘
                            │ HTTP POST /jobs/:uuid/update
                            ▼
┌─────────────────────────────────────────────────────────────────────┐
│                   [[ PeerTube Server ]]                             │
│                                                                     │
│   Receives incomplete segment ──► writes to disk ──► Nginx serves   │
│                                                                     │
└─────────────────────────────────────────────────────────────────────┘
```

Line 74 in `PeerTube/apps/peertube-runner/src/server/process/shared/process-live.ts` is the culprit:

```typescript
const tsWatcher = watch(this.outputPath, { ignored: ... })
// No awaitWriteFinish! File detected immediately while FFmpeg still writing!
```

**The Fix**

The fix is to add `awaitWriteFinish` to the watcher:

```typescript
const tsWatcher = watch(this.outputPath, {
  ignored: path => path !== this.outputPath && !path.endsWith('.ts'),
  awaitWriteFinish: {
    stabilityThreshold: 500,  // Wait 500ms for file size to stabilize
    pollInterval: 100
  }
})
```

This makes chokidar wait until FFmpeg finishes writing before triggering the event. Without this option, my AI tools says that chokidar triggers the 'add' event immediately when FFmpeg creates the segment file, before FFmpeg finishes writing to it. This could cause the runner to occasionally upload incomplete segments, which then leads to HTTP 416 "Requested Range Not Satisfiable" errors when players request byte ranges beyond what was actually written.

Thank you for considering my theory here. PeerTube is the best! 🤗

## Related issues

Fixes: #7328

## Has this been tested?

### DIFF
--- a/apps/peertube-runner/src/server/process/shared/process-live.ts
+++ b/apps/peertube-runner/src/server/process/shared/process-live.ts
@@ -71,7 +71,18 @@ export class ProcessLiveRTMPHLSTranscoding {
         const m3u8Watcher = watch(this.outputPath, { ignored: path => path !== this.outputPath && !path.endsWith('.m3u8') })
         this.fsWatchers.push(m3u8Watcher)
 
-        const tsWatcher = watch(this.outputPath, { ignored: path => path !== this.outputPath && !path.endsWith('.ts') })
+        // Fix: Wait for .ts files to be fully written before processing
+        // Without awaitWriteFinish, chokidar triggers immediately when FFmpeg creates the file,
+        // before FFmpeg finishes writing. This causes incomplete segments to be uploaded,
+        // leading to HTTP 416 errors when players request byte ranges beyond what was written.
+        // See: https://github.com/Chocobozzz/PeerTube/issues/7328
+        const tsWatcher = watch(this.outputPath, {
+          ignored: path => path !== this.outputPath && !path.endsWith('.ts'),
+          awaitWriteFinish: {
+            stabilityThreshold: 500,
+            pollInterval: 100
+          }
+        })
         this.fsWatchers.push(tsWatcher)
 
         m3u8Watcher.on('change', p => {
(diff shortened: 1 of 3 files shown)


Chocobozzz:
Hello,

Please see https://github.com/Chocobozzz/PeerTube/issues/7328#issuecomment-3645852564

Did you consider that we only send previous chunks to the peertube instance? So when we detect `segment-002.ts`, we send `segment-001.ts` and not `segment-002.ts` (that will be sent when `segment-003.ts` will be created)

ar9708:
> Hello,
> 
> Please see [#7328 (comment)](https://github.com/Chocobozzz/PeerTube/issues/7328#issuecomment-3645852564)
> 
> Did you consider that we only send previous chunks to the peertube instance? So when we detect `segment-002.ts`, we send `segment-001.ts` and not `segment-002.ts` (that will be sent when `segment-003.ts` will be created)

I saw it, but I didn't actually understand it, I had to re-read it a couple of times. Now, following your conversation with the issue creator, I finally understand what you mean. Your idea with just passing `temp_file` to FFmpeg is genius! Maybe I can be first with creating a PR for that instead, we'll see! 🙃

PR #7332, opened 2025-12-12
fix(scripts): make client build compatible with Bash 3.2 (macOS)

ar9708:
## Description

I started poking around here earlier this morning, to build and run tests and contribute something, but I quickly ran into the problem with Bash 4 requirements as I'm on MacOS. I was thinking this might pose an overly high barrier to entry for newcomers. Beautiful with the Bash 4+ associative arrays, but being able to check out the project and just run `npm run build` after the necessary Homebrew Node/npm/FFmpeg installs that's also beautiful! 😁 

## Has this been tested?

- [x] 👍 yes, I ~~added tests to the test suite~~ tested locally on my standard Bash 3.2 Apple Silicon MacBook laptop, with `npm run build` + then started Redis/Postgres + ran test commands

### DIFF
--- a/scripts/build/client.sh
+++ b/scripts/build/client.sh
@@ -4,9 +4,47 @@ set -eu
 
 defaultLanguage="en-US"
 
-# Supported languages - using parallel arrays for Bash 3.2 compatibility (macOS default)
-lang_keys=(ar sk is tr fa en vi hu th fi nl gd el es oc pt pt-PT sv pl ru zh-Hans zh-Hant fr ja eu ca gl cs hr eo de it uk sq tok nn nb kab)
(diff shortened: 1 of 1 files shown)


Chocobozzz:
I'm sorry but it's more difficult to read and maintain. Can you find a solution so we keep the association between the key and the value?

ar9708:
> I'm sorry but it's more difficult to read and maintain. Can you find a solution so we keep the association between the key and the value?

You're absolutely right, how about 200480f? This isn't too far from the original data structure. Tested with `npm run build` on my Bash 3.2 laptop.

Chocobozzz:
LGTM. Thank you!

PR #7333, opened 2025-12-12
fix(live): add FFmpeg `temp_file` HLS flag to prevent incomplete segment reads

ar9708:
Ping @Chocobozzz, here's your genius solution from https://github.com/Chocobozzz/PeerTube/issues/7328#issuecomment-3646543213, I was dying to try it out! 😎 

## Description

When FFmpeg writes HLS segments, the segment file is created before all data is written. This can potentially cause HTTP 416 (Range Not Satisfiable) errors when players request byte ranges that don't yet exist in the file. The `temp_file` flag makes FFmpeg write segments to a temporary file and atomically rename to the final name only when the segment is complete.

## Related issues

Resolves: #7328

## Has this been tested?

- [x] 🙋 no, because I need help, I don't know how we'd test this meaningfully!

### DIFF
--- a/packages/ffmpeg/src/ffmpeg-live.ts
+++ b/packages/ffmpeg/src/ffmpeg-live.ts
@@ -253,7 +253,7 @@ export class FFmpegLive {
 
     command.outputOption('-hls_time ' + segmentDuration)
     command.outputOption('-hls_list_size ' + segmentListSize)
-    command.outputOption('-hls_flags delete_segments+independent_segments+program_date_time')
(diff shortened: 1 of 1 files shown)


Chocobozzz:
@kontrollanten can you test this patch?

kontrollanten:
> @kontrollanten can you test this patch?

We tried it today and it worked well with ~200 simultan viewers and ~40 000 ts files downloaded. No 416 responses returned. Last time we had an issue there was ~300 viewers, so it was a bit less pressure this time but it looks promising.

Chocobozzz:
Thanks @kontrollanten & @ar9708

Issue #7326, opened 2025-12-11
[Security] Directory traversal vulnerability in example Nginx configuration (`support/nginx/peertube`)


ar9708:
### Describe the current behavior

According to @Copilot, the example nginx configuration file at `support/nginx/peertube` contains a **directory traversal vulnerability** in the client assets location block. I'm filing it here as a normal ticket instead of writing a security email, because it's a low hanging fruit for AI robots to catch. 

**Vulnerable configuration:**

```nginx
location ~ ^/client/(.*\.(js|css|png|svg|woff2|otf|ttf|woff|eot))$ {
    add_header Cache-Control "public, max-age=31536000, immutable";
    alias /var/www/peertube/peertube-latest/client/dist/$1;
}
```

Here's the pitch:

> The regex capture group `(.*)` allows any characters, including `../` sequences. When combined with the `alias` directive, this enables path traversal attacks.
> 
> **Attack example:**
> A malicious request to `/client/../../server/config.js` would:
> 1. Match the regex (captures `../../server/config.js`)
> 2. Substitute into alias: `/var/www/peertube/peertube-latest/client/dist/../../server/config.js`
> 3. Resolve to: `/var/www/peertube/server/config.js`
> 
> This could expose sensitive files outside the intended `client/dist` directory, including configuration files, credentials, or application source code.

### Steps to reproduce

This does not actually work when I try it, but here's the @Copilot repro steps:

> 1. Deploy PeerTube using the example nginx configuration from `support/nginx/peertube`
> 2. Send a request with path traversal: `curl https://your-instance.com/client/..%2F..%2Fserver/config/production.yaml.js`
> 3. The request may return files outside the intended client/dist directory

### Describe the expected behavior

Again, @Copilot copy-paste:

> The nginx configuration should prevent directory traversal and only serve files from within the intended `client/dist` directory.
> 
> **Suggested fix - replace the vulnerable block with:**
> 
> ```nginx
> location /client/ {
>     add_header Cache-Control "public, max-age=31536000, immutable";
>     root /var/www/peertube/peertube-latest;
>     try_files $uri =404;
> }
> ```
> 
> This uses `root` instead of `alias` with regex capture groups, which is inherently safer because nginx normalizes the path before mapping it to the filesystem.

### Additional information

More @Copilot copy-paste:

> **Severity:** Medium-High (depends on file permissions and directory structure)
> 
> **Impact:** 
> - Potential exposure of configuration files containing secrets
[...]

PR #7334, opened 2025-12-12
docs(nginx): add security comment for alias path traversal

ar9708:
## Description

Thank you for the explanation on #7326! How do you see adding a `# security` annotation to avoid having to repeat the information you provided yesterday? Just thinking out loud, feel free close! :)

(The `nosec` annotation helps suppress false positives from security scanners that flag the alias+regex pattern.)

## Has this been tested?

### DIFF
--- a/support/nginx/peertube
+++ b/support/nginx/peertube
@@ -200,6 +200,8 @@ server {
   }
 
   # Bypass PeerTube for performance reasons. Optional.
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thank you, but our security scanner doesn't detect it as a security issue so I think it's more a false positive bug of your IA tool they should fix :)

PR #7343, opened 2025-12-16
Fix typos in config files and Nginx template

ar9708:
## Description

Small nags I'm getting from @Copilot AI on my PRs, I think it's better to fix them upstream than diverge forks:
- Add missing semicolon in Content-Type header per [RFC 2045](https://datatracker.ietf.org/doc/html/rfc2045#section-5.1)
- Remove invalid trailing quote in `securitytxt` YAML blocks
- Fix yt-dlp/yt-dlp ARM binary URLs (`armv7l.zip` for ARMv7, `aarch64` for ARM64)

## Has this been tested?


Chocobozzz:
Thank you!

PR #7346, opened 2025-12-18
Update Dockerfile for VAAPI

nico198x:
added to install va-driver-all for Intel and AMD users of VAAPI

## Description

Insalling va-driver-all metapackage will install all the needed components for VAAPI support for Intel and AMD users.

## Has this been tested?

- [X ] 🙅 no, because this PR does not update server code

it's an exceedingly simply change of just installing some drivers. i do have a homespun build which includes this change and everything works as expected.


Chocobozzz:
Thank you!

PR #7355, opened 2025-12-22
add filter:feed.videos.list.result hook

### LINKED ISSUE #7314: Use videos filter in RSS feeds

In `peertube-plugin-premium-users` we're using `filter:api.video.get.result` to change the video files URL for non-paying users in order to provide premium content. But the RSS feed is still providing the "premium files" to all users.

Be able to filter the video files URL in RSS feeds as well.

@Chocobozzz what do you think of using `filter:api.video.get.result` in the RSS feed as well? We could provide a `context` parameter to the filter function with type `'details-api' | 'rss-feed'`.

kontrollanten:
## Description
Pass videos through `filter:feed.videos.list.result` before creating the podcast, XML och JSON feed.

## Related issues
closes #7314

## Has this been tested?

### DIFF
--- a/packages/models/src/plugins/server/server-hook.model.ts
+++ b/packages/models/src/plugins/server/server-hook.model.ts
@@ -149,7 +149,10 @@ export const serverFilterHookObject = {
 
   // Peertube >= 7.2
   'filter:email.subject.result': true,
-  'filter:email.template-path.result': true
+  'filter:email.template-path.result': true,
+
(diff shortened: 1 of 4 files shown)


kontrollanten:
It seems to have been an issue with external connections since requests to npm and peertube.cpy.re timed out. I don't think it's related to my changes.

Chocobozzz:
Thank you!

PR #7359, opened 2025-12-24
Update Dockerfile instructions in CHANGELOG

meljoann:
Clarified the usage of the PeerTube Dockerfile by specifying that the Debian suffix version is deprecated and must be replaced.

## Has this been tested?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@
 
 ### Docker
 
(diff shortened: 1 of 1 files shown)


meljoann:
We discussed this on Mastodon. I think it must have been updated elsewhere, not here...? Hope this helps :)

Chocobozzz:
Hi & thank you

Yes I fixed the changelog in https://github.com/Chocobozzz/PeerTube/commit/c5cb227e4f5097b7493303dec9c7b77cf74dfbb2

PR #7369, opened 2026-01-07
Fix openapi docs for playlist videos response

jakmeier:
This is the playlist element id, used to update or delete elements in the list.
The old description and example made it look like it is the video object id.


Chocobozzz:
Thank you!

Issue #7368, opened 2026-01-05
E-mail link color not set


afilina:
### Describe the current behavior

Because button.hbs does not set a link color from the theme, it can make the e-mail unreadable:

<img width="197" height="59" alt="Image" src="https://github.com/user-attachments/assets/abd7fe6c-9eac-4898-98f8-f5fa606d3639" />

People online suggest explicitly setting the inline style:
https://stackoverflow.com/a/13762730/544894

### Steps to reproduce

1. Create user using a Gmail account (haven't tested others)
2. Check e-mail

### Describe the expected behavior

Button foreground color same as on the website.

Muhammad-Sheik-Nauman:
I would like to work on this

Chocobozzz:
@afilina Can you share your instance URL so I can reproduce your configuration?

afilina:
https://phpc.tv

Chocobozzz:
Thank you! Fixed by https://github.com/Chocobozzz/PeerTube/commit/6ecd9823cfbe7bb3b1a7b6aa67438b590e1057cb

PR #7370, opened 2026-01-07
Add Button foreground color same as on the website.

Muhammad-Sheik-Nauman:
related to issue #7368


Chocobozzz:
Hello,

Where `buttonTextColor` is coming from?

PR #7374, opened 2026-01-10
Fix client overrides not overriding logo.svg

ayaadev:
## Description

If users wish to change PeerTube's logo to their own SVG logo, they only have one option: client overrides. This is because the `.svg` file extension is not supported by the custom logo configuration in the UI. This is problematic if users want their custom logo to appear clearly, without blurriness such as the result of uploading a `.png` file. I have first-hand experience with this frustration.

image

&nbsp;

The `logo.svg` file wasn't whitelisted in the NGINX configuration to be overridden by the client-overrides feature, therefore even if users placed the `logo.svg` file in the correct directory `/assets/images` the logo would not be overridden. This PR fixes this issue by simply whitelisting the `logo.svg` filename. 

## Related issues

https://framacolibri.org/t/override-logo-and-icon-with-client-overrides/9634

https://github.com/Chocobozzz/PeerTube/issues/3292 (this PR re-fixes overriding logo.svg)

https://github.com/Chocobozzz/PeerTube/pull/2897#issuecomment-672963723

## Has this been tested?

Yes, I have manually edited the NGINX configuration on my live production instance at https://peertube.tech. You can see the logo.svg file being replaced at: https://peertube.tech/client/assets/images/logo.svg

This option also applies, but I have tested it on my instance.

## Screenshots

The custom logo not becoming blurry when scaled 500% in the browser.
Custom logo not becoming blurry when scaled

When uploading a custom logo through the UI, it looks unclear and blurry at the same 500% scale like the previous screenshot.
image

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -140,6 +140,7 @@ storage:
   # Various admin/user uploads that are not suitable for the folders above
   uploads: 'storage/uploads/'
   # Overridable client files in client/dist/assets/images:
+  # - logo.svg
   # - default-avatar-account-48x48.png
   # - default-avatar-account.png
   # - default-avatar-video-channel-48x48.png
--- a/config/production.yaml.example
+++ b/config/production.yaml.example
@@ -138,6 +138,7 @@ storage:
   # Various admin/user uploads that are not suitable for the folders above
   uploads: '/var/www/peertube/storage/uploads/'
   # Overridable client files in client/dist/assets/images:
+  # - logo.svg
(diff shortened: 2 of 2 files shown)


Chocobozzz:
Hi and thank you, but I think we need to support SVG uploads instead :) Can you create an issue so we can track the feature request?

ayaadev:
Sure thing! That's very reasonable. Here's a link to the new issue I made: https://github.com/Chocobozzz/PeerTube/issues/7397

PR #7396, opened 2026-01-22
Live dvr

djblu2003dk-tech:
## Has this been tested?


djblu2003dk-tech:
Added a few modifications to PeerTube.

PIP
Live DVR, Skip, Scrub, pause and click to live.
Added an option to disable the right click context menu in embedded players.

client/src/standalone/player/src/peertube-player.ts
```diff
@@ -257,7 +263,15 @@ export class PeerTubePlayer {
         saveAverageBandwidth(Math.floor(data.bandwidthEstimate))
       })
 
-      this.player.contextMenu(this.getContextMenuOptions())
+      if (this.isInIframe()) {
+        // Disable custom and native context menus in embeds.
+        this.player.on('contextmenu', (event: Event) => {
```
Chocobozzz: Can you explain why?

Chocobozzz (commented):
Thank you. Can you split live dvr and pip control so it's easier to review?

Can you also add server tests for live dvr?

djblu2003dk-tech:
I have tested this branch and it completes without error. All other functions have been stripped and it works as expected. Added the ability to set the DVR window(max in .yaml) user defined in Live Stream admin page.

djblu2003dk-tech:
Just as an addition, the reason for my adding

force_path_style=false is due to config.js

having

FORCE_PATH_STYLE: config.get('object_storage.force_path_style')

There is no check on ENABLED: config.get('object_storage.enabled'). If this is false and the FORCE_PATH_STYLE object is empty it throws an error.

djblu2003dk-tech:
"Can you also add server tests for live dvr?"

I'm not familiar with writing these tests, it will take me a while but I'll add them as soon as I can.

djblu2003dk-tech:
> Thank you. Can you split live dvr and pip control so it's easier to review?
> 
> Can you also add server tests for live dvr?

Tests added.

djblu2003dk-tech:
Updated default.yaml to ensure tests work.

djblu2003dk-tech:
I've run all the tests again including simulating S3 etc, all are working and code now complies with 140 char limit.

config/production.yaml.example
```diff
@@ -179,9 +179,6 @@ object_storage:
 
   region: 'us-east-1'
 
-  # Use path style requests if the S3 provider does not support virtual hosting of buckets
```
Chocobozzz: Please revert these changes

client/src/standalone/player/src/peertube-player.ts
djblu2003dk-tech: Removed.

config/test.yaml
```diff
@@ -35,6 +35,9 @@ smtp:
   from_address: 'test-admin@127.0.0.1'
   username: null
   password: null
+  
+object_storage:
```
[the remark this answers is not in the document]
djblu2003dk-tech: Error: Configuration property "object_storage.force_path_style" is not defined
    at Config.get (.../node_modules/config/lib/config.js:146:11)
    at .../server/core/initializers/config.ts:267:30

property is not defined error thrown if it isn't set. It doesn't prevent the server running but the tests failed.

stevespaw:
Is this something we will see in a release soon? 
Thanks!

covalenco:
This would be very useful for our use case. We run 24/7 live streams on PeerTube and a DVR window (for example 1 hour) would allow viewers to rewind while keeping the live stream ongoing. I support this feature.

Chocobozzz:
Thank you

PR #7414, opened 2026-01-30
Add configurable default for live replay save option

### LINKED ISSUE #6304: Globally Toggle "Upload Replays of Lives" By Default

When creating a new live, the option to automatically publish a replay when the live is finished is always checked. For instances, like ours, that have storage quotas, I think having this unchecked by default would be the best option, that way livestreams aren't arbitrarily cut off and we don't have to remind people who are unfamiliar with PeerTube how to turn off the functionality.

I know that we can turn off the function by default, but there are some people who are allowed to use the function. I know this is kind of an edge case, but I figured I'd make a feature request for anyone else who might have this same problem.

Allow administrators to set the default option for "automatically upload a replay when the live ends" for all new lives.

ceasermikes002:
Fixes #6304

Allow administrators to configure the default state of the "automatically publish a replay when your live ends" checkbox via a new `live.save_replay_by_default` setting.

## Problem

When creating a new live video, the "save replay" checkbox is always checked by default. For instances with storage quotas, this causes livestreams to be cut off when storage fills up. While admins can disable the replay feature entirely, some users should be allowed to use it - just not by default.

## Solution

Added a new configuration option `live.save_replay_by_default` that:
- Defaults to `false` (unchecked)
- Can be configured by admins in the Live settings page
- Controls the default checkbox state when creating new live videos
- Still allows users to manually toggle the checkbox

### Backend
- Added `save_replay_by_default` config to YAML files
- Updated TypeScript models (`CustomConfig`, `HTMLServerConfig`)
- Exposed config via API endpoint
- Added validation for the new field

### Frontend
- Added admin UI checkbox in Live configuration page
- Modified live creation logic to use server config default

### Tests
- Updated config tests to validate the new field

## Testing

- [x] Admin UI shows new checkbox in Live settings
- [x] Setting saves and persists correctly
- [x] Unchecked admin setting → unchecked default in new live
- [x] Checked admin setting → checked default in new live
- [x] Users can still manually override the checkbox
- [x] Works for both normal and permanent lives

### DIFF
--- a/packages/models/src/server/custom-config.model.ts
+++ b/packages/models/src/server/custom-config.model.ts
@@ -342,18 +342,18 @@ export interface CustomConfig {
   }
 
   defaults: {
-      publish: {
-        downloadEnabled: boolean
-        commentsPolicy: VideoCommentPolicyType
-        privacy: VideoPrivacyType
-        licence: number
-      }
+    publish: {
+      downloadEnabled: boolean
+      commentsPolicy: VideoCommentPolicyType
+      privacy: VideoPrivacyType
+      licence: number
+    }
 
-      live: {
-        saveReplay: boolean
-      }
+    live: {
+      saveReplay: boolean
+    }
 
-      p2p: {
+    p2p: {
       webapp: {
         enabled: boolean
       }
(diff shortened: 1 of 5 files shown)


ceasermikes002:
@Chocobozzz

Chocobozzz:
Thank you but tests fail. You can reproduce by running `npm run build:tests` on your computer

ceasermikes002:
@Chocobozzz please review.

Chocobozzz:
Thank you!

PR #7424, opened 2026-02-02
Reverting va-driver-all changes

nico198x:
## Description

I deeply apologize, I got confused.

There was in fact nothing wrong with the original addition.

It just hadn't gone live yet in 8.0.2 and I assumed I did it wrong!

I re-checked the doc & re-tested on my system, and can confirm va-driver-all doesn't have any recs. Only deps and enh.

https://packages.debian.org/en/trixie/va-driver-all

So sorry for the hassle and confusion!

## Has this been tested?

Yes, on my machine.

- [ X ] 🙅 no, because this PR does not update server code

### DIFF
--- a/support/docker/production/Dockerfile
+++ b/support/docker/production/Dockerfile
@@ -4,9 +4,7 @@ ARG ALREADY_BUILT=0
 
 # Install dependencies
 RUN apt update \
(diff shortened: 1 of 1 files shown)


Chocobozzz:
No problem, thank you for reverting the change!

PR #7452, opened 2026-02-18
putVideo tags encoding fix

nosnickid:
## Description

When using openapi-generator to build an API from the spec, `tags` would have a `_collection_format` set to `csv`, which caused the API to submit eg "tags=tag1,tag2,tag3", instead of submitting multiple `tag` parameters, eg `tag=tag1,tag=tag2,tag=tag3`

Setting the encoding this way sets the `_collection_format` to `multi`, which works.

## Related issues

I didn't file a bug, but it happens.

## Has this been tested?


Chocobozzz:
Thank you!

Issue #7467, opened 2026-02-25
yt-dlp import fails when system has multiple Node.js versions (unsupported node found via PATH)


ClusterM:
### Describe the current behavior

When importing a video from YouTube, yt-dlp fails with:
                                                                                                                                                                                              
```
WARNING: [youtube+oauth2] ...: n challenge solving failed: Some formats may be missing.
Ensure you have a supported JavaScript runtime and challenge solver script distribution installed.
WARNING: Only images are available for download. use --list-formats to see them
ERROR: [youtube+oauth2] ...: Requested format is not available.
```

The root cause is that PeerTube hardcodes `--js-runtimes node` in `youtube-dl-cli.js`:
     
```js
wrapWithJSRuntimeOptions(args) {
    if (CONFIG.IMPORT.VIDEOS.HTTP.YOUTUBE_DL_RELEASE.NAME === 'yt-dlp') {
        return ['--js-runtimes', 'node'].concat(args);
    }
    return args;
}  
```

This causes yt-dlp to resolve `node` via PATH. On systems where PeerTube is launched with a supported Node.js version (e.g. v22 via nvm) but a different, older `node` binary exists in PATH (
e.g. `/usr/bin/node` v18 from a system package), yt-dlp finds the unsupported one.                                                                                                            
                                                                                                                                                                                              
yt-dlp's verbose log confirms:                                                                                                                                                                
                                                                                                                                                                                              
```                                                                                                                                                                                           
[debug] JS runtimes: node-18.20.4 (unsupported)                                                                                                                                               
```                                                                                                                                                                                           
                                                                                                                                                                                              
yt-dlp requires Node.js >= 20.0.0 for EJS challenge solving. With an unsupported version, the JS challenges fail silently, and no video formats are returned — only thumbnail images.

### Steps to reproduce

1. Install Node.js 22 via nvm (used to run PeerTube)                                                                                                                                          
2. Have Node.js 18 installed system-wide (`/usr/bin/node`)                                                                                                                                    
3. Run PeerTube via systemd (ExecStart pointing to nvm's node)
4. Enable HTTP video import     
5. Try to import a YouTube video
      

### Describe the expected behavior

yt-dlp should use the same Node.js binary that PeerTube is running on, or at least a supported version.   
                                                                                                                                                                                              
#### Actual behavior

yt-dlp finds the system-wide unsupported Node.js 18, JS challenge solving fails, and the import returns error 400.


### Additional information

#### Versions

[...]

PR #7468, opened 2026-02-25
yt-dlp import fix (unsupported node found via PATH)

ClusterM:
## Description

Fix for #7467

When importing a video from YouTube, yt-dlp fails with:
                                                                                                                                                                                              
```
WARNING: [youtube+oauth2] ...: n challenge solving failed: Some formats may be missing.
Ensure you have a supported JavaScript runtime and challenge solver script distribution installed.
WARNING: Only images are available for download. use --list-formats to see them
ERROR: [youtube+oauth2] ...: Requested format is not available.
```

The root cause is that PeerTube hardcodes `--js-runtimes node` in `youtube-dl-cli.js`:
     
```js
wrapWithJSRuntimeOptions(args) {
    if (CONFIG.IMPORT.VIDEOS.HTTP.YOUTUBE_DL_RELEASE.NAME === 'yt-dlp') {
        return ['--js-runtimes', 'node'].concat(args);
    }
    return args;
}  
```

This causes yt-dlp to resolve `node` via PATH. On systems where PeerTube is launched with a supported Node.js version (e.g. v22 via nvm) but a different, older `node` binary exists in PATH (
e.g. `/usr/bin/node` v18 from a system package), yt-dlp finds the unsupported one.

## Related issues

#7467

## Has this been tested?


Chocobozzz:
Thank you!

Issue #7474, opened 2026-03-03
Customizaton for color that remains orange after customizing Primary color


RSXEnthusiast:
### Describe the problem to be solved

Currently, even after customizing the primary color, some elements remain orange.

### Describe the solution you would like

Add another customization field for that color, allowing admin to change it.

RSXEnthusiast:
I'll have a PR up for this shortly.

Chocobozzz:
Fixed by https://github.com/Chocobozzz/PeerTube/commit/55afd5c4499edeb85d41c7f7967a1f1a3629bcf7

PR #7475, opened 2026-03-03
Adding ability to customize border primary color

RSXEnthusiast:
## Description

Adding ability to customize primary border color. It always bothered me that certain elements remained orange after setting the primary color to something else. Now users can customize that color as well.

## Related issues

#7474

## Has this been tested?

## Screenshots

Before:
image
After:
image

This is my first contribution, so make sure to use extra scrutiny :D

### DIFF
--- a/client/src/app/+admin/config/pages/admin-config-customization.component.ts
+++ b/client/src/app/+admin/config/pages/admin-config-customization.component.ts
@@ -56,6 +56,7 @@ type Form = {
 
     customization: FormGroup<{
       primaryColor: FormControl<string>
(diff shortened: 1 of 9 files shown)


Chocobozzz:
Thank you for the PR, but I fixed the bug in https://github.com/Chocobozzz/PeerTube/commit/55afd5c4499edeb85d41c7f7967a1f1a3629bcf7 where we prefer to inherit primary color instead of defining another specific color for borders.

RSXEnthusiast:
Works for me! I was split between doing something like that or what I did.

PR #7489, opened 2026-03-12
inform about necessity to shut down peertube first

Sakrecoer:
## Description

The script gives a heads up about peertube needing to be shut down for this command, but it shows for only a few seconds. Given how sensitive it is, it makes sense to inform about it in the docs.

### DIFF
--- a/support/doc/tools.md
+++ b/support/doc/tools.md
@@ -657,6 +657,8 @@ cd /var/www/peertube-docker; \
 
 ### Prune filesystem/object storage
 
+**PeerTube must be stopped before running this script**
+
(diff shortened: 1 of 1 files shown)


Sakrecoer:
Thanks, @spirillen ! I will amend. As for your question, the why: Never worked with this particular doc tool and didn't not know about the note box. 😅

Chocobozzz:
Thank you, but we updated the prune storage script so this is not needed anymore: https://github.com/Chocobozzz/PeerTube/commit/5f60f1488aee3502779e4edd1586add21c058884

Issue #7496, opened 2026-03-14
Upgrade to 8.1.0 fails on /absolute/path/to/server-certificates/root.crt


afilina:
### Describe the problem

The Changelog says:

> You need to manually execute a migration script after upgrading, while PeerTube is running and the database migration is complete

I _think_ I understood the full instructions correctly (classic installation):
```
cd /var/www/peertube/peertube-latest/scripts && sudo -H -u peertube ./upgrade.sh
sudo systemctl restart peertube
# this one is from the IMPORTANT NOTES
cd /var/www/peertube/peertube-latest && sudo -u peertube NODE_CONFIG_DIR=/var/www/peertube/config NODE_ENV=production node dist/scripts/migrations/peertube-8.1.js
```

The problem is that as soon as I finish upgrading, I can't start `peertube` for this last step. It's looking for a bogus cert path `/absolute/path/to/server-certificates/root.crt` and crashes.

I reverted the symlink and restarted to go back to the working version. This is my first PeerTube upgrade. I wonder whether some steps are assumed but not documented.

```
Error: ENOENT: no such file or directory, open '/absolute/path/to/server-certificates/root.crt'
    at readFileSync (node:fs:449:20)
    at file:///var/www/peertube/versions/peertube-v8.1.2/dist/core/initializers/database.js:93:33
    at ModuleJob.run (node:internal/modules/esm/module_job:325:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:606:24)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5) {
  "error": {
    "stack": "Error: ENOENT: no such file or directory, open '/absolute/path/to/server-certificates/root.crt'\n    at readFileSync (node:fs:449:20)\n    at file:///var/www/peertube/versions/peertube-v8.1.2/dist/core/initializers/database.js:93:33\n    at ModuleJob.run (node:internal/modules/esm/module_job:325:25)\n    at async ModuleLoader.import (node:internal/modules/esm/loader:606:24)\n    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5)",
    "message": "ENOENT: no such file or directory, open '/absolute/path/to/server-certificates/root.crt'",
    "errno": -2,
    "code": "ENOENT",
    "syscall": "open",
    "path": "/absolute/path/to/server-certificates/root.crt"
  },
[...]

PR #7511, opened 2026-03-21
Clearly mention production.yaml.new before asking to restart

afilina:
## Description

The step with `production.yaml.new` is mentioned much further in the upgrade steps, while the best time to _edit_ that config file is between there two upgrade steps:

```bash
cd /var/www/peertube/peertube-latest/scripts && sudo -H -u peertube ./upgrade.sh
sudo systemctl restart peertube # Or use your OS command to restart PeerTube if you don't use systemd
```

- Removed the restart step from this code block.
- Added a clear explanation of what to do about config.
- Added the restart step separately below, with additional comment so there's no chance to miss the config changes before restarting.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/7496

## Has this been tested?


Chocobozzz:
Thank you!

PR #7528, opened 2026-03-30
pass ip address to requires-approval.result hook

kontrollanten:
## Description
Add missing IP address param in filter hook.

## Related issues
#6691 

## Has this been tested?


Chocobozzz:
Thanks!

PR #7535, opened 2026-04-02
Fix missing stream error handling in web video object storage proxy

lawrence3699:
## Description

Was debugging intermittent crashes on a self-hosted instance using S3-compatible storage and traced the issue to `proxifyWebVideoFile` in `server/core/lib/object-storage/proxy.ts`.

The web video proxy uses `stream.pipe(res)` (line 30) without any stream error handler. If the S3 stream errors mid-transfer (connection drop, object deleted during download, S3 timeout), the error event goes unhandled, which can crash the Node.js process.

The HLS proxy in the same file already handles this correctly using `pipeline()` with an error callback (lines 60-69). The web video proxy should do the same.

**Before:** If the S3 connection drops during a web video download, the stream emits an unhandled error event. Depending on the Node.js configuration, this either crashes the process or silently corrupts the response.

**After:** Stream errors during web video proxying are caught and handled by `handleObjectStorageFailure`, which logs the error and returns an appropriate HTTP error response — the same behavior that already exists for HLS streaming.

Every other `.pipe()` call in the server codebase (`plugin-manager.ts:571`, `user-exporter.ts:122`) has an explicit error handler. This brings the web video proxy in line with the rest.

## Related issues

N/A — found by reading the code

## Has this been tested?

- [x] 👍 yes, I verified the fix resolves the inconsistency by reading both code paths; the change is minimal (replacing `stream.pipe(res)` with `pipeline(stream, res, errorCallback)`) and follows the exact pattern from `proxifyHLS` in the same file

## Screenshots

N/A

### DIFF
--- a/server/core/lib/object-storage/proxy.ts
+++ b/server/core/lib/object-storage/proxy.ts
@@ -27,7 +27,15 @@ export async function proxifyWebVideoFile (options: {
 
     setS3Headers(res, s3Response)
 
-    return stream.pipe(res)
+    return pipeline(
+      stream,
+      res,
+      err => {
+        if (!err) return
+
+        handleObjectStorageFailure(res, err)
+      }
+    )
   } catch (err) {
     return handleObjectStorageFailure(res, err)
   }


Chocobozzz:
Thanks!

PR #7539, opened 2026-04-02
Update OpenAPI library links

scottruona:
OpenAPI library links directed to repositories that no longer seem to exist. This update links to the relevant folders in the PeerTube repository.

## Description

I'm not able to see what the links used to direct to, but these seem to be relevant/correct replacements.

## Has this been tested?

### DIFF
--- a/support/doc/api/quickstart.md
+++ b/support/doc/api/quickstart.md
@@ -90,10 +90,4 @@ curl https://peertube.example.com/api/v1/videos
 
 ## Libraries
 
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Hi,

I think you can remove these links since we no longer generate these libraries?

scottruona:
Thanks - removed the links to specific language libraries. I kept the link to the OpenAPI specs and generator. Is it still valid to be using the OpenAPI generator, or is the OpenAPI spec out of date?

Chocobozzz:
Thank you! Yes, openapi generator should work. We just don't have enough resources to maintain the libraries.

PR #7574, opened 2026-04-21
Add smtp.bare_address config option. Solves #7569.

### LINKED ISSUE #7569: Provide option to strip display name in SMTP settings

I use ProtonMail as the backend for my PeerTube domain (self-hosted). They have a way to setup SMTP using tokens, but that is tied to the specific email address you use. I set everything up, but the emails still weren't going through.

It turns out that adding `${fromDisplayName}` in front of the email address causes Proton to think the from address doesn't match.

I will also reach out to Proton about this, but it would be nice to be able to strip the `${fromDisplayName}` part in the smtp config (something like `bareFromAddress: true/false`).

chiraag-nataraj:
## Description

Resolve #7569 by providing `smtp.bare_address`.

## Related issues

See #7569.

## Has this been tested?

### DIFF
--- a/server/core/initializers/config.ts
+++ b/server/core/initializers/config.ts
@@ -117,7 +117,8 @@ const CONFIG = {
     TLS: config.get<boolean>('smtp.tls'),
     DISABLE_STARTTLS: config.get<boolean>('smtp.disable_starttls'),
     CA_FILE: config.get<string>('smtp.ca_file'),
-    FROM_ADDRESS: config.get<string>('smtp.from_address')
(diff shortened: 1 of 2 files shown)


chiraag-nataraj:
@Chocobozzz Does this look okay?

chiraag-nataraj:
Whoops, fixed the silly issue @Chocobozzz :)

Chocobozzz:
Thank you, but as stated in https://github.com/Chocobozzz/PeerTube/issues/7569#issuecomment-4313251543 I'd prefer to not add another config key to peertube to fix this external provider bug.

PR #7588, opened 2026-05-05
Show real `count_view_after` in video stats Views help popup

### LINKED ISSUE #7580: Statistics -> Views popup help doesn't take account of custom count_view_after value

When looking at the statistics of a video, for "Views" area "?" help button says 

"A view means that someone watched the video for several seconds (10 seconds by default)":
Image

It would be nice if this displayed the value "count_view_after" from the `production.yaml`, in case it is set to a custom value.

### Steps to reproduce

1. Navigate to a video, click on Manage, and then Statistics
2. Click on the "?" in the "Views" section
3. Help text is displayed, which seems to be static and says views will be counted after "(10 seconds by default)"
4. This may not be the same as what is specified in `production.yaml`.
5. In my case, I set it to 30 seconds, but the help text still says "10"

It would be nice if the popup text displayed the actual value from the configuration file in case the default is changed.
For long videos, 10 seconds isn't a good measure of someone actually watching it, so I set it to 30 seconds, which is more reasonable for my set up.

### Additional information

* PeerTube instance:
  * URL: 
  * Version: 8.1.5
  * NodeJS version: v20.20.2
  * Ffmpeg version: 4.4.2-0ubuntu0.22.04.1+esm11

* Firefox 64 bit v136 on Ubuntu 64bit 20.04.6

mvanhorn:
Closes #7580

## Why

The Views card on the per-video statistics page rendered a help popup that hard-coded `(10 seconds by default)` regardless of the instance's `views.videos.count_view_after` setting in `production.yaml`. Operators who raised that threshold (commonly to 30s for longer-form videos, like the issue reporter) saw help text that contradicted what the server actually counted.

## What

Expose `views.videos.countViewAfter` (in seconds) on the public `ServerConfig` returned by `GET /api/v1/config` and rewrite the help text in `video-stats.component.ts` to interpolate the live value. The new copy also names the underlying setting (`count_view_after`) so operators can find it.

## Files

- `packages/models/src/server/server-config.model.ts` - add `views.videos.countViewAfter: number` to the `ServerConfig` interface (also flows into `HTMLServerConfig` via the existing `Omit`).
- `server/core/controllers/api/config.ts` - populate it from `CONFIG.VIEWS.VIDEOS.COUNT_VIEW_AFTER` (which is in ms internally), divided by 1000 for seconds-on-the-wire.
- `client/src/app/+videos-publish-manage/shared-manage/stats/video-stats.component.ts` - read `serverService.getHTMLConfig().views.videos.countViewAfter` in `buildOverallStatCard` and use it in the `$localize` help template.

## Notes

- Translation source string changes; existing `.xlf` entries for the old wording become stale and will retranslate via the normal Weblate pipeline.
- Value is in seconds on the wire because the help text reads in seconds; matches the unit operators see in `production.yaml` (`'10 seconds'`).

### DIFF
--- a/client/src/app/+videos-publish-manage/shared-manage/stats/video-stats.component.ts
+++ b/client/src/app/+videos-publish-manage/shared-manage/stats/video-stats.component.ts
@@ -333,11 +333,13 @@ export class VideoStatsComponent implements OnInit {
   }
 
   private buildOverallStatCard (overallStats: VideoStatsOverall) {
+    const countViewAfter = this.serverService.getHTMLConfig().views.videos.countViewAfter
+
     this.globalStatsCards = [
       {
         label: $localize`Views`,
         value: this.numberFormatter.transform(this.videoEdit.getVideoAttributes().views),
-        help: $localize`A view means that someone watched the video for several seconds (10 seconds by default)`
+        help: $localize`A view means that someone watched the video for at least ${countViewAfter} seconds (configured via "count_view_after" in the instance configuration).`
       },
       {
         label: $localize`Downloads`,
--- a/packages/models/src/server/server-config.model.ts
+++ b/packages/models/src/server/server-config.model.ts
@@ -295,6 +295,12 @@ export interface ServerConfig {
     }
   }
 
+  views: {
+    videos: {
+      countViewAfter: number
+    }
+  }
+
   import: {
     videos: {
       http: {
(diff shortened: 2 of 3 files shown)


mvanhorn:
Closing this myself - TypeScript build is failing because there is already a `views` field on `ServerConfig` (with `videos.remote.maxAge`, `videos.local.maxAge`, `videos.watchingInterval`). I should have merged `countViewAfter` into the existing `videos` block instead of creating a duplicate. Will reopen with a corrected version after verifying locally.

PR #7604, opened 2026-05-13
allow others to download original file

### LINKED ISSUE #6698: option for everyone to download the original video file

when `Keep a version of the input file ` is enabled, only administrators are allowed to download it
image

add an option to make it accessible for public

ac615223s5:
## Description

if keep original file is enabled and the uploader enables it for the video

## Related issues

fix #6698

## Has this been tested?

### DIFF
--- a/client/src/app/+videos-publish-manage/shared-manage/customization/video-customization.component.html
+++ b/client/src/app/+videos-publish-manage/shared-manage/customization/video-customization.component.html
@@ -59,6 +59,10 @@
       </div>
 
       <my-peertube-checkbox inputName="downloadEnabled" formControlName="downloadEnabled" i18n-labelText labelText="Enable download"></my-peertube-checkbox>
+
+      @if (form.get('downloadEnabled').value) {
+        <my-peertube-checkbox inputName="downloadOriginalFileEnabled" formControlName="downloadOriginalFileEnabled" i18n-labelText labelText="Allow downloading the original file"></my-peertube-checkbox>
+      }
     </div>
 
   </div>
--- a/client/src/app/+videos-publish-manage/shared-manage/customization/video-customization.component.html
+++ b/client/src/app/+videos-publish-manage/shared-manage/customization/video-customization.component.html
@@ -58,11 +58,17 @@
         </my-select-player-theme>
       </div>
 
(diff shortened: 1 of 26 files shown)


client/src/app/+videos-publish-manage/shared-manage/customization/video-customization.component.html
```diff
@@ -59,6 +59,10 @@ <h2 class="form-title">
       </div>
 
       <my-peertube-checkbox inputName="downloadEnabled" formControlName="downloadEnabled" i18n-labelText labelText="Enable download"></my-peertube-checkbox>
+
+      @if (form.get('downloadEnabled').value) {
```
Chocobozzz: Use `<ng-container ngProjectAs="extra">` inside `<my-peertube-checkbox inputName="downloadEnabled" formControlName="downloadEnabled" i18n-labelText labelText="Enable download"></my-peertube-checkbox>` and disable the block instead of using `@if` so users can see this option exists.

Still use `@if` if there's no original file for the video (in upload mode -> check the config, in update mode -> check the video has an original file)

server/core/middlewares/validators/videos/videos.ts
```diff
@@ -478,6 +478,10 @@ export function getCommonVideoEditAttributes () {
       .optional()
       .customSanitizer(toBooleanOrNull)
       .custom(isBooleanValid).withMessage('Should have downloadEnabled boolean'),
+    body('downloadOriginalFileEnabled')
```
Chocobozzz, line 481: Please also check `downloadOriginalFileEnabled` consistency if `downloadEnabled` is `false`

server/core/middlewares/validators/shared/videos.ts
```diff
@@ -277,6 +277,10 @@ export async function checkCanAccessVideoSourceFile (options: {
     return true
   }
```
Chocobozzz, line 279: Also check `downloadEnabled` is `true`

Chocobozzz (changes requested):
Please:  
 * Add config existence check in `server/core/initializers/checker-before-init.ts`
 * Also check config consistency in `server/core/initializers/checker-after-init.ts` to ensure download enabled is `true` if `download_original_file_enabled` is true
 * Update `production.yaml.example` with the new config
 * Support federation so other instances can also provide original file download

ac615223s5:
I have made most of the changes, but federation support is beyond my current understanding.

PR #7606, opened 2026-05-13
remux support for more formats

ac615223s5:
## Description

av1, vp9, opus (audio) are remuxed for the original resolution
other resolutions still use transcoding

## Has this been tested?


Chocobozzz (commented):
Opus is not supported by iOS < 18.4, VP9/AV1 on only a subset of iOS hardware.

I think it should be instead an admin opt-in, and we should keep consistency when transcoding the video. For example, add configuration so admins decide they transcode all resolutions above 1080p to av1, and other ones to x264. Same for audio codecs.

ac615223s5:
I have added an option to configure which formats are supported.
image

Chocobozzz:
Thank you, but I still don't understand the point of muxing only the input file. What's the use case of this PR? Is it to reduce transcoding time, or is it to add other codecs support to PeerTube?

ac615223s5:
remux is faster and higher quality than reencoding. so if possible, it is better to just remux the input file.

Chocobozzz:
> remux is faster and higher quality than reencoding. so if possible, it is better to just remux the input file.

How do you handle the case of broken or slow playback of a muxed resolution? How does hls.js handle this case?

Issue #7386, opened 2026-01-18
Store caption text in database for efficient transcript retrieval


okaziya:
### Describe the problem to be solved

Caption/subtitle text is currently stored only in VTT files (filesystem or object storage). The `videoCaption` database table only holds file references (`filename`, `fileUrl`, `storage`), not the actual text content.

This creates problems for two use cases:

### 1. JSON-LD transcript embedding (SEO)

Issue #7385 proposes adding a `transcript` property to the JSON-LD VideoObject schema. To embed transcript text in JSON-LD, PeerTube would need to read VTT files from disk or S3 on every video page render - inefficient and problematic for object storage setups.

### 2. Full-text search within video content

PeerTube's search currently only matches against:
- Video title (fuzzy trigram search via `pg_trgm`)
- Tags (exact match)
- UUID

**It cannot search video descriptions or what was said in videos.**

This is a significant discoverability gap. Users cannot:
- Find videos by words in the description
- Find videos by phrases spoken in the video

Note: `video.description` is already in the database but inexplicably not included in search. Transcript text would need to be stored in the database before it could be searched.

Reference: Search implementation is in `server/core/models/video/sql/video/videos-id-list-query-builder.ts:620-664`

### 3. Other potential uses

- AI-powered video summarization
- Automatic chapter generation based on transcript content
- Content moderation (scanning transcripts for policy violations)

### Describe the solution you would like

Add a `text` column to the `videoCaption` database table to store the actual VTT content.

### Database Migration

Create `server/core/initializers/migrations/0930-caption-text.ts`:

```typescript
import * as Sequelize from 'sequelize'

async function up (utils: {
  transaction: Sequelize.Transaction
  queryInterface: Sequelize.QueryInterface
  sequelize: Sequelize.Sequelize
}): Promise<void> {
  await utils.queryInterface.addColumn('videoCaption', 'text', {
    type: Sequelize.TEXT,
    defaultValue: null,
    allowNull: true
  }, { transaction: utils.transaction })
}

function down (options) {
  throw new Error('Not implemented.')
}

export { down, up }
```

### Update VideoCaptionModel

In `server/core/models/video/video-caption.ts`, add the column declaration:

```typescript
@AllowNull(true)
@Column(DataType.TEXT)
declare text: string
```

### Populate on caption creation/update

Modify `server/core/helpers/captions-utils.ts` to return the file content:

```typescript
import { readFile } from 'fs/promises'

[...]

PR #7612, opened 2026-05-18
Improve video searchability by allowing to search through descriptions

oaj349:
## Description

Currently PeerTube searches through videos' titles, tags and basic metadata about related playlists and channels. This PR focuses on adding a possibility to search through videos' descriptions, making it possible to find videos that would otherwise not appear in results at all. Not everything can fit into a 100-ish character limit imposed on titles and 5 related tags at most (with each tag having to be shorter than 30 characters).

This could be eventually improved further by searching through transcripts, but I am yet to find a solution which will not make the PeerTube's search engine visibly slower. I have tested these changes locally as well as on my production PeerTube instance (based upon version 8.1.5) and found out that search quality by this change has improved, because... more videos, potentially relevant to my search query, started to appear in results at all.

### Technical details

Originally this was supposed to be a trigram-based query (full details are available in the PR's edit history), but @Chocobozzz mentioned the necessity of migrating to PostgreSQL full-text search. 

So that's what I have did and during testing the changes I found out that things worked out quite well. Search results seem to differ compared to previous proposal, but still these are promising.

## Related issues

This is a partial solution to issue #7386 (point 2 to be precise).

## Has this been tested?


Chocobozzz:
Hello,

It will break search performance on big instances. We must implement full text instead.

oaj349:
Hello,

>It will break search performance on big instances. We must implement full text instead.

Which engine would be used to perform full text search queries? Will this be based upon [PostgreSQL's full text search](https://www.postgresql.org/docs/current/textsearch-tables.html) or is another solution considered?

Chocobozzz:
Use PostgreSQL full text search. Admins have already the ability to plug an external search system (such as https://sepiasearch.org/ powered by meilisearch)

oaj349:
Alright, I have migrated to the FTS-based query; while results seem to differ in some cases compared to the trigram-based query, these seem to be quite fast. There is a new migration available (`1040-video-search-vector.ts`), which creates a new GIN index.

If any other complications arise please let me know.

server/core/initializers/migrations/1040-video-search-vector.ts
```diff
@@ -0,0 +1,50 @@
+import * as Sequelize from 'sequelize'
+
+async function up (utils: {
+  transaction: Sequelize.Transaction
+  queryInterface: Sequelize.QueryInterface
+  sequelize: Sequelize.Sequelize
+}): Promise<void> {
+  const { transaction } = utils
+
+  await utils.sequelize.query(
```
Chocobozzz: Prefer to create another table to store the vectors (`videoSearch` for example):
 * `video` table can have mean writes
 * we may want to support specific languages in `to_tsvector` function (instead of using `simple`)
 * we may add more data in the future (tags, channel name, etc.)

server/core/initializers/migrations/1040-video-search-vector.ts
```diff
@@ -0,0 +1,50 @@
...
+    'ALTER TABLE "video" ADD COLUMN "searchVector" tsvector',
+    { transaction }
+  )
+
+  await utils.sequelize.query(
+    `UPDATE "video" SET "searchVector" = to_tsvector('simple', unaccent(coalesce(name, '') || ' ' || coalesce(description, '')))`,
```
Chocobozzz: Move this in a migration script (`server/scripts/migrations`), it could be long to fill the column on big instances

server/core/initializers/migrations/1040-video-search-vector.ts
```diff
@@ -0,0 +1,50 @@
...
+    { transaction }
+  )
+
+  await utils.sequelize.query(
+    'CREATE INDEX "video_search_vector_idx" ON "video" USING GIN ("searchVector")',
```
Chocobozzz: No need to create the index in the migration, sequelize will do it automatically at statup

server/core/initializers/migrations/1040-video-search-vector.ts
```diff
@@ -0,0 +1,50 @@
...
+    { transaction }
+  )
+
+  await utils.sequelize.query(
+    `CREATE OR REPLACE FUNCTION "video_search_vector_update"() RETURNS trigger AS $$
```
Chocobozzz: Prefer creating the trigger/function at server startup.

server/core/initializers/migrations/1040-video-search-vector.ts
```diff
@@ -0,0 +1,50 @@
...
+    BEGIN
+      NEW."searchVector" := to_tsvector('simple', unaccent(coalesce(NEW.name, '') || ' ' || coalesce(NEW.description, '')));
```
Chocobozzz: Prefer `:= setweight(to_tsvector('simple', coalesce(NEW.name, '')), 'A') || setweight(to_tsvector('simple', coalesce(NEW.description, '')), 'B');`

server/core/models/video/sql/video/videos-id-list-query-builder.ts
```diff
@@ -817,32 +817,29 @@ export class VideosIdListQueryBuilder extends AbstractRunQuery {
...
     const escapedSearch = this.sequelize.escape(search)
-    const escapedLikeSearch = this.sequelize.escape('%' + search + '%')
-
-    this.queryConfig = 'SET pg_trgm.word_similarity_threshold = 0.40;'
+    const escapedTsQuery = this.sequelize.escape(
+      search.trim().split(/\s+/).map(w => `${w}:*`).join(' & ')
+    )
 
     this.cte.push(
-      '"trigramSearch" AS (' +
```
Chocobozzz, line 833: Please keep trigram search to handle typos

Chocobozzz (changes requested):

Chocobozzz:
Any news on this PR?

oaj349:
Hello, yes. Please see recent changes I have made in commits 0ad50ce and 54aa3b0 attached to this PR.

server/core/models/video/sql/video/videos-id-list-query-builder.ts
oaj349: I don't know how to maintain both FTS and trigram search while maintaining reasonable search speeds. By the way, weren't [trigrams frowned upon](https://github.com/Chocobozzz/PeerTube/pull/7612#issuecomment-4488337488)?

Chocobozzz: Do you observe performance issues with your new join?
No trigram is okay for me. It seems the only way to handle typos in postgresql

PR #7615, opened 2026-05-21
In v8.1.6, a new checkUnicastOrThrow method was added to youtube-dl-w…

LinuxRenaissance:
In v8.1.6, a new checkUnicastOrThrow method was added to youtube-dl-wrapper.js that passes the full URL to isResolvingToUnicastOnly(), but that function expects just a hostname. It passes it directly to dns.lookup().

## Description

I upgraded 8.1.5 to 8.1.6 and YouTube sync stopped working. In logs I saw that dns is trying to resolve my youtube channel URL instead of hostname. I selected "doesn't update server code, below, because I believe youtube sync is not that. If I am mistaken about the definition please let me know. I applied this small change on my server already. Works.

## Has this been tested?

### DIFF
--- a/server/core/helpers/youtube-dl/youtube-dl-wrapper.ts
+++ b/server/core/helpers/youtube-dl/youtube-dl-wrapper.ts
@@ -269,7 +269,7 @@ export class YoutubeDLWrapper {
   }
 
   private async checkUnicastOrThrow (userLanguage: string) {
-    if (!await isResolvingToUnicastOnly(this.url)) {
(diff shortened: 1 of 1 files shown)


Chocobozzz:
Thank you very much! I added the commit directly in the release branch so I can publish a 8.1.7 version: https://github.com/Chocobozzz/PeerTube/commit/129d0d8c9ab3bc803daef5ff388eccf887ebbf87

spirillen:
Shouldn't it be added to v8.2.0-rc.2(3) as well?

PR #7619, opened 2026-05-22
Fix upgrade.sh: safer symlink swap and validate VERSION

### LINKED ISSUE #7617: upgrade.sh: non-atomic symlink swap and missing `VERSION` validation can leave `peertube-latest` broken, causing nginx to serve 404 / `text/html` for client assets
### Describe the problem

The `dist/scripts/upgrade.sh` script can leave the `peertube-latest` symlink in a broken state under two conditions:

1. **Non-atomic symlink swap** — the script does `rm -rf $PEERTUBE_PATH/peertube-latest` followed by `ln -s ...`. If the script is interrupted between these two operations (Ctrl+C, network glitch, OOM, kill), `peertube-latest` is deleted and never recreated. nginx will then return HTTP 404 with `Content-Type: text/html` for every `/client/*.js` and `/client/*.css` request, breaking the frontend completely (the browser blocks the modules with `disallowed MIME type "text/html"`).

2. **No validation that `VERSION` is non-empty** — `VERSION` is computed from `curl -s https://api.github.com/...`. If the GitHub API call fails or returns an unexpected payload (rate limit, transient outage, schema change), `VERSION` ends up as an empty string. `set -u` does **not** catch this because the variable is *defined* (just empty). The script then creates a symlink pointing to `versions/peertube-` (a non-existent path), with the same end-user-visible result as case 1.

I personally hit case 2 today while upgrading from 8.1.6 to 8.1.7: the resulting symlink was literally `peertube-latest -> versions/peertube-` and the entire frontend was broken until I manually re-pointed it.

### Steps to reproduce

**Case 1 (non-atomic):**
```sh
# Simulate interruption after rm but before ln:
rm -rf /var/www/peertube/peertube-latest
# (Ctrl+C / kill here)
```
nginx now serves 404 with `text/html` MIME for all `/client/...` paths.

**Case 2 (empty VERSION):**
```sh
VERSION=""
LATEST_VERSION_DIRECTORY="/var/www/peertube/versions/peertube-${VERSION}"
ln -s "$LATEST_VERSION_DIRECTORY" /var/www/peertube/peertube-latest
# Now peertube-latest -> versions/peertube- (broken)
```

### Expected behavior

The symlink swap should be atomic and never leave the system in an inconsistent state. `VERSION` should be validated as non-empty before any filesystem operation that uses it.

### Proposed fix

In `dist/scripts/upgrade.sh`, replace:

```sh
rm -rf $PEERTUBE_PATH/peertube-latest
ln -s "$LATEST_VERSION_DIRECTORY" $PEERTUBE_PATH/peertube-latest
```

with an atomic update and add a guard on `VERSION`:

```sh
# After computing VERSION:
if [ -z "$VERSION" ]; then
  echo "Error - could not determine the version to install (GitHub API returned empty)"
  exit 1
fi

# Later, replace the rm + ln block with an atomic swap:
ln -sfn "$LATEST_VERSION_DIRECTORY" "$PEERTUBE_PATH/peertube-latest"

# And verify the result before continuing:
if [ ! -d "$(readlink -f "$PEERTUBE_PATH/peertube-latest")" ]; then
  echo "Error - peertube-latest symlink is broken after update"
  exit 1
fi
```

`ln -sfn` performs an atomic `rename(2)` at the syscall level on Linux, eliminating the window during which `peertube-latest` does not exist.

### Environment

- PeerTube versions involved: upgrade from 8.1.6 to 8.1.7 (but the buggy code exists on `develop` and on all releases I checked)
- OS: Debian 13 (Trixie)
- Shell: `/bin/sh` (dash)

I'd be happy to submit a PR if the approach above is acceptable. Let me know if you'd prefer a different style (e.g. using `mv -T` with a temp link instead of `ln -sfn`).

### Additional information

* PeerTube instance URL: https://tube.nicfab.eu
* PeerTube version: 8.1.7
* NodeJS version: v22.22.2
* Ffmpeg version: 7.1.4-0+deb13u1
* Redis version: 8.0.2
* PostgreSQL version: 17.10

nicfab:
Two issues in `scripts/upgrade.sh` could leave `peertube-latest` in a broken state, causing nginx to return 404 with `text/html` MIME type for `/client/*.js` and `/client/*.css` requests, breaking the frontend.

1. **Non-atomic symlink swap**: `rm -rf` followed by `ln -s` left a window during which `peertube-latest` did not exist, and could leave it missing entirely if the script was interrupted. Replaced with `rm -f` + `ln -s`, reducing the window to two consecutive syscalls (POSIX `ln` has no `-n` option, so a truly atomic swap is not portably achievable in sh).

2. **Missing VERSION validation**: if the GitHub API call fails or returns an unexpected payload, `VERSION` ends up empty and the symlink ends up pointing to `versions/peertube-` (non-existent path). Added an explicit guard before any filesystem operation.

A final integrity check on the resolved symlink target (POSIX `cd -P` + `pwd`) was also added to fail fast if anything goes wrong.

All constructs are POSIX-compliant and tested with `dash`.

Fixes #7617

## Description

Fix two issues in `server/scripts/upgrade.sh` that can leave `peertube-latest` symlink in a broken state, breaking the nginx-served frontend after an upgrade.

## Related issues

Fixes #7617

## Has this been tested?

- [x] 🙂 yes, manually verified on a production instance (Debian 13 trixie, `/bin/sh -> dash`) after hitting the bug while upgrading from 8.1.6 to 8.1.7. The fix produces a consistent symlink in all tested interruption scenarios.
- [ ] 😐 no, because this PR is a draft and still needs work
- [ ] 🙁 no, because this PR does not update server code
- [ ] 😶 no, because I need help


server/scripts/upgrade.sh
```diff
@@ -82,9 +90,17 @@ npm run install-node-dependencies -- --production
 
 OLD_VERSION_DIRECTORY=$(readlink "$PEERTUBE_PATH/peertube-latest")
 
-# Switch to latest code version
-rm -rf $PEERTUBE_PATH/peertube-latest
-ln -s "$LATEST_VERSION_DIRECTORY" $PEERTUBE_PATH/peertube-latest
+# Switch to latest code version (atomic symlink swap to avoid leaving
+# peertube-latest in a broken state if the script is interrupted)
+ln -sfn "$LATEST_VERSION_DIRECTORY" "$PEERTUBE_PATH/peertube-latest"
```
Chocobozzz: Unfortunately `-n` option is not POSIX

Same for `-f` option of `readlink`

nicfab: Good catch, thanks. Replaced `ln -sfn` with `rm -f` + `ln -s`: the swap is no longer strictly atomic, but the window is reduced to two consecutive syscalls (vs the original `rm -rf` + `ln -s`), and the VERSION guard prevents the broken-symlink case reported in #7617. Also replaced `readlink -f` with a `cd -P` + `pwd` resolution check. Both verified with dash on Debian 13.

PR #7621, opened 2026-05-22
Fix upgrade pg_dump using socket conn Chocobozzz/PeerTube#7618

### LINKED ISSUE #7618: upgrade.sh fails when using postgres database socket

In config/production.yaml I have database.hostname: '' which makes peertube connect to postgres via socket instead of network port. Peertube itself works with this setup.

But it looks like the upgrade script doesn't check if hostname is populated and passes it to pg_dump as-is.

```
/var/www/peertube/peertube-latest/scripts# sudo -H -u peertube ./upgrade.sh
Backing up PostgreSQL database in /var/www/peertube/backup/sql-peertube_prod-20260522-1606.bak
pg_dump: error: too many command-line arguments (first is "peertube_prod")
pg_dump: hint: Try "pg_dump --help" for more **information.**
```

If I specify '127.0.0.1' as hostname the upgrade script does not fail.

### Steps to reproduce

1. Specify database:hostname as '' in production.yaml
2. Run instance upgrade command from https://docs.joinpeertube.org/install/any-os#upgrade
3. Script fails with pg_dump arguments error

Upgrade scipt doesn't fail when database:hostname is an empty string

### Additional information

* PeerTube instance:
  * URL:
  * Version: 8.1.5
  * NodeJS version:
  * Ffmpeg version:

Amlor:
## Description

Fixes #7618 

When empty hostname in database config in production.yaml upgrade.sh will error out with pg_dump complaining about unexpected arguments.
This PR makes it so on empty hostname pg_dump will be called without network connection related arguments.

## Related issues

#7618 

## Has this been tested?

### DIFF
--- a/server/scripts/upgrade.sh
+++ b/server/scripts/upgrade.sh
@@ -40,14 +40,29 @@ if [ -x "$(command -v pg_dump)" ]; then
 
   echo "Backing up PostgreSQL database in $SQL_BACKUP_PATH"
 
-  DB_USER=$(node -e "console.log(require('js-yaml').load(fs.readFileSync('$PEERTUBE_PATH/config/production.yaml', 'utf8'))['database']['username'])")
-  DB_PASS=$(node -e "console.log(require('js-yaml').load(fs.readFileSync('$PEERTUBE_PATH/config/production.yaml', 'utf8'))['database']['password'])")
   DB_HOST=$(node -e "console.log(require('js-yaml').load(fs.readFileSync('$PEERTUBE_PATH/config/production.yaml', 'utf8'))['database']['hostname'])")
-  DB_PORT=$(node -e "console.log(require('js-yaml').load(fs.readFileSync('$PEERTUBE_PATH/config/production.yaml', 'utf8'))['database']['port'])")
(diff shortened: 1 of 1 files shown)


server/scripts/upgrade.sh
```diff
@@ -40,14 +40,29 @@ if [ -x "$(command -v pg_dump)" ]; then
...
+  }
+
+  # Empty hostname means that we should connect using default postgres socket using peer auth
+  # For non-empty host we are using network connection with credentials
+  if [ "$DB_HOST" = "" ]; then
+
+    pg_dump_cmd
+  else
+
+    DB_USER=$(node -e "console.log(require('js-yaml').load(fs.readFileSync('$PEERTUBE_PATH/config/production.yaml', 'utf8'))['database']['username'])")
```
Amlor, line 58: @Chocobozzz I'm not sure if user/port/pass should be ignored altogether.

There are 3 ways to connect that I know:
1. System user "username" connects as the postgres user "username" via socket using "peer" auth, which is enabled by default.
2. Any user connects to some host's network port using credentials
3. Any user connects to socket using credentials. This is NOT enabled by default in postgres and will fail even with valid credentials unless allowed.

This PR covers cases 1 and 2 but doesn't support case 3.

Chocobozzz:
Thank you

Issue #1600, opened 2019-02-03
Recommended videos strategies


forteller:
Hi all. Thanks for creating an amazing product! It's much needed and well made!

I just recently tried it as a content creator for the first time, and although the experience was mostly really good, one thing also kindof upset me.

There's been a lot of talk for a while now of how YouTube's recommendation engine is ruining the world by pushing extreme conspiracy theories on everyone. You start out watching a video about some mainstream rightwing politician, and you get recomended a video about someone a bit further out. Then something even further out, and soon you're watching a documentary about how the jews are controlling the world economy trough reptilian clones or something just as crazy.

The same thing happens with islamic terrorist content and left wing nutjobs. You get fed the most crazy ideas very fast.

I tagged my first video uploaded to Peertube with "klima", the norwegian (and german) word for "climate". In the list of recommended "other videos", this showed up:

![screenshot from 2019-01-31 22-42-01](https://user-images.githubusercontent.com/343561/52170753-6fea3e00-2750-11e9-89bb-d2c5a1184953.png)

This is what your recomendation engine wanted people to watch next. Your engine, not you personally! But dangerous ideas like this gets spread automatically by a feature in your software. An unnecessary feature.

You are creating a tool for uncensored sharing of videos. Surely you know that people will use it for all kinds of crazy, dangerous shit. You're already talking about the so called "nazi problem" in #430, and discussing the possibility of automatic blacklisting in #1538.

You shouldn't want to have to go in and police everything people post. You shouldn't even have the option to do that. Certainly you will not have the manpower. I agree with that! But at the same time, you shouldn't help dangerous ideas more than necessary.

How can you reconcile these two things? By _turning off recomendations outside of the persons own videos_ by default!

That is: When someone installes Peertube on a new server, the default setting should be that the "Other videos" feature only shows other videos from the same account as the current video you're watching was posted by.

There could be a per user setting to change this, if you want. I don't know. But I do know that I had to remove the climate tag from my video to stop this dangerous chemtrails conspiracy theory from being propagated to everyone watching my video, and that I really don't want you to have the same bad impact on the world in this regard as YouTube has.

Thank you very much for your time and consideration!

PS: Via Mastodon I asked my instance, pe.ertu.be, if they could turn off this feature. They said they could not.

forteller:
[...]

PR #7633, opened 2026-05-28
Recommendations policy

immae:
## Description

This commit adds a new option attached to a video, that allows to chose  which kind of recommendation will appear below it in the viewer.  It currently allows four setups:
 - Any video (same as currently)
 - Local videos only
 - Owner videos only
 - Same channel videos only
    
This allows the user who owns the video to restrict the recommendations associated to his videos.

AI disclaimer: Tests were written in big part with AI help. It also sketched me a draft of the feature implementation but no line of this draft remain in this PR, it helped mostly with providing indications and relevant code locations since I’m not familiar with the project.

## Related issues

https://github.com/Chocobozzz/PeerTube/issues/1600

## Has this been tested?

- [X] 👍 yes, I added tests to the test suite

### DIFF
--- a/client/src/app/+video-watch/shared/recommendations/video-recommendation.service.ts
+++ b/client/src/app/+video-watch/shared/recommendations/video-recommendation.service.ts
@@ -5,7 +5,7 @@ import { Video } from '@app/shared/shared-main/video/video.model'
 import { VideoService } from '@app/shared/shared-main/video/video.service'
 import { AdvancedSearch } from '@app/shared/shared-search/advanced-search.model'
 import { SearchService } from '@app/shared/shared-search/search.service'
-import { HTMLServerConfig } from '@peertube/peertube-models'
+import { HTMLServerConfig, VideoRecommendationPolicy } from '@peertube/peertube-models'
 import { Observable, of } from 'rxjs'
 import { map, switchMap } from 'rxjs/operators'
 
@@ -59,11 +59,16 @@ export class VideoRecommendationService {
     return this.userService.getAnonymousOrLoggedUser()
       .pipe(
         switchMap(user => {
-          const defaultSubscription = this.videos.listVideos({
+          const defaultSubscription = this.videos.listRecommendationVideos({
             skipCount: true,
             videoPagination: pagination,
-            sort: '-publishedAt'
+            sort: '-publishedAt',
+            currentVideo
           }).pipe(map(v => v.data))
+
+          if (currentVideo.recommendationPolicy.id !== VideoRecommendationPolicy.ANY_VIDEOS) {
+            return defaultSubscription
+          }
 
           const searchIndexConfig = this.config.search.searchIndex
           if (searchIndexConfig.enabled === true && searchIndexConfig.disableLocalSearch === true) {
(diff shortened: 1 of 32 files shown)


Chocobozzz:
Thank you. Before reviewing this PR in detail, please:

 * Prefer using a `/api/v1/recommendations/videos` route that accepts a `strategy` query param. `strategy`, for now, would accept a `VideoRecommendationPolicy` (that we can rename `VideoRecommendationStrategy`. We can imagine more recommendation strategies in the future, the client will decide
 * Allow the admin to specify a default recommendation strategy, like we already do for some video metadata (download enabled, etc.)
 * Add federation support so remote instances can also choose which recommendation strategy they need to choose

immae:
>     * Prefer using a `/api/v1/recommendations/videos` route that accepts a `strategy` query param. `strategy`, for now, would accept a `VideoRecommendationPolicy` (that we can rename `VideoRecommendationStrategy`. We can imagine more recommendation strategies in the future, the client will decide

Just to make sure I understand this point correctly: The current endpoint format (as I implemented it) is '/api/v1/videos/{uuid}/recommendations', and you are talking about '/api/v1/recommendations/videos', is it just an omission in your comment and you actually meant '/api/v1/videos/{uuid}/recommendations/videos'? (because we already have uuid-namespaced endpoints, so it made sense to me to namespace it in this case too).

Then: is your request an addition to what I implemented? so by default we would use the video (or admin) policy, unless a query param is specified in which case we use the query param? Or is it in place of it? (so I would need to pass the preference to the client and then the client decides the strategy when it queries the recommendation, and the endpoint itself would be actually "stateless"?)

immae:
> Add federation support so remote instances can also choose which recommendation strategy they need to choose

Maybe I miss some federation feature, but I don’t see how this preference makes sense in a federation context, do you have an example scenario in mind?

Chocobozzz:
> is it just an omission in your comment

I would like to create a recommendation endpoint under `/api/v1/recommendations/videos`. This endpoint can be used, in the future, to also recommend videos to users based on their recent interactions (and so not related to a specific video ID).

> Then: is your request an addition to what I implemented? 

I suggest having a stateless REST API endpoint, accepting parameters to the client choose the recommendation algorithm it wants. Your PR implementation allows the uploader to decide the recommended algorithm, but we may envisage in the future that the viewer can also override the algorithm. It will be up to the client to decide which preference it wants to apply.
So yes: 
 1. client fetch recommendation strategy (from video or from instance)
 2. sends the strategy + needed query params (like the video ID) to the REST API
 3. the REST API responds video objects according to strategy and query params

> do you have an example scenario in mind?

You upload a video on peertube1.example.com and ask PeerTube to recommend only local videos to viewers. Viewers watching the video on peertube2.example.com should see videos of peertube1.example.com as recommendations

immae:
Understood, I’ll see how to do that

Chocobozzz:
Any news on this PR?

immae:
> Any news on this PR?

I’m lacking time to work on it for now, sorry.
If you don’t want to keep the MR open you may close it, I cannot give a timeline for when I’ll be back to it

(note: two weeks for the first review was a bit too long for me: I tried looking at the code after the review and since I’m not a regular contributor it was as if I saw it for the first time, so I gave up and didn’t have courage to go back to it since then)

Chocobozzz:
> (note: two weeks for the first review was a bit too long for me: I tried looking at the code after the review and since I’m not a regular contributor it was as if I saw it for the first time, so I gave up and didn’t have courage to go back to it since then)

I know and I'm sorry about that. I'll try to decrease review lag in the future (I was on holidays when you created the PR)

Issue #7654, opened 2026-06-07
Cross authentication between multiple auth plugins


asyncnomi:
### Describe the problem to be solved

When a user is registered with one authentication source it is automatically refused login from another auth plugin.
One reason is surely to avoid account takeover between multiples auth sources.

However in some cases where you control all sources of authentication it could be great to allow such behavior.
Some examples:
* You have multiples providers users can use to login on your infrastructure, you control all of them, but some users are present in two or more sources. In that case it would be nice that such users can use whatever providers they have an account on, to login (and not be limited to the one they first registered)
* For migrations purposes. Let's say you want to migrate from LDAP to OIDC, and all users got a new account on the OIDC provider (automatically created from the LDAP). Instead of having to manually update the db to allow the user to use the new provider. You can simply allow cross-authentication which will updates the user's pluginAuth attribute when they login using the new provider.

### Describe the solution you would like

A new advanced option that enable/disable cross-authentication between auth related plugins

I attempted an implementation in: #7655

Chocobozzz:
Implemented in https://github.com/Chocobozzz/PeerTube/pull/7655

PR #7655, opened 2026-06-08
Add cross-auth option

asyncnomi:
## Description

Add new advanced option that enable/disable cross-authentication between auth related plugins

When a user is registered with one authentication source it is automatically refused login from another auth plugin.
One reason is surely to avoid account takeover between multiples auth sources.

However in some cases where you control all sources of authentication it could be great to allow such behavior.
Further context is given in the associated issue.

## Related issues

#7654

## Has this been tested?

- [x] 🙋 kinda, but I need help 

I only added the basic static config test, it might be possible to add a dedicated test using openldap & keycloak, does it worth it ?
Nonetheless, the code was tested on a running server with LDAP, OIDC & SAML auth provider.
All three cases have been tested:
* Register with LDAP, then login with OIDC & SAML
* Register with OIDC, then login with LDAP & SAML
* Register with SAML then login with LDAP & OIDC

## Remarks

* It's the first time I edit the server code, there might be issues, or stuff I got wrong.
* I created a new top level option `advanced_auth` however, I'm not sure if this is the best place for the new `allow_cross_auth` option.
* I didn't update i18n, there was way more diff then what I was expecting.

## Screenshots

image


client/src/app/+admin/config/pages/admin-config-advanced.component.html
```diff
@@ -50,4 +50,38 @@ <h2 i18n>TWITTER/X</h2>
...
+      <h2 i18n>AUTHENTICATION</h2>
+      <div i18n class="inner-form-description">
+        Extra auth configuration behaviors
+      </div>
+    </div>
+
+    <div class="content-col">
+      <ng-container formGroupName="advancedAuth">
+        <div class="form-group">
+          <my-peertube-checkbox
```
Chocobozzz: Since it's a *very advanced* option with a big warning, I think we can keep it in the yaml/env configuration without exposing it in the web config

asyncnomi: Thanks for your review,
You're probably right, too few people will ever make use of it to justify the creation of a whole section in the web config.
I reverted the files related to that part

config/default.yaml
```diff
@@ -443,6 +443,11 @@ plugins:
     check_latest_versions_interval: '4 hours' # How often you want to check new plugins/themes versions
     url: 'https://packages.joinpeertube.org'
 
+advanced_auth:
```
Chocobozzz: `auth`

config/default.yaml
```diff
@@ -443,6 +443,11 @@ plugins:
...
+  # Enable cross authentication when using multiple auth plugins
+  # Never enable this unless you known what you're doing, it can leads to account takeover if not used correctly
+  allow_cross_auth: false
```
Chocobozzz: `allow_cross_provider_auth`

Chocobozzz (commented):
Can you add tests in `packages/tests/src/plugins/external-auth.ts`?

asyncnomi:
> Can you add tests in `packages/tests/src/plugins/external-auth.ts`?

I added a test, I used the already declared `cib` user in fixtures, If you rather used a dedicated one let me know

Issue #7645, opened 2026-06-01
Pickup Redis via Unix socket in Docker container


VolkerStolz:
### Describe the problem to be solved

The Docker `.env`-file currently only propagates _some_ Redis-related variables (controlled by `custom-environment-variables.yaml`) . It'd be nice to have the custom variable for a socket exposed as well via the single-line addition of `socket: PEERTUBE_REDIS_SOCKET` in the `redis:` section, since I like to mount my databases via sockets.

A quick test by editing `/app/support/docker/production/config/custom-environment-variables.yaml` inside the container followed by a `docker commit ... && <update compose-file to use new image> && docker-compose down && .. up` shows immediate success:
```
{"db":0,"host":"172.10.0.1","port":6379,"path":"/var/run/redis-server.sock","tags":["redis"],"level":"info","message":"Using standalone redis options", ...
```

### Describe the solution you would like

Add the required line to `custom-environment-variables.yaml`.

Should not have any impact anywhere unless folks go around actually using it in their `.env`.

Chocobozzz:
Could you please tell me what configuration is missing in https://github.com/Chocobozzz/PeerTube/blob/develop/support/docker/production/config/custom-environment-variables.yaml#L123?

Chocobozzz:
Thank you: https://github.com/Chocobozzz/PeerTube/commit/2d6f5777a3781dae7d389a792fb062280a5b433a

PR #7657, opened 2026-06-08
Expose socket

VolkerStolz:
See https://github.com/Chocobozzz/PeerTube/issues/7645


Chocobozzz:
Thank you, it will be included in 8.3 with this commit: https://github.com/Chocobozzz/PeerTube/commit/2d6f5777a3781dae7d389a792fb062280a5b433a

PR #7661, opened 2026-06-12
Add null guard for channelItem in VideoCustomizationComponent.ngOnInit to prevent crash when the video's channel is not 

### LINKED ISSUE #7660: Video manage "Customization" tab crashes with infinite loading when editing a video whose channel is not in the current user's channel list

Confirmed on PeerTube v8.2.0.

On the video manage page, the "Customization" tab (`/videos/manage/<id>/customization`) fails to render when the video's channel does not belong to the logged-in user; the typical case being an admin or moderator editing another user's video.

**Symptoms:**
- The top progress bar loads infinitely and never completes.
- Most of the tab's content disappears.
- The "Filename" field (which should be disabled and pre-filled with the source filename) renders empty and editable instead.
- The same tab works perfectly when editing a video from the user's own channel.

**Root cause:** in `VideoCustomizationComponent.ngOnInit()` (`video-customization.component.ts`, around line 73), the component looks up the video's channel in `userChannels` and reads `.name`/`.label` without a null-guard:

```ts
const channelItem = userChannels.find(c => c.id === videoEdit.toCommonFormPatch().channelId)
this.videoChannel = { name: channelItem.name, displayName: channelItem.label }
```

`userChannels` only contains the logged-in user's own channels (plus collaborations), built by the route resolver via `listUserChannelsForSelect`. When the video belongs to a channel not in that list, `userChannels.find(...)` returns `undefined`, and reading `channelItem.name` throws `TypeError: Cannot read properties of undefined (reading 'name')`.

**The cascade:** that exception aborts `ngOnInit` before `this.buildForm()` runs, so `this.form` stays `undefined`. The template uses `[formGroup]="form"`, which then fails, and the template also calls `hasPublicationDate()` (line 128):

```ts
hasPublicationDate () {
  return !!this.form.value.originallyPublishedAt
}
```

With `this.form` undefined, this throws `Cannot read properties of undefined (reading 'value')` on every change-detection cycle, in a loop. That loop is what keeps the loading bar spinning forever and leaves the tab half-rendered.

**Internal inconsistency (relevant):** the sibling component `video-main-info.component.ts` performs the identical `userChannels.find(c => c.id === channelId)` lookup but guards every use (e.g. `if (!channel) return ''`, `if (!newChannel) return`). So the project already handles this case elsewhere; the customization component simply omits the guard.

### Steps to reproduce

1. Log in as an admin (or moderator).
2. Open the manage page of a video that belongs to a different user's channel (a channel not in your own channel list) and go to the Customization tab: `/videos/manage/<videoId>/customization`
3. The tab never finishes loading (progress bar hangs), most content is missing, and the "Filename" field shows empty and editable. The browser console shows `TypeError: Cannot read properties of undefined (reading 'name')` at `video-customization.component.ts:73`, followed by a repeating loop of `(reading 'value')` from `hasPublicationDate` at line 128. The other manage tabs of the same video (Main info, Stats, etc.) load normally.

The Customization tab should load normally; like the other manage tabs already do for a non-owned video; without throwing, even when the video's channel is not in the logged-in user's channel list.

### Additional information

* PeerTube instance:
  * Version: 8.2.0
  * NodeJS version: v22.22.2
  * Ffmpeg version: 6.1.1 (Ubuntu build 6.1.1-3ubuntu5)
* Browser name, version and platforms on which you could reproduce the bug: Reproducible on any browser/platform; it is a client-side JavaScript logic error (missing null-guard), browser-independent.
* Link to browser console log if relevant: `TypeError: Cannot read properties of undefined (reading 'name')` at `VideoCustomizationComponent.ngOnInit` (`video-customization.component.ts:73`), cascading into a repeating `TypeError: Cannot read properties of undefined (reading 'value')` at `hasPublicationDate` (`video-customization.component.ts:128`) on every change-detection cycle.
* Link to server log if relevant: Not relevant; this is a front-end (Angular) error; nothing is logged server-side.

**Suggested fix:** guard `channelItem` before use (and/or derive `this.videoChannel` from the video's own channel attributes as a fallback so the player theme selector keeps working):

```ts
const channelItem = userChannels.find(c => c.id === videoEdit.toCommonFormPatch().channelId)
if (channelItem) {
  this.videoChannel = { name: channelItem.name, displayName: channelItem.label }
}
```

Federicorao:
Fixes #7660.

Add null guard for channelItem in VideoCustomizationComponent.ngOnInit to prevent crash when the video's channel is not in the user's channel list. The component assumes `userChannels.find(...)` always returns a channel, but when an admin/moderator edits a video belonging to another user's channel, `channelItem` is `undefined`. This causes a TypeError when accessing `.name` and `.label`, aborting `ngOnInit` before `buildForm()` runs, leading to infinite loading. The sibling component `video-main-info.component.ts` already guards such lookups. Adding a simple null check prevents the crash and allows the form to be built.

I ran the available lightweight check for the frontend change.

### DIFF
--- a/client/src/app/+videos-publish-manage/shared-manage/customization/video-customization.component.ts
+++ b/client/src/app/+videos-publish-manage/shared-manage/customization/video-customization.component.ts
@@ -70,7 +70,9 @@ export class VideoCustomizationComponent implements OnInit, OnDestroy {
     this.videoEdit = videoEdit
 
     const channelItem = userChannels.find(c => c.id === videoEdit.toCommonFormPatch().channelId)
-    this.videoChannel = { name: channelItem.name, displayName: channelItem.label }
+    if (channelItem) {
+      this.videoChannel = { name: channelItem.name, displayName: channelItem.label }
+    }
 
     this.buildForm()
   }


Chocobozzz:
Thank you but it's not the appropriate fix for this issue. The admin must populate the channels select component using the video owner channels.

PR #7696, opened 2026-06-24
Fix/spa sourcemap 404

### LINKED ISSUE #7665: Opening DevTools floods the backend: relative sourceMappingURL in inline component styles (and aggregated plugin CSS) resolves against the page URL, and the SPA catch-all answers every phantom .css.map with 200 + full index.html instead of 404

TL;DR: Any visitor who simply has browser DevTools open makes the browser request dozens of `*.css.map` URLs that can never exist (e.g. `GET /videos/header.component-PHGXGNMY.css.map`), and PeerTube answers each one with HTTP 200, text/html, the full index.html, and `Cache-Control: max-age=0, no-cache, must-revalidate`, so neither the browser nor any CDN can cache it. On our production instance this traffic reached 52% of all bytes served by the backend in one day. It reproduces on every public instance tested, including framatube.org (run by the maintainers).

MECHANISM (two core behaviors combine)

1. The production build emits unresolvable sourceMappingURL references in inline component styles.

The release build runs `ng build --configuration production --source-map=true` (`scripts/build/client.sh`, lines 62 to 66; the flag overrides `"sourceMap": false` of the production configuration in `client/angular.json`; sourcemaps were intentionally added to production builds in #3410). With that flag, Angular emits component styles inlined into the JS chunks, each style string ending with a relative comment such as `/*# sourceMappingURL=video-rate.component-ELADOIMD.css.map */` (extracted verbatim from `client/dist/<locale>/chunk-*.js` of the 8.2.0 release; one chunk alone carries 17 of these).

At runtime Angular injects these styles as inline `<style>` elements into the document. When DevTools is open, the browser tries to fetch each referenced sourcemap, and because the style is inline in the document, it resolves the relative path against the PAGE URL, not against `/client/<locale>/`. Visiting `/videos/browse` therefore triggers requests like `GET /videos/action-dropdown.component-2EN6PTCP.css.map` and `GET /videos/video-miniature.component-JKRN343J.css.map`.

The irony: the `.css.map` files DO exist in `client/dist/<locale>/`, they are simply unreachable at the URLs the browser is told to resolve. These references can never work for any visitor on any page.

2. The SPA catch-all answers those phantom URLs with 200 + full index.html, uncacheable.

The requests fall into the client catch-all (`server/core/controllers/client.ts`: the clientsRouter using `serveIndexHTML` as async middleware). `serveIndexHTML` (`server/core/lib/html/client-html.ts`) has an Accept guard: `if (req.accepts(ACCEPT_HEADERS) === 'html' || !req.headers.accept) { serve index.html }`. But DevTools fetches sourcemaps with `Accept: */*`, and Express's `req.accepts(['html', 'application/json', ...])` returns `'html'` for `*/*`, so the guard passes and every phantom `.css.map` receives HTTP 200, `Content-Type: text/html`, `Cache-Control: max-age=0, no-cache, must-revalidate`.

Because of the no-cache, a CDN in front marks every one of these as dynamic and forwards it to the origin every time, for every visitor, on every reload.

SECOND FLAVOR of the same problem (aggregated plugin CSS): the plugin CSS aggregator (`server/core/lib/plugins/plugin-manager.ts`, `concatFiles()`) concatenates each plugin's CSS into `/plugins/global.css` as a verbatim byte stream, never stripping sourceMappingURL comments. The official livechat plugin ships `dist/assets/styles/style.css` ending with `/*# sourceMappingURL=style.css.map */`, so every instance running it serves a `global.css` whose relative reference resolves to the phantom `/plugins/style.css.map`, producing the same 200 + full index.html. Confirmed on framatube.org as well (their `/plugins/global.css` carries exactly that comment, since they run the official livechat plugin), and absent on instances without it (tube.tchncs.de's `global.css` has none). In other words, any plugin CSS built with standard tooling triggers it, and the aggregator is the right place to strip these comments.

MEASURED IMPACT (production instance, v8.2.0):
- A single page reload with DevTools open: about 25 to 40 phantom `.css.map` requests, interleaved with the page's real API calls (they compete for the same HTTP/2 connection and rate-limit budget).
- Worst measured minute (one reload): 161 `.map` requests / 16.7 MB.
- One day with one admin browsing with DevTools open: 949 `.map` requests = 9.5% of all backend requests and 52.0% of all bytes served.
- A day with no DevTools usage: zero `.map` requests, perfect correlation.
- Side effects beyond bandwidth: it masks genuine 404s and pollutes logs/metrics with thousands of bogus 200s.

CONFIRMED ON THIRD-PARTY INSTANCES (not a local misconfiguration). Running `curl -H 'Accept: */*' https://<instance>/videos/header.component-FAKE123.css.map` (a deliberately nonexistent name):
- framatube.org (v8.2.1): 200, text/html, 28,387 bytes
- peertube.tv (v8.2.1): 200, text/html, 22,696 bytes
- tube.tchncs.de (v8.1.8): 200, text/html, 24,020 bytes
- our instance, bare Node on 127.0.0.1:9000, no reverse proxy (v8.2.0): 200, text/html, 71,223 bytes

### Steps to reproduce

1. Run any production PeerTube instance (reproduced on 8.1.8, 8.2.0 and 8.2.1).
2. Open any page (e.g. `/videos/browse`) in Chrome/Edge/Firefox with DevTools open and reload. The Network tab (and the server logs) show dozens of `GET /<current-route>/<component>-<hash>.css.map` requests, each answered 200 text/html, no-cache.
3. Or reproduce with a single command, no browser needed: run `curl -i -H 'Accept: */*' https://<your-instance>/videos/anything.css.map` and you get HTTP 200, `Content-Type: text/html`, `Cache-Control: max-age=0, no-cache, must-revalidate`, plus tens of KB of index.html.

1. A request for a path ending in a static-asset extension, at the very least `.map`, that doesn't match a real file should get a short 404 from the SPA catch-all, not 200 plus the full app HTML marked uncacheable. No legitimate Angular route ends in `.map`, so this is safe. For example, before the catch-all: `if (/\.map$/i.test(req.path)) { return res.status(HttpStatusCode.NOT_FOUND_404).end() }`. This could optionally be generalized to a broader asset-extension regex like `\.(map|js|css|png|jpe?g|svg|webp|woff2?|ttf|ico)$`, since none of those are SPA routes either.

2. The production build should not emit sourcemap references that can never resolve. Since the broken references are exclusively the inline component styles (the `chunk-*.js.map` ones resolve fine relative to `/client/<locale>/`), building with `"sourceMap": { "scripts": true, "styles": false }` instead of the blanket `--source-map=true` keeps the useful script sourcemaps and stops emitting the phantom style references. Alternatively, emit style maps as hidden sourcemaps (no trailing comment), or rewrite the comments to absolute `/client/<locale>/<file>.css.map` paths at build time.

3. The plugin CSS aggregator (`concatFiles()` in plugin-manager) should strip `/*# sourceMappingURL=... */` comments when concatenating each plugin's CSS into `global.css`, so a plugin shipping a normal built CSS doesn't produce a phantom `/plugins/style.css.map` request.

4. (Nice-to-have) Add `map` to the static-files regex in `support/nginx/peertube` (line ~203: `location ~ ^/client/(.*\.(js|css|png|svg|woff2|otf|ttf|woff|eot))$`, which has no `map`) so that real `.js.map` files are served from disk with proper caching instead of going through Node on every fetch.

### Additional information

* PeerTube instance:
  * Version: 8.2.0 (third-party confirmations on 8.1.8 and 8.2.1)
  * NodeJS version: v22.22.2
  * Ffmpeg version: 6.1.1
* Browser name, version and platforms on which you could reproduce the bug: Chrome 149 and Edge 149 on Windows 10/11 (any browser with DevTools open; also reproducible with a plain `curl -H 'Accept: */*'`, see steps)
* Link to browser console log if relevant: Nothing is printed to the console. The flood is only visible in the Network tab and in the server logs, which is precisely why it goes unnoticed.
* Link to server log if relevant: representative excerpt from journalctl, one reload with DevTools open: `GET /videos/action-dropdown.component-2EN6PTCP.css.map` 200 71382, `GET /videos/video-miniature.component-JKRN343J.css.map` 200 71382, `GET /videos/app.component-5FEL6MNA.css.map` 200 71358, `GET /plugins/style.css.map` 200 71326

Related (not duplicates, searched sourceMappingURL, css.map, source map: 0 issues on this): #3410 (added sourcemaps to production builds, which introduced the inline style references).

BakedSoups:
## Description

  Fix missing sourcemap asset requests being served by the SPA HTML fallback.

 When DevTools requests route-relative sourcemaps like `/videos/header.component-FAKE.css.map` with `Accept: */*`, PeerTube
  currently returns `200 text/html` with the full `index.html`. This PR makes `.map` requests that reach the SPA fallback
  return `404` instead.

  It also strips `sourceMappingURL` comments from aggregated plugin CSS so `/plugins/global.css` does not point browsers to
  phantom plugin sourcemap URLs.

  ## Related issues

  Fixes #7665

  ## Has this been tested?

  Added regression tests for:
  - missing `.css.map` requests not returning index HTML
  - normal SPA route fallback still returning index HTML
  - plugin global CSS stripping `sourceMappingURL` comments

  Could not run the test suite locally because this checkout is missing local binaries (`mocha`, `oxlint`). `git diff
  --check` passed.

  ## Screenshots

  Not relevant.

### DIFF
--- a/client/angular.json
+++ b/client/angular.json
@@ -244,7 +244,10 @@
             "production": {
               "optimization": true,
               "outputHashing": "all",
-              "sourceMap": false,
+              "sourceMap": {
+                "scripts": true,
+                "styles": false
+              },
               "namedChunks": false,
               "extractLicenses": true,
               "serviceWorker": "src/ngsw-config.json",
--- a/packages/tests/fixtures/peertube-plugin-test-css-sourcemap/main.js
+++ b/packages/tests/fixtures/peertube-plugin-test-css-sourcemap/main.js
@@ -1,0 +1,12 @@
+async function register () {
+  return
+}
+
+async function unregister () {
+  return
+}
+
+module.exports = {
+  register,
+  unregister
+}
--- a/packages/tests/fixtures/peertube-plugin-test-css-sourcemap/package.json
+++ b/packages/tests/fixtures/peertube-plugin-test-css-sourcemap/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "peertube-plugin-test-css-sourcemap",
+  "version": "0.0.1",
+  "description": "Plugin CSS sourcemap test",
+  "engine": {
+    "peertube": ">=1.3.0"
+  },
+  "keywords": [
+    "peertube",
+    "plugin"
+  ],
+  "homepage": "https://github.com/Chocobozzz/PeerTube",
+  "author": "Chocobozzz",
+  "bugs": "https://github.com/Chocobozzz/PeerTube/issues",
+  "library": "./main.js",
+  "staticDirs": {},
+  "css": [
+    "./style.css"
+  ],
+  "clientScripts": [],
+  "translations": {}
+}
--- a/packages/tests/fixtures/peertube-plugin-test-css-sourcemap/style.css
+++ b/packages/tests/fixtures/peertube-plugin-test-css-sourcemap/style.css
@@ -1,0 +1,5 @@
+.css-sourcemap-fixture {
+  color: red;
+}
+
+/*# sourceMappingURL=style.css.map */
--- a/packages/tests/src/client/index-html.ts
+++ b/packages/tests/src/client/index-html.ts
@@ -83,6 +83,18 @@ describe('Test index HTML generation', function () {
   })
 
   describe('Static assets fallback', function () {
+    it('Should serve index HTML for a client route with wildcard accept header', async function () {
+      const res = await makeGetRequest({
+        url: servers[0].url,
+        path: '/videos/browse',
+        accept: '*/*',
+        expectedStatus: HttpStatusCode.OK_200
+      })
+
+      expect(res.headers['content-type']).to.contain('text/html')
+      expect(res.text).to.contain('<!doctype html>')
+    })
+
     it('Should not serve index HTML for a missing sourcemap asset', async function () {
       const res = await makeGetRequest({
         url: servers[0].url,
--- a/packages/tests/src/client/index-html.ts
+++ b/packages/tests/src/client/index-html.ts
@@ -82,6 +82,20 @@ describe('Test index HTML generation', function () {
     })
   })
 
+  describe('Static assets fallback', function () {
+    it('Should not serve index HTML for a missing sourcemap asset', async function () {
+      const res = await makeGetRequest({
+        url: servers[0].url,
+        path: '/videos/header.component-FAKE.css.map',
+        accept: '*/*',
+        expectedStatus: HttpStatusCode.NOT_FOUND_404
+      })
+
+      expect(res.headers['content-type'] || '').to.not.contain('text/html')
+      expect(res.text || '').to.not.contain('<!doctype html>')
+    })
+  })
+
   describe('Canonical tags', function () {
     it('Should use the original video URL for the canonical tag', async function () {
       for (const basePath of getWatchVideoBasePaths()) {
--- a/packages/tests/src/client/index-html.ts
+++ b/packages/tests/src/client/index-html.ts
@@ -82,32 +82,6 @@ describe('Test index HTML generation', function () {
     })
   })
 
-  describe('Static assets fallback', function () {
-    it('Should serve index HTML for a client route with wildcard accept header', async function () {
-      const res = await makeGetRequest({
-        url: servers[0].url,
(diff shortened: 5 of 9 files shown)


Chocobozzz:
Thank you

BakedSoups:
of course!

PR #7698, opened 2026-06-25
Add TwelveLabs Pegasus transcription engine

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## Description

This PR adds an opt-in **TwelveLabs Pegasus** transcription engine to `@peertube/peertube-transcription`, alongside the existing local Whisper engines.

PeerTube can already auto-transcribe videos with `openai-whisper` / `whisper-ctranslate2`, but those require running a local model (and ideally a GPU). This adds a remote alternative for admins who'd rather offload transcription: when `video_transcription.engine` is set to `twelvelabs`, PeerTube uploads the extracted audio/video to the TwelveLabs platform and asks the Pegasus video-understanding model to return a WebVTT transcript, which is stored exactly like Whisper output. No model download or local inference needed.

What it adds:
- A new `'api'` engine type (the existing ones are `'binary'`), and a `TwelveLabsTranscriber` implementing the same `AbstractTranscriber` contract (`transcribe()` → `TranscriptFile`, no-op `install()`).
- Registration in the engine list + `TranscriberFactory`, so it's selected the same way as the other engines.
- Config docs in `default.yaml` / `production.yaml.example` (regenerated `config-schema.json`).
- The `twelvelabs-js` SDK dependency.

**Opt-in / non-breaking:** the default engine is unchanged (`whisper-ctranslate2`). Nothing changes unless an admin explicitly sets `engine: 'twelvelabs'`. The API key is read from the `TWELVELABS_API_KEY` environment variable so the secret never lives in PeerTube's config files; if it's missing the engine fails fast with a clear message.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

> Note: the contributing guide suggests discussing features in an issue first. I'm opening this directly as a concrete, tested proposal — happy to move the discussion to an issue or adjust the approach (config plumbing for the key, format handling, etc.) based on maintainer preference.

## Related issues

None yet — opening as a proposal.

## Has this been tested?

Added `packages/tests/src/transcription/twelvelabs/twelvelabs-transcriber.spec.ts`:
- No-network unit tests: unsupported-format rejection and the `TWELVELABS_API_KEY` guard (run in CI without a key).
- An integration smoke test that's skipped unless `TWELVELABS_API_KEY` is set.

Locally:
- `tsc -b packages/transcription` builds clean.
- `oxlint` passes on the changed files.
- The factory + unit specs pass (smoke test pending without a key).
- The integration smoke test passes end-to-end against the real API: it uploaded the `the_last_man_on_earth.mp4` fixture, ran Pegasus 1.5, and got back a valid `WEBVTT` transcript in ~14s.

### DIFF
--- a/config/config-schema.json
+++ b/config/config-schema.json
@@ -2104,7 +2104,7 @@
           "default": false
         },
         "engine": {
-          "description": "Choose engine for local transcription Supported: 'openai-whisper' or 'whisper-ctranslate2'",
+          "description": "Choose engine for transcription Local engines: 'openai-whisper' or 'whisper-ctranslate2' Remote engine: 'twelvelabs' (TwelveLabs Pegasus, requires the TWELVELABS_API_KEY env variable - https://twelvelabs.io)",
           "type": "string",
           "default": "whisper-ctranslate2"
         },
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -816,8 +816,9 @@ video_transcription:
   # Enable automatic transcription of videos
   enabled: false
 
-  # Choose engine for local transcription
-  # Supported: 'openai-whisper' or 'whisper-ctranslate2'
+  # Choose engine for transcription
+  # Local engines: 'openai-whisper' or 'whisper-ctranslate2'
+  # Remote engine: 'twelvelabs' (TwelveLabs Pegasus, requires the TWELVELABS_API_KEY env variable - https://twelvelabs.io)
   engine: 'whisper-ctranslate2'
 
   # You can set a custom engine path for local transcription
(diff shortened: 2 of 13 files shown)


Chocobozzz:
Hello,

Thank you for this PR but unfortunately we can't accept it.

Prefer:
 * To extend PeerTube plugin API so admins can install plugins that add transcription capabilities
 * Or develop a program that uses the [PeerTube runners API](https://docs.joinpeertube.org/api-rest-reference.html#tag/Runner-Jobs/paths/~1api~1v1~1runners~1jobs~1request/post) so that admins would just have to register a TwelveLabs Pegasus runner on their instance

mohit-twelvelabs:
Totally understand, @Chocobozzz — thanks for the quick and clear response. A plugin via PeerTube's plugin API is the right home for this; we'll look at shipping it that way as an installable transcription plugin rather than touching core. Appreciate the direction!

PR #7716, opened 2026-07-07
Add ability to disable video embed

### LINKED ISSUE #7690: Add ability for video owner to disable video embed

Currently, embedding-related information can still be displayed even when embedding is not actually available or usable. This can result in users seeing UI elements that do not provide any meaningful or actionable information.

While the frontend and official configuration flow typically require an embed URL when embedding restrictions are enabled, it is still possible at database level to activate embedding restrictions without providing a URL. This can result in a state where only the restriction flag is active. In such cases, embedding-related information is still shown even though embedding is effectively unavailable.

This may lead users to assume that embedding is available when it is not, creating confusion and reducing the clarity of the video information page.

Embedding-related UI elements should only be displayed when embedding functionality is actually available and relevant.

Embedding section behavior

If embedding is disabled, all embedding-related fields and sections should be completely hidden.

If embedding restrictions are enabled but no valid embed URL exists, the embed section should not be displayed. In this state, there is no meaningful information for users and showing the section only creates confusion.

This is especially relevant when videos are intentionally intended to be accessed only through direct links or QR codes and embedding is not meant to be used at all.

A useful improvement would be an administrator setting that completely disables embedding for a video. When enabled, all embedding-related UI elements should be hidden entirely.

If embedding restrictions are enabled and a valid embed URL is configured, the UI should clearly display the domains that are permitted to embed the video. This helps users understand the embedding policy and prevents misunderstandings when embedding fails on unauthorized domains.

Update embed privacy for non logged in admins

The "Update embed privacy" functionality is only relevant for administrators. It does not provide any meaningful functionality for guests or regular members and should therefore not be visible or accessible to them.

tiagodefendi:
## Description

Video owners could only restrict embedding to an allowlist of domains, with no way to disable it entirely. This adds a `DISABLED` value to the existing `VideoEmbedPrivacyPolicy` enum (no migration needed, reuses the current allowlist infrastructure).

When set, embedding is blocked everywhere, including on the origin instance and for the video owner/editor, with no bypass. The "Embed" tab is hidden from the share modal (URL and QR code sharing stay available), and the moderation form's allowlist checkbox becomes a 3-option policy selector (anyone / allowlist / nobody).

## Related issues

Fix #7690 

## Has this been tested?

### DIFF
--- a/client/src/app/+videos-publish-manage/shared-manage/common/video-edit.model.ts
+++ b/client/src/app/+videos-publish-manage/shared-manage/common/video-edit.model.ts
@@ -14,6 +14,7 @@ import {
   VideoDetails,
   VideoEmbedPrivacy,
   VideoEmbedPrivacyPolicy,
+  VideoEmbedPrivacyPolicyType,
   VideoEmbedPrivacyUpdate,
   VideoImportCreate,
   VideoPrivacy,
@@ -81,7 +82,7 @@ type StudioForm = {
 type PlayerSettingsForm = PlayerVideoSettingsUpdate
 
 type EmbedPrivacyForm = {
-  videoPrivacyEmbedEnableAllowlist?: boolean
+  videoPrivacyEmbedPolicy?: VideoEmbedPrivacyPolicyType
   videoPrivacyEmbedAllowlistDomains?: string
 }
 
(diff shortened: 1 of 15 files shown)


tiagodefendi:
Hey @Chocobozzz , this is ready for review! Just pinging you to take a look whenever you have a moment. Let me know if you need any changes.

Chocobozzz:
Thank you

PR #7725, opened 2026-07-16
Feature/control bar hide

### LINKED ISSUE #7714: hide control bar after a few seconds when in pause

Hi,

I wanted to take a clean screenshot of a video in pause ( in full screen ) and I noticed the control bar never hides when in pause ( peertube 8.2.1 in firefox and librewolf ). I can do this in VLC and with arte.tv website player but I don't know how widespread this behavior in web players/media player apps.

Currently when playing the video in full screen the control bar seems to hide after 3 seconds, if the user press the space key to pause the video the control bar reappear and never hides again while the video is paused. If the user takes a screenshot the control bar is present in the .png/.jpg file.

Thanks  

_No response_

sourav7274:
## Description

 Fix control bar not fading when video is paused and user becomes inactive.

When a video was paused, the control bar was supposed to fade away after 3 seconds of inactivity, but it stayed visible permanently. The root cause was that schedulePauseHide called player.userActive(false) after adding vjs-paused-inactive, which triggered video.js's internal checkUserActivity_ loop to re-fire the useractive event (video.js keeps users "active" while paused by design). This caused onUserActive to cancel the hide timeout and reschedule it in an infinite loop, so the class was never kept on the element long enough for the CSS fade to take effect.

The fix removes the userActive(false) call and replaces the useractive event listener with direct DOM event listeners (mousemove, keydown, touchstart, click) so only genuine user input resets the inactivity timer 

At file peertube-plugin.ts , two variables  PAUSED_INACTIVE_TIMEOUT_NORMAL  , PAUSED_INACTIVE_TIMEOUT_FULLSCREEN  , control the duration ( in ms  ) when the fade out starts  

## Has this been tested?

## Screenshots

I don't have screenshots, but I do have clips, attaching those for reference, I have covered both normal mode and full screen mode

https://github.com/user-attachments/assets/a9a8a3cc-c9f2-4c09-8d28-00019f1439ef

https://github.com/user-attachments/assets/127cb897-d148-42c0-8666-7fe819e92d8c

Closes #7714


Chocobozzz:
Thank you

sourav7274:
Thank you for the chance, had a blast while doing this

PR #7733, opened 2026-07-21
fix(client/embed): use correct CSS variable for big play button color

### LINKED ISSUE #7731: bigPlayBackgroundColor embed param sets wrong CSS variable (--pt-player-big-play-background-color instead of --pt-player-big-play-bg)

The `bigPlayBackgroundColor` embed URL parameter does not change the big play button's background color, even though the value is set correctly.

Inspecting the embed iframe in DevTools shows the parameter's value is applied to a CSS variable named:
`--pt-player-big-play-background-color`

However, the player's stylesheet (`client/src/sass/include/_variables.scss`) reads a different variable for that element's background:
`--pt-player-big-play-bg`

Since the two names don't match, the value set from the URL param is never read by the stylesheet, so the button always renders with its default background.

### Steps to reproduce

1. Open an embed URL with the param set:
   `https://your-instance.example.com/videos/embed/VIDEO_UUID?bigPlayBackgroundColor=red`
2. Open DevTools → inspect computed styles on the big play button / `:root`.
3. Note `--pt-player-big-play-background-color` is set to `red`, but the button still shows the default background because the CSS reads `--pt-player-big-play-bg` instead.

The big play button's background should render in the color passed via `bigPlayBackgroundColor`.

### Additional information

* PeerTube instance:
  * URL: https://video.frc-cappital.co/
  * Version: 8.2.2

- Docs: https://docs.joinpeertube.org/api/embed-player#bigplaybackgroundcolor
- CSS variable definition: https://github.com/Chocobozzz/PeerTube/blob/develop/client/src/sass/include/_variables.scss
- Likely source of the mismatch: `client/src/standalone/player/src/` (embed param parsing)
- Happy to submit a PR fixing the variable name once confirmed.

cappitaldeveloper:
## Description

The `bigPlayBackgroundColor` embed URL parameter did not change the big play button's background color, even though the value was being read and applied correctly.

Root cause: `client/src/standalone/videos/embed.ts` (line 398) set the color on a CSS variable named `--pt-player-big-play-background-color`, but the player's stylesheet (`client/src/standalone/player/src/sass/shared/peertube-skin.scss`, line 56) reads a different variable via `pvar()`: `--pt-player-big-play-bg`. Since the two names never matched, the value from the URL param was applied but never actually read by anything.

This PR corrects `embed.ts` to set `--pt-player-big-play-bg` instead, matching the variable defined in `_variables.scss` and `_css-variables.scss` and actually consumed by the stylesheet.

Note: `peertube-skin.scss` (line 11) still declares a default for the now-unused `--pt-player-big-play-background-color` variable. Left untouched here to keep this PR minimal and focused — flagging in case maintainers prefer a follow-up cleanup.

## Related issues

Fixes #7731

## Has this been tested?

Manually verified: `?bigPlayBackgroundColor=red` on an embed URL now correctly tints the big play button, where it previously had no effect.

### DIFF
--- a/client/src/standalone/player/src/sass/shared/peertube-skin.scss
+++ b/client/src/standalone/player/src/sass/shared/peertube-skin.scss
@@ -8,8 +8,6 @@ body {
   --pt-player-overlay-bg: rgba(0, 0, 0, 0.6);
   --pt-player-overlay-bg-hover: rgba(255, 255, 255, 0.15);
 
-  --pt-player-big-play-background-color: rgba(0, 0, 0, 0.8);
-
   // Prevent z-index bug with Firefox PiP button
   // See:
   // * https://bugzilla.mozilla.org/show_bug.cgi?id=1742585
--- a/client/src/standalone/videos/embed.ts
+++ b/client/src/standalone/videos/embed.ts
@@ -395,7 +395,7 @@ export class PeerTubeEmbed {
     const body = document.getElementById('custom-css')
 
     if (this.playerOptionsBuilder.hasBigPlayBackgroundColor()) {
-      body.style.setProperty('--pt-player-big-play-background-color', this.playerOptionsBuilder.getBigPlayBackgroundColor())
+      body.style.setProperty('--pt-player-big-play-bg', this.playerOptionsBuilder.getBigPlayBackgroundColor())
     }
 
     if (this.playerOptionsBuilder.hasForegroundColor()) {


Chocobozzz:
Thank you

cappitaldeveloper:
@Chocobozzz Can you confirm in which release version this fix will be made live ?

Chocobozzz:
8.3

PR #7738, opened 2026-07-23
feat: add lastmod to sitemap entries for videos, channels and accounts

### LINKED ISSUE #7704: Lastmod in Sitemap.xml

PeerTube's XML sitemap currently does not include the <lastmod> element for URLs. As a result, search engines such as Google receive less information about when a video or page was last updated. This can reduce the efficiency of crawling, since crawlers cannot easily prioritize recently changed content and may recrawl unchanged pages unnecessarily while delaying updated ones.

For larger PeerTube instances with thousands of videos, this may negatively impact crawl efficiency and the speed at which updated metadata, descriptions, thumbnails, or other changes are reflected in search results.

Please add support for the <lastmod> element to all relevant URLs in the XML sitemap.

The value should reflect the resource's latest modification timestamp (for example, the video's updated date or the page's last modification date) and be formatted according to the XML Sitemap specification (ISO 8601 / W3C Datetime).

Adding <lastmod> would:

Improve crawl efficiency for search engines.
Help search engines prioritize recently updated content.
Potentially speed up re-crawling after metadata or content changes.
Bring PeerTube's sitemap closer to XML Sitemap best practices without affecting existing functionality.

This would be a relatively small enhancement with meaningful SEO benefits, especially for medium and large PeerTube instances.

valeron90:
Adds `lastmod` to sitemap entries for videos, video channels and accounts.  
fixes #7704 
`lastmod` is set to `updatedAt`, which Sequelize updates on every `save()` call.  
This covers relevant content changes (title, description, tags, category, thumbnail)  
but also some non-content changes (download settings, comment policy, privacy).  

A more precise solution would be a dedicated `contentUpdatedAt` field updated  
only on relevant content changes — but that would require a database migration.  
Happy to implement that if the maintainers prefer it.  
  
## Has this been tested?  
  
- [X] 👍 Tested manually on a live PeerTube instance — `lastmod` appears correctly in `/sitemap.xml`  

Screenshot:
Screenshot 2026-07-23 222348


Chocobozzz:
Yes please add `contentUpdatedAt` field

valeron90:
The sitemap lastmod value for videos is contentUpdatedAt.
publishedAt as a fallback for videos published before this migration or if nothing have been changed. The lastmod value for channel and account sitemap entries is set to updatedAt.

Currently, channels do not have a description, so they still use "updatedAt".

Migrationfile:

1075-video-content-updated-at.ts adds the contentUpdatedAt column.

Chocobozzz:
Thank you

PR #7746, opened 2026-08-02
Use ICU variables instead of hardcoded text

oaj349:
## Description

This PR focuses on using ICU variables instead of hardcoded text in the notifications regarding blocked servers and accounts. Should be treated independently of my bigger PR (#7612).

## Has this been tested?


Chocobozzz:
Thank you, added in https://github.com/Chocobozzz/PeerTube/commit/9fd2659c592d7f9902966e6d72cd73b51fa03b06

PR #7748, opened 2026-08-03
feat(client/plugins): pass router to plugin.register

kontrollanten:
## Description
Pass angular router to the plugins `register` function in order to let plugins navigate within the app.

## Has this been tested?


Chocobozzz:
Hello,

Prefer using an helper that would abstract angular router to navigate into the app

kontrollanten:
> Prefer using an helper that would abstract angular router to navigate into the app

I'm not sure how you want it exposed, byt I updated with an API with only `navigateByUrl` only accepting a string as parameter. I guess it will handle the most scenarios for plugins.

PR #7750, opened 2026-08-05
Feature/dynamic channel stats

### LINKED ISSUE #7711: Channel stats; raise the views-per-day chart window from 30 to 90 days

In *My Library → Channels* (`/my-library/video-channels`), each channel displays a "views per day" chart. The time window is a fixed rolling 30 days: `VideoChannelModel.listByAccountForAPI` hardcodes `statsDaysPrior: 30` when `withStats=true` (https://github.com/Chocobozzz/PeerTube/blob/3aa7ce442d7852f61ac832fe39e3d4de662ac215/server/core/models/video/video-channel.ts#L407), and the API validator only accepts the `withStats` boolean; there is no way to request a different range.

Since the window slides every day, channel owners permanently lose visibility of anything older than 30 days, even though the data is still in the database: `videoStat` rows for local videos are kept forever by default (`views.videos.local.max_age` defaults to `-1`). On my instance I have per-day stats going back months, but the UI can only ever show the last 30 days.

Video-level stats already support custom date ranges on the video stats page, so channel-level stats feel behind; creators expect to see how their channel is trending over a longer period (90 days ≈ a quarter is a common analytics preset).

**Raise the window from 30 to 90 days.** As far as I can tell it is a one-line change: the SQL already parameterizes the window (`statsDaysPrior` in `video-channel-list-query-builder.ts` builds `generate_series(now() - '<N> day', now())`), the response shape stays identical (same `viewsPerDay` field, just a longer array), and the client simply maps whatever array it receives; the line chart handles 90 points fine.

Cost-wise it looks negligible: I benchmarked the exact stats subquery on my instance and 90 days runs in ~3 ms per channel, adding ~1.5 KB (pre-gzip) per channel to the payload. This query also only runs for the logged-in owner's own channel list (`withStats=true`), never on public pages.

Going further would of course be even better; e.g. an optional validated query parameter (`statsDays=`), a small time-range selector above the chart (30 / 90 / all time, Studio-style), or an unbounded window with weekly/monthly aggregation; and I'd be happy with any of those too. But simply bumping the hardcoded 30 to 90 would already solve the main pain.

Related (not duplicates): #6835 asks for a full channel analytics overview (watch time, countries, devices…) this issue is a much smaller incremental step, only about the time range of the chart that already exists. #3843 is about instance-wide stats in the admin area.

sourav7274:
## Description

Extends the channel views chart on My Library → Channels beyond the hardcoded 30-day window.

Previously listByAccountForAPI always used statsDaysPrior: 30, so owners could not see older videoStat history even though it is kept in the DB. This PR adds:

A Stats range selector: 30 days / 90 days / last year / all time
API query param statsDays (30 | 90 | 365 | 0, default 30) passed through validator → controller → query builder (not client-side slicing)
Aggregation by effective history span (not the preset alone): day (< 60 days), week (60–400), month (> 400), so a young channel on “All time” does not get one coarse monthly bar
Bar chart with axes + hover tooltip; weekly buckets show “Week of …” (monthly stays like Nov 2025)
totalViews (when withStats=true) is the sum of views in the selected range (from videoStat), matching the chart
New response field viewsGroupInterval (day | week | month) so the client can label buckets correctly
Chart sits in the channel row (avatar | meta | chart | buttons) for a more balanced layout

## Related issues
N/A

## Has this been tested?

## Screenshots

https://github.com/user-attachments/assets/616e0404-c59d-4c19-82dd-08c5cecf728d

Screenshot 2026-08-05 211516
Screenshot 2026-08-05 211553

Closes #7711


Chocobozzz:
Thank you

Issue #7391, opened 2026-01-19
Docker timeout when stopping peertube


Myriade-Medieval:
### Describe the current behavior

This is better seen in docker compose as it displays the time a container takes to stop, when stopping peertube, it always takes around 10 secs, which is the default timeout for docker compose until it ungracefully stops a container.

I had this issue previously in one of my projects, and fixed it by using exec form on CMD and the exec command. But looking at your Dockerfile, you seem to be doing the same thing but still get a timeout

### Steps to reproduce

1. Follow the steps to get peertube running in docker compose
2. Run peertube as a daemon: `docker compose run -d`
3. Stop it `docker compose down`

Watch the main container take 10 seconds to stop, whereas the db and mail containers take under a second

Note: docker compose is not strictly required, I'm just following peertube's docker guide setup

### Describe the expected behavior

When stopping the docker container it is expected to shut down relatively quickly and not timeout

### Additional information

As this is a backend thing, it was tested locally on my machine and on my personal peertube instance.

Here is a screenshot of it happening

<img width="707" height="194" alt="Image" src="https://github.com/user-attachments/assets/dc3570e2-92e2-4d62-aa37-cb81e0f7e124" />

Chocobozzz:
It seems we have the same issue with the webserver service. It should be fixed by https://github.com/Chocobozzz/PeerTube/commit/2e56f31b954ed37ebde55fe529017bcf94306267

I tried to debug the peertube service shutdown, but I don't understand why it doesn't correctly exit on SIGTERM. PeerTube correctly exit on SIGTERM signal, and I don't see anything in the entrypoint/cmd that could break it.

hexsm:
I tested `chocobozzz/peertube:v8.1.8` and did `docker events --filter container=<container-name>`. This is the result in a short form:

```
11:17:46  signal=15  -> SIGTERM
11:17:56  signal=9   -> SIGKILL
exitCode=137
```

---

Additional finding:

`SIGTERM` does not stop the container, even when sent manually:

`docker kill --signal=SIGTERM <container-name>`

The container keeps running.

`SIGINT`, however, stops the container immediately and exits with code 0:

`docker kill --signal=SIGINT <container-name>`

As a workaround, setting the following in compose.yaml works:

```
services:
  peertube:
    stop_signal: SIGINT
```

---

A minimal fix could be to handle SIGTERM the same way as SIGINT in `server/server.ts`:

```
process.on('SIGINT', () => process.exit(0))
[...]

PR #7760, opened 2026-08-15
Fix docker timeout on stop

Myriade-Medieval:
## Description

The docker container of peertube doesn't stop properly and instead timeouts. as discussed in #7391 it is caused by the wrong signal being sent to the container. This commit changes the STOPSIGNAL in the Dockerfile.

## Related issues

#7391 

## Has this been tested?


Chocobozzz:
See https://github.com/Chocobozzz/PeerTube/issues/7391#issuecomment-5325527221

PR #7769, opened 2026-08-24
feat: add support for plugins to disable account description 

### LINKED ISSUE #7768: Plugin ability to disable account description

We've a lot of spam accounts who add links in their account description.

We're using [peertube-plugin-premium-users](https://www.npmjs.com/package/peertube-plugin-premium-users) and would like to try to make it a premium only feature to add account description.

This could be done by adding an API filter at the `getAccount` API and client plugin selector to hide the description field in the form.

kontrollanten:
## Description
Add support for plugins to disable account description. Also added plugin selector to username description.

## Related issues
closes #7768 

## Has this been tested?

### DIFF
--- a/client/src/app/+my-account/my-account-settings/my-account-profile/my-account-profile.component.html
+++ b/client/src/app/+my-account/my-account-settings/my-account-profile/my-account-profile.component.html
@@ -6,7 +6,7 @@
   <div class="form-group">
     <label i18n for="username">Username</label>
     <input type="text" id="username" class="form-control" formControlName="username" readonly>
-    <div id="plugin-selector-profile-username-description" class="form-group-description" i18n>
+    <div myPluginSelector pluginSelectorId="profile-form-username-description" class="form-group-description" i18n>
       People can find you using &#64;{{ user().username }}&#64;{{ instanceHost }}
     </div>
   </div>
@@ -25,7 +25,7 @@
 
     <my-help helpType="markdownText" supportRelMe="true"></my-help>
 
-    <div id="plugin-selector-profile-description">
+    <div myPluginSelector pluginSelectorId="profile-description">
       <my-markdown-textarea
         inputId="description"
         formControlName="description"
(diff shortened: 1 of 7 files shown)


client/src/app/+my-account/my-account-settings/my-account-profile/my-account-profile.component.html
```diff
@@ -6,7 +6,7 @@
   <div class="form-group">
     <label i18n for="username">Username</label>
     <input type="text" id="username" class="form-control" formControlName="username" readonly>
-    <div class="form-group-description" i18n>
+    <div id="plugin-selector-profile-username-description" class="form-group-description" i18n>
```
Chocobozzz: Use `myPluginSelector` directive

client/src/app/+my-account/my-account-settings/my-account-profile/my-account-profile.component.html
```diff
@@ -25,14 +25,16 @@
...
 
-    <my-markdown-textarea
-      inputId="description"
-      formControlName="description"
-      markdownType="enhanced"
-      [formError]="formErrors['description']"
-      withEmoji="true"
-      withHtml="true"
-    ></my-markdown-textarea>
+    <div id="plugin-selector-profile-description">
```
Chocobozzz: I think it would make more sense to have this plugin selector in the `form-group` `div` above no?

client/src/app/+my-account/my-account-settings/my-account-profile/my-account-profile.component.html
kontrollanten: Sure! Maybe a better name is `profile-form-username-description`?

client/src/app/+my-account/my-account-settings/my-account-profile/my-account-profile.component.html
kontrollanten: Sure. It's the textarea that we're interested in, but we can put the selector at a higher level.

kontrollanten:
A more consistent way of solving this would be to add `profile-form-[name]-field-group` to `username` and `description`. Then we could select their children with `.form-group-description` and `textarea` selectors. But that depends on that those selectors will be stable during updates, do you have any thoughts about this?

kontrollanten:
> A more consistent way of solving this would be to add `profile-form-[name]-field-group` to `username` and `description`. Then we could select their children with `.form-group-description` and `textarea` selectors. But that depends on that those selectors will be stable during updates, do you have any thoughts about this?

I updated the PR according to this comment. Please let me know what you think.

PR #7775, opened 2026-08-30
Improve Redis connection resilience for BullMQ job queue

### LINKED ISSUE #7774: No jobs seem to be running after some time running the instance

After running for some time (less than an hour) some requests take long and actually end up with a server error.
One of those requests is the one to check jobs, it seems no new job is run (I have transcoding jobs waiting and no ffmpeg is running on the server so I guess they aren't launched anymore). Also some jobs like activitypub-http-fetcher are timing out according to the logs.

I also have errors about no being able to renew a job lock so it might be related.

### Steps to reproduce

1. run for some time
2. check if jobs page is still accessible
3. once it isn't check logs for timeouts or other things showing no job is launched

_No response_

### Additional information

* PeerTube instance (found in https://instanceURL/about/instance/tech):
  * URL: https://peertube.qontinuum.space
  * Version: 8.2.4
  * NodeJS version: 24.20.0
  * Ffmpeg version: 7.1.5

* Link to server log if relevant (`journalctl` or `/var/www/peertube/storage/logs/`): https://termbin.com/oxyao

m-altaifi:
### Summary
This PR improves Redis connection resilience for BullMQ workers and queues, addressing issues where idle TCP connections get dropped (causing `read ETIMEDOUT`) and workers subsequently fail background lock renewal or stop processing jobs (e.g. #7774).

### Changes
1. **TCP Keepalive**: Configured `keepAlive: 30000` in `Redis.getRedisClientOptions()` to send periodic TCP keep-alive probes across Redis connections.
2. **BullMQ Connection Options**: Passed `maxRetriesPerRequest: null` and `enableReadyCheck: false` to BullMQ `Worker`, `Queue`, `QueueEvents`, and `FlowProducer` connections as recommended by BullMQ to allow graceful reconnection and prevent worker command rejection during transient socket blips.

Closes #7774

### DIFF
--- a/server/core/lib/job-queue/job-queue.ts
+++ b/server/core/lib/job-queue/job-queue.ts
@@ -44,6 +44,7 @@ import {
   Worker,
   WorkerOptions
 } from 'bullmq'
+import { RedisOptions } from 'ioredis'
 import { createLogger } from '../../helpers/logger.js'
 import { JOB_ATTEMPTS, JOB_CONCURRENCY, JOB_REMOVAL_OPTIONS, JOB_TTL, REPEAT_JOBS, WEBSERVER } from '../../initializers/constants.js'
 import { Hooks } from '../plugins/hooks.js'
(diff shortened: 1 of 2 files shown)


server/core/lib/job-queue/job-queue.ts
```diff
@@ -230,7 +230,7 @@ class JobQueue {
     }
 
     this.flowProducer = new FlowProducer({
-      connection: Redis.getRedisClientOptions('FlowProducer'),
+      connection: Redis.getRedisClientOptions('FlowProducer', { maxRetriesPerRequest: null, enableReadyCheck: false }),
```
Chocobozzz: Please wrap `{ maxRetriesPerRequest: null, enableReadyCheck: false }` in an object to not repeat it across the file, and add a small comment explaining why we chose these options

Chocobozzz (commented):
Can you target `release/v8.3.x` branch?

Chocobozzz:
Thank you

PR #7779, opened 2026-09-01
Document how to reinstall a plugin during development

### LINKED ISSUE #7765: Plugin installation during development unclear
### Describe the desired information/correction

I've been working on a custom plugin, but I was having a bit of trouble installing subsequent versions during development, with peertube running in docker.

I used the docs at https://docs.joinpeertube.org/contribute/plugins#test-your-plugin-theme to install my plugin, by using a volume mount to get the plugin source inside the docker container and then running:

```
(cd /path/to/peertube-plugin-hypha-festival-import; npm run build) && peertube-cli plugins install -u 'http://localhost:9000' --path /plugin-src/peertube-plugin-hypha-festival-import
```

However, when I then change the plugin code and rerun the above command, it does not update to the latest version, but just reinstalls the existing version. Inititially I had an error in package.json, so the installation failed, but rerunning the above command after fixing that would try to reinstall the first version instead of the new version. I confirmed that the new code was indeed present in the container, but adding some debug output shows that the package is installed from `/data/plugins/node_modules/peertube-plugin-hypha-festival-import`. It looks like the install runs a `pnpm add` on the updated source in `/plugin-src`, but somehow that does not update the copy in `/data/plugins/...`.

I thought that maybe the version number would need to be bumped, but this did not seem to help either.

Maybe this is super obvious to someone with npm experience, but I am not too familiar with how npm packages work.

So the main question is: What would be the typical development flow to update a plugin after making a change?

A related question is whether this can be done without restarting the docker container, to shorten the iterations a bit?

For now, I ended up just removing the `/data/plugins` version, e.g. updating with:

```
(cd /home/matthijs/docs/src/peertube-plugin-hypha-festival-import; npm run build) && docker compose exec peertube rm -rf /data/plugins/node_modules/peertube-plugin-hypha-festival-import && peertube-cli plugins install -u 'http://localhost:9000' --path /plugin-src/peertube-plugin-hypha-festival-import && docker compose restart peertube
```

If you can help me figure out how this can be done, I would be happy to make a PR for the docs to clarify this.

m-altaifi:
## Description

Fixes #7765.

The plugin development guide tells you to install a local plugin with `peertube-cli plugins install --path …`, but not what to do next time you rebuild it. Re-running the same command reinstalls the plugin the instance already has, so it looks like the new code is being ignored — which is what @matthijskooijman ran into.

The cause is that `installNpmPluginFromDisk` runs `pnpm add file:<path>` (`server/core/lib/plugins/package-manager.ts:28`), which copies the directory into the plugin storage rather than linking to it, and `install()` records `packageJSON.version` on the plugin row. So editing the source has no effect until the plugin is reinstalled under a new version.

`peertube-cli plugins update --path` already exists for exactly this, but the plugin guide never mentions it — the section only documents `install`. This adds that command, explains why rebuilding alone does nothing, and links the CLI tools guide from the section, since the second half of the issue was that the CLI docs were hard to find from here.

No behaviour change, documentation only.

## Related issues

#7765

## Has this been tested?

### DIFF
--- a/support/doc/plugins/guide.md
+++ b/support/doc/plugins/guide.md
@@ -1213,7 +1213,7 @@ You built files are in the `dist/` directory. Check `package.json` to correctly 
 You need to have a local PeerTube instance with an administrator account.
 If you're using dev server on your local computer, test your plugin on `localhost:9000` using `npm run dev` because plugin CSS is not injected in Angular webserver (`localhost:3000`).
 
-Install PeerTube CLI (can be installed on another computer/server than the PeerTube instance):
+Install [PeerTube CLI](https://docs.joinpeertube.org/maintain/tools) (can be installed on another computer/server than the PeerTube instance):
 
 ```bash
 npm install -g @peertube/peertube-cli
@@ -1232,6 +1232,16 @@ If the PeerTube instance is running on another server/computer, you must copy yo
 ```sh
 peertube-cli plugins install --path /your/absolute/plugin-or-theme/path
 ```
+
+`plugins install` copies the directory into the instance plugin storage instead of linking to it,
+so rebuilding your plugin/theme afterwards does not change what the instance runs.
+To load a new build, update the `version` key in your `package.json` and reinstall it:
+
+```sh
+peertube-cli plugins update --path /your/absolute/plugin-or-theme/path
+```
+
+Reusing the same version reinstalls the copy the instance already has, so your changes look like they were ignored.
 
 ### Publish
 


support/doc/plugins/guide.md
```diff
@@ -1233,6 +1233,16 @@ If the PeerTube instance is running on another server/computer, you must copy yo
 peertube-cli plugins install --path /your/absolute/plugin-or-theme/path
 ```
 
+`plugins install` copies the directory into the instance plugin storage instead of linking to it,
+so rebuilding your plugin/theme afterwards does not change what the instance runs.
```
Chocobozzz, line 1237: I think it would be simpler to specify the plugin developer should uninstall and reinstall the plugin instead of updating the package version

